PR #672, opened 2023-07-17
[Doc] support front matter

wangweipeng2:
## What do these changes do?
Currently we use "---" as front matter to describe the doc order and title. But in GitHub markdown preview, it does not work fine. So we use the comment "" as front matter which can be previewed rightly in GitHub. When syncing to Fury site, we replace the comment front matter back to "---".


chaokunyang (approved):
LGTM

PR #763, opened 2023-07-25
[Rust] Several Rust idioms

tisonkun:
BTW, you can publish fury-derive to hold the crate name.

Also, [fury](https://crates.io/crates/fury) is taken by @carllerche so we may have a communication with them for possibly passing the name.


chaokunyang (approved):
LGTM

chaokunyang:
> BTW, you can publish fury-derive to hold the crate name.
> 
> Also, [fury](https://crates.io/crates/fury) is taken by @carllerche so we may have a communication with them for possibly passing the name.

Thanks for reminding this, seems [rust fury](https://crates.io/crates/fury) hasn't been updating for three years, will talking to  @carllerche for naming issue

PR #768, opened 2023-07-26
[Java] remove requireClassRegistration API

chaokunyang:
## What do these changes do?
remove requireClassRegistration API

## Related issue number

Closes #xxxx

## Check code requirements


chaokunyang:
@tisonkun Could you help review this PR?

tisonkun (commented):
While `withSecureMode(false)` is vague and even ambiguous, is it possible with use `requireClassRegistration` to replace all the usage of `withSecureMode` instead and document `requireClassRegistration(false)`'s safety requirement (i.e., "your environment is indeed secure").

I don't know if `withSecureMode` has other meaning - it reads vague and ambiguous to me. Here we can expose the real effect `requireClassRegistration` that is not an internal concept but what users modify the option need to know.

chaokunyang:
Good suggestion, `withSecureMode` did read vague and ambiguous, it didn't show what it does.

docs/guide/java_object_graph_guide.md
```diff
@@ -94,7 +94,7 @@ System.out.println(fury.deserialize(bytes));
 ```
 
 ### Security & Class Registration
-`FuryBuilder#requireClassRegistration`/`FuryBuilder#withSecureMode` can be used to disable class registration, this will allow to deserialize objects unknown types, more flexible but **less secure**.
+`FuryBuilder#requireClassRegistration` can be used to disable class registration, this will allow to deserialize objects unknown types, more flexible but **less secure**.
```
tisonkun, line 97: Perhaps we can elaborate a bit why it's "less secure" and thus users can check if their env is "indeed secure".

tisonkun (approved):
LGTM. One comment inline that can be improved later.

docs/guide/java_object_graph_guide.md
chaokunyang: Good suggestion, added detailed insecurity reason

PR #786, opened 2023-07-27
Corrected grammar and paraphrasing to make the README easier to read

hieu-ht:
## What do these changes do?
Corrected grammar and paraphrasing to make the README easier to read

## Related issue number

Closes #736 

## Check code requirements

- [ ] tests added / passed (if needed)
- [ ] Ensure all linting tests pass, see [here](https://github.com/alipay/fury/blob/main/CONTRIBUTING.rst) for how to run them

PR #787, opened 2023-07-27
Don't let CI run when the changed files are not the code

hieu-ht:
## What do these changes do?
I created this PR #786 to change README.md, it is not the code but github action CI still running. To avoid unnecessary CI runs, it's best to prevent them from triggering when only non-code files have been modified.

## Related issue number

Closes #xxxx

## Check code requirements


chaokunyang (approved):
Good catch, thanks for this

PR #908, opened 2023-09-11
Optimize StringBuilder/StringBuffer serialization

### LINKED ISSUE #94: [Java] Optimize StringBuilder/StringBuffer serialization
**Is your feature request related to a problem? Please describe.**
In #93, we implement StringBuilder/StringBuffer serialization by converting to/from java.lang.String, which have some copy cost. A better solution is tackle inner data structure to avoid this copy.

**Additional context**
#92 #93

pandalee99:
## What do these changes do?

To optimize the code, I avoiding the unnecessary string conversion and directly operating on the internal character array and length of the StringBuffer.  This would involve modifying the write() and read() methods to work directly with the StringBuffer's internal data structure.

## Related issue number

Closes #94 

## Check code requirements


java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -420,12 +420,21 @@ public StringBuilderSerializer(Fury fury) {
 
     @Override
     public void write(MemoryBuffer buffer, StringBuilder value) {
-      stringSerializer.writeJavaString(buffer, value.toString());
+      int length = value.length();
+      buffer.writeInt(length);
```
chaokunyang, line 424: use `buffer.writePositiveVarint` can svae 1~3 bytes

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -420,12 +420,21 @@ public StringBuilderSerializer(Fury fury) {
...
+      for (int i = 0; i < length; i++) {
+        buffer.writeChar(value.charAt(i));
```
chaokunyang, line 426: `charAt` is expensive in JDK9+ since internal `java.lang.AbstractStringBuilder` use `byte[]` instead for jdk9+.

Can we extract internal `char[]` for jdk8 and `byte[]` for jdk9+ to serialize instead?

chaokunyang: We did same thing in `StringSerializer`

chaokunyang:
Hi @pandalee99 , thanks for taking time to contribute to fury, I left some comments, could you please take a look at it?

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
pandalee99: cool

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
pandalee99: yup, i try to refer

java/fury-core/src/main/java/io/fury/serializer/StringSerializer.java
```diff
@@ -705,4 +704,117 @@ public String readUTF8String(MemoryBuffer buffer) {
...
+  }
+
+  private StringBuilder readJDK9PlusStringBuilder(MemoryBuffer buffer) {
+    int length = buffer.readInt();
+    byte[] bytes = new byte[length];
+    buffer.readBytes(bytes);
+    return new StringBuilder(new String(bytes, StandardCharsets.UTF_8));
+  }
+
+  public boolean isJDK8StringBuilder() {
```
chaokunyang, line 808: `io.fury.util.Platform.JAVA_VERSION` can be used for check jdk version

pandalee99: thanks,i try to make it and fix error

chaokunyang:
StringBuilder didn't provide a zero-copy constructor, can we optimize write in this PR only to reduce complexities.

For example, we can keep protocol consistent with StringSerializer by :

```java

  static Tuple2<ToByteFunction, Function> builderCache;

  private static synchronized Tuple2<ToByteFunction, Function> getBuilderFunc() {
    if (builderCache == null) {
      Function getValue = (Function) Functions.makeGetterFunction(
        StringBuilder.class.getSuperclass(), "getValue");
      if (Platform.JAVA_VERSION > 8) {
        ToByteFunction getCoder = (ToByteFunction) Functions.makeGetterFunction(
          StringBuilder.class.getSuperclass(), "getCoder");
        builderCache = Tuple2.of(getCoder, getValue);
      } else {
        builderCache = Tuple2.of(null, getValue);
      }
    }
    return builderCache;
  }

  public static abstract class AbstractStringBuilderSerializer<T> extends Serializer<T> {
    protected final ToByteFunction getCoder;
    protected final Function getValue;
    protected final StringSerializer stringSerializer;

    public AbstractStringBuilderSerializer(Fury fury, Class type) {
      super(fury, type);
      Tuple2<ToByteFunction, Function> builderFunc = getBuilderFunc();
      getCoder = builderFunc.f0;
      getValue = builderFunc.f1;
      stringSerializer = fury.getStringSerializer();
    }

    @Override
    public void write(MemoryBuffer buffer, T value) {
      if (Platform.JAVA_VERSION > 8) {
        byte coder = getCoder.applyAsByte(value);
        byte[] v = (byte[]) getValue.apply(value);
        buffer.writeByte(coder);
        buffer.writeBytesWithSizeEmbedded(v);
      } else {
        char[] v = (char[]) getValue.apply(value);
        if (StringSerializer.isAscii(v)) {
          stringSerializer.writeJDK8Ascii(buffer, v, v.length);
        } else {
          stringSerializer.writeJDK8UTF16(buffer, v, v.length);
        }
      }
      stringSerializer.writeJavaString(buffer, value.toString());
    }
  }

  public static final class StringBuilderSerializer extends AbstractStringBuilderSerializer<StringBuilder> {
    
    public StringBuilderSerializer(Fury fury) {
      super(fury, StringBuilder.class);
    }

    @Override
    public StringBuilder read(MemoryBuffer buffer) {
      return new StringBuilder(stringSerializer.readJavaString(buffer));
    }
  }

```

Same thing can be done for `StringBuffer`

pandalee99:
>         char[] v = (char[]) getValue.apply(value);

Thank you, but I found a problem, the problem is, if using this way to get char[], then the final result is wrong, the length of char[] will be 19, the content will be "strnullnullnullnull..." This could be a jdk bug

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,17 +415,89 @@ public Double read(MemoryBuffer buffer) {
...
+      getValue = builderFunc.f1;
       stringSerializer = new StringSerializer(fury);
     }
 
     @Override
-    public void write(MemoryBuffer buffer, StringBuilder value) {
-      stringSerializer.writeJavaString(buffer, value.toString());
+    public void write(MemoryBuffer buffer, T value) {
+      if (Platform.JAVA_VERSION > 8) {
+        byte[] v = new byte[0];
```
chaokunyang, line 452:
```suggestion
        byte[] v = (byte[])(getValue.apply(value))
```

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,17 +415,89 @@ public Double read(MemoryBuffer buffer) {
...
+        byte coder = getCoder.applyAsByte(value);
+        try {
+          Method getValueMethod =
```
chaokunyang, line 455: Can we use `byte[] v = (byte[])(getValue.apply(value))` to extract value?

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -444,7 +521,11 @@ public short getXtypeId() {
 
     @Override
     public void write(MemoryBuffer buffer, StringBuffer value) {
-      stringSerializer.writeJavaString(buffer, value.toString());
+      int length = value.length();
```
chaokunyang, line 524: ditto

java/fury-core/src/main/java/io/fury/serializer/StringSerializer.java
```diff
@@ -705,4 +705,110 @@ public String readUTF8String(MemoryBuffer buffer) {
       return new String(tmpArray, 0, numBytes, StandardCharsets.UTF_8);
     }
   }
+
+  public void writeJavaStringBuilder(MemoryBuffer buffer, StringBuilder value) {
```
chaokunyang, line 709: Seems unnecessary, we can provide `public void writeJDK8Ascii(MemoryBuffer buffer, char[] chars, int strLen)` and `public void writeJDK8UTF16(MemoryBuffer buffer, char[] chars, int strLen)` in StringSerializer and cann those function in this class

pandalee99: Yes, it is unnecessary

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
pandalee99: I didn't do well in my tests. I'm not sure why

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -429,38 +493,82 @@ public StringBuilder read(MemoryBuffer buffer) {
...
+                Functions.makeGetterFunction(StringBuffer.class.getSuperclass(), "getCoder");
+        bufferCache = Tuple2.of(getCoder, getValue);
+      } else {
+        bufferCache = Tuple2.of(null, getValue);
+      }
+    }
+    return bufferCache;
+  }
+
+  public abstract static class AbstractStringBufferSerializer<T> extends Serializer<T> {
```
chaokunyang, line 514: This seems unnecessary, `StringBufferSerializer` can just extends `AbstractStringBuilderSerializer`. Both `StringBuffer` and `StringBuilder` share same parent class

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,17 +415,76 @@ public Double read(MemoryBuffer buffer) {
...
-    public void write(MemoryBuffer buffer, StringBuilder value) {
-      stringSerializer.writeJavaString(buffer, value.toString());
+    public void write(MemoryBuffer buffer, T value) {
+      if (Platform.JAVA_VERSION > 8) {
+        byte coder = getCoder.applyAsByte(value);
+        byte[] v = (byte[]) getValue.apply(value);
+        buffer.writeByte(coder);
+        buffer.writeBytesWithSizeEmbedded(v);
+      } else {
+        char[] v = new char[0];
```
chaokunyang, line 457: Does `char[] v = (char[]) getValue.apply(value)` work?

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
pandalee99: thanks,it should be

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
pandalee99: and then, It is difficult to test jdk9+ issues locally

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,17 +414,60 @@ public Double read(MemoryBuffer buffer) {
...
+    public void write(MemoryBuffer buffer, T value) {
+      if (Platform.JAVA_VERSION > 8) {
+        byte coder = getCoder.applyAsByte(value);
+        byte[] v = (byte[]) getValue.apply(value);
+        buffer.writeByte(coder);
+        buffer.writeBytesWithSizeEmbedded(v);
+      } else {
+        char[] v = (char[]) getValue.apply(value);
+        if (StringSerializer.isAscii(v)) {
+          stringSerializer.writeJDK8Ascii(buffer, v);
```
chaokunyang, line 458:
```suggestion
          stringSerializer.writeJDK8Ascii(buffer, v, v.lengh());
```
Can we pass length to reduce extra data transfer?

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,17 +414,60 @@ public Double read(MemoryBuffer buffer) {
...
+      } else {
+        builderCache = Tuple2.of(null, getValue);
+      }
+    }
+    return builderCache;
+  }
 
-    public StringBuilderSerializer(Fury fury) {
-      super(fury, StringBuilder.class);
+  public abstract static class AbstractStringBuilderSerializer<T> extends Serializer<T> {
```
chaokunyang, line 435:
```suggestion
  public abstract static class AbstractStringBuilderSerializer<T extends CharSequence> extends Serializer<T> {
```

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
pandalee99: Yup,I to modify

chaokunyang:
@pandalee99 `io.fury.util.function.ToByteFunction` is not defined in jdk library, we can't use it to generate accessor lambda for jdk classes, otherwise jdk will raise ClassNotFoundException. 

Can we add a new `makeFunction` in `Functions`:
```java
  public static Object makeGetterFunction(Method method) {
    return makeGetterFunction(method, method.getReturnType());
  }
  
  public static Object makeGetterFunction(Method method, Class<?> returnType) {
    MethodHandles.Lookup lookup = _JDKAccess._trustedLookup(method.getDeclaringClass());
    try {
      // Why `lookup.findGetter` doesn't work?
      // MethodHandle handle = lookup.findGetter(field.getDeclaringClass(), field.getName(),
      // field.getType());
      MethodHandle handle = lookup.unreflect(method);
      return _JDKAccess.makeGetterFunction(lookup, handle, returnType);
    } catch (IllegalAccessException ex) {
      throw new RuntimeException(ex);
    }
  }
```

then creating accessor by :
```java
    Method getCoder = StringBuilder.class.getSuperclass().getDeclaredMethod("getCoder");
    ToIntFunction<CharSequence> o = (ToIntFunction<CharSequence>) makeGetterFunction(lookup, lookup.unreflect(getCoder), int.class);
    System.out.println(o);
    System.out.println(o.applyAsInt(new StringBuilder("abc")));
    System.out.println(o.applyAsInt(new StringBuilder("abc你好")));
```

pandalee99:
thanks.        
 but,Some of the names of the Functions are repeated.and the `lookup` in the `java.lang.CharacterName` is private .maybe have other way

pandalee99:
thanks.
but,Some of the method of the Functions are repeated.  and the `lookup` in the `java.lang.CharacterName` is private .maybe have other way, and the ToIntFunction maybe not able to `Type conversion`

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,17 +414,69 @@ public Double read(MemoryBuffer buffer) {
...
   }
 
-  public static final class StringBuilderSerializer extends Serializer<StringBuilder> {
-    private final StringSerializer stringSerializer;
+  static Tuple2<Function, Function> builderCache;
+
+  private static synchronized Tuple2<Function, Function> getBuilderFunc() {
+    if (builderCache == null) {
+      Function getValue =
+          (Function) Functions.makeGetterFunction(StringBuilder.class.getSuperclass(), "getValue");
```
chaokunyang, line 422: Can we pass `int.class` to `Functions.makeGetterFunction` to get `ToIntFunction<CharSequence>`, otherwise following exception wil lbe raised:
image

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,17 +414,69 @@ public Double read(MemoryBuffer buffer) {
...
+      if (Platform.JAVA_VERSION > 8) {
+        Method getCoderMethod;
+        try {
+          getCoderMethod = StringBuilder.class.getSuperclass().getDeclaredMethod("getCoder");
+        } catch (NoSuchMethodException e) {
+          throw new RuntimeException(e);
+        }
+        //        ToIntFunction<CharSequence> o = (ToIntFunction<CharSequence>)
```
chaokunyang, line 430: maybe we need to use this, otherwise following exception wil lbe raised:
image

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -429,38 +485,17 @@ public StringBuilder read(MemoryBuffer buffer) {
...
+  public static final class StringBufferSerializer
+      extends AbstractStringBuilderSerializer<StringBuffer> {
 
     public StringBufferSerializer(Fury fury) {
       super(fury, StringBuffer.class);
-      stringSerializer = new StringSerializer(fury);
-    }
-
-    @Override
-    public short getXtypeId() {
```
chaokunyang, line 441: Can we add this to `AbstractStringBuilderSerializer`? Otherwise cross-language StringBuilder/Buffer will raise error?

pandalee99: yes,i can to do

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,57 +416,90 @@ public Double read(MemoryBuffer buffer) {
...
 
     @Override
-    public StringBuilder read(MemoryBuffer buffer) {
-      return new StringBuilder(stringSerializer.readJavaString(buffer));
+    public void write(MemoryBuffer buffer, T value) {
+      if (Platform.JAVA_VERSION > 8) {
+        int coder = getCoder.applyAsInt(value);
+        byte[] v = (byte[]) getValue.apply(value);
+        buffer.writeByte(coder);
+        buffer.writeBytesWithSizeEmbedded(v);
```
chaokunyang, line 467:
```suggestion
        if (coder == 0) {
          buffer.writePrimitiveArrayWithSizeEmbedded(v, Platform.BYTE_ARRAY_OFFSET, value.length());
        } else {
          if (coder != 1) {
            throw new UnsupportedOperationException("Unsupported coder " + coder);
          }
          buffer.writePrimitiveArrayWithSizeEmbedded(
              v, Platform.BYTE_ARRAY_OFFSET, value.length() << 1);
        }
```

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,57 +416,90 @@ public Double read(MemoryBuffer buffer) {
...
+    protected final Function getValue;
+    protected final StringSerializer stringSerializer;
+
+    public AbstractStringBuilderSerializer(Fury fury, Class<T> type) {
+      super(fury, type);
+      Tuple2<ToIntFunction, Function> builderFunc = getBuilderFunc();
+      getCoder = builderFunc.f0;
+      getValue = builderFunc.f1;
       stringSerializer = new StringSerializer(fury);
     }
```
chaokunyang, line 455: @Override
    public void xwrite(MemoryBuffer buffer, T value) {
      stringSerializer.writeUTF8String(buffer, value.toString());
    }

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,57 +416,90 @@ public Double read(MemoryBuffer buffer) {
...
+      super(fury, StringBuilder.class);
     }
 
     @Override
-    public short getXtypeId() {
-      return (short) -Type.STRING.getId();
+    public StringBuilder read(MemoryBuffer buffer) {
+      return new StringBuilder(stringSerializer.readJavaString(buffer));
     }
+  }
```
chaokunyang, line 491: @Override
    public StringBuilder xread(MemoryBuffer buffer) {
      return new StringBuilder(stringSerializer.readUTF8String(buffer));
    }

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,46 +416,102 @@ public Double read(MemoryBuffer buffer) {
...
+          if (coder != 1) {
+            throw new UnsupportedOperationException("Unsupported coder " + coder);
+          }
+          buffer.writePrimitiveArrayWithSizeEmbedded(
+              v, Platform.BYTE_ARRAY_OFFSET, value.length() << 1);
+        }
+      } else {
+        char[] v = (char[]) getValue.apply(value);
+        if (StringSerializer.isAscii(v)) {
+          stringSerializer.writeJDK8Ascii(buffer, v, v.length);
```
chaokunyang, line 484:
```suggestion
          stringSerializer.writeJDK8Ascii(buffer, v, value.length());
```

java/fury-core/src/main/java/io/fury/serializer/Serializers.java
```diff
@@ -410,46 +416,102 @@ public Double read(MemoryBuffer buffer) {
...
+        } else {
+          stringSerializer.writeJDK8UTF16(buffer, v, v.length);
```
chaokunyang, line 486:
```suggestion
          stringSerializer.writeJDK8UTF16(buffer, v, value.length());
```

pandalee99:
so good!

chaokunyang (approved):
LGTM, thank you so much for your contribution. It's not an easy work for such optimization

PR #1028, opened 2023-10-25
[Doc] add scala user doc

### LINKED ISSUE #1024: [Java] serializing Scala Enumerations is complicated
**Describe the bug**
https://www.scala-lang.org/api/2.13.x/scala/Enumeration.html is not a popular feature in Scala but it exists. Scala 3 has a new enum class (as well as the scala.Enumeration). Enumeratum is a popular 3rd party lib.

I found serializing scala.Enumeration instances awkward.

I get an exception unless I register scala.Enumeration.Val class and this is not public. I had to create a hacky class in Java to access it.

https://github.com/pjfanning/fury-scala-test/blob/a6585dc407ac2f15f09352aec1f85025c24adf2f/src/main/java/scala/FuryUtils.java

I also had to register the scala.runtime.ModuleSerializationProxy

https://github.com/pjfanning/fury-scala-test/blob/a6585dc407ac2f15f09352aec1f85025c24adf2f/src/test/scala/io/github/pjfanning/fury/FuryTest.scala#L26-L33

**Expected behavior**

If and when Fury adds specialized Scala support, making this easier should be considered.

chaokunyang:
## What do these changes do?
add scala user doc

## Related issue number

Closes #1024 

## Check code requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ In addition to cross-language serialization, Fury also features at:
   It can greatly improve the efficiency of high-performance RPC calls, data transfer and object persistence.
 - JDK serialization **100% compatible**: JDK `writeObject/readObject/writeReplace/readResolve/readObjectNoData/Externalizable` serialization API supported, **JDK8~21** supported.
 - Supports shared and circular reference object serialization for golang.
+- Supports [scala object serialization](https://github.com/alipay/fury/blob/main/docs/guide/scala.md)
 - Supports automatic object serialization for golang.
 
 ## Protocols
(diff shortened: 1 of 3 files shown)


chaokunyang:
Hi @pjfanning , I just add a simple scala doc. It would be great if you can take a look at it.

pjfanning:
I did a quick test with Scala List and it looks like a lot of internal Scala classes are needed under the hood and they all need to be explicitly registered (unless you disable the need for registration).

I can come back to this but it could take a while. I suspect that the list of Scala library classes that need to be registered is going to be very large.

To make matters worse, Scala 2.13 rewrote a lot of Scala code relating to the collections - so the internal classes will be different from Scala 2.12. Any code to register a curated list of Scala library classes will need to try to register classes but move on if some of them fail.

This could be a fair amount of work.

chaokunyang:
That would be an issue. Registered should be stable across scala versions.

Do we need to support registration check for some clas pattern such as scala.collection?

chaokunyang:
Hi @pjfanning , do you think we should merge this pr now, or hold it until we have a more scala-native serialization support?

There are some fury scala users already: https://github.com/alipay/fury/issues/1029

pjfanning:
It's fine to merge as is.

PR #1034, opened 2023-10-28
[Java] Allow for `InputStream` not reading entire length

knutwannheden:
The `InputStream#read()` API allows for an implementation to only partially read the requested number of bytes. Update the calls in `Fury#deserialize()` and `Fury#deserializeFromStream()` to account for this.

### DIFF
--- a/java/fury-core/src/main/java/io/fury/Fury.java
+++ b/java/fury-core/src/main/java/io/fury/Fury.java
@@ -703,19 +703,7 @@ public final class Fury {
   public Object deserialize(InputStream inputStream, Iterable<MemoryBuffer> outOfBandBuffers) {
     buffer.readerIndex(0);
     try {
-      int read = inputStream.read(buffer.getHeapMemory(), 0, 4);
-      Preconditions.checkArgument(read == 4);
-      int size = buffer.readInt();
-      buffer.ensure(size + 4);
-      read = 0;
-      while (read < size) {
-        int count;
-        if ((count = inputStream.read(buffer.getHeapMemory(), read + 4, size - read)) == -1) {
-          break;
-        }
-        read += count;
-      }
-      Preconditions.checkArgument(read == size);
+      transferObjectFromStream(inputStream, buffer);
       return deserialize(buffer, outOfBandBuffers);
     } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


java/fury-core/src/main/java/io/fury/Fury.java
```diff
@@ -707,7 +707,14 @@ public Object deserialize(InputStream inputStream, Iterable<MemoryBuffer> outOfB
       Preconditions.checkArgument(read == 4);
       int size = buffer.readInt();
       buffer.ensure(size + 4);
-      read = inputStream.read(buffer.getHeapMemory(), 4, size);
+      read = 0;
```
chaokunyang, line 710: Do we need to extract this part as a private method? It's repeated twice.

knutwannheden: I will refactor and add a unit test.

knutwannheden: Let me know if you want any more changes or feel free to update the code so that it fits your personal style.

java/fury-core/src/main/java/io/fury/Fury.java
```diff
@@ -703,12 +703,7 @@ public Object deserialize(InputStream inputStream) {
   public Object deserialize(InputStream inputStream, Iterable<MemoryBuffer> outOfBandBuffers) {
     buffer.readerIndex(0);
```
chaokunyang, line 704: Can we move this line into `transferObjectFromStream`?

java/fury-core/src/main/java/io/fury/Fury.java
```diff
@@ -1146,6 +1136,22 @@ private Object deserializeFromStream(
     }
   }
 
+  private static void transferObjectFromStream(InputStream inputStream, MemoryBuffer buffer) throws IOException {
```
chaokunyang, line 1139: `readToBufferFromStream` seems more intuitive to me, since we are copy binary instead of object

java/fury-core/src/main/java/io/fury/Fury.java
knutwannheden: Done.

PR #1044, opened 2023-10-30
[Java] Use `ReflectionUtils.getCtrHandle()` in `ExternalizableSerializer`

knutwannheden:
When deserializing an `Externalizable` object, it may end up throwing an exception, as the class may be declared as non-public in another package and thus isn't accessible from the serializer. Fix this by using the `ReflectionUtils.getCtrHandle()` utility.

### DIFF
--- a/java/fury-core/src/main/java/io/fury/serializer/ExternalizableSerializer.java
+++ b/java/fury-core/src/main/java/io/fury/serializer/ExternalizableSerializer.java
@@ -41,6 +41,7 @@ public class ExternalizableSerializer<T extends Externalizable> extends Serializ
     super(fury, cls);
     try {
       constructor = cls.getConstructor();
+      constructor.setAccessible(true);
     } catch (NoSuchMethodException e) {
       Utils.ignore(e);
     }
--- a/java/fury-core/src/main/java/io/fury/serializer/ExternalizableSerializer.java
+++ b/java/fury-core/src/main/java/io/fury/serializer/ExternalizableSerializer.java
@@ -21,11 +21,10 @@ import io.fury.io.FuryObjectInput;
 import io.fury.io.FuryObjectOutput;
 import io.fury.memory.MemoryBuffer;
 import io.fury.util.Platform;
-import io.fury.util.Utils;
+import io.fury.util.ReflectionUtils;
 import java.io.Externalizable;
 import java.io.IOException;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
+import java.lang.invoke.MethodHandle;
 
 /**
  * Serializer for class implements {@link Externalizable}.
@@ -33,18 +32,13 @@ import java.lang.reflect.InvocationTargetException;
  * @author chaokunyang
  */
 public class ExternalizableSerializer<T extends Externalizable> extends Serializer<T> {
-  private Constructor<T> constructor;
+  private final MethodHandle constructor;
   private final FuryObjectInput objectInput;
   private final FuryObjectOutput objectOutput;
 
   public ExternalizableSerializer(Fury fury, Class<T> cls) {
     super(fury, cls);
-    try {
-      constructor = cls.getConstructor();
-      constructor.setAccessible(true);
-    } catch (NoSuchMethodException e) {
-      Utils.ignore(e);
-    }
+    constructor = ReflectionUtils.getCtrHandle(cls, false);
 
     objectInput = new FuryObjectInput(fury, null);
     objectOutput = new FuryObjectOutput(fury, null);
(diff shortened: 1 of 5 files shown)


java/fury-core/src/main/java/io/fury/serializer/ExternalizableSerializer.java
```diff
@@ -41,6 +41,7 @@ public ExternalizableSerializer(Fury fury, Class<T> cls) {
     super(fury, cls);
     try {
       constructor = cls.getConstructor();
+      constructor.setAccessible(true);
```
chaokunyang, line 44: Good catch! I missed this before. Can we use `io.fury.util.ReflectionUtils#getCtrHandle(java.lang.Class<?>, boolean)` instead, since `setAccessible` won't work for JDK17 if the class isn't accessible from this module

java/fury-core/src/test/java/io/fury/serializer/ExternalizableSerializerTest.java
```diff
@@ -18,49 +18,17 @@
...
-    }
-  }
-
   @Test
   public void testExternalizable() {
-    A a = new A();
-    a.x = 1;
-    a.y = 1;
-    a.bytes = "bytes".getBytes();
+    Externalizable a = Factory.newA(1, 1, "bytes".getBytes());
```
chaokunyang, line 31: Seems `Factory` class not committed

java/fury-core/src/main/java/io/fury/serializer/ExternalizableSerializer.java
knutwannheden: Oh, I didn't even know. Thanks for pointing that out. Will update the PR.

knutwannheden: Not sure about the `checked` flag, but I set it to `false`...

java/fury-core/src/test/java/io/fury/serializer/ExternalizableSerializerTest.java
chaokunyang, line 31: Is there some reason for using `Factory` to create `A` instead of invoking constructor directly?

knutwannheden: In my code the issue was only reproducible when `A` was in a different package and had `package` visibility. So from the package of the test the constructor would not be callable.

chaokunyang: I see, thanks. How about adding a factory method in `A`, just move `newA` to `A`

knutwannheden: The class `A` would still not be public and thus not visible to the test.

Please note that in my case I think the problem is that my serialized objects contain objects where the class implements `Externalizable` and there also the `writeReplace()` and `readResolve()` methods. This may explain why the class is not public.

chaokunyang: I see

java/fury-core/src/main/java/io/fury/serializer/ExternalizableSerializer.java
```diff
@@ -65,12 +60,13 @@ public void xwrite(MemoryBuffer buffer, T value) {
...
 
   @Override
+  @SuppressWarnings("unchecked")
   public T read(MemoryBuffer buffer) {
     T t;
     if (constructor != null) {
       try {
-        t = constructor.newInstance();
-      } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
+        t = (T) constructor.invokeWithArguments();
```
chaokunyang, line 68:
```suggestion
        t = (T) constructor.invoke();
```

java/fury-core/src/test/java/io/fury/serializer/ExternalizableSerializerTest.java
chaokunyang: Could we add some comments on `A` and `Factory` to indicate why we need the factory and a inassessible class?

knutwannheden:
@chaokunyang I think to support the combination of `Externalizable` and `writeReplace()` / `readResolve()` I will have to hand over to you. I am not sure how that would best be done, as I see that you already have some support for the latter. Feel free to close this PR, if you think you want to fix that together with this change.

java/fury-core/src/test/java/io/fury/serializer/ExternalizableSerializerTest.java
knutwannheden: Certainly.

chaokunyang:
> @chaokunyang I think to support the combination of `Externalizable` and `writeReplace()` / `readResolve()` I will have to hand over to you. I am not sure how that would best be done, as I see that you already have some support for the latter. Feel free to close this PR, if you think you want to fix that together with this change.

`writeReplace() / readResolve()` has higher priority than `Externalizable`, currently we ignored that. This PR can be merged first, the support for combination can be put in a new PR

chaokunyang (approved):
LGTM

knutwannheden:
@chaokunyang Thanks for merging this fix! I hope that I will soon be able to successfully serialize and deserialize our object graphs. Do you want me to create a separate issue for the combination of `Externalizable` and `writeReplace()` / `readResolve()`?

knutwannheden:
Here it is in any case: https://github.com/alipay/fury/issues/1045

chaokunyang:
@knutwannheden Thanks for creating new issue, I'll fix it soon

PR #1048, opened 2023-10-31
[Java] support writeReplace/readResolve for Externalizable

### LINKED ISSUE #1045: [Java] Support combination of `Externalizable` and `writeReplace()` / `readResolve()`
Currently Fury supports both `Externalizable` and `writeReplace()` / `readResolve()`, but it looks like the combination of these two Java features is not supported and results in a `ClassCastException` when deserializing. E.g.

```
java.lang.ClassCastException: com.fasterxml.jackson.databind.node.NodeSerialization cannot be cast to com.fasterxml.jackson.databind.JsonNode
```

It looks like the serialization did use `writeReplace()` and then during the deserialization the `readResolve()` method was not called on the result before returning it.

chaokunyang:
## What do these changes do?
This PR supports Externalizable which has  a `writeReplace/readResolve` defined by spec: https://docs.oracle.com/javase/8/docs/platform/serialization/spec/output.html. The `writeReplace` can return same type object or different tyoe object, or just return itself.

## Related issue number

Closes #1045 

## Check code requirements

### DIFF
--- a/java/fury-core/src/main/java/io/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/io/fury/resolver/ClassResolver.java
@@ -840,8 +840,6 @@ public class ClassResolver {
         return TimeSerializers.ZoneIdSerializer.class;
       } else if (TimeZone.class.isAssignableFrom(cls)) {
         return TimeSerializers.TimeZoneSerializer.class;
-      } else if (Externalizable.class.isAssignableFrom(cls)) {
-        return ExternalizableSerializer.class;
       } else if (ByteBuffer.class.isAssignableFrom(cls)) {
         return BufferSerializers.ByteBufferSerializer.class;
       }
@@ -888,6 +886,9 @@ public class ClassResolver {
       }
       if (useReplaceResolveSerializer(cls)) {
         return ReplaceResolveSerializer.class;
+      }
+      if (Externalizable.class.isAssignableFrom(cls)) {
+        return ExternalizableSerializer.class;
       }
       if (requireJavaSerialization(cls)) {
         return getJavaSerializer(cls);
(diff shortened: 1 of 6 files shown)


chaokunyang:
@knutwannheden Could you help review this PR and test whether it works for your object graph?

chaokunyang:
Compared to kryo, this implementation won't forward the implementation to JDK serialization, thus have much better performance. And reference tracking are all tracked by fury instead of part in kryo, part in JDK. No reference tracking error will happen in fury.

knutwannheden:
Looks good to me and I can confirm that this allows me to correctly serialize and deserialize `Externalizable` objects using `writeReplace()` / `readResolve()`.

One thing that worries me a little bit is the static `ClassValue` caches in `JavaSerializer` (possibly also in other places). Maybe it would be possible to have these scoped to the `Fury` instance instead. I am just a bit worried about app servers with a lot of dynamic class loading. But this is of course a different issue.

knutwannheden (approved):

chaokunyang:
> Looks good to me and I can confirm that this allows me to correctly serialize and deserialize `Externalizable` objects using `writeReplace()` / `readResolve()`.
> 
> One thing that worries me a little bit is the static `ClassValue` caches in `JavaSerializer` (possibly also in other places). Maybe it would be possible to have these scoped to the `Fury` instance instead. I am just a bit worried about app servers with a lot of dynamic class loading. But this is of course a different issue.

ClassValue is not a strong reference, java gc can recaime it when necessary

chaokunyang:
> Looks good to me and I can confirm that this allows me to correctly serialize and deserialize `Externalizable` objects using `writeReplace()` / `readResolve()`.
> 
> One thing that worries me a little bit is the static `ClassValue` caches in `JavaSerializer` (possibly also in other places). Maybe it would be possible to have these scoped to the `Fury` instance instead. I am just a bit worried about app servers with a lot of dynamic class loading. But this is of course a different issue.

I'll merge this PR first. We have some tests to check GC, see `io.fury.FuryTest#testClassGC`. If class leak does happen, please open a new issue for it. Or you can submit a new PR now if you like. You can replace all usage of `ClassValue` by a lightweight cache using `SoftReference`.

knutwannheden:
> ClassValue is not a strong reference, java gc can recaim it when necessary

Good to know. Thanks for clarifying!

PR #1063, opened 2023-11-03
Empty ListExpression#genCode will throw npe

farmerworking:
## What do these changes do?

Empty ListExpression invoke genCode will throw npe. this pull request fix this bug.

### DIFF
--- a/java/fury-core/src/main/java/io/fury/codegen/Expression.java
+++ b/java/fury-core/src/main/java/io/fury/codegen/Expression.java
@@ -182,12 +182,16 @@ public interface Expression {
           hasCode = true;
         }
       }
-      ExprCode lastExprCode = last.genCode(ctx);
       String code = codeBuilder.toString();
       if (!hasCode) {
         code = null;
       }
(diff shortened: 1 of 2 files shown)


java/fury-core/src/main/java/io/fury/codegen/Expression.java
```diff
@@ -182,12 +182,16 @@ public ExprCode doGenCode(CodegenContext ctx) {
...
       String code = codeBuilder.toString();
       if (!hasCode) {
         code = null;
       }
-      return new ExprCode(code, lastExprCode.isNull(), lastExprCode.value());
+      if (code != null) {
+        ExprCode lastExprCode = last.genCode(ctx);
+        return new ExprCode(code, lastExprCode.isNull(), lastExprCode.value());
+      } else {
+        return new ExprCode(code);
```
chaokunyang, line 193: This may cause `Reference` expr value lost.

how aboutting add a if branch at NPE location instead :

```java
 if (last == null) {
        return new ExprCode(null, null, null);
      }
```

chaokunyang (approved):
LGTM, thanks for your contribution @farmerworking

PR #1075, opened 2023-11-04
[Scala] get build to work with Scala 3

pjfanning:
* Code needs small hack for spliterator - won't compile in Scala 3 otherwise - possible Scala 3 compiler bug
* removes Scala 2.12 build option beause existing code uses Scala `Factory` class that was only added in Scala 2.13 - if you want to support Scala 2.12, new Scala 2.12 specific code is needed
* updates CI to use `+test` which tests with all the `crossScalaVersions` and gets rid of setup-scala action that is deprecated in favour of setup-java - sbt build tool installs the Scala version it needs anyway so no need to pre-install Scala

## Check code requirements

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,11 +80,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: olafurpg/setup-scala@v11
+      - name: Set up JDK8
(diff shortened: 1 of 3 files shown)


chaokunyang (approved):
LGTM, thanks you @pjfanning.

chaokunyang:
`sbt +test` for all scala version tests. new scala tips for me 😄

PR #1082, opened 2023-11-09
[Java] When user disable class registration, turn off frequent and annoying log warnings

wtis-cj:
## What do these changes do?

When user decides to disable class register, we have enough log to show risk, this PR turn off frequent and annoying log warnings in high concurrency environment.

## Related issue number

none

## Check code requirements

### DIFF
--- a/java/fury-core/src/main/java/io/fury/config/Config.java
+++ b/java/fury-core/src/main/java/io/fury/config/Config.java
@@ -48,6 +48,7 @@ public class Config implements Serializable {
   private final boolean compressLong;
   private final LongEncoding longEncoding;
   private final boolean requireClassRegistration;
+  private final boolean classRegistrationWarningIgnored;
   private final boolean registerGuavaTypes;
   private final boolean shareMetaContext;
   private final boolean asyncCompilationEnabled;
@@ -66,6 +67,7 @@ public class Config implements Serializable {
     longEncoding = builder.longEncoding;
     compressLong = longEncoding != LongEncoding.LE_RAW_BYTES;
     requireClassRegistration = builder.requireClassRegistration;
+    classRegistrationWarningIgnored = builder.classRegistrationWarningIgnored;
     registerGuavaTypes = builder.registerGuavaTypes;
     codeGenEnabled = builder.codeGenEnabled;
     checkClassVersion = builder.checkClassVersion;
@@ -148,6 +150,10 @@ public class Config implements Serializable {
 
   public boolean requireClassRegistration() {
     return requireClassRegistration;
+  }
+
+  public boolean classRegistrationWarningIgnored() {
+    return classRegistrationWarningIgnored;
   }
 
   public boolean registerGuavaTypes() {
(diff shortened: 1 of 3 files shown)


chaokunyang:
@wtis-cj Thanks for your contribution, the class registration log can be used for security audit. Is is possible to add a new switch in `FuryBuilder` for disabling log instead of just skipping it?

wtis-cj:
@chaokunyang add config: classRegistrationWarningIgnored, default false.

Only takes effect when using `FuryBuilder#requireClassRegistration(false)`

java/fury-core/src/main/java/io/fury/config/FuryBuilder.java
```diff
@@ -212,6 +213,12 @@ public FuryBuilder requireClassRegistration(boolean requireClassRegistration) {
     return this;
   }
 
+  /** Only takes effect when using `FuryBuilder#requireClassRegistration(false)`. */
+  public FuryBuilder classRegistrationWarningIgnored(boolean classRegistrationWarningIgnored) {
```
chaokunyang, line 217:
```suggestion
  public FuryBuilder suppressClassRegistrationWarning(boolean suppressClassRegistrationWarning) {
```

java/fury-core/src/main/java/io/fury/config/FuryBuilder.java
```diff
@@ -68,6 +68,7 @@ public final class FuryBuilder {
   boolean checkJdkClassSerializable = true;
   Class<? extends Serializer> defaultJDKStreamSerializerType = ObjectStreamSerializer.class;
   boolean requireClassRegistration = true;
+  boolean classRegistrationWarningIgnored = false;
```
chaokunyang, line 71:
```suggestion
  boolean isClassRegistrationWarningSuppressed = false;
```

java/fury-core/src/main/java/io/fury/config/FuryBuilder.java
```diff
@@ -212,6 +213,12 @@ public FuryBuilder requireClassRegistration(boolean requireClassRegistration) {
     return this;
   }
 
+  /** Only takes effect when using `FuryBuilder#requireClassRegistration(false)`. */
```
chaokunyang, line 216:
```suggestion
  /** Supresss class registration check warnings, only takes effect when using `#requireClassRegistration(false)`. */
```

java/fury-core/src/main/java/io/fury/config/Config.java
```diff
@@ -150,6 +152,10 @@ public boolean requireClassRegistration() {
     return requireClassRegistration;
   }
 
+  public boolean classRegistrationWarningIgnored() {
```
chaokunyang, line 155:
```suggestion
  public boolean isClassRegistrationWarningIgnored() {
```

Also better to add java doc for this method here

wtis-cj:
These suggestions make no sense

chaokunyang:
@wtis-cj Thanks for your suggestion and contribution. The method naming is a history issue. We have two kinds of naming in `FuryBuilder`:
- `withXXX`: a builder style naming
- `verbXXX`: `ignoreTimeRef/requireClassRegistration`

I try to keep naming less messy, but I must admit that we should keep this from new contributor to reduce their work.

PR #1088, opened 2023-11-15
[Java] FuryPooledObjectFactory getFury refactor, remove redundant recursive call

mof-dev-3:
## What do these changes do?
Before this refactor, when pooled fury cache not present:
1. double check cache present
2. put in cache
3. recursive call get function

In order to avoid recursive call, we could return cached obj or newly added cache object
thus rename addCache to getOrAddCache


chaokunyang:
LGTM, thanks for taking time to optimize this, @mof-dev-3

The code style check has some failure, could you fix this?
image

chaokunyang:
You can use `mvn spotless:apply && mvn checkstyle:check` to check code style

chaokunyang (approved):

mof-dev-3:
> You can use `mvn spotless:apply && mvn checkstyle:check` to check code style

Done.

PR #1092, opened 2023-11-15
Fix undefined behavior due to use of uninitialized field in Buffer

PragmaTwice:
## What do these changes do?

There are two point about this change:
- In the constructor `Buffer::Buffer`, `own_data_` is used but not initialized which is an undefined behavior, so that it may contain arbitrary value rather than `false`.
- It is no necessary to check `own_data_` since it is a constructor and the buffer is always constructed without owning any `data_`.


chaokunyang (approved):
LGTM, thanks for this fix @PragmaTwice

PR #1093, opened 2023-11-15
Make sure the c++ standard is set to 17

### LINKED ISSUE #1089: [C++] About the confusing C++ standard used in fury
I notice that there is some inconsistency regarding to the C++ standard used in this library:
- in `.bazelrc`, `-std=c++17` is added in cxxflags,
- in `bazel/fury.bzl`, there is a command line option `-std=c++11`.

If the library uses some post-c++11 features, there may be confusion during the build: 
one may be able to compile according to the `.bazelrc` settings, but another may not.

May I ask if it is possible to clarify the standard?

PragmaTwice:
## What do these changes do?

As clarified in https://github.com/alipay/fury/issues/1089#issuecomment-1812515055 by @chaokunyang , the C++ standard is expected to be set to 17.

## Related issue number

It closes #1089.

### DIFF
--- a/bazel/fury.bzl
+++ b/bazel/fury.bzl
@@ -1,1 +1,1 @@
-COPTS = ["-pthread","-std=c++11", "-D_GLIBCXX_USE_CXX11_ABI=0"]
+COPTS = ["-pthread","-std=c++17", "-D_GLIBCXX_USE_CXX11_ABI=0"]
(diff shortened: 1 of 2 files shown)


chaokunyang (approved):
LGTM, thanks for your contribution, @PragmaTwice

PR #1096, opened 2023-11-15
Simplify endian utility functions and `IsOneOf`

PragmaTwice:
## What do these changes do?

- Simplify the implementation of `IsOneOf`
- Simplify the implementation of `To/FromBig/LittleEndian` via `if constexpr`

BTW I think maybe there's no difference between `ToXXXEndian` and `FromXXXEndian`.

### DIFF
--- a/src/fury/util/util.h
+++ b/src/fury/util/util.h
@@ -22,6 +22,7 @@
 #include <memory>
 #include <sstream>
 #include <string>
+#include <type_traits>
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


chaokunyang:
@PragmaTwice Thanks for this work! It looks much better than before. `FromXXXEndian` is used to convert from big/little endian format to the machine's native endian format. It's not used anymore, and can be deleted directly

chaokunyang (approved):
LGTM

PR #1122, opened 2023-11-21
[Java] DateTimeUtils minor refactor, reuse floorDiv to calculate floorMod

mof-dev-3:
## What do these changes do?

we can reuse Math.floorDiv result to calculate floorMod according to the code implement from Math.floorMod


chaokunyang (approved):
LGTM, your implementation is faster than `java.lang.Math#floorMod(long, long)`, good optimization!

PR #1134, opened 2023-11-23
[C++] Add move ctor/assign op to Status

PragmaTwice:
## What do these changes do?

- Make assignment operator return reference
- Add move ctor and move assignment operator for Status


chaokunyang (approved):
Looks great

PR #1136, opened 2023-11-23
[C++] Remove useless overload of Writer::WriteString

PragmaTwice:
## What do these changes do?

- Make Writer::WriteString accept const reference
- Remove the rvalue reference one since it's same


chaokunyang (approved):
lGTM, thanks for this optimization, @PragmaTwice

PR #1158, opened 2023-11-28
[C++] Support string type in RowEncoder

PragmaTwice:
## What do these changes do?

We support `std::string` and `std::string_view` in RowEncoder, e.g.
```c++
struct B {
  int num;
  std::string str;
};

FURY_FIELD_INFO(B, num, str);

RowWriter writer(encoder::RowEncodeTrait<B>::Schema());
writer.Reset();

B b{233, "hello"};
encoder::RowEncodeTrait<B>::Write(b, writer);
```

Besides, I move `IsOneOf` from `util` to `meta`.

And I think maybe we can rename `util.h` to something like `bit_util.h`, `time_util.h` and other `xxx_util.h`, so that it will not become a very messy source file that all developer want to put something into it since the name `util` is too general. cc @chaokunyang 

## Check code requirements

### DIFF
--- a/src/fury/encoder/row_encoder.h
+++ b/src/fury/encoder/row_encoder.h
@@ -20,6 +20,7 @@
 #include "fury/meta/type_traits.h"
 #include "fury/row/row.h"
 #include "src/fury/row/writer.h"
+#include <string_view>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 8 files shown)


chaokunyang:
> And I think maybe we can rename util.h to something like bit_util.h, time_util.h and other xxx_util.h, so that it will not become a very messy source file that all developer want to put something into it since the name util is too general.

Agreed!  those methods should be divied into different files for better maintainability. the util methods are little previously, so we put all things here. We should avoid such things in future

PR #1193, opened 2023-12-01
[C++] Fix ownership problem for children writers by visitor

PragmaTwice:
## What do these changes do?

Here we use visitor pattern to solve the ownership issue for children of writers.

## Check code requirements


chaokunyang (approved):
LGTM

PR #1200, opened 2023-12-03
[C++] Add RowEncoder wrapper to RowEncodeTrait

PragmaTwice:
## Related issue number

We added a simpler interface to use `RowEncodeTrait` named `RowEncoder`, and renamed `row_encoder.h` to `row_encode_trait.h`

## Check code requirements

### DIFF
--- a/src/fury/util/buffer.cc
+++ b/src/fury/util/buffer.cc
@@ -58,7 +58,7 @@ Buffer &Buffer::operator=(Buffer &&buffer) noexcept {
 
 Buffer::~Buffer() {
   if (own_data_) {
-    delete data_;
+    free(data_);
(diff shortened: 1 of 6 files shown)


src/fury/util/buffer.cc
```diff
@@ -58,7 +58,7 @@ Buffer &Buffer::operator=(Buffer &&buffer) noexcept {
 
 Buffer::~Buffer() {
   if (own_data_) {
-    delete data_;
+    free(data_);
```
PragmaTwice, line 61: We need to use `free` since it is allocated from `malloc`

chaokunyang: Good catch

PR #1208, opened 2023-12-05
[JavaScript] Support partial record

### LINKED ISSUE #1207: [JavaScript] Use Type.object to define protocol will cause error
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/alipay/issues) and found no similar issues.

### Version

0.0.18

### Component(s)

JavaScript

### Minimal reproduce step

For example, I define a protolcol like this:

```ts
import Fury, { Type } from '@furyjs/fury';

export { PlatformBuffer } from '@furyjs/fury/dist/lib/platformBuffer';

/**
 * @furyjs/hps use v8's fast-calls-api that can be called directly by jit, ensure that the version of Node is 20 or above.
 * Experimental feature, installation success cannot be guaranteed at this moment
 **/
// import hps from '@furyjs/hps';

const hps = undefined;

// Now we describe data structures using JSON, but in the future, we will use more ways.
const description = Type.object('ws-channel-protocol', {
  kind: Type.string(),
  clientId: Type.string(),
  id: Type.uint64(),
  path: Type.string(),
  content: Type.string(),
  code: Type.uint16(),
  reason: Type.string(),
});

const fury = new Fury({ hps });
const { serialize, deserialize } = fury.registerSerializer(description);

serialize({
  kind: '1234',
//  ^ error occurs here: Argument of type '{ kind: string; }' is not assignable to parameter of type '{ kind: string | null; clientId: string | null; id: number | null; path: string | null; content: string | null; code: number | null; reason: string | null; }'.
});

export { serialize, deserialize };
```
and this will also cause error:

image

I think the infered object's typings should have a question mark on its all properties.

for example:

```ts
const description = Type.object('ws-channel-protocol', {
  kind: Type.string(),
  clientId: Type.string(),
  id: Type.uint64(),
  path: Type.string(),
  content: Type.string(),
  code: Type.uint16(),
  reason: Type.string(),
});

// ->

{ kind?: string | null; clientId?: string | null; id?: number | null; path?: string | null; content?: string | null; code?: number | null; reason?: string | null; }
```

### What did you see instead?

{ kind: string | null; clientId: string | null; id: number | null; path: string | null; content: string | null; code: number | null; reason: string | null; }

### Anything Else?

_No response_

wangweipeng2:
## What do these changes do?
1. JavaScript is a dynamic language, so the object being serialized may lack certain properties. We should handle partial objects by adding default values for non-referencable types and making the description of object field types optional.

2. Sometimes, webpack may mock the process object, which can break environment detection. To mitigate this, we can detect the process.version.node property, as it is usually not mocked.

## Related issue number

Closes #1207

## Check code requirements

### DIFF
--- a/javascript/packages/fury/lib/description.ts
+++ b/javascript/packages/fury/lib/description.ts
@@ -59,7 +59,7 @@ type Props<T> = T extends {
     };
 }
     ? {
-        [P in keyof T2]: (ToRecordType<T2[P]> | null);
+        [P in keyof T2]?: (ToRecordType<T2[P]> | null);
     }
     : unknown;
 
--- a/javascript/packages/fury/lib/internalSerializer/bool.ts
+++ b/javascript/packages/fury/lib/internalSerializer/bool.ts
@@ -26,7 +26,7 @@ export default (fury: Fury) => {
         ...referenceResolver.deref(() => {
             return readUInt8() === 0 ? false : true
         }),
-        write: referenceResolver.withNotNullableWriter(InternalSerializerType.BOOL, (v: boolean) => {
+        write: referenceResolver.withNotNullableWriter(InternalSerializerType.BOOL, false, (v: boolean) => {
             writeUInt8(v ? 1 : 0)
         }),
         writeWithoutType: (v: boolean) => {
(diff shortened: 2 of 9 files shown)


chaokunyang (approved):
LGTM

chaokunyang:
I see referenceResolver naming, maybe we can name it to 'refResolver' for consistency with other languages

wangweipeng2:
> I see referenceResolver naming, maybe we can name it to 'refResolver' for consistency with other languages

Good idea, I'll correct it in the next update

PR #1211, opened 2023-12-06
[JavaScript] Fix register a description twice will get undefined serializer

bytemain:
## What do these changes do?

I found an issue that if we register a description twice, in the second times, the result of `genSerializer` is undefined.

## Related issue number

Closes #xxxx

## Check code requirements


wangweipeng2 (approved):
LGTM

PR #1212, opened 2023-12-06
[C++] Support iterable types in RowEncodeTrait

PragmaTwice:
## What do these changes do?

In this PR we support any iterable types for `RowEncodeTrait` and its combination with struct types, e.g.
```c++
std::vector<int>

// nested array type
std::vector<std::vector<int>>

// class type in array type
struct A { ... }
std::vector<A>

// array in class type
struct B {
  std::vector<...> field1;
  ...
}
```

## Check code requirements

### DIFF
--- a/src/fury/encoder/row_encode_trait.h
+++ b/src/fury/encoder/row_encode_trait.h
@@ -20,6 +20,7 @@
 #include "fury/meta/type_traits.h"
 #include "fury/row/writer.h"
 #include <string_view>
+#include <type_traits>
 #include <utility>
 
 namespace fury {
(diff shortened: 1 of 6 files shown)


chaokunyang (approved):
LGTM. one more thing, do we need to add an `ArrayEncoder` in array_encoder.h ?

chaokunyang:
Those meta programing is really impressive, learned a lot from your code 😄

PragmaTwice:
> do we need to add an ArrayEncoder in array_encoder.h ?

I plan to extend `RowEncoder<T>` rather than adding a new class.

PR #1216, opened 2023-12-08
[JavaScript] Implements tuple serializer

bytemain:
## What do these changes do?

Implements tuple serializer

In many language, there is a data structure called tuple, but the fury specification has not defined tuple behaviours.

so I am using Array as Tuple.

and also thanks to @wangweipeng2 for his teaching, he also has a lot contribution for this pr.

## Related issue number

Closes #xxxx

## Check code requirements

### DIFF
--- a/javascript/packages/fury/lib/codeGen.ts
+++ b/javascript/packages/fury/lib/codeGen.ts
@@ -97,7 +97,7 @@ function typeHandlerDeclaration(fury: Fury) {
 
                 const name = `tuple_${names.join('_')}`;
                 return addDeclar(name, `
-                    const ${name} = tupleSerializer(fury, ${names.join(', ')})`
+                const ${name} = tupleSerializer(fury, [${names.join(', ')}])`
                 )
             }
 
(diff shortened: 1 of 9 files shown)


chaokunyang:
Well done! this is great @bytemain

javascript/packages/fury/lib/codeGen.ts
```diff
@@ -87,6 +88,19 @@ function typeHandlerDeclaration(fury: Fury) {
...
         }
         if (description.type === InternalSerializerType.ARRAY) {
+            const tupleOptions = Cast<TupleTypeDescription>(description).options;
+            if (tupleOptions && tupleOptions.isTuple) {
+                const names = [] as string[];
+                Cast<TupleTypeDescription>(description).options.inner.forEach(v => {
+                    names.push(genDeclaration(v));
+                })
+
+                const name = `tuple_${names.join('_')}`;
```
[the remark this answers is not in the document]
bytemain: after some investigation, it seem the `addDeclar` function can avoid declare same variable twice.

and if two tuple has the same name, that means it can serialize the same data.

and also I have add an code gen snapshot to check the generated code.

javascript/packages/fury/lib/internalSerializer/tuple.ts
```diff
@@ -0,0 +1,55 @@
...
+            const result = new Array(len);
+            pushReadObject(result);
+            for (let i = 0; i < result.length; i++) {
+                const item = serializers[i];
+                result[i] = item.read();
+            }
+            return result;
+        }),
+        write: referenceResolver.withNullableOrRefWriter(InternalSerializerType.TUPLE, (v: any[]) => {
+            writeVarInt32(v.length);
```
wangweipeng2, line 40: should be serializers.length

javascript/packages/fury/lib/internalSerializer/tuple.ts
```diff
@@ -0,0 +1,55 @@
...
+    const { varInt32: readVarInt32 } = binaryReader;
+
+    return {
+        ...referenceResolver.deref(() => {
+            const len = readVarInt32();
+            const result = new Array(len);
+            pushReadObject(result);
+            for (let i = 0; i < result.length; i++) {
+                const item = serializers[i];
+                result[i] = item.read();
```
wangweipeng2, line 35: item may be undefined, if values is larger then serializers, maybe we can use any.read if values is too large

wangweipeng2 (commented):
great work😊

javascript/packages/fury/lib/internalSerializer/tuple.ts
bytemain: if we save the serialzier length, this problem is gone

wangweipeng2 (approved):

PR #1231, opened 2023-12-15
[JavaScript] Add more methods for BinaryReader

bytemain:
## What do these changes do?

Add more methods for BinaryReader

In real world scene, we need do many opreation on binary, it is better to have an buffer ref method to get a buffer reference instead of get a copy of the original buffer.

## Check code requirements

### DIFF
--- a/javascript/packages/fury/lib/reader.ts
+++ b/javascript/packages/fury/lib/reader.ts
@@ -99,6 +99,13 @@ export const BinaryReader = (config: Config) => {
     return result;
   }
 
+  function stringUtf8OfVarint32() {
+    const len = varInt32();
+    const result = buffer.utf8Slice(cursor, cursor + len);
+    cursor += len;
+    return result;
+  }
+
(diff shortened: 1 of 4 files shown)


javascript/packages/fury/lib/reader.ts
```diff
@@ -99,6 +99,13 @@ export const BinaryReader = (config: Config) => {
     return result;
   }
 
+  function stringUtf8OfVarint32() {
```
wangweipeng2, line 102: Perhaps we should rename it to stringOfVarInt32, as it corresponds to the stringOfVarInt32 method in the writer. This change would also correct the naming mistake I made a long time ago.

wangweipeng2: Accordingly, readVarInt32 in `packages/fury/lib/internalSerializer/string.ts` should also be moved to this function

wangweipeng2 (approved):
great work

PR #1233, opened 2023-12-15
[C++] Improve FormatTimePoint by removing sstream

PragmaTwice:
## What do these changes do?

In FormatTimePoint we don't need to use std::stringstream and even std::string.

## Check code requirements


chaokunyang (approved):
LGTM

PR #1234, opened 2023-12-15
[C++] Improve Status by using unique_ptr

PragmaTwice:
## What do these changes do?

We can use std::unique_ptr to make Status more easy to implement.

## Check code requirements

### DIFF
--- a/src/fury/util/status.cc
+++ b/src/fury/util/status.cc
@@ -34,22 +34,6 @@ namespace fury {
 #define STATUS_CODE_IO_ERROR "IOError"
 #define STATUS_CODE_UNKNOWN_ERROR "Unknown error"
 
-Status::Status(StatusCode code, const std::string &msg) {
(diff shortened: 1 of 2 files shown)


chaokunyang:
If we use this Status for every serialization operation, will unique_ptr introduce extra cost? Or we should use Exception for that error check such as bound check in deserialization? `std::unique_ptr` is much simpler, but I'm not sure whether the extra indirection will introduce some cost?

PragmaTwice:
> If we use this Status for every serialization operation, will unique_ptr introduce extra cost? Or we should use Exception for that error check such as bound check in deserialization? `std::unique_ptr` is much simpler, but I'm not sure whether the extra indirection will introduce some cost?

`std::unqiue_ptr` does not bring any runtime cost, especially when the default deleter is used.

chaokunyang (approved):
LGTM

Issue #1229, opened 2023-12-10
[Java] Use collection serialization v2 for array


chaokunyang:
## Is your feature request related to a problem? Please describe.
In #923 , we implemented the collection protocol in #927 , which can be 1X faster at most. This can improve  space cost by 1X at most. Such optimization should be applied to array too.

## Describe the solution you'd like
Refactor `io.fury.serializer.ArraySerializers.ObjectArraySerializer` to forward serialization to `FuryArrayAsListSerializer`.
- Wrap array into ArrayAsList`, note this wrapped object can't be reused if nested serialization invoke to this place too.
- Push array component generics to `FuryArrayAsListSerializer`
- If array component doesn't have nested generics, reimplement the serialization in `ObjectArraySerializer` for better peroformance.

## Additional context
#1228

david1437:
Hello! Is this issue still available?

I would like to contribute to this issue as it's my first issue I may need some help getting into it.

Thanks

david1437:
Great so I have started looking at the code for ObjectArraySerializer and have a couple of questions:

Wrap array into ArrayAsList, note this wrapped object can't be reused if nested serialization invoke to this place too.
 - Seems to be the same logic as used in StringArraySerializer methods is this accurate?

Push array component generics to FuryArrayAsListSerializer
 - Does this mean using setElementSerializer() method with genric serializer? ie. `this.componentTypeSerializer = fury.getClassResolver().getSerializer(componentType);`

If array component doesn't have nested generics, reimplement the serialization in ObjectArraySerializer for better peroformance.
 - I am not sure I understand what nested generics means in this case? Is it something along the line of given a T[] where T is a <k, v> Map<k, v> for example? This would mean in case of T[] where T is just some class with no generics it should continue to use the previous serialization approach?

chaokunyang:
Yes, you are right. wrapped object can't be reused if nested serialization invoke to this place too.

setElementSerializer can not be used. Since it can only push one layer serializer. Push generics is enough. It's just like normal collection serializer.


For Nested generics, It's the array component type info. For example,  we should push `List<List<Foo>>` For `Foo[][]`.

For more simple array type without nested generics, we could implement a faster version in a special serializer

david1437:
Thanks!

Based on your comment I am understanding it as for nested generics we need to take into account dimensions of the array which is already calculated in the array serialization code.
```
int dimension = 0;
while (t != null && t.isArray()) {
  dimension++;
   t = t.getComponentType();
[...]

Issue #925, opened 2023-10-01
[Java][Protocol] Chunk by chunk predictive map serialization protocol


chaokunyang:
## Is your feature request related to a problem? Please describe.
Optimize Collection/Map serialization by potential homogenization in elements:
- Collection: elements are all not-empty mostly, and all elements are same type
- Map: all keys are not empty mostly, all keys are same type; all values are not empty mostly, all values are same type.

By using those information, the serialization performance can be enhanced and the size of serialized binary can be smaller.

For collection, we can compute header before serializing elements, since iterating collection is cheap. But for map iteration, it's expensive, it takes same cost as serialization for `Map<Integer, Integer>`.

We need to finish kv writing and header writing in one-round iteration.

## Describe the solution you'd like
Users can use `MapFieldInfo` annotation to provide header in advance. Otherwise Fury will use first key-value pair to
predict header optimistically, and update the chunk header if predict failed at some pair.

Fury will serialize map chunk by chunk, every chunk
has 127 pairs at most.

```
+----------------+----------------+~~~~~~~~~~~~~~~~~+
| chunk size: N  |    KV header   |   N*2 objects   |
+----------------+----------------+~~~~~~~~~~~~~~~~~+
```

KV header:

- If track key ref, use first bit `0b1` of header to flag it.
- If key has null, use second bit `0b10` of header to flag it. If ref tracking is enabled for this
  key type, this flag is invalid.
- If map key type is not declared type, use 3rd bit `0b100` of header to flag it.
- If map key type different, use 4rd bit `0b1000` of header to flag it.
- If track value ref, use 5rd bit `0b10000` of header to flag it.
- If value has null, use 6rd bit `0b100000` of header to flag it. If ref tracking is enabled for this
  value type, this flag is invalid.
- If map value type is not declared type, use 7rd bit `0b1000000` of header to flag it.
- If map value type different, use 8rd bit `0b10000000` of header to flag it.

If streaming write is enabled, which means Fury can't update written `chunk size`. In such cases, map key-value data
format will be:

```
+----------------+~~~~~~~~~~~~~~~~~+
|    KV header   |   N*2 objects   |
+----------------+~~~~~~~~~~~~~~~~~+
```

`KV header` will be header marked by `MapFieldInfo` in java. For languages such as golang, this can be computed in
advance for non-interface type mostly.

## Additional context
#923

chaokunyang:
map iteration and serialization benchmark:
```java
  @Benchmark
[...]

PR #1240, opened 2023-12-20
docs(java): standardizing fury java spec

### LINKED ISSUE #1239: [Java][Doc] Java object graph specification
Java object graph spec documentation is missing a specification, many users and developers want to see fury spec for better understanding fury serialization machinery. 

We should have a formal spec document for this, and it can be used by future contributors for better improvement of the prototol and hte implementation.

- Add Fury Java object graph spec documentation
- Ignore the existing implementation if the format can be more efficient
- #925 
- #1243 

## Additional context
#1017

chaokunyang:
## What do these changes do?
This PR refine fury java serialization format spec. The  cross-language object graph serialization spec is similar and will be added in a later PR, but it needs more discuss.

This PR added some new spec which hasn't been implemented in current java implementation:
- chunk-by-chunk predictive map serialization: #925
- layed class meta
- new class meta encoding
- #1229 
- object serialization with schema evolution support by auto meta share.

Some parts has been omitted in this spec:
- object serialization with schema evolution support by write field in a KV like pattern: this will be replaced by schema evolution  mode described in this spec in the future.

Currently fury doesn't provide binary compatibility, the spec may be revised in the future.

## Related issue number

Closes #1239 

#1238

## Check code requirements


docs/.DS_Store
PragmaTwice, line 1: Remove these files?

chaokunyang: removed

chaokunyang:
@wangweipeng2 @PragmaTwice Could you help review this PR?

wangweipeng2 (approved):

docs/protocols/java_object_graph_spec.md
```diff
@@ -69,19 +84,49 @@ If schema evolution mode is enabled globally or enabled for current class, class
 - If meta share mode is enabled, class will be written as a unsigned int.
```
wangweipeng2, line 84: The sentences feel a little confusing. Perhaps we can change it to:
```
If meta share mode is not enabled, class meta will be written as scheme consistent mode. Additionally， field meta such as field type
  and name will be written when the object value is being serialized using a key-value like layout.
```

docs/protocols/java_object_graph_spec.md
```diff
@@ -53,8 +64,12 @@ If schema consistent mode is enabled globally or enabled for current class, clas
 
 - If class is registered, it will be written as a little-endian unsigned int: `class_id << 1` using fury unsigned int
```
wangweipeng2, line 65: The concept of 'class_id' was not mentioned or defined elsewhere

chaokunyang: Added a description about class id:

Fury support register class by an optional id, the registration can be used to security check, and identify the class.
If the class is registered, the class will have a users provided or a auto-growing id unsigned int `class_id`.

docs/protocols/java_object_graph_spec.md
chaokunyang: updated

wangweipeng2 (approved):

docs/protocols/java_object_graph_spec.md
```diff
@@ -2,15 +2,26 @@
...
+much more easy to use, but
+also introduce more complexities compared to static serialization frameworks. So the format will be more complex.
+
+Here is the overall format:
 
 ```
 | fury header | object ref meta | object class meta | object value data |
 ```
 
+The data are serialized using little endian order overall. If bytes swap is costly, the byte order will be encoded as a
```
bytemain, line 22: I think the statement should like this, should be more clear:

```diff
- the byte order will be encoded as a flag in data.
+ we can define the endian flag in Fury header
```

chaokunyang: How about this:
```
The data are serialized using little endian byte order overall. If bytes swap is costly for some object, 
Fury will write the byte order for that object into the data instead of converting it to little endian.
```

bytemain (changes requested):

docs/protocols/java_object_graph_spec.md
```diff
@@ -64,24 +83,62 @@ If schema consistent mode is enabled globally or enabled for current class, clas
...
-- If meta share mode is enabled, class will be written as a unsigned int.
+- If meta share mode is enabled, class meta will be written as a meta-share encoded binary if class hasn't been written
+  before, otherwise unsigned it will be written as an unsigned varint which references to previous written class meta.
 
 ## Meta share
+
 > This mode will forbid streaming writing since it needs to look back for update the offset after the whole object graph
 > writing and mete collecting is finished.
-> TODO: We have plan to streamline meta writing but not started yet.
+> We have plan to streamline meta writing but haven't started yet.
```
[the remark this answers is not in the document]
chaokunyang: updated

chaokunyang:
Hi @wangweipeng2 @PragmaTwice @bytemain, do you have other comments about this PR? I hope all concerns has been addressed.

bytemain (approved):

docs/protocols/java_object_graph_spec.md
```diff
@@ -64,24 +83,62 @@ If schema consistent mode is enabled globally or enabled for current class, clas
...
+
+```
+| meta header: hash + num classes | current class meta | parent class meta | ... |
+```
+
+#### Meta header
+
+Meta header is a 64 bits number value encoded in little endian order.
+
+- Lowest 4 digits `0b0000~0b1110` are used to record num classes. `0b1111` is preserved to indicate that Fury need to
```
PragmaTwice, line 114: max number of classes is 14?

docs/protocols/java_object_graph_spec.md
```diff
@@ -220,11 +293,11 @@ In most cases, all collection elements are same type and not null, elements head
...
 which will be encoded by elements header, each use one bit:
 
-- Whether track elements ref, use first bit `0b1` of header to flag it.
-- Whether collection has null, use second bit `0b10` of header to flag it. If ref tracking is enabled for this
+- If track elements ref, use first bit `0b1` of header to flag it.
+- If collection has null, use second bit `0b10` of header to flag it. If ref tracking is enabled for this
   element type, this flag is invalid.
-- Whether collection elements type is not declare type, use 3rd bit `0b100` of header to flag it.
-- Whether collection elements type different, use 4rd bit `0b1000` of header to flag it.
+- If collection elements type is not declare type, use 3rd bit `0b100` of header to flag it.
```
PragmaTwice, line 299: what's "declare type" here?

docs/protocols/java_object_graph_spec.md
```diff
@@ -220,11 +293,11 @@ In most cases, all collection elements are same type and not null, elements head
...
+- If collection elements type different, use 4rd bit `0b1000` of header to flag it.
```
PragmaTwice, line 300:
```suggestion
- If collection element types are different, use 4rd bit `0b1000` of header to flag it.
```

docs/protocols/java_object_graph_spec.md
chaokunyang: Java support polymorphism, if we declare a field with type `List<SomeClass>`, the element type may be `ChildClass`

PragmaTwice: I see. Maybe you can change it to declared type or compile-time type.

chaokunyang: Updated to `declared type`

docs/protocols/java_object_graph_spec.md
chaokunyang: If number of classes  exceeds 14, `0b1111` will be used and another byte will be write next to indicate num classes

PragmaTwice (approved):

docs/protocols/java_object_graph_spec.md
PragmaTwice: Ahh I see.

docs/protocols/java_object_graph_spec.md
```diff
@@ -45,16 +56,24 @@ field of a class, only `NULL FLAG` and ` NOT_NULL VALUE FLAG` will be used.
```
theweipeng, line 56: The REF FLAG description should indicate whether the 'unsigned ref id' is an unsigned variable integer or an fixed-length integer.

chaokunyang: updated

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,457 @@
...
++----------------+----------------+~~~~~~~~~~~~~~~~~+
+```
+
+KV header:
+
+- If track key ref, use first bit `0b1` of header to flag it.
+- If key has null, use second bit `0b10` of header to flag it. If ref tracking is enabled for this
+  key type, this flag is invalid.
+- If map key type is not declared type, use 3rd bit `0b100` of header to flag it.
+- If map key type different, use 4rd bit `0b1000` of header to flag it.
```
bytemain, line 366: how to understand this sentence? does it means the key is different from all other key?

chaokunyang: Every key has a type, for example, if we have a map of type `Map<Foo, String> map`, and we have a subclass of type `class Bar extends Bar`, and map has value `Map.of(new Foo(), "v1", new Bar(), "v2")`. In such cases, map key will have type `Foo` and `Bar`, Fury will need to write key type for every key instead of write it only once.

bytemain: thanks, now I know that. so actually, the `map key type` should only compare to it's previous sibling key type.

and in some scene, we can use this flag to determine that should we create a new serializer or not.

chaokunyang: Yep, and if key type of map is different from previous key, this chunk will be finished, and next chunk may choose to assump every key type of map is different if same key types are not too much

chaokunyang:
Hi @theweipeng @PragmaTwice @bytemain @pjfanning , the java serialization specification has be finished finally, coule you please help review it again? I hope all concerns has been addressed. The  java serialization specification  in this PR is not final, we can still improve it in future. But I hope it sets up a good foundation for further discussion.

theweipeng:
> Hi @theweipeng @PragmaTwice @bytemain @pjfanning , the java serialization specification has be finished finally, coule you please help review it again? I hope all concerns has been addressed. The java serialization specification in this PR is not final, we can still improve it in future. But I hope it sets up a good foundation for further discussion.

Everything looks great. All my previous concerns have been addressed. I have no further issues

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,541 @@
...
+Fury header consists starts one byte:
+
+```
+|     4 bits    | 1 bit | 1 bit | 1 bit  | 1 bit |          optional 4 bytes          |
++---------------+-------+-------+--------+-------+------------------------------------+
+| reserved bits |  oob  | xlang | endian | null  | unsigned int for meta start offset |
+```
+
+- null flag: 1 when object is null, 0 otherwise. If an object is null, other bits won't be set.
+- endian flag: 1 when the system uses little endian, 0 otherwise.
```
PragmaTwice, line 36:
```suggestion
- endian flag: 1 when data is encoded by little endian, 0 otherwise.
```

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,541 @@
...
+
+Reference tracking handles whether the object is null, and whether to track reference for the object by writing
+corresponding flags and maintaining internal state.
+
+Reference flags:
+
+| Flag                | Byte Value | Description                                                                                                                              |
+|---------------------|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| NULL FLAG           | `-3`       | This flag indicates the object is a not-null value. We don't use another byte to indicate REF, so that we can save one byte.             |
+| REF FLAG            | `-2`       | This flag indicates the object is written before, and fury will write a ref id with unsigned varint format instead of serialize it again |
```
PragmaTwice, line 52:
```suggestion
| REF FLAG            | `-2`       | This flag indicates the object is already serialized previously, and fury will write a ref id with unsigned varint format instead of serialize it again |
```

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,541 @@
...
+## Reference Meta
+
+Reference tracking handles whether the object is null, and whether to track reference for the object by writing
+corresponding flags and maintaining internal state.
+
+Reference flags:
+
+| Flag                | Byte Value | Description                                                                                                                              |
+|---------------------|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| NULL FLAG           | `-3`       | This flag indicates the object is a not-null value. We don't use another byte to indicate REF, so that we can save one byte.             |
```
PragmaTwice, line 51: is a null value?

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,541 @@
...
+| REF FLAG            | `-2`       | This flag indicates the object is already serialized previously, and fury will write a ref id with unsigned varint format instead of serialize it again |
+| NOT_NULL VALUE FLAG | `-1`       | This flag indicates the object is a non-null value and fury doesn't track ref for this type of object.                                   |
+| REF VALUE FLAG      | `0`        | This flag indicates the object is a referencable and first write.                                                                        |
```
PragmaTwice, line 54:
```suggestion
| REF VALUE FLAG      | `0`        | This flag indicates the object is referencable and the first time to serialize.                                                                        |
```

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,541 @@
...
+
+When reference tracking is disabled globally or only for some type, or for some type under some context such as some
+field of a class, only `NULL FLAG` and ` NOT_NULL VALUE FLAG` will be used.
+
+## Class Meta
+
+Fury support register class by an optional id, the registration can be used to security check, and identify the class.
+If the class is registered, the class will have a user-provided or an auto-growing unsigned int `class_id`.
```
PragmaTwice, line 62:
```suggestion
Fury supports to register class by an optional id, the registration can be used for security check and class identification.
If a class is registered, it will have a user-provided or an auto-growing unsigned int i.e. `class_id`.
```

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,541 @@
...
+- If meta share mode is enabled, class meta will be written as a meta-share encoded binary if class hasn't been written
+  before, otherwise an unsigned varint id which references to previous written class meta will be written.
+
+## Meta share
+
+> This mode will forbid streaming writing since it needs to look back for update the start offset after the whole object
+> graph
+> writing and meta collecting is finished. Only in this way we can ensure deserialization failure doesn't lost shared
+> meta.
+> Meta streamline will be supported in the future for enclosed meta sharing without cross multiple serializations.
```
PragmaTwice, line 101: what's "cross multiple serializations"?

docs/protocols/java_object_graph_spec.md
chaokunyang: This flag is the system endian flag instead of the data. Most data are encoded in little endian. Some arrays such as `long[]` array may use native endian under the hood, we didn't change it to little endian. So if the peer has same endian, we can save the endian conversion cost

docs/protocols/java_object_graph_spec.md
chaokunyang: Good catch, updated

docs/protocols/java_object_graph_spec.md
chaokunyang: How about changing to:
```
Meta streamline will be supported in the future for enclosed meta sharing without cross multiple serializations of different objects.
```

chaokunyang:
cc @LiangliangSui

docs/protocols/java_object_graph_spec.md
PragmaTwice: Good to me 👍

docs/protocols/java_object_graph_spec.md
PragmaTwice: Ahh I see. Thank you.

PragmaTwice: > Ahh I see. Thank you.

Previously I think that actually there are other endian types rather than big and little, so we'd better describe explicitly.

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,543 @@
...
+Object will be serialized by writing its fields data in fury order.
+
+Depending on schema compatibility, objects will have different formats.
+
+#### Field order
+
+Field will be ordered as following, every group of fields will have its own order:
+
+- primitive fields: larger size type first, smaller later, variable size type last.
+- boxed primitive fields: same sort as primitive fields
```
PragmaTwice, line 483:
```suggestion
- boxed primitive fields: same order as primitive fields
```

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,543 @@
...
+- final fields: same type together, then sorted by field name lexicographically.
+- collection fields: same sort as final fields
+- map fields: same sort as final fields
+- other fields: same sort as final fields
```
PragmaTwice, line 487:
```suggestion
- collection fields: same order as final fields
- map fields: same order as final fields
- other fields: same order as final fields
```

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,543 @@
...
+|    KV header   |   N*2 objects   |
+```
+
+`KV header` will be a header marked by `MapFieldInfo` in java. For languages such as golang, this can be computed in
+advance for non-interface types in most times.
+
+### Enum
+
+Enums are serialized as an unsigned var int. If the order of enum values change, the deserialized enum value may not be
+the value users expect. In such cases, users must register enum serializer by make it write enum value as a enumerated
```
PragmaTwice, line 468:
```suggestion
the value users expect. In such cases, users must register enum serializer by make it write enum value as an enumerated
```

docs/protocols/java_object_graph_spec.md
```diff
@@ -0,0 +1,543 @@
...
+
+#### Char
+
+- size: 2 byte
+- byte order: little endian order
+
+#### Unsigned int
+
+- size: 1~5 byte
+- Format: The first bit in every byte indicates whether to have the next byte. If first bit is set
```
PragmaTwice, line 298:
```suggestion
- Format: The most significant bit (MSB) in every byte indicates whether to have the next byte. If first bit is set
```

PragmaTwice: `first` is unclear.

PR #1241, opened 2023-12-21
[JavaScript] Improve tag write performance

bytemain:
## What do these changes do?

Previously, everytime we write tag, we create new BufferWriter, we alloc new Buffer pool, this is no need.

## Check code requirements

### DIFF
--- a/javascript/packages/fury/lib/classResolver.ts
+++ b/javascript/packages/fury/lib/classResolver.ts
@@ -24,9 +24,8 @@ import boolSerializer from "./internalSerializer/bool";
 import { uInt16Serializer, int16Serializer, int32Serializer, uInt32Serializer, uInt64Serializer, floatSerializer, doubleSerializer, uInt8Serializer, int64Serializer, int8Serializer } from "./internalSerializer/number";
 import { InternalSerializerType, Serializer, Fury, BinaryReader, BinaryWriter } from "./type";
 import anySerializer from "./internalSerializer/any";
-import { PlatformBuffer, fromUint8Array } from "./platformBuffer";
+import { PlatformBuffer, fromString } from "./platformBuffer";
 import { x64hash128 } from "./murmurHash3";
-import { utf8Encoder } from "./util";
 
 const USESTRINGVALUE = 0;
 const USESTRINGID = 1;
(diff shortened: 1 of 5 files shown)


javascript/packages/fury/lib/classResolver.ts
```diff
@@ -36,7 +36,8 @@ export default class SerializerResolver {
   };
 
   private readStringPool: string[] = [];
-  private writeStringIndex: string[] = [];
+  private writeStringCount = 0;
+  private writeStringIndex: Record<string, number> = {};
```
wangweipeng2, line 40: I have a idea, Perhaps, we can generate tag accessors code at runtime, 
```
```


I have an idea. Perhaps we can generate tag accessor code at runtime. Here's how we might do it:


1. Add a counter to classResolver:
```
writeStringCount = 0;
```

2. Generate a closure when registering a tag:

```

function createTagWriter() {
    retindex = -1;

    return function () {
        // store the index
        if (index > -1) {
            writeTagIndex(); // USESTRINGID

        } else {
            writeTagContent(); // USESTRINGVALUE
            index = writeStringCount++;
        }
    } 
}

```

3. When writing the tag, we can directly call the result of the createTagWriter function, which can help us avoid the hashmap lookup.

javascript/packages/fury/lib/classResolver.ts
```diff
@@ -74,7 +75,7 @@ export default class SerializerResolver {
 
   reset() {
     this.readStringPool = [];
-    this.writeStringIndex = [];
+    this.writeStringIndex = {};
```
wangweipeng2, line 78: Should clear the writeStringCount

javascript/packages/fury/lib/classResolver.ts
```diff
@@ -94,30 +94,30 @@ export default class SerializerResolver {
...
   }
 
-  tagToBuffer(tag: string) {
-    const tagBuffer = fromUint8Array(new TextEncoder().encode(tag));
-    let tagHash = x64hash128(tagBuffer, 47).getBigUint64(0);
-    if (tagHash === BigInt(0)) {
-      tagHash = BigInt(1);
-    }
+  createTagWriter(tag: string) {
+    let tagIndex = -1;
```
wangweipeng2, line 98: Each tag call createTagWriter one time, so the tagIndex should be reset before start serialize. Perhaps we can store 
 the tagIndex on classResolver, and reset it to zero when start serialize.
The code maybe like this:
```javascript
 this.writeStringIndex = [];
 reset() {
    this.writeStringIndex.fill(-1);
 }
 createTagWriter(tag: string) {
    this.writeStringIndex.push(-1);
    let idx = this.writeStringIndex.length - 1;
    return () =>  {
        const tagIndex = this.writeStringIndex[idx];
       // when update
       this.writeStringIndex[idx] = this.writeStringCount++;
    }
}
```

bytemain: done

theweipeng (approved):
Great work

Issue #1113, opened 2023-11-20
[Java] remove guava dependency


chaokunyang:
## Is your feature request related to a problem? Please describe.
Fury use guava for generic reflection, cache. But guava is used so widely, it's very likely to version conflict with users code.

## Describe the solution you'd like
Remove guava dependency, implement all releated methods by fury self.

## Additional context
Shade guava is an option too, but introduce complexities  for fury GuavaSerializers and guava CVE

nandakumar131:
@chaokunyang 
If you haven't started working on the guava removal part 2, I can take it up and post a PR.

chaokunyang:
Hi @nandakumar131, thanks for the willingness to contribute to Fury. I haven't start part2 yet, you can take up this part work.

FYI, Currently Fury use guava TypeToken for generics inspection and use guava cache for some static data cache. These are the biggest dependency on guava

nandakumar131:
Thanks @chaokunyang!
If the change becomes huge, I will split it into two PRs to make it easy to review.

nandakumar131:
@chaokunyang  do we still want to support `GuavaCollectionSerializers`?

chaokunyang:
> @chaokunyang do we still want to support `GuavaCollectionSerializers`?

I prefer to still support it, but may use a way without source api dependency. For example, we use methodhandle to get all method for invocation

doom369:
Hello. It seems like all dependencies are removed, so you can now remove the guava from the required dependencies during the build? Currently, the fury jar is 2.3 mb and guava brings an additional 3 mb.

Munoon:
This also helps in reducing JAR file size. Let me know if I can help you here.

chaokunyang:
> This also helps in reducing JAR file size. Let me know if I can help you here.

That would be really great. Actually I am on the work of first release of Fury under asf and don't have enough time for this feature

Munoon:
> @Munoon That would be really great. Actually I am on the work of first release of Fury under asf and don't have enough time for this feature

Can you please specify which exactly help you need? To complete @nandakumar131 PR or to do some further steps?

chaokunyang:
> > @Munoon That would be really great. Actually I am on the work of first release of Fury under asf and don't have enough time for this feature
> 
> Can you please specify which exactly help you need? To complete @nandakumar131 PR or to do some further steps?

Here are the things which may need to be finished:
[...]

PR #1244, opened 2023-12-23
[Java] Remove guava part 2.

nandakumar131:
## What do these changes do?

Remove guava part 2:

Removed guava `Preconditions` usage
Removed guava `Objects` usage
Removed guava `MoreObjects` usage
Removed guava `ImmutableList`
Removed guava `ImmutableSet` & `ImmutableSortedSet`

## Related issue number
#1113 

## Check code requirements


java/fury-core/src/test/java/io/fury/serializer/collection/CollectionSerializersTest.java
```diff
@@ -477,7 +476,7 @@ public void testJavaSerialization() {
 
     assertSame(
         fury.getClassResolver().getSerializer(setClass).getClass(),
-        GuavaCollectionSerializers.ImmutableSortedSetSerializer.class);
+        UnmodifiableSerializers.UnmodifiableCollectionSerializer.class);
```
chaokunyang, line 479: Hi @nandakumar131, guava usage in tests is OK. actually we need the input object to be a guava list to test GuavaSerializers. In the long run, we may mark guava as an maven optional dependency and only mark itas an compile dependency for test

java/fury-core/src/main/java/io/fury/collection/ImmutableSet.java
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+
+package io.fury.collection;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.TreeSet;
+
+public class ImmutableSet {
```
chaokunyang, line 23: Could we use java.util.collection at the invoke point instead? Add a new public class for guava only may make new users and developers confused.

java/fury-core/src/test/java/io/fury/serializer/collection/CollectionSerializersTest.java
nandakumar131: Sure, will revert this change!

java/fury-core/src/main/java/io/fury/collection/ImmutableSet.java
nandakumar131: You don't want to introduce `ImmutableSet` & `ImmutableList`? Instead use `java.util.collection` directly?

chaokunyang: Yep, ImmutableSet and ImmutableList is used widely in test code, but barely used in source code. We can use jdk collection instead. In the end, there will no guava dependency in fury-core module.

java/fury-core/src/test/java/io/fury/serializer/collection/CollectionSerializersTest.java
```diff
@@ -72,7 +71,7 @@ public void testBasicList(boolean referenceTrackingConfig) {
             .withRefTracking(referenceTrackingConfig)
             .requireClassRegistration(false)
             .build();
-    List<String> data = new ArrayList<>(ImmutableList.of("a", "b", "c"));
+    List<String> data = new ArrayList<>(Collections.unmodifiableList(Arrays.asList("a", "b", "c")));
```
chaokunyang, line 74: We may need to use guava list instead, otherwise we can't cover guava collection serialization.

We can ignore all guava usage in test module.

nandakumar131: Reverted the changes.

java/fury-format/src/main/java/io/fury/format/row/binary/BinaryMap.java
```diff
@@ -134,10 +135,12 @@ public void writeTo(ByteBuffer buffer) {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this)
-        .add("keys", keys)
-        .add("values", values)
-        .add("sizeInBytes", sizeInBytes)
-        .toString();
+    final Map<String, Object> elements = new HashMap<>();
```
chaokunyang, line 138: How about using IDE-generated `toString` method
```java
  @Override
  public String toString() {
    return "BinaryMap{" +
      "keys=" + keys +
      ", values=" + values +
      ", sizeInBytes=" + sizeInBytes +
      '}';
  }
```

chaokunyang (approved):
LGTM overall,  I left a minor comment. Thank you @nandakumar131

java/fury-format/src/main/java/io/fury/format/row/binary/BinaryMap.java
nandakumar131: Fixed!

chaokunyang (approved):

chaokunyang:
@nandakumar131 Code style check failed, could you fix it?

You can use mvn spotless:apply to format code

nandakumar131:
@chaokunyang  fixed code style issue!

chaokunyang (approved):

PR #1252, opened 2023-12-26
chore: clear alipay refs

tisonkun:


PragmaTwice (approved):

SECURITY.md
tisonkun, line 1: Just use ASF's SECURITY file. It will be auto conveyed from apache/.github.

PR #1254, opened 2023-12-26
add NOTICE and DISCLAIMER; update SECURITY policy

pjfanning:
* Incubating projects need a DISCLAIMER
* add basic NOTICE (it's nearly 2024, so might as well add that year already)
* Code of Conduct defaults in Apache projects to https://github.com/apache/incubator-pekko?tab=coc-ov-file#readme (ie omit the .md file and GitHub will show this instead)
* ASF have a standarised security policy


pjfanning:
I see this clashes with #1253

PR #1263, opened 2023-12-26
[Java] Format integration tests code

chaokunyang:
-  Format integration tests code
- Upgrade spotless plugin to support JDK 17&21
- Format java code using upgraded spotless plugin


theweipeng (approved):

java/pom.xml
```diff
@@ -246,7 +246,8 @@
         <configuration>
           <java>
             <googleJavaFormat>
-              <version>1.7</version>
+              <!-- 1.19.1 support JDK21 -->
+              <version>1.19.1</version>
               <style>GOOGLE</style>
```
tisonkun, line 251: I'll suggest we use the plantir style for longer line length and better breaking lines strategy.

tisonkun: You can check this build https://github.com/apache/curator/blob/111f3a6bdcd562db0493e52243e5cec933cc36ef/pom.xml#L933-L954.

chaokunyang: Thanks, I will do this in another PR

chaokunyang: plantir style seems doesn't support 2 spaces: https://github.com/palantir/palantir-java-format/issues/951. I suggest continuing to use google code style

nandakumar131:
Spotless plugin version upgrade breaks `spotless:apply` in Java 8. (#1278)

PR #1264, opened 2023-12-27
ci: Release Fury Java Snapshot 

tisonkun:


chaokunyang (approved):
LGTM

java/pom.xml
```diff
@@ -261,35 +254,4 @@
       </plugin>
     </plugins>
   </build>
-
-  <profiles>
-    <profile>
-      <id>release</id>
```
chaokunyang, line 267: Just found that remove this part will skip jar release in sub modules.

Fury use properties to skip module deploy, deploy plugin won't be enabled without this profile

tisonkun: No. They are released - https://repository.apache.org/content/groups/snapshots/org/apache/fury/

We now inherit with Apache Maven Parent POM which contains an` apache-release` profile to help.

chaokunyang: I see, thanks

PR #1280, opened 2023-12-29
[Java] Support registration in thread safe fury

### LINKED ISSUE #1279: [Java] support class registration in ThreadSafeFury
`Fury` provide a `register` API to register class and serializer, but `ThreadSafeFury` doesn't provide such API.

Although we provide such doc for registration:
```java
ThreadSafeFury fury =
    new ThreadLocalFury(
        classLoader -> {
          Fury f = Fury.builder().requireClassRegistration(true).build();
          // register and generate serializer code.
          f.register(Foo.class, true);
          return f;
        });
System.out.println("Init fury at build time");
```
Users still missed this doc often and ask for help how to register class using `ThreadSafeFury`. Also it's confusing that `Fury` and `ThreadSafeFury` doesn't provide consistent API.

Provide `register` API for `ThreadSafeFury`

## Additional context
#704

chaokunyang:
Closes #1279

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/AbstractThreadSafeFury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/AbstractThreadSafeFury.java
@@ -1,0 +1,33 @@
+package org.apache.fury;
+
+import java.util.function.Consumer;
+import org.apache.fury.serializer.Serializer;
+
+public abstract class AbstractThreadSafeFury implements ThreadSafeFury {
+  @Override
+  public void register(Class<?> clz) {
+    processCallback(fury -> fury.register(clz));
+  }
+
+  @Override
+  public void register(Class<?> cls, boolean createSerializer) {
+    processCallback(fury -> fury.register(cls, createSerializer));
+  }
+
+  @Override
+  public void register(Class<?> cls, Short id) {
+    processCallback(fury -> fury.register(cls, id));
+  }
+
+  @Override
+  public void register(Class<?> cls, Short id, boolean createSerializer) {
+    processCallback(fury -> fury.register(cls, id, createSerializer));
+  }
+
+  @Override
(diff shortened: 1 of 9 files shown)


java/fury-core/src/main/java/org/apache/fury/AbstractThreadSafeFury.java
```diff
@@ -0,0 +1,33 @@
+package org.apache.fury;
```
tisonkun, line 1: Please properly add license header or include necessary info if it's copied.

tisonkun (commented):
LGTM for the technical part,

java/fury-core/src/main/java/org/apache/fury/AbstractThreadSafeFury.java
chaokunyang: header added.  Thanks for this catch, I always forget add header for new file.

tisonkun: I don't look into details but CI should fail if license headers missing.

tisonkun (approved):

java/fury-core/src/main/java/org/apache/fury/AbstractThreadSafeFury.java
pjfanning: * the code style check was failing until the header was added
* the Graal VM CI is failing still

pjfanning (changes requested):
Graal VM check is failing

java/fury-core/src/main/java/org/apache/fury/AbstractThreadSafeFury.java
chaokunyang: CI succeed now

chaokunyang:
CI has been fixed

pjfanning (approved):
lgtm

PR #1282, opened 2023-12-29
[CODE HEADER] Remove all fury author tag

### LINKED ISSUE #1281: [CODE HEADERS] Can we remove the author tags from the Fury code covered by the code grant?
It appears to be unusual to have author tags in ASF code bases. There are exceptions for code that is borrowed from 3rd party sources - if it was not explicitly granted to the ASF.

chaokunyang:
Closes #1281


pjfanning (approved):
lgtm

Issue #1275, opened 2023-12-28
[LICENSE] need proper licensing of Preconditions class?


pjfanning:
org.apache.fury.util.Preconditions seems to be basically derived from Guava's class.

If so. we need
* to replace the license header with the license header from the original Guava source file
* ideally add a comment with a link to the Guava source file
* update the LICENSE file to acknowledge the Guava license and copyright

I think we need a very good look at https://github.com/apache/incubator-fury/pull/1244 and https://github.com/apache/incubator-fury/pull/1114 to see if there are other source files where we have borrowed from Guava.

chaokunyang:
Hi @pjfanning , thanks for bring this license issue up, I'll take a look later

pjfanning:
I'll close this for now. Updating the project LICENSE/NOTICE can be done later.

PR #1283, opened 2023-12-29
[JAVA] add guava license header

pjfanning:
relates to #1275


java/fury-core/src/main/java/org/apache/fury/util/Preconditions.java
```diff
@@ -1,25 +1,24 @@
...
 
 package org.apache.fury.util;
 
-/** Check utils. */
+/**
+ * Check utils.
+ *
+ * derived from https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Preconditions.java
+ */
 public class Preconditions {
```
chaokunyang, line 22: We may need to add this file to `licenserc.toml`

chaokunyang (approved):
LGTM

PR #1284, opened 2023-12-29
[JAVA] remove license header from MurmurHash3

pjfanning:
feels wrong to add an ASL when the code is borrowed and the source is not Apache licensed

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/util/MurmurHash3.java
+++ b/java/fury-core/src/main/java/org/apache/fury/util/MurmurHash3.java
@@ -1,28 +1,11 @@
-/*
- * Licensed to the Apache Software Foundation (ASF) under one
- * or more contributor license agreements.  See the NOTICE file
- * distributed with this work for additional information
- * regarding copyright ownership.  The ASF licenses this file
- * to you under the Apache License, Version 2.0 (the
- * "License"); you may not use this file except in compliance
- * with the License.  You may obtain a copy of the License at
- *
- *   http://www.apache.org/licenses/LICENSE-2.0
- *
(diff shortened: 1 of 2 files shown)


pjfanning:
@tisonkun do you think this makes sense?

chaokunyang (approved):
LGTM

PragmaTwice (approved):

java/fury-core/src/main/java/org/apache/fury/util/MurmurHash3.java
```diff
@@ -1,28 +1,11 @@
...
- * under the License.
- */
-
 package org.apache.fury.util;
 
 // Derived from
 // https://github.com/yonik/java_util/blob/435ae306d2f2c077d981ab4de5c9ac3c45f92a4b/src/util/hash/MurmurHash3.java.
 
-// This implementtion is 3.5x faster than Guava's Hashing.murmur3_128.
+// yonik/java_util file has no license header
```
tisonkun, line 6: This would be redundant but no harm to include.

tisonkun (approved):
Thank you!

java/fury-core/src/main/java/org/apache/fury/util/MurmurHash3.java
pjfanning: I've spent the last year having Pekko releases blocked because of missing headers and having to explain that there is a good reason why there is no header

tisonkun: 🤣 well. Just if the reporter take a few second look at the referred original file..

Issue #1274, opened 2023-12-28
[LICENSE] update Fury LICENSE to list all the licenses for classes where we've borrowed non-Fury code


pjfanning:
https://github.com/apache/incubator-fury/blob/e0ba03b2f7462a63ee102d42fafce83790a7a392/licenserc.toml seems to have a list of source files that have contain code that is not originally from Fury.

These licenses will need to be acknowledged in the Fury license. And the Fury NOTICE may also need to be updated if some of the code comes from ASL licensed projects that have their own NOTICE files. We still need to do this even if the other code is in ASF projects (eg Arrow and Flink).

We may also need to update some source file headers.

chaokunyang:
@pjfanning We removed such headers and license acknowledgement in #783 , maybe we need to add them back and do more checks

pjfanning:
@chaokunyang @tisonkun the [LICENSE](https://github.com/apache/incubator-fury/blob/main/LICENSE) makes no mention of any borrowed code. There appears to be lots of borrowed code in Fury and it should all be mentioned in our LICENSE. This issue has nothing to do with #783.

pjfanning:
But the Fury NOTICE also makes zero mention of the borrowed code.

https://github.com/apache/incubator-fury/blob/main/NOTICE

With Pekko, we were forced by Incubator team to mention all the borrowed code in our LICENSEs. I will need to track down the references and the rules that forced us to do this. The problem with ASF projects is that not all ASF projects are compliant with the licensing requirements.

In the mean time, we can start updating the NOTICE.

tommyettinger:
Speaking only for myself, I was quite pleased to see my name in an `@author Tommy Ettinger` tag in FuryObjectMap. That was part of Kryo initially; some later enhancements to that class were applied to a different repo ([jdkgdxds](https://github.com/tommyettinger/jdkgdxds), in its ObjectObjectMap class), but they might not improve performance here, or in Kryo. If you want to pull code from jdkgdxds, be my guest; the nextPowerOfTwo(), place(), and resize() methods are different, for instance.

Noting where code came from can be a hassle, especially in a large project, but I think it's important to do.

Fury is doing very well in my benchmarks, and it's definitely been a nice experience using it. I look forward to seeing it advance in the future!

chaokunyang:
[...]

PR #1285, opened 2023-12-29
Update NOTICE to mention all the licenses for source files where code is borrowed from different project.

nandakumar131:
#1274

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -28,6 +28,10 @@ The text of each license is the standard Apache 2.0 license.
       java/fury-core/src/main/java/org/apache/fury/util/Platform.java
       java/fury-format/src/main/java/org/apache/fury/format/vectorized/ArrowWriter.java
 
+* flink (https://github.com/apache/flink)
+    Files:
+      java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+
 * commons-io (https://github.com/apache/commons-io)
     Files:
       java/fury-core/src/main/java/org/apache/fury/io/ClassLoaderObjectInputStream.java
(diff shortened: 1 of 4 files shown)


pjfanning:
Seems like a good start to me. I suspect we might need tidy up and possibly some more additions but this could probably be merged as is to get us started.

NOTICE
```diff
@@ -3,3 +3,100 @@ Copyright 2023 The Apache Software Foundation
...
+      java/fury-benchmark/src/main/java/org/apache/fury/benchmark/data/MediaContent.java
+      java/fury-benchmark/src/main/java/org/apache/fury/benchmark/data/Sample.java
+      java/fury-benchmark/src/main/java/org/apache/fury/benchmark/state/BufferType.java
+      java/fury-benchmark/src/main/java/org/apache/fury/benchmark/state/KryoState.java
+      java/fury-core/src/main/java/org/apache/fury/collection/FuryObjectMap.java
+      java/fury-core/src/main/java/org/apache/fury/collection/IdentityMap.java
+      java/fury-core/src/main/java/org/apache/fury/collection/IdentityObjectIntMap.java
+      java/fury-core/src/main/java/org/apache/fury/collection/LongMap.java
+      java/fury-core/src/main/java/org/apache/fury/collection/ObjectIntMap.java
+      java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
```
chaokunyang, line 65: This is derived from apache flink, could we put it into apache license part

PragmaTwice, line 65: It is hard to maintain the list to me. Could we just use wildcard matcher? e.g. `fury-benchmark/*`

nandakumar131: Fixed.

nandakumar131: Moved `MemoryBuffer` under the Apache license.
Thanks for the info.

chaokunyang: @PragmaTwice I suggest maintaining exact file list, not all of those files comes from kryo

nandakumar131: For now, I will add the exact file names. In the future, if we feel it's getting hard to maintain, we can replace it with wildcard.

chaokunyang: Sounds good to me

chaokunyang (approved):
LGTM

licenses/LICENSE-node.txt
tisonkun, line 1: @nandakumar131 can you elaborate how this file generated?

nandakumar131: This is a copy of the LICENSE from [nodejs/node](https://github.com/nodejs/node).
https://github.com/nodejs/node/blob/main/LICENSE


This is mentioned in the NOTICE file

> The following components are provided under the Node.js License. See project link for details.
The text of each license is also included in licenses/LICENSE-[project].txt.

PR #1286, opened 2023-12-31
[JavaScript] Implement the standard protocol

theweipeng:
1. Make the tag string lazy to avoid creating strings.
2. Add the serializeVolatile API to take ownership of the bufferWriter and avoid buffer copying, which is 30% faster than the serialize operation. There is a usage limitation; the bufferWriter remains inaccessible until the dispose method from the result of serializeVolatile has been called.
4. Remove the useLatin1 configuration, which is a standard in the Fury protocol.

Before: 
image

After:
image


chaokunyang (approved):
LGTM

Issue #1287, opened 2024-01-01
[Java]  refine Collection util data structure


mof-dev-3:
1. use local reference when possible
2. make tuple value field final

PR #1288, opened 2024-01-01
[Java] refine Collection util data structure (#1287)

mof-dev-3:
1. use local reference when possible
2. make tuple value field final

### DIFF
--- a/java/fury-benchmark/src/main/java/org/apache/fury/benchmark/Identity2IdMap.java
+++ b/java/fury-benchmark/src/main/java/org/apache/fury/benchmark/Identity2IdMap.java
@@ -26,16 +26,16 @@ import org.apache.fury.util.Platform;
 public class Identity2IdMap {
   private static final int RESERVE = 4;
   private static final int MAX_DEPTH = 4;
-  static int[] prim = {
+  static final int[] PRIM = {
     3, 5, 7, 11, 13, 17, 19, 23, 29, 37, 67, 97, 139, 211, 331, 641, 1097, 1531, 2207, 3121, 5059,
     7607, 10891, 15901, 19993, 30223, 50077, 74231, 99991, 150001, 300017, 1000033, 1500041, 200033,
     3000077, 5000077, 10000019
   };
 
   static int adjustSize(int size) {
-    for (int i = 0; i < prim.length - 1; i++) {
-      if (size < prim[i]) {
-        return prim[i] + RESERVE;
+    for (int i = 0; i < PRIM.length - 1; i++) {
+      if (size < PRIM[i]) {
+        return PRIM[i] + RESERVE;
       }
     }
     return size + RESERVE;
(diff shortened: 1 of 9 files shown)


java/fury-benchmark/src/main/java/org/apache/fury/benchmark/Identity2IdMap.java
```diff
@@ -26,16 +26,16 @@
 public class Identity2IdMap {
   private static final int RESERVE = 4;
   private static final int MAX_DEPTH = 4;
-  static int[] prim = {
```
chaokunyang, line 29: This class is copied from fst, it's unnecessary to update it

java/fury-core/src/main/java/org/apache/fury/collection/FuryObjectMap.java
```diff
@@ -180,6 +180,7 @@ public V put(K key, V value) {
   /** Skips checks for existing keys, doesn't increment size. */
   private void putResize(K key, V value) {
     K[] keyTable = this.keyTable;
+    V[] valueTable = this.valueTable;
```
chaokunyang, line 183: This is accessed only once, it's unnecessary to make it as a local variable

java/fury-core/src/main/java/org/apache/fury/collection/ObjectIntMap.java
```diff
@@ -92,6 +92,7 @@ public void put(K key, int value) {
 
   private void putResize(K key, int value) {
     K[] keyTable = this.keyTable;
+    int[] valueTable = this.valueTable;
```
chaokunyang, line 95: Accessed only once, unnecessary to use a local variable

java/fury-core/src/main/java/org/apache/fury/collection/Tuple2.java
```diff
@@ -46,6 +48,43 @@ public Tuple2(T0 value0, T1 value1) {
     this.f1 = value1;
   }
 
+  public static class NonFinal<T0, T1> implements Serializable {
```
chaokunyang, line 51: Make this as a top level class, maybe name it as MutableTuple2?

java/fury-core/src/main/java/org/apache/fury/collection/Tuple3.java
```diff
@@ -50,6 +52,70 @@ public Tuple3(T0 value0, T1 value1, T2 value2) {
...
+   * nonFinal.
+   *
+   * @param value0 The value for field 0
+   * @param value1 The value for field 1
+   * @param value2 The value for field 2
+   */
+  public static <T0, T1, T2> NonFinal ofNonFinal(T0 value0, T1 value1, T2 value2) {
+    return new NonFinal<>(value0, value1, value2);
+  }
+
```
chaokunyang, line 66: Ditto

java/fury-benchmark/src/main/java/org/apache/fury/benchmark/Identity2IdMap.java
mof-dev-3: Done. revert all change in this class

chaokunyang (approved):
LGTM

PR #1290, opened 2024-01-02
[JavaScript] compress number

theweipeng:
1. Compress the numbers so that they correspond to standard protocol
2. Remove the arrow dependency in cargo, it is unnecessary at this stage


chaokunyang (approved):
LGTM

Issue #812, opened 2023-08-01
[Python] Mac M1/M2 python wheel release


chaokunyang:
## Is your feature request related to a problem? Please describe.
Suppport install and release fury python for Mac M1/M2

## Additional context
#799

esafak:
Is this not done yet?

chaokunyang:
This is done too, we have macos-14, whitch is arm based

PR #1295, opened 2024-01-03
[CI] Add macos C++ build to ci.yml

PragmaTwice:
macOS C++ build is added to CI workflow in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,14 +43,14 @@ jobs:
       MY_VAR: "PATH"
     strategy:
       matrix:
-        java-version: ["8", "11", "17"]
+        java-version: ["8", "11", "17", "21"]
     steps:
       - uses: actions/checkout@v3
       - name: Set up JDK ${{ matrix.java-version }}
         uses: actions/setup-java@v3
         with:
           java-version: ${{ matrix.java-version }}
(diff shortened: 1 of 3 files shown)


tisonkun (approved):

.github/workflows/ci.yml
```diff
@@ -148,9 +132,13 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@nightly
       - run: ./ci/run_ci.sh rust
+
   cpp:
     name: C++ CI
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
chaokunyang, line 140: Does this include bont Mac Intel and Mac silicon?

PragmaTwice: No. Currently it's x86.

PragmaTwice: Actually it's a hard problem for arm64-based GitHub Action runners (it is currently not supported by GitHub official runners, but supported only in self-hosted runners). cc @tisonkun may have some information.

chaokunyang: We never release fury python for Mac M1/M2, seems it's not easy to automate it. It's beyoud the scope of this PR, we can discuss it in #812

PR #1297, opened 2024-01-03
[C++] Use python3 instead of python in bazel build

PragmaTwice:
Some linux distributions do not have `python` by default, and some make `python` point to python2 instead of python3.


tisonkun (approved):

chaokunyang (approved):
LGTM

PR #1303, opened 2024-01-04
ci(c++): Implement C++ CI scripts using python

LiangliangSui:
Implement C++ CI scripts using python.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,11 +141,8 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.8
-      - name: Install bazel
-        run: ./ci/run_ci.sh install_bazel
       - name: Run C++ CI with Bazel
-        run: ./ci/run_ci.sh cpp
-
+        run: python ./ci/run_ci.py cpp
   python:
     name: Python CI
     # Fix python 3.6 install issue, see
--- a/ci/run_ci.py
+++ b/ci/run_ci.py
@@ -1,0 +1,136 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
+import argparse
+import subprocess
+import platform
+import urllib.request as ulib
+import os
+import logging
+import importlib
+
+
+BAZEL_VERSION = "6.3.2"
+
+PYARROW_VERSION = "14.0.0"
+
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s"
+)
+
+
+def _exec_cmd(cmd: str):
+    logging.info(f"running command: {cmd}")
+    try:
+        result = subprocess.check_output(cmd, shell=True, universal_newlines=True)
+    except subprocess.CalledProcessError as error:
+        logging.error(error.stdout)
+        raise
+
+    logging.info(f"command result: {result}")
+    return result
+
+
+def _get_os_name_lower():
+    return platform.system().lower()
+
+
+def _is_windows():
+    return _get_os_name_lower() == "windows"
+
+
+def _get_os_machine():
+    machine = platform.machine().lower()
+    # Unified to x86_64(Windows return AMD64, others return x86_64).
+    return machine.replace("amd64", "x86_64")
+
+
+def _get_bazel_download_url():
+    download_url_base = (
+        f"https://github.com/bazelbuild/bazel/releases/download/{BAZEL_VERSION}"
+    )
+    suffix = "exe" if _is_windows() else "sh"
+    return (
+        f"{download_url_base}/bazel-{BAZEL_VERSION}{'' if _is_windows() else '-installer'}-"
+        f"{_get_os_name_lower()}-{_get_os_machine()}.{suffix}"
+    )
+
+
+def _run_cpp():
+    _install_cpp_deps()
+    # run test
+    query_result = _exec_cmd("bazel query //...")
+    _exec_cmd(
+        "bazel test {}".format(query_result.replace("\n", " ").replace("\r", " "))
+    )
+
+
+def _install_cpp_deps():
+    _exec_cmd(f"pip install pyarrow=={PYARROW_VERSION}")
+    _exec_cmd("pip install psutil")
+    _install_bazel()
+
+
+def _install_bazel():
+    local_name = "bazel.exe" if _is_windows() else "bazel-installer.sh"
+    bazel_download_url = _get_bazel_download_url()
+    logging.info(bazel_download_url)
+    ulib.urlretrieve(bazel_download_url, local_name)
+    os.chmod(local_name, 0o777)
+
+    if _is_windows():
+        bazel_path = os.path.join(os.getcwd(), local_name)
+        _exec_cmd(f'setx path "%PATH%;{bazel_path}"')
+    else:
+        _exec_cmd(f"./{local_name} --user")
+        os.remove(local_name)
+
+    # bazel install status check
+    _exec_cmd("bazel --version")
+
+    # default is byte
+    psutil = importlib.import_module("psutil")
+    total_mem = psutil.virtual_memory().total
+    limit_jobs = int(total_mem / 1024 / 1024 / 1024 / 3)
+    with open(".bazelrc", "a") as file:
+        file.write(f"\nbuild --jobs={limit_jobs}")
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    parser.set_defaults(func=parser.print_help)
+    subparsers = parser.add_subparsers()
+
+    cpp_parser = subparsers.add_parser(
+        "cpp",
+        description="Run C++ CI",
+        help="Run C++ CI",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
+    cpp_parser.set_defaults(func=_run_cpp)
+
+    args = parser.parse_args()
+    arg_dict = dict(vars(args))
+    del arg_dict["func"]
+    args.func(**arg_dict)
+
+
+if __name__ == "__main__":
+    _parse_args()


LiangliangSui:
Code Style Check CI continues to error, but I cannot get the cause of the error from the log.

LiangliangSui:
PTAL, thanks.

chaokunyang:
Could we add windows CI in this PR?

LiangliangSui:
Windows CI has not been added yet, windows bazel requires some additional configuration, I am trying it locally based on the current `run_ci.py`.

For the current linux and mac os C++ CI, it can work. Maybe we can merge the pr first.

ci/run_ci.py
```diff
@@ -0,0 +1,99 @@
...
+        # TODO: Implement other CI Tests.
+        pass
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser()
+    choices = ["java", "cpp"]
+    parser.add_argument(
+        "-i", "--item", choices=choices, help="Specify an item that needs to be run"
+    )
```
PragmaTwice, line 83: I think we should use `add_subparsers` instead of this.

ci/run_ci.py
```diff
@@ -0,0 +1,99 @@
...
+    return platform.system().lower()
+
+
+def _install_bazel():
+    os_name = _get_os_name_lower()
+    if os_name == "windows":
+        # TODO: Implement Windows download bazel.
+        bazel_cmd = "./bazel"
+    else:
+        URL = f"https://github.com/bazelbuild/bazel/releases/download/6.3.2/bazel-6.3.2-installer-{os_name}-x86_64.sh"
```
PragmaTwice, line 44: we can detect the architecture here, not only for x86_64.

ci/run_ci.py
```diff
@@ -0,0 +1,99 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
+import argparse
+import subprocess
+import platform
+import psutil
+import wget
```
PragmaTwice, line 23: We should use urllib instead of wget to minimize dependencies. Additionally, it appears that this package has been unmaintained for a long time.

ci/run_ci.py
```diff
@@ -0,0 +1,99 @@
...
+import os
+
+
+def _exec_cmd(cmd: list, stdout=None):
+    print(f"run command: {cmd}")
+    ret = subprocess.run(cmd, stdout=stdout)
+    assert ret.returncode == 0
+    return ret
```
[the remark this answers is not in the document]
PragmaTwice: Ahhh I see.

ci/run_ci.py
LiangliangSui:
``` python
def _run(item):
    if item.cpp:
        _run_cpp()


def _parse_args():
    parser = argparse.ArgumentParser()

    subparsers = parser.add_subparsers(title="subcommands", dest='subcommand', help="all subcommands")
    run_parser = subparsers.add_parser("run", help="Run the specified CI")
    run_parser.add_argument("--cpp", action="store_true", default=False, help="Run C++ CI")
    # TODO: Add other types of CI here

    parser.add_argument(
        "--install-bazel",
        action="store_true",
        default=False,
        help="Install bazel on the current machine",
    )

    return parser.parse_args()


if __name__ == "__main__":
    args = _parse_args()

    if args.subcommand == "run":
        _run(args)
    elif args.install_bazel:
        _install_bazel()
```

Like this ?

PragmaTwice: Nope. The subcommand should be `cpp`, `java`, etc.

You can refer to https://github.com/apache/kvrocks/blob/unstable/x.py

LiangliangSui: Done

ci/run_ci.py
```diff
@@ -0,0 +1,104 @@
...
+import subprocess
+import platform
+import psutil
+import urllib.request as ulib
+import os
+
+
+def _exec_cmd(cmd: list, stdout=None):
+    print(f"run command: {cmd}")
+    ret = subprocess.run(cmd, stdout=stdout)
```
chaokunyang, line 29: use `checked_call` is enough?

LiangliangSui: Use `check_output`, if the returncode is not 0, an exception will be thrown, otherwise the command output will be returned.

ci/run_ci.py
```diff
@@ -0,0 +1,107 @@
...
+import argparse
+import subprocess
+import platform
+import psutil
+import urllib.request as ulib
+import os
+
+
+def _exec_cmd(cmd: str):
+    print(f"run command: {cmd}")
```
PragmaTwice, line 28:
```suggestion
    logging.info(f"running command: {cmd}")
```

ci/run_ci.py
```diff
@@ -0,0 +1,107 @@
...
+    result = subprocess.check_output(cmd, shell=True).decode("utf-8")
+    print(result)
```
PragmaTwice, line 30:
```suggestion
    logging.info(f"command result: {result}")
```

ci/run_ci.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # default is byte
+    total_mem = psutil.virtual_memory().total
+    limit_jobs = int(total_mem / 1024 / 1024 / 1024 / 3)
+    with open(".bazelrc", "a") as file:
+        file.write(f"\nbuild --jobs={limit_jobs}")
+
+
+def _run_cpp():
+    bazel_cmd = "./bazel" if _get_os_name_lower() == "windows" else "bazel"
```
PragmaTwice, line 66: Why should it be `./bazel` in windows? Could bazel be in other directories?

ci/run_ci.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _get_os_machine():
+    return platform.machine()
+
+
+def _install_bazel():
+    os_name = _get_os_name_lower()
+    if os_name == "windows":
+        # TODO: Implement Windows download bazel.
```
PragmaTwice, line 45: I don't think it is hard to implement, you can see there are many executable for windows in this page:

https://github.com/bazelbuild/bazel/releases/6.3.2/

ci/run_ci.py
```diff
@@ -0,0 +1,107 @@
...
+        bazel_cmd = "./bazel"
+    else:
+        URL = f"https://github.com/bazelbuild/bazel/releases/download/6.3.2/bazel-6.3.2-installer-{os_name}-{_get_os_machine()}.sh"
+        local_name = "installer.sh"
```
PragmaTwice, line 49:
```suggestion
        local_name = "bazel-installer.sh"
```

ci/run_ci.py
```diff
@@ -0,0 +1,107 @@
...
+        ulib.urlretrieve(URL, local_name)
+        os.chmod(local_name, 0o777)
+        _exec_cmd(f"./{local_name} --user")
```
PragmaTwice, line 52: Could we delete this file after installing?

ci/run_ci.py
```diff
@@ -0,0 +1,107 @@
...
+    return platform.machine()
+
+
+def _install_bazel():
+    os_name = _get_os_name_lower()
+    if os_name == "windows":
+        # TODO: Implement Windows download bazel.
+        bazel_cmd = "./bazel"
+    else:
+        URL = f"https://github.com/bazelbuild/bazel/releases/download/6.3.2/bazel-6.3.2-installer-{os_name}-{_get_os_machine()}.sh"
```
PragmaTwice, line 48:
```suggestion
        bazel_version = "6.3.2"
        bazel_download_url = f"https://github.com/bazelbuild/bazel/releases/download/{bazel_version}/bazel-{bazel_version}-installer-{os_name}-{_get_os_machine()}.sh"
```

ci/run_ci.py
LiangliangSui: Requires additional installation of Visual C++ Build Tool and setting environment variables.

https://bazel.build/configure/windows#build_cpp

ci/run_ci.py
LiangliangSui: It may also be `bazel.exe`, I will modify it when adding Windows C++ CI.

ci/run_ci.py
PragmaTwice: We can default to these things already being installed.

ci/run_ci.py
LiangliangSui: Can be promoted to become a global variable.

ci/run_ci.py
PragmaTwice: You can pass a cli option for it.

ci/run_ci.py
LiangliangSui: Currently the download and installation logic is implemented here ?

PragmaTwice: I mean we can assume MSVC are already installed.

LiangliangSui:
```python
def _install_bazel():
    is_windows = _get_os_name_lower() == "windows"
    download_url_base = f"https://github.com/bazelbuild/bazel/releases/download/{BAZEL_VERSION}"
    local_name = "bazel.exe" if is_windows else "bazel-installer.sh"
    exe_tail = "exe" if is_windows else "sh"
    bazel_download_url = f"{download_url_base}/bazel-{BAZEL_VERSION}{'' if is_windows else '-installer'}-" \
                         f"{os_name}-{_get_os_machine()}.{exe_tail}"
    ulib.urlretrieve(bazel_download_url, local_name)

    if is_windows:
        bazel_path = os.path.join(os.curdir, local_name)
        _exec_cmd(f"setx path \"%PATH%;{bazel_path}\"")
    else:
        os.chmod(local_name, 0o777)
        _exec_cmd(f"./{local_name} --user")
        os.remove(local_name)

    # bazel install status check
    _exec_cmd("bazel --version")

    # default is byte
    total_mem = psutil.virtual_memory().total
    limit_jobs = int(total_mem / 1024 / 1024 / 1024 / 3)
    with open(".bazelrc", "a") as file:
        file.write(f"\nbuild --jobs={limit_jobs}")


def _run_cpp():
    # run test
    query_result = _exec_cmd("bazel query //...")
    _exec_cmd("bazel test {}".format(query_result.replace("\n", " ").replace("\r", " ")))
```

Windows download has been implemented, PTAL.

PragmaTwice: looks good to me.

LiangliangSui:
Hi @PragmaTwice, It has been revised according to the comments mentioned.

In addition, the local adaptation of Windows C++ CI has been completed in the past two days. After this PR is merged, another PR will be created to implement the final adaptation of Windows C++ CI.

.github/workflows/ci.yml
```diff
@@ -141,11 +141,13 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.8
-      - name: Install bazel
-        run: ./ci/run_ci.sh install_bazel
+      - name: Install Bazel
+        run: |
+          pip install psutil
+          pip install pyarrow==14.0.0
```
chaokunyang, line 147: Do we need to install pyarrow in py script?

LiangliangSui: We can do this by moving install into py and then calling the shell through py to execute install.

chaokunyang: Sounds good to me. Can we install `psutil` and `pyarrow` in py script in this PR? The import for `psutil` can be moved into function to avoid import error

LiangliangSui: Done.

chaokunyang (approved):
LGTM

LiangliangSui:
Can this PR be merged?

PR #1304, opened 2024-01-04
[JavaScript] Serialize any type

bytemain:
remove the limit: root type must be an object


theweipeng:
I have refactored the code generator in another pr #1313 , which had been remove the limit.

PR #1311, opened 2024-01-05
Keep head metadata as is in docs

PragmaTwice:
We can just keep these head metadata as it is instead of replacing with html comments.

It will display in a normal manner on GitHub.


tisonkun (approved):
More explicit. Thank you!

PR #1313, opened 2024-01-06
feat(javascript): Refactor &  Compress Long

theweipeng:
## Refactor the code generator
1. Currently, JavaScript can only register a Object by Type.object, because the generator is too simple
2. Primitive generator of types like string、number can't be inlined in the holder collection, which cause necessary polymorphic that affect the performance

## Enhancement performance
1. Currently, Buffer.latin1Slice would transfer from native code generated by JIT to v8 runtime, This has a significant impact on performance, so we use String.fromCharCode to create little string when the string length is less than 15.  It is a magic number, but I have tested it and it works fine.

## Improve
When useSliceString is disabled, the performance of deserialization improves by 100%. It is 3 times faster than native JSON and twice as fast as protobuf.

When useSliceString is enabled, the performance of deserialization improves by 30%. It is 5 times faster than native JSON and 3 times faster than protobuf.

#### Disable useSliceString
image

#### Enable useSliceString
image

#### Disable useSliceString
image

#### Enable useSliceString
image


theweipeng:
@bytemain  Hi, could you please take a review, It remove the limit which mentioned in #1304

bytemain:
image

I checked the snapshot and donot sure is this a problem?

bytemain:
fantasitic work, this is the thing I wanted of a serialization framework

theweipeng:
> it seem that this is not a problem, all the call will happen in this closure
> 
> image
> I checked the snapshot and donot sure is this a problem?
This is not a problem, but declare the readInner before the read is better. I will change it

javascript/packages/fury/lib/classResolver.ts
```diff
@@ -158,26 +184,15 @@ export default class SerializerResolver {
...
-
   readTag(binaryReader: BinaryReader) {
     const flag = binaryReader.uint8();
     if (flag === USESTRINGVALUE) {
       binaryReader.skip(8); // The tag hash is not needed at the moment.
-      const start = binaryReader.getCursor();
       const len = binaryReader.int16();
+      const start = binaryReader.getCursor();
       binaryReader.skip(len);
       this.readStringPool.push(Lazystring.fromPair(start, len));
```
bytemain, line 197: I think the class name should use camel case

```diff
- class Lazystring
+ class LazyString
```

javascript/packages/fury/lib/gen/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    throw new Error(`${description.type} generator not exists`);
+  }
+  const scope = new Scope();
+  const generator = new InnerGeneratorClass(description, new Builder(scope, fury), scope);
+
+  const funcString = generator.toSerializer();
+  return new Function(funcString);
+};
+
+function regDependences(fury: Fury, description: TypeDescription) {
```
bytemain, line 50:
```diff
- regDependences
+ regDependencies
```

javascript/packages/fury/lib/type.ts
```diff
@@ -28,7 +29,7 @@ export type BinaryReader = ReturnType<typeof BinaryReader>;
 export enum InternalSerializerType {
   STRING = 13,
   ARRAY = 25,
-  TUPLE = 25,
+  TUPLE = 25.1,
```
bytemain, line 32: after the enum value change, we can remove the redundant code now:
image

bytemain:
> @bytemain Hi, could you please take a review, It remove the limit which mentioned in #1304

I think you can add some test case in the #1304:

- https://github.com/apache/incubator-fury/pull/1304/files#diff-9d33aa5fce03357373ad7c54b9636452600284f3e7153064438e277158a3e9cfR60

bytemain:
Looks good to me, but has some typo.

And we can improve the style of generated code, this can be done in further

bytemain (approved):

theweipeng:
> Looks good to me, but has some typo.
> 
> And we can improve the style of generated code, this can be done in further

I fixed it and added an afterCodeGenerated hook to allow users to format the code. Formatting is not a default action, as I believe we should keep things simple and avoid including too many dependencies.

javascript/packages/fury/lib/gen/number.ts
```diff
@@ -0,0 +1,105 @@
...
+    builder => builder.reader.sliLong()
+  )
+);
+Register.reg(InternalSerializerType.FLOAT,
+  buildNumberSerializer(
+    (builder, accessor) => builder.writer.float(accessor),
+    builder => builder.reader.float()
+  )
+);
+Register.reg(InternalSerializerType.DOUBLE,
```
chaokunyang, line 100:
```suggestion
Registry.register(InternalSerializerType.DOUBLE,
```

javascript/packages/fury/lib/gen/router.ts
```diff
@@ -17,28 +17,22 @@
...
-    config: () => {
-      return {
-        reserve: 4,
-      };
-    },
-  };
-};
+type SerializerGeneratorConstructor = new (description: TypeDescription, builder: Builder, scope: Scope) => SerializerGenerator;
+
+export class Register {
```
chaokunyang, line 28: This can be named as `Registry`, so `reg` can be renamed to `register`

javascript/packages/fury/lib/reader.ts
```diff
@@ -144,21 +190,67 @@ export const BinaryReader = (config: Config) => {
...
+    }
+    return result;
+  }
+
+  function bigUInt8() {
+    return BigInt(uint8() >>> 0);
+  }
+
+  function varUInt64() {
+    let byte_ = bigUInt8();
```
chaokunyang, line 224: If remaing bytes is greater than 8, we can get a long first, and check its bits. This will have better performance than read byte by byte

javascript/packages/fury/lib/classResolver.ts
```diff
@@ -17,87 +17,102 @@
...
-    this.internalSerializer[InternalSerializerType.FURY_STRING_ARRAY] = stringArraySerializer(fury);
-    this.internalSerializer[InternalSerializerType.FURY_PRIMITIVE_BOOL_ARRAY] = boolArraySerializer(fury);
-    this.internalSerializer[InternalSerializerType.FURY_PRIMITIVE_SHORT_ARRAY] = shortArraySerializer(fury);
-    this.internalSerializer[InternalSerializerType.FURY_PRIMITIVE_INT_ARRAY] = intArraySerializer(fury);
-    this.internalSerializer[InternalSerializerType.FURY_PRIMITIVE_LONG_ARRAY] = longArraySerializer(fury);
-    this.internalSerializer[InternalSerializerType.FURY_PRIMITIVE_FLOAT_ARRAY] = floatArraySerializer(fury);
-    this.internalSerializer[InternalSerializerType.FURY_PRIMITIVE_DOUBLE_ARRAY] = doubleArraySerializer(fury);
-    this.internalSerializer[InternalSerializerType.BINARY] = binarySerializer(fury);
+    this.internalSerializer[InternalSerializerType.ANY] = anySerializer(fury);
+    this.regSerializer(fury, Type.string());
```
chaokunyang, line 91: Why not name it as `registerSerializer`, `regSerializer` seems a little vague.

javascript/packages/fury/lib/gen/router.ts
chaokunyang: Seems it's codegen related, maybe we can name it as `CodegenRegistry` or something like

chaokunyang: In java we have a class named as `CodeGenerator`, and has a class named `CodecBuilder`. Hope it can give some inspiration

javascript/packages/fury/lib/reader.ts
theweipeng: Great suggestion! Currently, the maximum bit length that the Number type can safely represent in JavaScript is 53 bits, which is insufficient to decode our varUInt64. BigInt serves as a workaround; however, its performance in JavaScript is significantly suboptimal. I plan to address the performance issues associated with handling 64-bit numbers comprehensively in subsequent pull requests.

chaokunyang (approved):
LGTM

PR #1317, opened 2024-01-06
ci: add PR title lint following the conventional commits

PragmaTwice:
We can add a CI workflow to check if the PR title follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).


tisonkun (approved):
Let's give it a try.

PR #1324, opened 2024-01-09
fix(java): fix polymorphic array serialization

### LINKED ISSUE #1323: [Java] Component types on generic arrays are lost during (de-)serialization
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-fury/issues) and found no similar issues.

### Version

* 0.5.0-SNAPSHOT from the nightly builds (current HEAD is a80dce6)
* OS: Linux x64
* Java: fails on JDK 11/17/21 

### Component(s)

Java

### Minimal reproduce step

```java
public class App {

        final Fury fury = Fury.builder().requireClassRegistration(false).build();
        final ArrayWrapper<String> wrapper = new ArrayWrapper<>(String.class, 2);

        wrapper.array[0] = "Hello";

        final byte[] bytes = fury.serialize(wrapper);
        final ArrayWrapper<String> deserialized = (ArrayWrapper<String>) fury.deserialize(bytes);

        deserialized.array[1] = "World";
    }

    static class ArrayWrapper<T> {

        private final T[] array;

        public ArrayWrapper(Class<T> clazz, int capacity) {
            this.array = (T[]) Array.newInstance(clazz, capacity);
        }
    }
}
```

A successfull run of the program

### What did you see instead?

```
Exception in thread "main" java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [Ljava.lang.String; ([Ljava.lang.Object; and [Ljava.lang.String; are in module java.base of loader 'bootstrap')
```

### Anything Else?

It appears that the information about the component type (here `String.class`) are lost and the (de-) serializer simply assumes "Generic Type? => Object!". The problem here is that one cannot simply replace the `Array.newInstance(clazz, capacity);` with `new Object[capacity];` because then the `wrapper.array[0] = "Hello";` statement fails with the same error.

chaokunyang:
`Object[]` can be assigned from `String[]` and other arrays, we take it as a final type before, so we skip writing classinfo for such types and lost object type too.

This PR fix this polymorphic array serialization bug by checking array type polymorphism by component type.

Closes #1323

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -515,17 +515,11 @@ public class ClassResolver {
    */
   public boolean isMonomorphic(Class<?> clz) {
     if (fury.getConfig().shareMetaContext()) {
-      if (!ReflectionUtils.isMonomorphic(clz)) {
-        return false;
-      }
-      boolean isInnerClass = isInnerClass(clz);
-      if (!isInnerClass) {
-        return false;
-      } else {
-        // can't create final map/collection type using TypeUtils.mapOf(TypeToken<K>,
-        // TypeToken<V>)
-        return !Map.class.isAssignableFrom(clz) && !Collection.class.isAssignableFrom(clz);
-      }
+      // can't create final map/collection type using TypeUtils.mapOf(TypeToken<K>,
+      // TypeToken<V>)
+      return ReflectionUtils.isMonomorphic(clz)
+          && isInnerClass(clz)
+          && (!Map.class.isAssignableFrom(clz) && !Collection.class.isAssignableFrom(clz));
     }
     return ReflectionUtils.isMonomorphic(clz);
   }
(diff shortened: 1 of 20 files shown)


java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -514,22 +513,21 @@ public List<Class<?>> getRegisteredClasses() {
...
+      }
+      boolean isInnerClass = isInnerClass(clz);
+      if (!isInnerClass) {
+        return false;
       } else {
-        return true;
+        // can't create final map/collection type using TypeUtils.mapOf(TypeToken<K>,
+        // TypeToken<V>)
+        return !Map.class.isAssignableFrom(clz) && !Collection.class.isAssignableFrom(clz);
       }
```
LiangliangSui, line 528: Can it be simplified to `return ReflectionUtils.isMonomorphic(clz) && isInnerClass(clz) && (!Map.class.isAssignableFrom(clz) && !Collection.class.isAssignableFrom(clz))` ?

theweipeng (approved):

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
chaokunyang: updated

theweipeng (approved):

PR #1329, opened 2024-01-11
chore(cpp): rename src to cpp

### LINKED ISSUE #1327: Rename src to cpp
Can we rename src to cpp so that it can be unified with the existing naming method (such as Java, javascript, python, etc.).

chaokunyang:
rename src to cpp

Closes #1327


chaokunyang:
cc @PragmaTwice @theweipeng

theweipeng (approved):
LGTM

PR #1334, opened 2024-01-11
feat(java): refine collection builder util

chaokunyang:
refine collection builder util to avoid array creation cost and generic array creation warnings


theweipeng (approved):
LGTM

PR #1336, opened 2024-01-12
[refactor] threadSafeFury use thread pool

### LINKED ISSUE #1335: [Java] Use `org.apache.fury.pool.ThreadPoolFury` as default ThreadSafeFury implementation
Currently we use `org.apache.fury.ThreadLocalFury` as the default implementation of `ThreadSafeFury`. It create a Fury instance for every thread. But threading model is not controlable by Fury, if users create threads frequently, Fury creation may become bottleneck or introduce extra latency.

Use `org.apache.fury.pool.ThreadPoolFury` as default ThreadSafeFury implementation.

We may neeed to benchmark `ThreadPoolFury` vs `ThreadLocalFury`, and optimize `ThreadPoolFury` if it's slow

zzzk1:
close #1335 

**Use org.apache.fury.pool.ThreadPoolFury as default ThreadSafeFury implementation.**

**failed test case** 
- *org.apache.fury.ThreadSafeFuryTest#testClassDuplicateName*
- *org.apache.fury.ThreadSafeFuryTest#testClassGC*
- *org.apache.fury.ThreadSafeFuryTest#testSerializeWithMetaShare*

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
+++ b/java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
@@ -340,8 +340,8 @@ public final class FuryBuilder {
   }
 
   /** Build thread safe fury. */
-  public ThreadSafeFury buildThreadSafeFury() {
-    return buildThreadLocalFury();
+  public ThreadSafeFury buildThreadSafeFury(int minPoolSize, int maxPoolSize) {
+    return buildThreadSafeFuryPool(minPoolSize, maxPoolSize);
   }
 
   /** Build thread safe fury backed by {@link ThreadLocalFury}. */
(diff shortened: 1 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
```diff
@@ -340,8 +340,8 @@ public Fury build() {
   }
 
   /** Build thread safe fury. */
-  public ThreadSafeFury buildThreadSafeFury() {
```
chaokunyang, line 343: Why remove this API? We can have a default configuration, such as use cpu cores as min instance, 2 * cpu cores as max instances

zzzk1: Okay

Issue #1330, opened 2024-01-11
[JavaScript] Specification implementation progress


bytemain:
Now we have a new java spec:
https://github.com/apache/incubator-fury/blob/main/docs/protocols/java_object_graph.md

and there have been some changes compared to the previous version.

Changes:
- Array/Map

Has not implemented:
- Schema Evolution

bytemain:
`Enum` has been done in #1321

theweipeng:
`Array` has been done in  #1337 .

PR #1337, opened 2024-01-12
feat(JavaScript): Implement the collection protocol

theweipeng:
Now we have introduced a new binary layout for Collection, which is utilized by both Array and Set. The Collection is defined in the [SPEC](https://github.com/apache/incubator-fury/blob/3da32797beb25fd39b06c42f8c8b9e67eec5184d/docs/protocols/java_object_graph_spec.md#unsigned-int). This PR implements the Collection in JavaScript.

1. Array and Set will extend CollectionSerializerGenerator to generate code.
2. The Xlang Collection is not stable yet; it may undergo changes in the future, particularly regarding polymorphism, although the changes are expected to be minimal.
3. We have refined the reference checks, which were previously disorganized.
4. The refactor of collection types has resulted in performance improvements. In our sample benchmark, we observed an increase in performance of nearly 20%.

#1330


theweipeng:
@bytemain  The implementation of the collection mentioned in #1330 is complete.

javascript/packages/fury/lib/gen/collection.ts
```diff
@@ -0,0 +1,365 @@
...
+          serializer!.write(item);
+        }
+      } else {
+        for (const item of value) {
+          const serializer = this.fury.classResolver.getSerializerByData(item);
+          this.fury.binaryWriter.uint16(serializer!.meta.type);
+          serializer!.write(item);
+        }
+      }
+    }
```
LiangliangSui, line 131: Can something be streamlined here? A lot of repetition.

theweipeng: Indeed, there is a lot of duplicate code; I'll think about how to simplify it.

javascript/packages/fury/lib/gen/collection.ts
```diff
@@ -0,0 +1,365 @@
...
+          if (RefFlags.NullFlag === refFlag) {
+            accessor(result, index, null);
+          } else {
+            const serializer = this.fury.classResolver.getSerializerById(this.fury.binaryReader.int16());
+            accessor(result, index, serializer.read());
+          }
+        }
+      }
+      return result;
+    }
```
LiangliangSui, line 180: ditto

javascript/test/__snapshots__/codeGen.test.ts.snap
```diff
@@ -8,13 +8,13 @@ return function(fury, external) {
   const bw = fury.binaryWriter;
   const cr = fury.classResolver;
   const rr = fury.referenceResolver;
-  const tag_ser_3 = fury.classResolver.getSerializerByTag("example.foo.1");
-  const tag_ser_4 = fury.classResolver.getSerializerByTag("example.foo.2");
-  const tag_ser_12 = fury.classResolver.getSerializerByTag("example.bar.1");
-  const tag_ser_13 = fury.classResolver.getSerializerByTag("example.bar.2");
-  const tagWriter_19 = cr.createTagWriter("tuple-object-wrapper");
+  const tag_ser_4 = fury.classResolver.getSerializerByTag("example.foo.1");
```
chaokunyang, line 11: Could we avoid to commit `ts.snap` to repo?

bytemain: .snap is used in unittest, it can ensure the generated code to match our expectation

javascript/test/__snapshots__/codeGen.test.ts.snap
```diff
@@ -25,136 +25,178 @@ return function(fury, external) {
       tuple2: null,
       tuple2_: null
     };
-    rr.pushReadObject(result_0)
     
+    if (fromRef) {
+      rr.reference(result_0)
+    } else {
+
```
bytemain, line 32: shall we remove this empty else condition statement

javascript/test/__snapshots__/codeGen.test.ts.snap
```diff
@@ -25,136 +25,178 @@ return function(fury, external) {
...
+    if (fromRef) {
+      rr.reference(result_0)
+    } else {
+      
+    }
+    
+    
+    let fromRef_1 = false;
     switch (br.int8()) {
+      case 0:
```
bytemain, line 38: here we have two `case 0`

bytemain (changes requested):

javascript/packages/fury/lib/gen/collection.ts
theweipeng: Whether to write the 2-byte type and the 1-byte refflag is determined by whether the element contains null or is entirely of the same type. I can't think of a better way to simplify this without avoiding the determination of 'isSame' and 'includeNone' on each iteration.

javascript/packages/fury/lib/gen/collection.ts
theweipeng: Same situation above

javascript/packages/fury/lib/gen/collection.ts
```diff
@@ -0,0 +1,364 @@
...
+            continue;
+          }
+          this.fury.binaryWriter.uint8(RefFlags.NotNullValueFlag);
+          serializer!.write(item);
+        }
+      }
+    } else {
+      if (includeNone) {
+        for (const item of value) {
+          const serializer = this.fury.classResolver.getSerializerByData(item);
```
chaokunyang, line 115: this can be moved after `if` branch

javascript/packages/fury/lib/gen/collection.ts
```diff
@@ -0,0 +1,364 @@
...
+    let isSame = true;
+    let serializer: Serializer | null = null;
+    let includeNone = false;
+
+    for (const item of arr) {
+      if ((item === undefined || item === null) && !includeNone) {
+        includeNone = true;
+      }
+      if (isSame) {
+        const current = this.fury.classResolver.getSerializerByData(item);
```
chaokunyang, line 58: This can be cached, the hash look up and method call is costly. We can cache it, and check whether element types are same as previous element to reduce this cose

javascript/test/__snapshots__/codeGen.test.ts.snap
theweipeng: Also used for reviewing the generated code on GitHub and tracing the code changes with Git.

theweipeng: Also used for reviewing the generated code on GitHub and tracing the code changes with Git.

chaokunyang: Can we use unittest to ensure the correctness of generated code. If we have a complicated class, we may generate thounds of lines of code. In fury java, I have see some big objects generated 6000 line of code.

javascript/packages/fury/lib/gen/collection.ts
theweipeng: The `getSerializerByData` function can only check the 8 base types of a JavaScript object because objects in JavaScript usually do not have a type. Therefore, it does not include a hash lookup, just an array index. 
Additionally, the`writeElementsHeader` function is used only in compatible mode, where the user declares the type by using `Type.any()` rather than a specific type like `Type.string()`.  If a specific type is declared,`readStmtSpecificType` and `writeStmtSpecificType` will generate code that does not require checking if the elements are of the same type.

theweipeng: I will add a collection snapshot for you to review the generated code.

javascript/test/__snapshots__/collection.test.ts.snap
```diff
@@ -0,0 +1,159 @@
...
+    const flags_2 = br.uint8();
+    br.skip(2);
+    const len_1 = br.varUInt32();
+    const result_0 = new Array(len_1);
+    
+    if (fromRef) {
+      rr.reference(result_0)
+    }
+    
+    if (flags_2 & 1) {
```
theweipeng, line 23: @chaokunyang  This snapshot is generated by a typed collection declared with `Type.array(Type.string())`, which is equivalent to `Array<String>` in Java.

javascript/packages/fury/lib/gen/collection.ts
LiangliangSui:
```javascript
    for (const item of value) {
      if (includeNone) {
        if (item === null || item === undefined) {
          this.fury.binaryWriter.uint8(RefFlags.NullFlag);
          continue;
        }
        this.fury.binaryWriter.uint8(RefFlags.NotNullValueFlag);
      }

      let finalSerializer = serializer;
      if (!isSame) {
        finalSerializer = this.fury.classResolver.getSerializerByData(item);
        this.fury.binaryWriter.uint16(finalSerializer!.meta.type);
      }
      finalSerializer!.write(item);
    }
```

Can it become like this?

chaokunyang: LGTM. Although this introduces `isSame` and `includeNone` check overhead for every element, but compared to other memory writing overhead, I think the overhead acceptable

chaokunyang (approved):
LGTM

PR #1342, opened 2024-01-13
feat(java): refine collection builder util (#1334), fix collection init size typo

mof-dev-3:
refine collection builder util (#1334), fix collection init size typo


chaokunyang (approved):
LGTM, thanks

PR #1343, opened 2024-01-13
feat(java): Specify JPMS module names

### LINKED ISSUE #1341: [Java] Specify JPMS module names

When using fury in a modular (JPMS) application, the `maven-compiler-plugin` warns that

```
[WARNING] ****************************************************************************************************************************************************
[WARNING] * Required filename-based automodules detected: [fury-core-0.5.0-SNAPSHOT.jar]. Please don't publish this project to a public artifact repository! *
[WARNING] ****************************************************************************************************************************************************
```

With this feature request I propose to claim and configure module names to get rid of this warning.

I basically see two feasible approaches to achieve this goal:

1. The least invasive approach would be to use the `Automatic-Module-Name` entries in the `MANIFEST.MF`s of the respective `.jar` files. This does not affect the build process in any way and can be configured relatively easy via configurations in the `maven-jar-plugin` (or the `maven-shade-plugin` if you use it).
2. To fully support *named* modules, you would have to provide a full `module-info.java` descriptor. Since this is a Java9+ feature, you would have to extend the build process to either use multiple compile phases (one for `module-info.java` on Java9+ and one for the remaining code on Java8 to maintain backwards compatibility) or build a multi-release JAR. This would make sense if you want to support tools such as `jlink` or `jpackage` in the future. However, this would also require your dependencies to be fully modularized. So if that is not the case, it is probably not worth the effort for now.

## Additional context

I could lend a hand and provide an initial PR for the first solution. The only thing worth discussing beforehand is which module names you want to claim. You could either go with the direct translation of the maven coordinates, e.g. `org.apache.fury.fury_core`, `org.apache.fury.fury_format`, etc. or go with more idiomatic names such as `org.apache.fury.core`, `org.apache.fury.format`, etc.

mtf90:
This PR adds support for JPMS modules by adding the `Automatic-Module-Name` entry to the `MANIFEST.MF` files of the constructed `.jar` files.

Closes #1341.


chaokunyang (approved):
LGTM, one minor thing, do we need to add a test in integration_test dir?

mtf90:
I added an integration test that compiles some arbitrary code that references the new module names in a `module-info.java` file.

ci/run_ci.sh
```diff
@@ -118,6 +118,9 @@ integration_tests() {
   export JAVA_HOME="$ROOT/zulu21.28.85-ca-jdk21.0.0-linux_x64"
   export PATH=$JAVA_HOME/bin:$PATH
   mvn -T10 -B --no-transfer-progress clean test
+  echo "Start JPMS tests"
+  cd "$ROOT"/integration_tests/jpms_tests
+  mvn -T10 -B --no-transfer-progress clean compile
```
chaokunyang, line 123:
```suggestion
  mvn -T10 -B --no-transfer-progress clean test
```

integration_tests/jpms_tests/src/main/java/org/apache/fury/integration_tests/Test.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.fury.Fury;
+import org.apache.fury.benchmark.Benchmark;
+import org.apache.fury.format.encoder.Encoders;
+import org.apache.fury.test.bean.Foo;
+
+/**
+ * A test class that simply references classes from the various Fury artifacts to check whether
+ * they specify the module names referenced in the module-info descriptor.
+ */
+public class Test {
```
chaokunyang, line 31: Could We add some testng unit test?

mtf90: What should the unit test look like? To my understanding, this integration test should only check that the module names are configured correctly and not that the code does anything reasonable (this is what the unit tests of the actual core, format, etc. modules are for). This already happens at the compile level, because if the compiler cannot find the modules, it can't link against the referenced classes  (you can check this by using an older `maven-compiler-plugin` version (e.g., `3.3.0`) which does not put the dependencies on the module-path correctly).

I could move this class to `src/test/java` but it still wouldn't really be a `@Test` class. It would only move the potential error detection from the `default-compile` execution to the `default-testCompile` execution.

chaokunyang: I see, thanks for detailed explanation

chaokunyang (approved):
LGTM

PR #1346, opened 2024-01-17
fix(java): fix ClassLoader npe in loadOrGenCodecClass

chaokunyang:
fix ClassLoader npe in `loadOrGenCodecClass` mentioned in #1340

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/builder/BaseObjectCodecBuilder.java
+++ b/java/fury-core/src/main/java/org/apache/fury/builder/BaseObjectCodecBuilder.java
@@ -479,15 +479,18 @@ public abstract class BaseObjectCodecBuilder extends CodecBuilder {
       Class<? extends Serializer> serializerClass =
           visitFury(f -> f.getClassResolver().getSerializerClass(cls));
       Preconditions.checkNotNull(serializerClass, "Unsupported for class " + cls);
-      ClassLoader beanClassClassLoader =
-          beanClass.getClassLoader() == null
-              ? Thread.currentThread().getContextClassLoader()
-              : beanClass.getClassLoader();
       if (!ReflectionUtils.isPublic(serializerClass)) {
         // TODO(chaokunyang) add jdk17+ unexported class check.
         // non-public class can't be accessed in generated class.
         serializerClass = Serializer.class;
       } else {
+        ClassLoader beanClassClassLoader = beanClass.getClassLoader();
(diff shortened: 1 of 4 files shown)


Maurice-Betzel (commented):
That was my initial idea as well, get the Fury class loader. But I am still curious why a thread does not have a context class loader. The context class loader is set at the time of the creation of a thread. If not set explicitly, it defaults to the context class loader of the parent thread. So somewhere (JIT?) it gets actively nulled out.
Edit, on second thought, if the parent of the thread was the main application thread having a null class loader pointing to the Java boot class loader, it makes sense to get a null context class loader.
I just have not see that happen before in my career.
(removed part)

Maurice-Betzel:
There is another location having the same problem, after adding the solution here as well all seems up and running.
I observe see a strange phenomenon, if I changed Fury code I had to delete my complete runtime data folder because it was holding a reference to previous Fury code somewhere which I did not find yet.

```
  protected Expression getOrCreateSerializer(Class<?> cls) {
    // Not need to check cls final, take collection writeSameTypeElements as an example.
    // Preconditions.checkArgument(isFinal(cls), cls);
    Reference serializerRef = serializerMap.get(cls);
    if (serializerRef == null) {
      // potential recursive call for seq codec generation is handled in `getSerializerClass`.
      Class<? extends Serializer> serializerClass =
          visitFury(f -> f.getClassResolver().getSerializerClass(cls));
      Preconditions.checkNotNull(serializerClass, "Unsupported for class " + cls);
      ClassLoader beanClassClassLoader =
          beanClass.getClassLoader() == null
              ? Thread.currentThread().getContextClassLoader()
              : beanClass.getClassLoader();
      if (beanClassClassLoader == null) beanClassClassLoader = Fury.class.getClassLoader();
```

chaokunyang:
> There is another location having the same problem, after adding the solution to BaseObjectCodecBuilder as well all seems up and running.
> 
> ```
>   protected Expression getOrCreateSerializer(Class<?> cls) {

Good catch, thanks. I just pushed a fix for it.

chaokunyang:
> That was my initial idea as well, get the Fury class loader. But I am still curious why a thread does not have a context class loader. The context class loader is set at the time of the creation of a thread. If not set explicitly, it defaults 

This breaks my assumption too, thread classloader will be app classloader by default, maybe someone set it to null by accident

theweipeng (approved):

PR #1347, opened 2024-01-19
feat(JavaScript): Support esm

theweipeng:
ESM is the next-generation module system for javascript. We should support it.


javascript/benchmark/index.mjs
```diff
@@ -17,17 +17,19 @@
...
+import Fury from "@furyjs/fury";
+import Hps from '@furyjs/hps';
+import Benchmark from 'benchmark';
+import protobuf from 'protobufjs';
+import path, { dirname } from 'path';
+import assert from 'assert';
+import { fileURLToPath } from 'url';
+import { spawn } from 'child_process';
 
+const fury = new Fury.default({ hps: Hps, refTracking: false, useSliceString: true });
```
bytemain, line 30: if we use esm format, here we can remove this default

javascript/packages/fury/package.json
```diff
@@ -1,10 +1,11 @@
 {
   "name": "@furyjs/fury",
-  "version": "0.5.6-beta",
+  "version": "0.5.7-beta",
   "description": "A blazing fast multi-language serialization framework powered by jit and zero-copy",
   "main": "dist/index.js",
+  "module": "dist/esm/index.js",
   "scripts": {
-    "build": "tsc",
+    "build": "tsc -p tsconfig.json && tsc -p tsconfig.esm.json",
```
bytemain, line 8: this may produce two typings, and we indeed need one.

we can set `declaration: false` in the tsconfig.esm.json, and use:
```
    "declaration": true,
    "declarationDir": "types"
```

in the tsconfig.json.

finally, set `"types": "types/index.d.ts"` in package.json

bytemain (changes requested):

esafak:
@chaokunyang Is this OK?
@theweipeng Doesn't this need some tests?

PR #1348, opened 2024-01-19
feat(JavaScript): Support oneof

theweipeng:
### 1. Fixed a performance bug that caused writeInt32 to slow down
Before:
```JavaScript
// reader/index.ts
function writeVarInt32() {
    buffer.byteLength
}
```
After:
```JavaScript
// reader/index.ts
const byteLength = buffer.byteLength;
function writeVarInt32() {
    byteLength
}
```
The byteLength property in a Buffer is slow to access. It appears that when we access byteLength, the V8 engine processes it using a hash lookup. so we store it in closure.

### 2.  Support Oneof
Sometimes, the data we want to serialize does not have a confirmed type; instead, it could be one of several confirmed types. If we use an object to handle this situation, the size of the resulting binary will be too large, as it will contain much unused information.

usage:
```JavaScript

      const oneOfThree = Type.oneof({
          option1: Type.string(),
          option2: Type.object("foo", {
              a: Type.int32()
          }),
          option3: Type.int32(),
      });
      const { serialize, deserialize } = fury.registerSerializer(oneOfThree);
      const obj = {
          option1: "hello"
      }
      const input = serialize(obj);
      const result = deserialize(
          input
      );
      expect(result).toEqual(obj.option1)
```

### DIFF
--- a/javascript/packages/fury/lib/gen/any.ts
+++ b/javascript/packages/fury/lib/gen/any.ts
@@ -124,13 +124,13 @@ class AnySerializerGenerator extends BaseSerializerGenerator {
 
     const declare = `
       const readInner = (fromRef) => {
-        throw new Error("Type oneof readInner can't call directly");
+        throw new Error("Type Any readInner can't call directly");
       };
       const read = () => {
         ${this.toReadEmbed(expr => `return ${expr}`)}
       };
       const writeInner = (v) => {
-        throw new Error("Type oneof writeInner can't call directly");
+        throw new Error("Type Any writeInner can't call directly");
       };
       const write = (v) => {
         ${this.toWriteEmbed("v")}
--- a/javascript/packages/fury/lib/gen/any.ts
+++ b/javascript/packages/fury/lib/gen/any.ts
@@ -94,12 +94,12 @@ class AnySerializerGenerator extends BaseSerializerGenerator {
     this.description = description;
   }
 
-  writeStmt(accessor: string): string {
-    return `${this.builder.furyName()}.anySerializer.writeInner(${accessor})`;
+  writeStmt(): string {
+    throw new Error("Type Any writeStmt can't inline");
   }
 
-  readStmt(accessor: (expr: string) => string): string {
-    return `${this.builder.furyName()}.anySerializer.readInner(${accessor})`;
+  readStmt(): string {
+    throw new Error("Type Any readStmt can't inline");
   }
 
   toReadEmbed(accessor: (expr: string) => string, excludeHead = false): string {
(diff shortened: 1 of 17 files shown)


bytemain:
I think here has a wrong behavior, the deserialize result should has a kind property also, which indicates that the kind of the raw obj

theweipeng:
> I think here has a wrong behavior, the deserialize result should has a kind property also, which indicates that the kind of the raw obj

The protocol natively supports polymorphism, making a 'kind' field to describe the object type unnecessary.

bytemain:
ok, It still can meet my needs.

javascript/packages/fury/lib/gen/any.ts
```diff
@@ -115,6 +115,41 @@ class AnySerializerGenerator extends BaseSerializerGenerator {
...
+
+  toSerializer() {
+    this.scope.assertNameNotDuplicate("read");
+    this.scope.assertNameNotDuplicate("readInner");
+    this.scope.assertNameNotDuplicate("write");
+    this.scope.assertNameNotDuplicate("writeInner");
+
+    const declare = `
+      const readInner = (fromRef) => {
+        throw new Error("Type oneof readInner can't call directly");
```
bytemain, line 127: here maybe you mean `Type any`?

javascript/packages/fury/lib/gen/oneof.ts
```diff
@@ -0,0 +1,115 @@
...
+                ${this.builder.writer.int8(RefFlags.NullFlag)};
+            }`,
+    ];
+    Object.entries(this.description.options.inner).forEach(([key, value]) => {
+      const InnerGeneratorClass = CodegenRegistry.get(value.type);
+      if (!InnerGeneratorClass) {
+        throw new Error(`${value.type} generator not exists`);
+      }
+      const innerGenerator = new InnerGeneratorClass(value, this.builder, this.scope);
+      stmts.push(` if (${CodecBuilder.safeString(key)} in ${accessor}) {
```
bytemain, line 61: here we implicitly implies that the user can only select the first options in the obj, but if user pass 
```
{
  option1: xxx,
  options2: xxx,
}
```

we still write these two options into the final binary result

bytemain (changes requested):

javascript/packages/fury/lib/gen/oneof.ts
theweipeng: The statements are joined by an 'else' on line 70.

bytemain (approved):

chaokunyang (approved):
LGTM

PR #1349, opened 2024-01-21
chore(java): Remove redundant code

LiangliangSui:
Both `false` and `null` will determine `ENABLE_CLASS_REGISTRATION_FORCIBLY` as `false`, so there is no need to take this extra step.


chaokunyang (approved):
LGTM

PR #1355, opened 2024-01-23
chore(JavaScript): Fix JavaScript test CI

theweipeng:
1. JavaScript codebase contains a c++ module which requires the nodejs version greater than 20, we should ignore it when the version is not satisfied to prevent it from breaking test ci
2. Remove coverage threshold  temporarily before the xlang protocol release
3. Add 14.x and 16.x to the docker envs

#1350

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.x, 20.x]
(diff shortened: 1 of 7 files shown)


bytemain (approved):

chaokunyang (approved):
LGTM

LiangliangSui:
node 16 still fails in this pr ci， see https://github.com/apache/incubator-fury/actions/runs/7628690034/job/20780349832#step:4:105

PR #1357, opened 2024-01-25
chore(JavaScript): Fix using npm run build error

LiangliangSui:
When using `npm run build` or `npm run test`, the hps module will still be compiled, resulting in an error.


theweipeng (approved):
LGTM

Issue #1364, opened 2024-02-06
[java] Java Proxy classes: Reference missing when InvocationHandler holds reference to the Proxy itself.


cn-at-osmit:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-fury/issues) and found no similar issues.

### Version

fury 0.4.1, Java 21, macOS 14.3

### Component(s)

Java

### Minimal reproduce step

Run this testcase: 

```java

	public static class MyInvocationHandler implements InvocationHandler, Serializable {

		private static final long	serialVersionUID	= 1L;

		private Callable<String>	myProxy;

		@Override
		public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
			return "Hello world from " + (myProxy == null ? "null" : proxy.getClass().getName() + "@" + System.identityHashCode(proxy));
		}

		public void setMyProxy(Callable<String> myProxy) {
			this.myProxy = myProxy;
		}

		public Callable<String> getMyProxy() {
			return myProxy;
		}
	}

	public static void main(String[] args) throws Exception {

		MyInvocationHandler handler = new MyInvocationHandler();
		@SuppressWarnings("unchecked")
		Callable<String> proxy = (Callable<String>) Proxy.newProxyInstance(Class.class.getClassLoader(), new Class<?>[] { Callable.class }, handler);
		handler.setMyProxy(proxy);

		System.out.println(proxy.call());
		assert handler.getMyProxy() == proxy;

		Fury fury = Fury.builder().withRefTracking(true).requireClassRegistration(false).build();
		byte[] serData = fury.serialize(proxy);
		@SuppressWarnings("unchecked")
		Callable<String> proxyDeser = (Callable<String>) fury.deserialize(serData);

		System.out.println(proxyDeser.call());
		MyInvocationHandler handlerDeser = (MyInvocationHandler) Proxy.getInvocationHandler(proxyDeser);
		if (handlerDeser.getMyProxy() != proxyDeser) {
			throw new IllegalStateException("Inner reference not set");
		}
	}
```

### What did you expect to see?

I expect the `myProxy` object to have a reference to the proxy instance after deserialization

### What did you see instead?

A `null` value

### Anything Else?

PR is on the way...

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

chaokunyang:
Thanks for reporting this issue. This is tricky. 

[...]

PR #1365, opened 2024-02-06
fix(java): References within InvocationHandler, Issue #1364

cn-at-osmit:
This is the fix for the issue #1364 including a test case ;)

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
@@ -34,15 +34,20 @@ import org.apache.fury.util.ReflectionUtils;
 public class JdkProxySerializer extends Serializer {
 
   // Make offset compatible with graalvm native image.
-  private static final long PROXY_HANDLER_FIELD_OFFSET =
-      Platform.objectFieldOffset(Proxy.class, InvocationHandler.class);
-  private static final InvocationHandler STUB_HANDLER =
-      new InvocationHandler() {
-        @Override
-        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
-          throw new IllegalStateException("Deserialization stub handler still active");
-        }
-      };
+  private static final long PROXY_HANDLER_FIELD_OFFSET;
+  private static final InvocationHandler STUB_HANDLER;
+
+  static {
+    // Make offset compatible with graalvm native image.
+    PROXY_HANDLER_FIELD_OFFSET = Platform.objectFieldOffset(Proxy.class, InvocationHandler.class);
+    STUB_HANDLER =
+        new InvocationHandler() {
+          @Override
+          public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+            throw new IllegalStateException("Deserialization stub handler still active");
+          }
+        };
+  }
 
   public JdkProxySerializer(Fury fury, Class cls) {
     super(fury, cls);
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
@@ -20,11 +20,11 @@
 package org.apache.fury.serializer;
 
 import java.lang.reflect.InvocationHandler;
-import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import org.apache.fury.Fury;
 import org.apache.fury.memory.MemoryBuffer;
 import org.apache.fury.resolver.RefResolver;
+import org.apache.fury.util.GraalvmSupport;
 import org.apache.fury.util.Platform;
 import org.apache.fury.util.Preconditions;
 import org.apache.fury.util.ReflectionUtils;
@@ -37,16 +37,23 @@ public class JdkProxySerializer extends Serializer {
   private static final long PROXY_HANDLER_FIELD_OFFSET;
 
   static {
-    // Make offset compatible with graalvm native image.
-    PROXY_HANDLER_FIELD_OFFSET = Platform.objectFieldOffset(Proxy.class, InvocationHandler.class);
+    if (GraalvmSupport.isGraalBuildtime()) {
+      try {
+        // Make offset compatible with graalvm native image.
+        PROXY_HANDLER_FIELD_OFFSET = Platform.objectFieldOffset(Proxy.class.getDeclaredField("h"));
+      } catch (NoSuchFieldException e) {
+        throw new RuntimeException(e);
+      }
+    } else {
+      // not all JVM implementations use 'h' as internal InvocationHandler name
+      PROXY_HANDLER_FIELD_OFFSET =
+          ReflectionUtils.getFieldOffset(Proxy.class, InvocationHandler.class);
+    }
   }
 
   private static final InvocationHandler STUB_HANDLER =
-      new InvocationHandler() {
-        @Override
-        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
-          throw new IllegalStateException("Deserialization stub handler still active");
-        }
+      (proxy, method, args) -> {
+        throw new IllegalStateException("Deserialization stub handler still active");
       };
 
   public JdkProxySerializer(Fury fury, Class cls) {
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
@@ -20,6 +20,7 @@
 package org.apache.fury.serializer;
 
 import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import org.apache.fury.Fury;
 import org.apache.fury.memory.MemoryBuffer;
@@ -33,16 +34,15 @@ import org.apache.fury.util.ReflectionUtils;
 public class JdkProxySerializer extends Serializer {
 
   // Make offset compatible with graalvm native image.
-  private static final long PROXY_HANDLER_FIELD_OFFSET;
-
-  static {
-    try {
-      // Make offset compatible with graalvm native image.
-      PROXY_HANDLER_FIELD_OFFSET = Platform.objectFieldOffset(Proxy.class.getDeclaredField("h"));
-    } catch (NoSuchFieldException e) {
-      throw new RuntimeException(e);
-    }
-  }
+  private static final long PROXY_HANDLER_FIELD_OFFSET =
+      Platform.objectFieldOffset(Proxy.class, InvocationHandler.class);
+  private static final InvocationHandler STUB_HANDLER =
+      new InvocationHandler() {
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+          throw new IllegalStateException("Deserialization stub handler still active");
+        }
+      };
 
   public JdkProxySerializer(Fury fury, Class cls) {
     super(fury, cls);
@@ -63,8 +63,7 @@ public class JdkProxySerializer extends Serializer {
     final int refId = resolver.lastPreservedRefId();
     final Class<?>[] interfaces = (Class<?>[]) fury.readRef(buffer);
     Preconditions.checkNotNull(interfaces);
-    final Class<?> proxyClass = Proxy.getProxyClass(fury.getClassLoader(), interfaces);
-    Object proxy = Platform.newInstance(proxyClass);
+    Object proxy = Proxy.newProxyInstance(fury.getClassLoader(), interfaces, STUB_HANDLER);
     resolver.setReadObject(refId, proxy);
     InvocationHandler invocationHandler = (InvocationHandler) fury.readRef(buffer);
     Preconditions.checkNotNull(invocationHandler);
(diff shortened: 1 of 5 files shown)


java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
```diff
@@ -23,13 +23,27 @@
...
 
+  // Make offset compatible with graalvm native image.
+  private static final long PROXY_HANDLER_FIELD_OFFSET;
+
+  static {
+    try {
+      // Make offset compatible with graalvm native image.
+      PROXY_HANDLER_FIELD_OFFSET = Platform.objectFieldOffset(Proxy.class.getDeclaredField("h"));
+    } catch (NoSuchFieldException e) {
+      throw new RuntimeException(e);
```
chaokunyang, line 43: The name `h` is not a common abbr, could we use offset from a field with type `InvocationHandler` in catch block? We can add a method in `ReflectionUtils` which check a class has only one filed with specified type an return the offset of that type

java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
```diff
@@ -39,17 +53,23 @@ public JdkProxySerializer(Fury fury, Class cls) {
...
   @Override
   public Object read(MemoryBuffer buffer) {
-    InvocationHandler invocationHandler = (InvocationHandler) fury.readRef(buffer);
-    Preconditions.checkNotNull(invocationHandler);
+    final RefResolver resolver = fury.getRefResolver();
+    final int refId = resolver.lastPreservedRefId();
     final Class<?>[] interfaces = (Class<?>[]) fury.readRef(buffer);
     Preconditions.checkNotNull(interfaces);
-    return Proxy.newProxyInstance(fury.getClassLoader(), interfaces, invocationHandler);
+    final Class<?> proxyClass = Proxy.getProxyClass(fury.getClassLoader(), interfaces);
```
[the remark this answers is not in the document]
chaokunyang: After a detailed look at JDK `Proxy` code, seems there is no easy way to workaround this.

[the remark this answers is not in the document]
marvkis: What do you think about initializing the proxy with `Proxy.newProxyInstance(...)` but using a dummy / static *invocationHandler*? The invocation handler would be changed later, after the real invocationHandler is ready. This way we could get rid of the deprecation warning.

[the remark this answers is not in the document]
cn-at-osmit: Sorry, used had the wrong account active. marvkis is my private/personal account, this is my company account ;)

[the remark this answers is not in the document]
chaokunyang: Good point, this works:
```java
    RefResolver refResolver = fury.getRefResolver();
    int id = refResolver.lastPreservedRefId();
    final Class<?>[] interfaces = (Class<?>[]) fury.readRef(buffer);
    Preconditions.checkNotNull(interfaces);
    Object proxy = Proxy.newProxyInstance(fury.getClassLoader(), interfaces, STUB_HANDLER);
    refResolver.setReadObject(id, proxy);
    // create InvocationHandler again.
    Object handler = fury.readRef(buffer);
    Platform.putObject(proxy, PROXY_HANDLER_FIELD_OFFSET, handler);
    return proxy;
  }
```

java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
cn-at-osmit: I tried to implement it this way in my latest commit ;)

java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
```diff
@@ -20,16 +20,30 @@
...
+import org.apache.fury.util.Platform;
 import org.apache.fury.util.Preconditions;
 import org.apache.fury.util.ReflectionUtils;
 
 /** Serializer for jdk {@link Proxy}. */
 @SuppressWarnings({"rawtypes", "unchecked"})
 public class JdkProxySerializer extends Serializer {
 
+  // Make offset compatible with graalvm native image.
+  private static final long PROXY_HANDLER_FIELD_OFFSET =
```
chaokunyang, line 37: The try catch block is still needed, graalvm only capture this pattern and replace the offset. This invocation can be placed in catch block

marvkis: Should I leave the InvocationHandler stub initialization as is or move it to the static block as well? I'm new to grallvm but it's on my todo list 👍

cn-at-osmit: Should I leave the InvocationHandler stub initialization as is or move it to the static block as well? I'm new to grallvm but it's on my todo list 👍

chaokunyang: Leave it as it looks good to me

cn-at-osmit: done 👍

cn-at-osmit: Okay, I just stumbled over the try...catch mention. Is a try / catch really necessary or is the static block okay? 🤔

chaokunyang: Some jdk may have different field name for `InvocationHandler` instead of using `h`

java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
```diff
@@ -20,16 +20,35 @@
...
 
+  // Make offset compatible with graalvm native image.
+  private static final long PROXY_HANDLER_FIELD_OFFSET;
+
+  static {
+    // Make offset compatible with graalvm native image.
+    PROXY_HANDLER_FIELD_OFFSET = Platform.objectFieldOffset(Proxy.class, InvocationHandler.class);
+  }
+
+  private static final InvocationHandler STUB_HANDLER =
```
chaokunyang, line 44: Using lambda is more concise:
```java
  private static final InvocationHandler STUB_HANDLER =
    (proxy, method, args) -> {
      throw new IllegalStateException("Deserialization stub handler still active");
    };
```

java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
cn-at-osmit: I'm asking because this new [Platform.objectFieldOffset(clazz, fieldType)](https://github.com/apache/incubator-fury/pull/1365/files#diff-7675c883d6e1c72d376ffb7941552f0b28f6821b491bce97d2838bce9bd8992cR155) variant doesn't throw a `NoSuchFieldException`

chaokunyang: Or we can leave it in another PR.

If we implement in this PR, we may need an another inner class to hold offset like:
```java
 private static class Offset {
    private static final long ARRAY_LIST_SIZE_OFFSET;
    private static final long ARRAY_LIST_ARRAY_OFFSET;

    static {
      try {
        ARRAY_LIST_SIZE_OFFSET =
            Platform.objectFieldOffset(ArrayList.class.getDeclaredField("size"));
        ARRAY_LIST_ARRAY_OFFSET =
            Platform.objectFieldOffset(ArrayList.class.getDeclaredField("elementData"));
      } catch (NoSuchFieldException e) {
      }
    }
  }
```

Then use that offset in `org.apache.fury.serializer.JdkProxySerializer#read` with a try catch block.

chaokunyang: > I'm asking because this new [Platform.objectFieldOffset(clazz, fieldType)](https://github.com/apache/incubator-fury/pull/1365/files#diff-7675c883d6e1c72d376ffb7941552f0b28f6821b491bce97d2838bce9bd8992cR155) variant doesn't throw a `NoSuchFieldException`

This pattern can be captured at graalvm native image build time. It should be used only in fallback block.

chaokunyang: see more in:
-  https://developers.redhat.com/articles/2022/05/09/using-unsafe-safely-graalvm-native-image#unsafe_proves_to_be_unsafe
- https://www.graalvm.org/latest/reference-manual/native-image/dynamic-features/Reflection/#unsafe-accesses

java/fury-core/src/main/java/org/apache/fury/util/Platform.java
```diff
@@ -143,6 +143,40 @@ public static long objectFieldOffset(Field f) {
...
+  /**
+   * Gets the offset of the only field in the class matching the required type.
+   *
+   * @param clazz the class in which the field should be declared
+   * @param fieldType the required type of the field
+   * @return offset of the field
+   * @throws IllegalStateException if there are multiple fields of the required type
+   * @throws IllegalArgumentException if there are no fields of the required type
+   */
+  public static long objectFieldOffset(Class<?> clazz, Class<?> fieldType) {
```
chaokunyang, line 155: Could we move this method to `ReflectionUtils`? Platform is a more internal API which expose jdk internals. We try to make it simple and leave more logic under ReflectioUtils

java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
cn-at-osmit: Do you have something like this in mind?
```java
static {
    try {
      // Make offset compatible with graalvm native image.
      PROXY_HANDLER_FIELD_OFFSET =
          Platform.objectFieldOffset(Proxy.class.getDeclaredField("h"));
    } catch (NoSuchFieldException e) {
      PROXY_HANDLER_FIELD_OFFSET = Platform.objectFieldOffset(Proxy.class, InvocationHandler.class);
    }
}
```
But I think i have to work with a temp assignment...

cn-at-osmit: This compiles but I'm not sure if it works for graalvm?
```java
  static {
    long proxyHanlderFieldOffset;
    try {
      // Make offset compatible with graalvm native image.
      proxyHanlderFieldOffset = Platform.objectFieldOffset(Proxy.class.getDeclaredField("h"));
    } catch (NoSuchFieldException e) {
      proxyHanlderFieldOffset = Platform.objectFieldOffset(Proxy.class, InvocationHandler.class);
    }
    PROXY_HANDLER_FIELD_OFFSET = proxyHanlderFieldOffset;
  }
```

java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
cn-at-osmit: done

java/fury-core/src/main/java/org/apache/fury/util/Platform.java
cn-at-osmit: done

java/fury-core/src/main/java/org/apache/fury/serializer/JdkProxySerializer.java
cn-at-osmit: Okay, I studied the redhat documentation in the link provided. So it seems for graalvm we assume that the name of the field is alway's 'h' and for other JVM's it could differ.

I updated the PR with this variant:
```java
  static {
    if (GraalvmSupport.isGraalBuildtime()) {
      try {
        // Make offset compatible with graalvm native image.
        PROXY_HANDLER_FIELD_OFFSET = Platform.objectFieldOffset(Proxy.class.getDeclaredField("h"));
      } catch (NoSuchFieldException e) {
      }
    } else {
      // not all JVM implementations use 'h' as internal InvocationHandler name
      PROXY_HANDLER_FIELD_OFFSET =
          ReflectionUtils.getFieldOffset(Proxy.class, InvocationHandler.class);
    }
  }
```
My intention is to use the graalvm 'automatic fix' pattern when it's executed during graalvm build time and the other variant when executed during normal JVM execution.
Could this work?

chaokunyang: I'm not sure, may need a test. Could we add an test in integration_tests/graalvm_tests?
FYI, I tested following pattern, it works too:
```java
  private static final Field field;
  private static final long offset;

  static {
    field = ReflectionUtils.getField(Example.class, "f2");
    offset = Platform.objectFieldOffset(field);
  }
```

chaokunyang: Proxy in graalvm is a little complex, may need to add a json config file: https://www.graalvm.org/latest/reference-manual/native-image/guides/configure-dynamic-proxies/

chaokunyang:
Hi @cn-at-osmit , thanks for contribute this fix, I'll merge it first and add graalvm tests in a following-up PR

chaokunyang (approved):
LGTM

Issue #1363, opened 2024-02-05
[Java] CodeGen Issue: Assignment conversion not possible from type "java.lang.Object" to type "org.zkoss.zhtml.Object"


cn-at-osmit:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-fury/issues) and found no similar issues.

### Version

fury 0.4.1, Java 21, macOS 14.3

### Component(s)

Java

### Minimal reproduce step

Import these dependencies in pom.xml:
```xml
	<dependencies>
		<dependency>
			<groupId>org.furyio</groupId>
			<artifactId>fury-core</artifactId>
			<version>0.4.1</version>
		</dependency>
		<dependency>
			<groupId>org.zkoss.zk</groupId>
			<artifactId>zhtml</artifactId>
			<version>9.6.0.2-jakarta</version>
		</dependency>
		<!-- Requried to make slf4j and to see the stacktrace -->
		<dependency>
			<groupId>org.slf4j</groupId>
			<artifactId>slf4j-simple</artifactId>
			<version>1.7.30</version>
		</dependency>
	</dependencies>
```
And start codegen for any class from `org.zkoss.zhtml`:
```java
  Fury fury = Fury.builder().build();
  fury.register(org.zkoss.zhtml.Text.class, true);
```

### What did you expect to see?

A working fury instance ;)

### What did you see instead?

A stacktrace:
```
[main] INFO io.fury.Fury - Created new fury io.fury.Fury@1e965684
[main] WARN io.fury.serializer.ObjectStreamSerializer - class org.zkoss.zhtml.Text customized jdk serialization, which is inefficient. Please replace it with a io.fury.serializer.Serializer or implements java.io.Externalizable
[main] INFO io.fury.codegen.CompileUnit - Generate code for org.zkoss.zhtml.TextFuryCompatibleCodec_1_414493378_1883919084 took 75 ms.
Exception in thread "main" io.fury.codegen.CodegenException: Compile error: 
org.zkoss.zhtml.TextFuryCompatibleCodec_1_414493378_1883919084:
/* 0001 */ package org.zkoss.zhtml;
/* 0002 */ 
/* 0003 */ import java.util.List;
/* 0004 */ import java.util.Map;
/* 0005 */ import java.util.Set;
/* 0006 */ import io.fury.Fury;
/* 0007 */ import io.fury.memory.MemoryBuffer;
/* 0008 */ import io.fury.resolver.NoRefResolver;
[...]

PR #1366, opened 2024-02-06
fix(java): CodeGen: Name conflicts when omitting java.lang prefix #1363

cn-at-osmit:
Hi. This is my initial approach to provide a PR for fixing #1363 

I'm not sure what you mean
> Maybe we should move it to CodeGenerator
as the CodeGenerator is not available in `CodegenContext`. So for the first approach I use the same way to determine the classloader as in the sample.

As we only have 'java.lang' classes here - wouldn't the classloader of java.lang.X  always be null?

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/codegen/CodegenContext.java
+++ b/java/fury-core/src/main/java/org/apache/fury/codegen/CodegenContext.java
@@ -307,7 +307,7 @@ public class CodegenContext {
                     return Boolean.FALSE;
                   }
                 });
-        return conflictRes.booleanValue() ? clz.getName() : simpleName;
+        return conflictRes ? clz.getName() : simpleName;
       }
     }
     if (imports.contains(type)) {
--- a/java/fury-core/src/main/java/org/apache/fury/codegen/CodegenContext.java
+++ b/java/fury-core/src/main/java/org/apache/fury/codegen/CodegenContext.java
@@ -289,8 +289,9 @@ public class CodegenContext {
     if (type.startsWith("java.lang")) {
       if (!type.substring("java.lang.".length()).contains(".")) {
         String simpleName = clz.getSimpleName();
+        boolean hasPackage = StringUtils.isNotBlank(pkg);
         Map<String, Boolean> packageMap =
-            nameConflicts.computeIfAbsent(pkg, p -> new ConcurrentHashMap<>());
+            nameConflicts.computeIfAbsent(hasPackage ? pkg : "", p -> new ConcurrentHashMap<>());
         Boolean conflictRes =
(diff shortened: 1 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/codegen/CodegenContext.java
```diff
@@ -285,7 +288,26 @@ public String type(Class<?> clz) {
...
+                        clz.getClassLoader() == null
+                            ? Thread.currentThread().getContextClassLoader()
+                            : clz.getClassLoader();
+                    beanClassClassLoader.loadClass(hasPackage ? pkg + "." + sn : sn);
+                    return Boolean.TRUE;
+                  } catch (ClassNotFoundException e) {
+                    return Boolean.FALSE;
+                  }
+                });
+        return conflictRes.booleanValue() ? clz.getName() : simpleName;
```
chaokunyang, line 310:
```suggestion
        return conflictRes ? clz.getName() : simpleName;
```

Unbox is unnecessary

chaokunyang (approved):
LGTM, thanks @cn-at-osmit.

chaokunyang:
> Hi. This is my initial approach to provide a PR for fixing #1363
> 
> I'm not sure what you mean
> 
> > Maybe we should move it to CodeGenerator
> > as the CodeGenerator is not available in `CodegenContext`. So for the first approach I use the same way to determine the classloader as in the sample.
> 
> As we only have 'java.lang' classes here - wouldn't the classloader of java.lang.X always be null?

You are right, the classloader of java.lang.X is always null. There are many places which uses `Thread.currentThread().getContextClassLoader()` ,  but sometimes the context classloader is null. We may need to create a method to return a not null classloader. But this not related to this PR, we can do it in a new PR

cn-at-osmit:
Regarding the classloder: What do you think, would this be a good approach:

```java
String className = hasPackage ? pkg + "." + sn : sn)
ClassLoader tcl =Thread.currentThread().getContextClassLoader();
if (tcl == null)
   Class.forName(className);
else
   tcl.loadClass(className);
```
This would solve the issue when the Thread's classloader is null. Shall I update the PR with this?

chaokunyang:
> Regarding the classloder: What do you think, would this be a good approach:
> 
> ```java
> String className = hasPackage ? pkg + "." + sn : sn)
> ClassLoader tcl =Thread.currentThread().getContextClassLoader();
> if (tcl == null)
>    Class.forName(className);
> else
>    tcl.loadClass(className);
> ```
> 
> This would solve the issue when the Thread's classloader is null. Shall I update the PR with this?

We can update all related code in a next PR

PR #1372, opened 2024-02-14
fix(Rust): Fix memory errors caused by casting

theweipeng:
In the stable version of Rust, the `from_raw_parts_mut` API does not check for memory alignment, but in the 1.78.0-nightly version, alignment checking has been added to the preconditions.  To ensure memory safety, we have replaced direct memory writes with the `LittleEndian::write_xx` methods from the appropriate library. While this may introduce additional overhead for memory checks, robustness is the priority before releasing the software.

### DIFF
--- a/rust/fury/src/buffer.rs
+++ b/rust/fury/src/buffer.rs
@@ -15,27 +15,12 @@
 // specific language governing permissions and limitations
 // under the License.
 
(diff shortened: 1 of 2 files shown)


chaokunyang (approved):
LGTM

PR #1373, opened 2024-02-18
feat(JavaScript): Make PlatformBuffer available if has Buffer polyfill

bytemain:
If user has a polyfill(process/buffer) in browser, furyjs cannot work.


javascript/packages/fury/lib/platformBuffer.ts
```diff
@@ -17,23 +17,38 @@
...
+    return this.utf8Slice(start, end);
+  }
+
   static alloc(size: number) {
     return new BrowserBuffer(new Uint8Array(size));
   }
 
   latin1Write(string: string, offset: number) {
-    for (let index = 0; index < string.length; index++) {
+    let index = 0;
```
bytemain, line 50: here I originally want return the index, because the `Buffer.prototype.write` will return the writes counts

bytemain: but i found it not necessary to do this, so i remove the return statement

theweipeng (approved):
LGTM

javascript/benchmark/platform-buffer-draw.py
```diff
@@ -0,0 +1,102 @@
...
+
+# 绘制第二部分比较：browser write 和 native write
+axs[1].bar(
+    ["browser write", "native write"], [browser_write_1, native_write], color=["g", "r"]
+)
+axs[1].set_title("Browser Write vs Native Write")
+axs[1].set_xlabel("Operation Type")
+axs[1].set_xticklabels(["browser write", "native write"])
+
+# 在柱形图上添加数值标签
```
chaokunyang, line 65: Could we change those comments to english?

bytemain: ok, I will send a pr later

PR #1382, opened 2024-02-20
chore(java): Extract BaseFury interface

LiangliangSui:
Extract the BaseFury interface based on Fury and ThreadSafeFury, and extract the common methods in Fury and ThreadSafeFury into BaseFury.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/BaseFury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/BaseFury.java
@@ -1,0 +1,114 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/BaseFury.java
```diff
@@ -0,0 +1,117 @@
...
+  <T> T deserializeJavaObject(byte[] data, Class<T> cls);
+
+  /**
+   * Deserialize java object from binary by passing class info, serialization should use {@link
+   * #serializeJavaObject}.
+   */
+  <T> T deserializeJavaObject(MemoryBuffer buffer, Class<T> cls);
+
+  /** Returns classLoader of serializer. */
+  ClassLoader getClassLoader();
```
chaokunyang, line 116: Could we remove this API, it's unstable yet

LiangliangSui: Move these two APIs back to `Fury` and `ThreadSafeFury`?

LiangliangSui: Maybe move `ClassLoader getClassLoader()` this API back to `Fury` and `ThreadSafeFury`.

chaokunyang: Yep, let's do this

LiangliangSui: Done

chaokunyang (approved):
LGTM

PR #1392, opened 2024-03-01
test(java): Add test for ClassLoaderFuryPooled

LiangliangSui:
Add test for `ClassLoaderFuryPooled`.

### DIFF
--- a/java/fury-core/src/test/java/org/apache/fury/pool/ClassLoaderFuryPooledTest.java
+++ b/java/fury-core/src/test/java/org/apache/fury/pool/ClassLoaderFuryPooledTest.java
@@ -1,0 +1,127 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


LiangliangSui:
@chaokunyang ptal, thanks!

java/fury-core/src/test/java/org/apache/fury/pool/ClassLoaderFuryPooledTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+  @Test
+  public void testReturnFury() {
+    try {
+      Function<ClassLoader, Fury> furyFactory = getFuryFactory();
+      Fury fury = furyFactory.apply(getClass().getClassLoader());
+      ClassLoaderFuryPooled pooled = getPooled(4, 8, furyFactory);
+      pooled.returnFury(fury);
+    } catch (RuntimeException e) {
+      Assert.fail();
```
chaokunyang, line 122: Seems we can just remove catch block

LiangliangSui: Updated

chaokunyang (approved):
LGTM

PR #1395, opened 2024-03-04
chore(java): Use existing API to streamline code

LiangliangSui:


LiangliangSui:
@chaokunyang ptal, thanks!

PR #1396, opened 2024-03-06
refactor(java): Improve code readability and fix an issue

LiangliangSui:
1. Refactor the code structure to increase code readability.

2. Fix a problem. When `classInfo == null || typeTag != null || classId != classInfo.classId` is true, and already registered, we should also update `registeredId2ClassInfo`.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -714,12 +714,15 @@ public class ClassResolver {
         typeTagToClassXLangMap.put(typeTag, type);
       }
     }
+
+    // 1. Try to get ClassInfo from `registeredId2ClassInfo` and
+    // `classInfoMap` or create a new `ClassInfo`.
     ClassInfo classInfo;
     Short classId = extRegistry.registeredClassIdMap.get(type);
+    boolean registered = classId != null;
     // set serializer for class if it's registered by now.
-    if (classId != null) {
+    if (registered) {
       classInfo = registeredId2ClassInfo[classId];
-      classInfo.serializer = serializer;
     } else {
       if (serializer instanceof ReplaceResolveSerializer) {
         classId = REPLACE_STUB_ID;
@@ -728,13 +731,17 @@ public class ClassResolver {
       }
       classInfo = classInfoMap.get(type);
     }
+
     if (classInfo == null || typeTag != null || classId != classInfo.classId) {
-      classInfo = new ClassInfo(this, type, typeTag, serializer, classId);
-    } else {
-      classInfo.serializer = serializer;
+      classInfo = new ClassInfo(this, type, typeTag, null, classId);
+      classInfoMap.put(type, classInfo);
+      if (registered) {
+        registeredId2ClassInfo[classId] = classInfo;
+      }
     }
-    // make `extRegistry.registeredClassIdMap` and `classInfoMap` share same classInfo instances.
-    classInfoMap.put(type, classInfo);
+
+    // 2. Set `Serializer` for `ClassInfo`.
+    classInfo.serializer = serializer;
   }
 
   @SuppressWarnings("unchecked")


java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -728,13 +731,17 @@ private void addSerializer(Class<?> type, Serializer<?> serializer) {
...
-      classInfo = new ClassInfo(this, type, typeTag, serializer, classId);
-    } else {
-      classInfo.serializer = serializer;
+      classInfo = new ClassInfo(this, type, typeTag, null, classId);
+      classInfoMap.put(type, classInfo);
+      if (registered) {
+        registeredId2ClassInfo[classId] = classInfo;
+      }
     }
-    // make `extRegistry.registeredClassIdMap` and `classInfoMap` share same classInfo instances.
```
LiangliangSui, line 736: Here it should be that `classInfoMap` and `registeredId2ClassInfo` share the same ClassInfo.

The ClassInfo in `registeredId2ClassInfo` and `classInfoMap` should always be shared. The `ClassInfo` obtained in `registeredId2ClassInfo` can be the same in `classInfoMap`, so there should be redundancy here, only whether to create ClassInfo to `classInfoMap `Just put it in

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -728,13 +731,17 @@ private void addSerializer(Class<?> type, Serializer<?> serializer) {
...
     }
+
     if (classInfo == null || typeTag != null || classId != classInfo.classId) {
-      classInfo = new ClassInfo(this, type, typeTag, serializer, classId);
-    } else {
-      classInfo.serializer = serializer;
+      classInfo = new ClassInfo(this, type, typeTag, null, classId);
+      classInfoMap.put(type, classInfo);
+      if (registered) {
+        registeredId2ClassInfo[classId] = classInfo;
```
LiangliangSui, line 739: `registered == true` means that ClassInfo has been re-created, so `registeredId2ClassInfo` needs to be updated.

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -728,13 +731,17 @@ private void addSerializer(Class<?> type, Serializer<?> serializer) {
       }
       classInfo = classInfoMap.get(type);
     }
+
     if (classInfo == null || typeTag != null || classId != classInfo.classId) {
-      classInfo = new ClassInfo(this, type, typeTag, serializer, classId);
-    } else {
-      classInfo.serializer = serializer;
+      classInfo = new ClassInfo(this, type, typeTag, null, classId);
```
LiangliangSui, line 736: Perhaps when `typeTag != null`, we do not need to re-create `ClassInfo`, we can just reassign the `typeTagBytes` in `ClassInfo`, which can save the time of creating `ClassInfo` object.

LiangliangSui:
@chaokunyang ptal, thanks!

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -714,12 +714,15 @@ private void addSerializer(Class<?> type, Serializer<?> serializer) {
...
+    // 1. Try to get ClassInfo from `registeredId2ClassInfo` and
+    // `classInfoMap` or create a new `ClassInfo`.
     ClassInfo classInfo;
     Short classId = extRegistry.registeredClassIdMap.get(type);
+    boolean registered = classId != null;
     // set serializer for class if it's registered by now.
-    if (classId != null) {
+    if (registered) {
       classInfo = registeredId2ClassInfo[classId];
-      classInfo.serializer = serializer;
```
chaokunyang, line 722: Why remove this line

chaokunyang (approved):
LGTM

LiangliangSui:
BTW, `ClassResolverTest` has low test coverage for `ClassResolver`, so can we add some test cases to `ClassResolverTest`?

PR #1397, opened 2024-03-06
feat(java): avoid big object graph cause buffer take up too much memory

chaokunyang:
There has been cases where Fury use too much memory:
image

This is caused by Fury held buffer. Fury holds a MemoryBuffer and write to it when users doesn't provide its own buffer. This held buffer can save memory copy since we don't have to grow buffer every time.  But Fury never clear this buffer. If a big object tree is passed, the held buffer will take much memory even later object tree is small. 

This PR fix it by reset buffer when it's lager than 128Kb.


theweipeng (approved):
LGTM

PR #1398, opened 2024-03-08
fix(java): Blacklist detection is also performed when the Class is registered.

LiangliangSui:
1. When a Class has been registered using `Fury#register`, it currently escapes blacklist detection, so correct it to always perform blacklist detection.

2. `FuryBuilder#suppressClassRegistrationWarnings` should remain the same as `FuryBuilder#requireClassRegistration`. The warning log will only be suppressed when `requireClassRegistration` is true, that is, it will only be output when `requireClassRegistration` is not turned on and the Class is not registered.

3. Integrate some code into `isSecure()`.


java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
```diff
@@ -303,7 +303,7 @@ private void finish() {
               + "`ClassResolver#setClassChecker`");
     }
     if (suppressClassRegistrationWarnings == null) {
-      suppressClassRegistrationWarnings = !requireClassRegistration;
+      suppressClassRegistrationWarnings = requireClassRegistration;
```
chaokunyang, line 306: This is just to set default value, we set it to true when ClassRegistration is disabled to reduce the verbose logs which may bother users.

java/fury-core/src/main/java/org/apache/fury/resolver/BlackList.java
```diff
@@ -50,7 +51,9 @@ class BlackList {
     }
   }
 
-  static Set<String> getDefaultBlackList() {
-    return DEFAULT_BLACKLIST_SET;
+  static void checkNotInBlackList(String clsName) {
```
chaokunyang, line 54: Could we document this method that it will throws exception when class is in blacklist

java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
LiangliangSui: By default ClassRegistration is enabled, so the warning log is printed.

Should we only print the warning log when ClassRegistration is disabled?

So we should keep ClassRegistration and suppressClassRegistrationWarnings(suppressClassRegistrationWarnings is true, which means the warning log is not printed.) at the same value.

java/fury-core/src/main/java/org/apache/fury/resolver/BlackList.java
LiangliangSui: Updated.

java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
chaokunyang: How about set the `suppressClassRegistrationWarnings` to true by default. Users can set it to false if they want to do some security audit

chaokunyang (commented):
LGTM overall, only some minor comments

chaokunyang (approved):

java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
LiangliangSui: Looks good to me, updated.

PR #1400, opened 2024-03-11
ci: Add arm64 macos to CI workflow

PragmaTwice:
- `macos-12`: x86
- `macos-14`: arm64

Refer to https://github.com/actions/runner-images for more details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,10 +156,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.8
+      - name: Set up Python 3.11
         uses: actions/setup-python@v2
         with:
-          python-version: 3.8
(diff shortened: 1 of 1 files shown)


PragmaTwice:
@chaokunyang Python 3.8 seems to be too old to have an arm64 darwin version. I think maybe we can try 3.11?

https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

PragmaTwice:
> > @chaokunyang Python 3.8 seems to be too old to have an arm64 darwin version. I think maybe we can try 3.11?
> > https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
> 
> Can we use python 3.11 for c++/rust, and keep using current python version matrix for python test, and add an extra python 3.11 arm ci? Python 3.8 is used commonly, we should not skip it in ci

Done. Now it is ready for review.

chaokunyang (approved):
LGTM

PR #1406, opened 2024-03-13
refactor(java): Optimize Class ID allocation.

LiangliangSui:
1. `registeredClassIdCounter` is incremented every time `register(Class<?> cls, int classId)` is successfully called, so it can be used directly in `register(Class<?> cls)` without recalculation.

2. Delete `registeredId2Classes` and use `registeredId2ClassInfo` to achieve the corresponding function.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -228,6 +228,8 @@ public class ClassResolver {
   private final ShimDispatcher shimDispatcher;
 
   private static class ExtRegistry {
+    // Here we set it to 1 because `NO_CLASS_ID` is 0 to avoid calculating it again in
+    // `register(Class<?> cls)`.
     private short registeredClassIdCounter = 1;
     private SerializerFactory serializerFactory;
     private final IdentityMap<Class<?>, Short> registeredClassIdMap =
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -88,7 +88,6 @@ import org.apache.fury.codegen.Expression.Invoke;
 import org.apache.fury.codegen.Expression.Literal;
 import org.apache.fury.collection.IdentityMap;
 import org.apache.fury.collection.IdentityObjectIntMap;
-import org.apache.fury.collection.LongMap;
 import org.apache.fury.collection.ObjectMap;
 import org.apache.fury.collection.Tuple2;
 import org.apache.fury.config.CompatibleMode;
@@ -229,8 +228,7 @@ public class ClassResolver {
   private final ShimDispatcher shimDispatcher;
 
   private static class ExtRegistry {
-    private short registeredClassIdCounter = 0;
-    private final LongMap<Class<?>> registeredId2Classes = new LongMap<>(initialCapacity);
+    private short registeredClassIdCounter = 1;
     private SerializerFactory serializerFactory;
     private final IdentityMap<Class<?>, Short> registeredClassIdMap =
         new IdentityMap<>(initialCapacity);
@@ -296,6 +294,8 @@ public class ClassResolver {
     registerWithCheck(Class.class, CLASS_CLASS_ID);
     registerWithCheck(Object.class, EMPTY_OBJECT_ID);
     // Register common class ahead to get smaller class id for serialization.
+    // TODO(Liangliang Sui): Clean up duplicately registered Classes and throw exceptions when
+    // Classes are duplicately registered.
     registerCommonUsedClasses();
     addDefaultSerializers();
     registerDefaultClasses();
@@ -392,7 +392,8 @@ public class ClassResolver {
   /** register class. */
   public void register(Class<?> cls) {
     if (!extRegistry.registeredClassIdMap.containsKey(cls)) {
-      while (extRegistry.registeredId2Classes.containsKey(extRegistry.registeredClassIdCounter)) {
+      while (extRegistry.registeredClassIdCounter < registeredId2ClassInfo.length
+          && registeredId2ClassInfo[extRegistry.registeredClassIdCounter] != null) {
         extRegistry.registeredClassIdCounter++;
       }
       register(cls, extRegistry.registeredClassIdCounter);
@@ -439,12 +440,11 @@ public class ClassResolver {
                 "Class %s with name %s has been registered, registering class with same name are not allowed.",
                 extRegistry.registeredClasses.get(cls.getName()), cls.getName()));
       }
-      Class<?> idToClass = extRegistry.registeredId2Classes.get(id);
-      if (idToClass != null) {
+      if (id < registeredId2ClassInfo.length && registeredId2ClassInfo[id] != null) {
         throw new IllegalArgumentException(
             String.format(
                 "Class %s with id %s has been registered, registering class %s with same id are not allowed.",
-                idToClass, id, cls.getName()));
+                registeredId2ClassInfo[id].getCls(), id, cls.getName()));
       }
       extRegistry.registeredClassIdMap.put(cls, id);
       if (registeredId2ClassInfo.length <= id) {
@@ -465,7 +465,6 @@ public class ClassResolver {
       registeredId2ClassInfo[id] = classInfo;
       extRegistry.registeredClasses.put(cls.getName(), cls);
       extRegistry.registeredClassIdCounter++;
-      extRegistry.registeredId2Classes.put(id, cls);
     }
   }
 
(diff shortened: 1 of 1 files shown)


LiangliangSui:
@chaokunyang ptal, thanks!

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -229,8 +228,7 @@ public class ClassResolver {
   private final ShimDispatcher shimDispatcher;
 
   private static class ExtRegistry {
-    private short registeredClassIdCounter = 0;
-    private final LongMap<Class<?>> registeredId2Classes = new LongMap<>(initialCapacity);
+    private short registeredClassIdCounter = 1;
```
chaokunyang, line 231: why set default value to 1?

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -391,12 +391,7 @@ private void registerDefaultClasses() {
 
   /** register class. */
   public void register(Class<?> cls) {
-    if (!extRegistry.registeredClassIdMap.containsKey(cls)) {
-      while (extRegistry.registeredId2Classes.containsKey(extRegistry.registeredClassIdCounter)) {
-        extRegistry.registeredClassIdCounter++;
-      }
-      register(cls, extRegistry.registeredClassIdCounter);
-    }
+    register(cls, extRegistry.registeredClassIdCounter);
```
chaokunyang, line 394: What will happen if there are some hold in `registeredId2ClassInfo`? 
For example:
```java
// registeredClassIdCounter = 99
fury.register(class1); // registeredClassIdCounter = 100 now
fury.register(class2, 101); // registeredClassIdCounter = 101 now
fury.register(class3); // throw exception since 101 has been used
```

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
LiangliangSui: `registeredClassIdCounter` will increase automatically after each successful registration. When we use the built-in ClassID to register and use `register(Class<?> cls)` again, the `registeredClassIdCounter` can be recalculated once.

```java
// registeredClassIdCounter = 0
register(LambdaSerializer.ReplaceStub.class, LAMBDA_STUB_ID); // registeredClassIdCounter = 1 now
register(class); // registeredClassIdCounter = 1 has been registered at this time, and registeredClassIdCounter needs to be recalculated
```

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
LiangliangSui: The situation you mentioned above cannot be covered because the user can specify the classId.

Duplicate registration of ClassID will be detected.

LiangliangSui: We may be able to provide an interface to users in Fury, for example named `allocateClassId`, so that users can easily use ClassID to prevent the provided ClassID from being registered repeatedly (because the user does not know the allocation of ClassId within ClassResolver, so the ClassID provided by the user There is also the risk of being registered. ClassID is only maintained by the system, and obtaining ClassID is transparent to users).

What about this idea ?

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
chaokunyang: `NO_CLASS_ID` used `0`, is that the reason for use `1` here. Could we add a comment here

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
chaokunyang: Yes, registered class id should be exposed to users, so they can know which ids they could use to avoid the conflict

chaokunyang: > What will happen if there are some hold in `registeredId2ClassInfo`? For example:
> 
> ```java
> // registeredClassIdCounter = 99
> fury.register(class1); // registeredClassIdCounter = 100 now
> fury.register(class2, 101); // registeredClassIdCounter = 101 now
> fury.register(class3); // throw exception since 101 has been used
> ```

I think for `org.apache.fury.resolver.ClassResolver#register(java.lang.Class<?>)` , it should always succeed. The implemention shoukd handle class id conflict inside

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
LiangliangSui: > `NO_CLASS_ID` used `0`, is that the reason for use `1` here? Could we add a comment here?

Yes, Comment added.

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
LiangliangSui: > Yes, registered class id should be exposed to users, so they can know which ids they could use to avoid the conflict

We can abstract a separate class for ClassID allocation, provide an interface for users to obtain ClassID, and then reference this class in `ClassResolver` (Currently, the ClassResolver class is responsible for a lot of content, and this can also reduce the burden on the `ClassResolver` class).

In this way, we have one more class(For example named `ClassIdAllocator`) responsible for maintaining ClassID allocation, which can be used internally (`ClassResolver`) or externally (`Fury#allocateClassId`).

LiangliangSui: > > What will happen if there are some hold in `registeredId2ClassInfo`? For example:
> > ```java
> > // registeredClassIdCounter = 99
> > fury.register(class1); // registeredClassIdCounter = 100 now
> > fury.register(class2, 101); // registeredClassIdCounter = 101 now
> > fury.register(class3); // throw exception since 101 has been used
> > ```
> 
> I think for `org.apache.fury.resolver.ClassResolver#register(java.lang.Class<?>)` , it should always succeed. The implemention shoukd handle class id conflict inside

Therefore, there is no need to deal with Classd conflicts in `org.apache.fury.resolver.ClassResolver#register(java.lang.Class<?>)`, because there will be no conflicts in ClassId allocated using `ClassIdAllocator`

LiangliangSui: If you think there is nothing wrong with abstracting a `ClassIdAllocator`, and there is nothing unusual about the above suggestions, I will code and implement a `ClassIdAllocator` in the next few days. @chaokunyang

chaokunyang: Could you elabrate more the `ClassIdAllocator`? 
I'm not sure how it handle such an situation: there are three libraries and one client of three libraries. Every libraries has some classes to register, but they can't assume the register order since they don't know which will be registered first, so they use specified class id for such registration. If the client has two versions, the first version use library1 and library2, and the second version use library1 and library3. Then the registered ids in library2 should be left here, those ids shouldn't be used by library3. Will the `ClassIdAllocator` handle such cases?

chaokunyang: I agree with you, the current `ClassResolver` is a little complex, it would be great if we can refactor some code out.

LiangliangSui: Can we create a new issue to discuss `ClassIdAllocator`? I think the content related to `ClassIdAllocator` should not be related to the current PR.

Maybe we can merge this PR first.

chaokunyang: Yes, could you open an issue for `ClassIdAllocator`? For this PR, maybe we still need to ensure `Fury.register` doesn't raise exception for id conflict

chaokunyang: `registeredClassIdCounter` maybe can be renamed to `classIdGenerator` for better intuition

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -229,8 +228,9 @@ public class ClassResolver {
   private final ShimDispatcher shimDispatcher;
 
   private static class ExtRegistry {
-    private short registeredClassIdCounter = 0;
-    private final LongMap<Class<?>> registeredId2Classes = new LongMap<>(initialCapacity);
+    // Here we set it to 1 because `NO_CLASS_ID` is 0 to avoid calculating it again in
+    // `register(Class<?> cls)`.
+    private short registeredClassIdCounter = 1;
```
chaokunyang, line 233:
```suggestion
    private short classIdGenerator = 1;
```

LiangliangSui: Updated.

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
LiangliangSui: I have created an issue and expressed my thoughts on the question you raised above. https://github.com/apache/incubator-fury/issues/1412

chaokunyang (approved):
LGTM

PR #1408, opened 2024-03-14
chore(java): Move default Serializer registration location

LiangliangSui:
Move all default Serializer registrations to specific Serializers.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -57,7 +57,6 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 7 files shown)


java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -305,33 +304,26 @@ public void initialize() {
...
-    addDefaultSerializer(String.class, new StringSerializer(fury));
+    StringSerializer.registerDefaultSerializers(fury);
     PrimitiveSerializers.registerDefaultSerializers(fury);
     Serializers.registerDefaultSerializers(fury);
     ArraySerializers.registerDefaultSerializers(fury);
     TimeSerializers.registerDefaultSerializers(fury);
     OptionalSerializers.registerDefaultSerializers(fury);
     CollectionSerializers.registerDefaultSerializers(fury);
     MapSerializers.registerDefaultSerializers(fury);
-    addDefaultSerializer(Locale.class, new LocaleSerializer(fury));
```
chaokunyang, line 316: those files has only one serializer, maybe it's OK to register such serializers directly here?

LiangliangSui: Yes, but it may be better in design to move them into their own Serializers, which are only responsible for calling their respective `Serializer#registerDefaultSerializers`.

chaokunyang: It has only one serializer, I pprefer not using plurality.

chaokunyang: And the reason`registerDefaultSerializers` exists is becuase there are multiple deeply-related serializers in one file

LiangliangSui: Ok, get it.

PR #1411, opened 2024-03-16
feat(java): avoid recompilation when gc happens for memory pressure

### LINKED ISSUE #1410: [Java] Serialized got recompilation when full gc happens
When there is a memory pressure, serializer class got recompiled, which makes the serialization slow.

`CodeGenerator` use soft reference to cache the generated serializer. When there is a memory pressure, the `CodeGenerator` will be gc. 

We can hold strong reference in `Fury` instance if possible, and use a mixed `WeakReference` and `SoftReference` to delay the reference be set to null by garbage collector.

chaokunyang:
Closes #1410


theweipeng (approved):
LGTM

PR #1413, opened 2024-03-18
feat(spec): standardizing fury cross-language serialization specification

### LINKED ISSUE #1418: [Spec][Doc] fury cross-language serialization specification proposal
We've standardized java serialization spec in #1240, but the cross-language serialzation spec has never been formulized.

The current implementation of fury xlang serialization across multiple languages are all based on the code in one of the languages. It's not complete and pone to inconsistencies. 

And if some one want to implement Fury for a new language such as Fury C# in #686, he must read all Fury java serialization code. This would be a huge burden for new developers. Not even to say someone may don't write java either.

Another thing is that our xlang serialization is not standardized, we can't have a foundation to discuss how to improve our protocol too.

And our current  xlang serialization has many places to improve, such as it didn't resolve the type inconsistencies between languages. Such things should be resolved too.

We should design a new protocol for Fury and standardized it as a document.

## Additional context
Serialization frameworks such as arrow/avro/hession/thrift/flatbuffer/msgpack all have a serialization spec:
- https://github.com/apache/arrow/blob/main/format/Schema.fbs
- https://numpy.org/doc/stable/user/basics.types.html
- https://thrift.apache.org/docs/types
- https://www.caucho.com/resin-3.1/doc/hessian-2.0-spec.xtp
- https://github.com/msgpack/msgpack/blob/master/spec.md

chaokunyang:
## What does this PR do?

This PR standardizes fury cross-language serialization specification. It comes with following changes:
- Remove type tag from the protocol since it introduce space and performance overhead to the implementation. The `type tag` version can be seen in https://github.com/apache/incubator-fury/blob/6ea2e0b83d5449d63ca62296ff0dfd67b96c5bc5/docs/protocols/xlang_object_graph_spec.md .
- Fury preserves `0~63` for internal types, but let users register type by id from `0`(added by 64 automatically） to setup type mapping between languages. 
- Streamline the type systems, only `bool/byte/i16/i32/i64/half-float/float/double/string/enum/list/set/map/Duration/Timestamp/decimal/binary/array/tensor/sparse/tensor/arrow/record/batch/arrow/table` are allowed.
- Formulized the binary format for above types.
- Add type disambiguation: the deserialization are determined by data type in serialized binary and target type jointly.
- Introduce meta string encoding algorithm for field name to reduce space cost by 3/8.
- Introduce schema consist mode format for struct.
- Introduce schema envolution mode for struct: 
  - this mode can embeed meta in the data or share across multiple messages,
  - it can avoid the cost of type tag comparison in frameworks like protobuf

This protocol also supports object inheriance for xlang serializaiton. This is a feature request that users has been discussed for a long time in protobuf/flatbuffer:
- https://github.com/google/flatbuffers/issues/4006
- https://github.com/protocolbuffers/protobuf/issues/5645

Although there are some languages such as `rust/golang` doesn't support inheriance, there are many cases only langauges like `java/c#/python/javascript` are involved, and the support for inheriance is not complexed in the protocol level, so we added the inheriance support in the protocol. And in languages such as `rust/golang`, we can use some annotation to mark composition field as parent class for serialization layout, or we can disable inheriance foor such languages at the protocol level.
 
The protocol support polymorphic natively by type id, so I don't include types such as `OneOf/Union`. With this protocol, you can even serialize multiple rust `dyn trait` object which implement same trait., and get exactly the same objects when deserialization.

## Related issue
This PR Closes #1418

### DIFF
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -471,7 +471,7 @@ Primitive array are taken as a binary buffer, serialization will just write the 
 then copy the whole buffer into the stream.
 
 Such serialization won't compress the array. If users want to compress primitive array, users need to register custom
-serializers for such types.
+serializers for such types or mark it as list type.
 
 #### object array
 
@@ -488,7 +488,7 @@ Format:
 | length(unsigned varint) | key value chunk data | ... | key value chunk data |
 ```
 
-#### map key-value data
+#### map key-value chunk data
 
 Map iteration is too expensive, Fury won't compute the header like for list since it introduce
 [considerable overhead](https://github.com/apache/incubator-fury/issues/925).
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -24,6 +24,8 @@ also introduce more complexities compared to static serialization frameworks. So
 - float: A 32-bit floating point number.
 - double: A 64-bit floating point number including NaN and Infinity.
 - string: A text string encoded using Latin1/UTF16/UTF-8 encoding.
+- enum: a data type consisting of a set of named values. Rust enum with non-predefined field values are not supported as
+  an enum
 - list: A sequence of objects.
 - set: An unordered set of unique elements.
 - map: A map of key-value pairs.
@@ -50,7 +52,7 @@ also introduce more complexities compared to static serialization frameworks. So
 
 ### Type ambiguities
 
-Due to differences between type systems of languages, those types can't mapped one-to-one between languages. When
+Due to differences between type systems of languages, those types can't be mapped one-to-one between languages. When
 deserializing, Fury use the target data structure type and the data type in the data jointly to determine how to
 deserialize and populate the target data structure. For example:
 
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -345,9 +345,14 @@ for data.
 #### int16
 
 - size: 2 byte
-- byte order: little endian order
+- byte order: raw bytes of little endian order
 
 #### unsigned int32
+
+- size: 4 byte
+- byte order: raw bytes of little endian order
+
+#### unsigned varint32
 
 - size: 1~5 byte
 - Format: The most significant bit (MSB) in every byte indicates whether to have the next byte. If first bit is set
@@ -356,11 +361,21 @@ for data.
 
 #### signed int32
 
+- size: 4 byte
+- byte order: raw bytes of little endian order
+
+#### signed varint32
+
 - size: 1~5 byte
 - Format: First convert the number into positive unsigned int by `(v << 1) ^ (v >> 31)` ZigZag algorithm, then encode
   it as an unsigned varint.
 
 #### unsigned int64
+
+- size: 8 byte
+- byte order: raw bytes of little endian order
+
+#### unsigned varint64
 
 - size: 1~9 byte
 - Fury SLI(Small long as int) Encoding:
@@ -371,6 +386,11 @@ for data.
       i.e. `b & 0x80 == 0x80`, then the next byte should be read until the first bit is unset.
 
 #### signed int64
+
+- size: 8 byte
+- byte order: raw bytes of little endian order
+
+#### signed varint64
 
 - size: 1~9 byte
 - Fury SLI(Small long as int) Encoding:
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -190,12 +190,28 @@ varint of `type_id`.
 
 If schema evolution mode is enabled globally or enabled for current type, type meta will be written as follows:
 
-- If meta share mode is not enabled, type meta will be written as schema consistent mode. Additionally, field meta such
-  as field type and name will be written with the field value using a key-value like layout.
-- If meta share mode is enabled, type meta will be written as a meta-share encoded binary if type hasn't been written
-  before, otherwise an unsigned varint id which references to previous written type meta will be written.
+- If meta share mode is not enabled:
+    - If type meta hasn't been written before, the data will be written as:
+      ```
+      | unsigned varint: 0b11111111 | type def |
+      ```
+    - If type meta has been written before, the data will be written as:
+      ```
+      | unsigned varint: written index << 1 | type def |
+      ```
+- If meta share mode is enabled:
+    - If type meta hasn't been written before, add `type def` to `MetaContext`, then get
+      index of the meta in `MetaContext`, write that index as an unsigned varint.
+    - The type meta in `MetaContext` will be carried into the serialized data when the type is first serialized in a
+      batch. That what's `optional 4 bytes` used for. It records the start offset of serialized type defs which not in
+      provided `MetaContext` in current object graph serialization. Type defs
+      ```python
+      buffer.write_var_uint32(len(writting_type_metas))
+      for type_meta in writting_type_metas:
+          type_meta.write_type_def(buffer)
+      ```
 
-## Meta share
+#### Type Def
 
 > This mode will forbid streaming writing since it needs to look back for update the start offset after the whole object
 > graph
@@ -204,8 +220,7 @@ If schema evolution mode is enabled globally or enabled for current type, type m
 > Meta streamline will be supported in the future for enclosed meta sharing which doesn't cross multiple serializations
 > of different objects.
 
-For Schema consistent mode, type will be encoded as an enumerated string by full type name. Here we mainly describe
-the meta layout for schema evolution mode:
+Here we mainly describe the meta layout for schema evolution mode:
 
 ```
 |      8 bytes meta header      |   variable bytes   |  variable bytes   | variable bytes |
@@ -215,7 +230,7 @@ the meta layout for schema evolution mode:
 
 Type meta are encoded from parent type to leaf type, only type with serializable fields will be encoded.
 
-### Meta header
+##### Meta header
 
 Meta header is a 64 bits number value encoded in little endian order.
 
@@ -225,10 +240,8 @@ Meta header is a 64 bits number value encoded in little endian order.
   only, num classes will be 1.
 - The 5th bit is used to indicate whether this type needs schema evolution.
 - Other 56 bits are used to store the unique hash of `flags + all layers type meta`.
-- Inheritance: Fury
-    - For languages
 
-### Single layer type meta
+##### Single layer type meta
 
 ```
 | unsigned varint | var uint |  field info: variable bytes   | variable bytes  | ... |
@@ -243,7 +256,8 @@ Meta header is a 64 bits number value encoded in little endian order.
       consistent fields will be serialized first, then compatible fields will be serialized next. At deserialization,
       Fury will use fields info of those fields which aren't annotated by tag id for deserializing schema consistent
       fields, then use fields info in meta for deserializing compatible fields.
-- Field info:
+- type id: the registered id for the current type, which will be written as an unsigned varint.
+- field info:
     - Header(8 bits):
         - Format:
             - `reserved 1 bit + 3 bits field name encoding + polymorphism flag + nullability flag + ref tracking flag + tag id flag`.
@@ -282,7 +296,7 @@ Field order are left as implementation details, which is not exposed to specific
 resort fields based on Fury field comparator. In this way, fury can compute statistics for field names or types and
 using a more compact encoding.
 
-### Other layers type meta
+##### Other layers type meta
 
 Same encoding algorithm as the previous layer.
 
@@ -573,8 +587,8 @@ field value of non-final type without ref tracking:
 
 Schema evolution have similar format as schema consistent mode for object except:
 
-- For this object type itself, `schema consistent` mode will write type by id/name, but `schema evolution` mode will
-  write type field names, types and other meta too, see [Type meta](#type-meta).
+- For the object type, `schema consistent` mode will write type by id only, but `schema evolution` mode will
+  write type consisting of field names, types and other meta too, see [Type meta](#type-meta).
 - Type meta of `final custom type` needs to be written too, because peers may not have this type defined.
 
 ### Type
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -511,6 +511,17 @@ Field will be ordered as following, every group of fields will have its own orde
 
 #### schema consistent
 
+Object will be written as:
+
+```
+|    4 byte     |  variable bytes  |
++---------------+------------------+
+|   type hash   |   field values   |
+```
+
+Type hash is used to check the type schema consistency across languages. Type hash will be the first 32 bits of 56 bits
+value of the type meta.
+
 Object fields will be serialized one by one using following format:
 
 ```
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -197,7 +197,7 @@ If schema evolution mode is enabled globally or enabled for current type, type m
       ```
     - If type meta has been written before, the data will be written as:
       ```
-      | unsigned varint: written index << 1 | type def |
+      | unsigned varint: written index << 1 |
       ```
 - If meta share mode is enabled:
     - If type meta hasn't been written before, add `type def` to `MetaContext`, then get
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -15,26 +15,26 @@ also introduce more complexities compared to static serialization frameworks. So
 
 ### Data Types
 
-- bool: A boolean value (true or false).
-- int8: An 8-bit signed integer.
-- int16: A 16-bit signed integer.
-- int32: A 32-bit signed integer.
-- int64: A 64-bit signed integer.
-- float16: A 16-bit floating point number.
-- float32: A 32-bit floating point number.
-- float64: A 64-bit floating point number including NaN and Infinity.
-- string: A text string encoded using Latin1/UTF16/UTF-8 encoding.
-- enum: A data type consisting of a set of named values. Rust enum with non-predefined field values are not supported as
+- bool: a boolean value (true or false).
+- int8: a 8-bit signed integer.
+- int16: a 16-bit signed integer.
+- int32: a 32-bit signed integer.
+- int64: a 64-bit signed integer.
+- float16: a 16-bit floating point number.
+- float32: a 32-bit floating point number.
+- float64: a 64-bit floating point number including NaN and Infinity.
+- string: a text string encoded using Latin1/UTF16/UTF-8 encoding.
+- enum: a data type consisting of a set of named values. Rust enum with non-predefined field values are not supported as
   an enum.
-- list: A sequence of objects.
-- set: An unordered set of unique elements.
-- map: A map of key-value pairs.
+- list: a sequence of objects.
+- set: an unordered set of unique elements.
+- map: a map of key-value pairs.
 - time types:
-    - Duration: an absolute length of time, independent of any calendar/timezone, as a count of nanoseconds.
-    - Timestamp: a point in time, independent of any calendar/timezone, as a count of nanoseconds. The count is relative
+    - duration: an absolute length of time, independent of any calendar/timezone, as a count of nanoseconds.
+    - timestamp: a point in time, independent of any calendar/timezone, as a count of nanoseconds. The count is relative
       to an epoch at UTC midnight on January 1, 1970.
 - decimal: exact decimal value represented as an integer value in two's complement.
-- binary: An variable-length array of bytes.
+- binary: an variable-length array of bytes.
 - array type: only allow numeric component. Other arrays will be taken as List. The implementation should support the
   interoperability between array and list.
     - array: multidimensional array which every sub-array can have different size but all have same type.
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -19,7 +19,12 @@ also introduce more complexities compared to static serialization frameworks. So
 - int8: a 8-bit signed integer.
 - int16: a 16-bit signed integer.
 - int32: a 32-bit signed integer.
+- var_int32: a 32-bit signed integer which use fury var_int32 encoding.
+- fixed_int32: a 32-bit signed integer which use two's complement encoding.
 - int64: a 64-bit signed integer.
+- var_int64: a 64-bit signed integer which use fury PVL encoding.
+- sli_int64: a 64-bit signed integer which use fury SLI encoding.
+- fixed_int64: a 64-bit signed integer which use two's complement encoding.
 - float16: a 16-bit floating point number.
 - float32: a 32-bit floating point number.
 - float64: a 64-bit floating point number including NaN and Infinity.
@@ -49,6 +54,10 @@ also introduce more complexities compared to static serialization frameworks. So
 - sparse tensor: a multidimensional array whose elements are almost all zeros.
 - arrow record batch: an arrow [record batch](https://arrow.apache.org/docs/cpp/tables.html#record-batches) object.
 - arrow table: an arrow [table](https://arrow.apache.org/docs/cpp/tables.html#tables) object.
+
+Note:
+
+- Unsigned int/long are not added here, since not every language support those types.
 
 ### Type disambiguation
 
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -208,9 +208,10 @@ using one of the following mode. Which mode to use is configured when creating f
         - Firstly, set current to `meta start offset` of fury header
         - Then write `captured_type_defs` one by one:
           ```python
-          buffer.write_var_uint32(len(writting_type_defs))
+          buffer.write_var_uint32(len(writting_type_defs) - len(schema_consistent_type_def_stubs))
           for type_meta in writting_type_defs:
-              type_meta.write_type_def(buffer)
+              if not type_meta.is_stub():
+                  type_meta.write_type_def(buffer)
           writing_type_defs = copy(schema_consistent_type_def_stubs)
           ```
 - Meta share mode: the writing steps are same as the normal mode, but `captured_type_defs` will be shared across
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -545,9 +545,7 @@ Type will be serialized using type meta format.
 - Try to use one varint/long to write flags and length together to save one byte cost and reduce memory io.
 - Condition branches are less expensive compared to memory IO cost unless there are too many branches.
 
-### Fast deserialization for languages without runtime codegen support
-
-The Rust and C++ don't support dynamic codegen. We need to generate all code are compile-time use meta programing.
+### Fast deserialization for static languages without runtime codegen support
 
 For type evolution, the serializer will encode the type meta into the serialized data. The deserializer will compare
 this meta with class meta in current process, and use the diff to determine how to deserialize the data.
@@ -555,10 +553,60 @@ this meta with class meta in current process, and use the diff to determine how 
 For java/javascript/python, we can use the diff to generate serializer code at runtime and load it as class/function for
 deserialization. In this way, the type evolution will be as fast as type consist mode.
 
-For C++/Rust, we can't generate the serializer code at runtime. So we need to generate the code at compile-time. But at
-that time, we don't know the type schema in other processes. So we can't generate the serializer code for such
-inconsistent types. We may need to generate the code which has a loop and compare field name one by one to decide
-whether deserialize and assign the field or deserialize and skip the field. One lucky thing is that we can optimize the
-string comparison into long comparison: generate **a 64-bit id** for every field name of a type at build time,
-cache the type meta, and **generate a 64-bit id** for its every field name too, then we can convert the field name
-comparison into long comparison, and it will be fast.
+For C++/Rust, we can't generate the serializer code at runtime. So we need to generate the code at compile-time using
+meta programing. But at that time, we don't know the type schema in other processes, so we can't generate the serializer
+code for such inconsistent types. We may need to generate the code which has a loop and compare field name one by one to
+decide whether deserialize and assign the field or skip the field value.
+
+One fast way is that we can optimize the string comparison into `jump` instructions:
+
+- Assume current type has `n` fields, peer type has `n1` fields.
+- Generate an auto growing `field id` from `0` for every sorted field in current type at the compile time.
+- Compare the received type meta with current type, generate same id if the field name is same, otherwise generate an
+  auto growing id starting from `n`, cache this meta at runtime.
+- Iterate the fields of received type meta, use a `switch` to compare the `field id` to deserialize data
+  and `assign/skip` field value. **Continuous** field id will be optimized into `jump` in `switch` block, so it will
+  very fast.
+
+Here is an example, suppose process A has a class `Foo` with version 1 defined as `Foo1`, process B has a class `Foo`
+with version 2 defined as `Foo2`:
+
+```c++
+// class Foo with version 1
+class Foo1 {
+  int32_t v1; // id 0
+  std::string v2; // id 1
+}
+// class Foo with version 2
+class Foo2 {
+  // id 0, but will have id 2 in process A
+  bool v0;
+  // id 1, but will have id 0 in process A
+  int32_t v1;
+  // id 2, but will have id 3 in process A
+  int64_t long_value;
+  // id 3, but will have id 1 in process A
+  std::string v2;
+  // id 4, but will have id 4 in process A
+  std::vector<std::string> list;
+}
+```
+
+When process A received serialized `Foo2` from process B, here is how it deserialize the data:
+
+```c++
+Foo1 &foo1 = xxx;
+std::vector<fury::FieldInfo> &field_infos = type_meta.field_infos;
+for (auto &field_info : field_infos) {
+  switch (field_info.field_id) {
+    case 0:
+      foo1.v1 = buffer.read_varint32();
+      break;
+    case 1:
+      foo1.v2 = fury.read_string();
+      break;
+    default:
+      fury.skip_data(field_info);
+  }
+}
+```
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -57,13 +57,16 @@ When reference tracking is disabled globally or for specific types, or for certa
 context(e.g., a field of a type), only the `NULL` and `NOT_NULL VALUE` flags will be used for reference meta.
 
 For languages which doesn't support reference such as rust, reference tracking must be disabled for correct
-deserialization
-by fury rust implementation.
+deserialization by fury rust implementation.
 
-In languages which doesn't support GC, Fury use `optional` to indicate nullability:
+For languages whose object values are not null by default:
 
 - In rust, Fury takes `Option:None` as a null value
 - In c++, Fury takes `std::nullopt` as a null value
+- In golang, Fury takes `null interface/pointer` as a null value
+
+If one want to deserialize in languages like `Java/Python/JavaScript`, he should mark the type with all fields
+not-null by default, or using schema-evolution mode to carry the not-null fields info in the data.
 
 ## Type Meta
 
@@ -292,11 +295,6 @@ If string has been written before, the data will be written as follows:
 - format: write as pure byte.
 
 #### Short
-
-- size: 2 byte
-- byte order: little endian order
-
-#### Char
 
 - size: 2 byte
 - byte order: little endian order
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -190,7 +190,36 @@ varint of `type_id`.
 
 If schema evolution mode is enabled globally or enabled for current type, type meta will be written as follows:
 
-- If meta share mode is not enabled:
+- Normal mode(meta share not enabled):
+    - If type meta hasn't been written before, add `type def`
+      to `captured_type_defs`: `captured_type_defs[type def] = map size`
+    - Get index of the meta in `captured_type_defs`, write that index as an unsigned varint.
+    - The meta will be written as:
+      ```
+      | unsigned varint: written index |
+      ```
+    - After finished the serialization of the object graph, fury will start to write `captured_type_defs`:
+        - Firstly, set current to `meta start offset` of fury header
+        - Then write `captured_type_defs` one by one:
+          ```python
+          buffer.write_var_uint32(len(writting_type_defs))
+          for type_meta in writting_type_defs:
+          type_meta.write_type_def(buffer)
+          ```
+- Meta share mode: the writing steps are same as the normal mode, but `captured_type_defs` will be shared across
+  multiple serializations of different objects. For example, suppose we have a batch to serialize:
+    ```python
+    captured_type_defs = {}
+    stream = ...
+    # add `Type1` to `captured_type_defs` and write `Type1`
+    fury.serialize(stream, [Type1()])
+    # add `Type2` to `captured_type_defs` and write `Type2`, `Type1` is written before.
+    fury.serialize(stream, [Type1(), Type2()])
+    # `Type1` and `Type2` are written before, no need to write meta.
+    fury.serialize(stream, [Type1(), Type2()])
+    ```
+
+- Streaming mode(streaming mode doesn't support meta share):
     - If type meta hasn't been written before, the data will be written as:
       ```
       | unsigned varint: 0b11111111 | type def |
@@ -199,26 +228,14 @@ If schema evolution mode is enabled globally or enabled for current type, type m
       ```
       | unsigned varint: written index << 1 |
       ```
-- If meta share mode is enabled:
-    - If type meta hasn't been written before, add `type def` to `MetaContext`, then get
-      index of the meta in `MetaContext`, write that index as an unsigned varint.
-    - The type meta in `MetaContext` will be carried into the serialized data when the type is first serialized in a
-      batch. That what's `optional 4 bytes` used for. It records the start offset of serialized type defs which not in
-      provided `MetaContext` in current object graph serialization. Type defs
-      ```python
-      buffer.write_var_uint32(len(writting_type_metas))
-      for type_meta in writting_type_metas:
-          type_meta.write_type_def(buffer)
-      ```
+      `written index` is the id in `captured_type_defs`.
+    - With this mode, `meta start offset` can be omitted.
+
+> The normal mode and meta share mode will forbid streaming writing since it needs to look back for update the start
+> offset after the whole object graph writing and meta collecting is finished. Only in this way we can ensure
+> deserialization failure in meta share mode doesn't lost shared meta.
 
 #### Type Def
-
-> This mode will forbid streaming writing since it needs to look back for update the start offset after the whole object
-> graph
-> writing and meta collecting is finished. Only in this way we can ensure deserialization failure doesn't lost shared
-> meta.
-> Meta streamline will be supported in the future for enclosed meta sharing which doesn't cross multiple serializations
-> of different objects.
 
 Here we mainly describe the meta layout for schema evolution mode:
 
--- a/docs/protocols/xlang_object_graph_spec.md
+++ b/docs/protocols/xlang_object_graph_spec.md
@@ -1,5 +1,8 @@
 # Cross language object graph serialization
 
+> Format Version History:
+> - Version 0.1 - serialization spec formalized
+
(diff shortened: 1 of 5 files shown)


chaokunyang:
The Rust and C++ need more consideration, since they don't support dynamic codegen. We need to genertate all code are compilet-time use meta programing. The generate code may not be the best code for execution.

For type evolution, the serializer will encode the type meta into the serialized data. The deserializer can compare this meta with class in current process, and use the diff to generate the serializer for deserializaiton.
- For java/javascript/python, we can use the diff to generate serializer code at runtime and load it as `class/function` for deserialization. In this way, the type evolution will be as fast as type consist mode.
- For C++/Rust, we can't generate the serializer code at runtime. So we need to generate the code at compile-time. But at that time, we don't know the type schema in other processes. So we can't generate the serializer code for such inconsistent types. We may need to generate the code which has a loop and compare field name one by one to decide whether deserialize and assign the field or deserialize and skip the field. One lucky thing is that we can cache the type meta, and genetate a 64-bit id for every for field name, and genetate the id for field names for current type at build time, so we can convert the field name comparasion into long comparasion. And it will be fast too.
- 
- 
-  use field pointer to

chaokunyang:
The spec must allow the languages like rust/c++ which don't have runtime reflection work and fast, and let languages like javasccript which has a weak type systems to work too.

chaokunyang:
I updated the implementation guide, the implementation for static languages can use `switch/jump` for type evolution deserialization:

- Assume current type has `n` fields, peer type has `n1` fields.
- Generate an auto growing `field id` from `0` for every sorted field in current type at the compile time.
- Compare the received type meta with current type, generate same id if the field name is same, otherwise generate an
  auto growing id starting from `n`, cache this meta at runtime.
- Iterate the fields of received type meta, use a `switch` to compare the `field id` to deserialize data
  and `assign/skip` field value. **Continuous** field id will be optimized into `jump` in `switch` block, so it will
  very fast.

Here is an example, suppose process A has a class `Foo` with version 1 defined as `Foo1`, process B has a class `Foo`
with version 2 defined as `Foo2`:

```c++
// class Foo with version 1
class Foo1 {
  int32_t v1; // id 0
  std::string v2; // id 1
}
// class Foo with version 2
class Foo2 {
  // id 0, but will have id 2 in process A
  bool v0;
  // id 1, but will have id 0 in process A
  int32_t v1;
  // id 2, but will have id 3 in process A
  int64_t long_value;
  // id 3, but will have id 1 in process A
  std::string v2;
  // id 4, but will have id 4 in process A
  std::vector<std::string> list;
}
```

When process A received serialized `Foo2` from process B, here is how it deserialize the data:

```c++
Foo1 &foo1 = xxx;
std::vector<fury::FieldInfo> &field_infos = type_meta.field_infos;
for (auto &field_info : field_infos) {
  switch (field_info.field_id) {
    case 0:
      foo1.v1 = buffer.read_varint32();
      break;
    case 1:
      foo1.v2 = fury.read_string();
      break;
    default:
      fury.skip_data(field_info);
  }
}
```

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+The serialization format is a dynamic binary format. The dynamics and reference/polymorphism support make Fury flexible,
+much more easy to use, but
+also introduce more complexities compared to static serialization frameworks. So the format will be more complex.
+
+## Type Systems
+
+### Data Types
+
+- bool: A boolean value (true or false).
+- byte: An 8-bit signed integer.
```
PragmaTwice, line 19: Should `byte` be unsigned? Or, should `byte` have no relationship with sign?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+- i16: A 16-bit signed integer.
+- i32: A 32-bit signed integer.
+- i64: A 64-bit signed integer.
+- half-float: A 16-bit floating point number.
+- float: A 32-bit floating point number.
+- double: A 64-bit floating point number including NaN and Infinity.
```
PragmaTwice, line 25: About naming:

How about fp16, fp32, fp64 or float16, float32, float64?

docs/protocols/xlang_object_graph_spec.md
chaokunyang: It's int8 here

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+- string: A text string encoded using Latin1/UTF16/UTF-8 encoding.
```
PragmaTwice, line 26: Is Latin1 still widely used? And how about UTF32?

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Maybe we should use int8 instead

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+- enum: a data type consisting of a set of named values. Rust enum with non-predefined field values are not supported as
+  an enum
```
PragmaTwice, line 28: How about the underlying type of enum? i.e. how the enum is stored? 32bit? 64bit?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+- list: A sequence of objects.
+- set: An unordered set of unique elements.
+- map: A map of key-value pairs.
```
PragmaTwice, line 31: Is there any restriction ion the key type?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+- time types:
+    - Duration: an absolute length of time independent of any calendar/timezone, as a count of seconds and
```
PragmaTwice, line 33:
```suggestion
    - Duration: an absolute length of time, independent of any calendar/timezone, as a count of seconds and
```

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+      fractions of seconds at nanosecond resolution.
+    - Timestamp: a point in time independent of any calendar/timezone, as a count of seconds and fractions of
+      seconds at nanosecond resolution. The count is relative to an epoch at UTC midnight on January 1, 1970.
```
PragmaTwice, line 36:
```suggestion
    - Duration: an absolute length of time, independent of any calendar/timezone, as a count of nanoseconds.
    - Timestamp: a point in time independent of any calendar/timezone, as a count of nanoseconds, relative to January 1st, 1970 at 00:00:00 UTC.
```

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+- decimal: exact decimal value represented as an integer value in two's complement.
```
PragmaTwice, line 37: What's kind of decimal? decimal32/decimal64 in IEEE754? or an unlimited-precison decimal?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+- binary: binary data.
```
PragmaTwice, line 38: What's binary data? An array of bytes?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+- array type: only allow numeric component. Other arrays will be taken as List. The implementation should support the
+  interoperability between array and list.
+    - array: multiple dimension array which every subarray can have have different size.
```
PragmaTwice, line 41: Could subarray have different type?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+    - int16_array: one dimension int16 array.
+    - int32_array: one dimension int32 array.
+    - int64_array: one dimension int64 array.
+    - half_float_array: one dimension half_float_16 array.
+    - float_array: one dimension float32 array.
+    - double_array: one dimension float64 array.
```
PragmaTwice, line 47: How about bool, byte and enum?

docs/protocols/xlang_object_graph_spec.md
chaokunyang: I think we can use float16, float32, float64 instead. It's the type in golang, arrow, numpy. 

 fp16, fp32, fp64 are more used in deeplearning

docs/protocols/xlang_object_graph_spec.md
chaokunyang: > Is Latin1 still widely used? And how about UTF-32?

UTF-32 use 4 byte for a char, which will bloat the data a lot. 

Actually most chars can be expressed using Latin1, and there are many langauges such as java/python/javascript support `Latin1/UTF-16` natively, so we add such encoding here. When the language support does support the `Latin1/UTF-16` encoding, we can skip the encoding/decoding cost, and using a memory copy to create a string object.

Languages like rust/golang using utf-8 for string encoding, they can still use a copy to create a string object from the serialized data if the data is encoded using utf-8. But if the peer language is Java, a conversion from latin1/utf-16 to utf8 would be needed. But it's Ok, because if we use utf-8 encoding only, the java would need to encode the latin1/utf16 to utf8 when serialization, the cost didn't go way.

chaokunyang: > If you wan to support ISO 8859 variants, you should probably name them with names like ISO-8859-1. Unicode encodings have reduced use of ISO 8859 family.

@pjfanning Java/pytho/nodejs all use latin1, how about name it as `Latin1(ISO-8859-1)`

https://docs.python.org/3/library/codecs.html
image
https://nodejs.org/api/buffer.html
image

docs/protocols/xlang_object_graph_spec.md
chaokunyang: it's serialized as an unsigned var int, we detailed the format for it in `Value Format` section

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Nope, it should be immutable, but this is nit related to the protocol

docs/protocols/xlang_object_graph_spec.md
chaokunyang: An unlimited-precison decimal, the precison will  be written in the data.

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Yep

docs/protocols/xlang_object_graph_spec.md
chaokunyang: nope, array component can be numeric only, it's final.

If you want to store a polymorphic object in array, it will be encoded as a list type.

chaokunyang: for array, we use a memory copy for its serialization.

The memory layout is fixed for array in fury

docs/protocols/xlang_object_graph_spec.md
chaokunyang: I missed bool array, will add it later. 
byte array is just binary, seems unnecessary to add it again.

enum array should use list type

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+
+The serialization format is a dynamic binary format. The dynamics and reference/polymorphism support make Fury flexible,
+much more easy to use, but
+also introduce more complexities compared to static serialization frameworks. So the format will be more complex.
+
+## Type Systems
+
+### Data Types
+
+- bool: A boolean value (true or false).
```
LiangliangSui, line 18: Can we unify the capitalization style of the first letter of the first word after the type, and use it in uppercase or lowercase?

chaokunyang: thanks, updated

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Arrow support decimal128/decimal256 only. I'm not sure whether fury should add such limit.

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+```
+
+`intArray` has `int32_array` type. But both `objects` and `objectList` field in the serialize data have `list` data
+type. When deserializing, the implementation will create an `Object` array for `objects`, but create a `ArrayList`
+for `objectList` to populate it's elements. And the serialized data of `Foo` can be deserialized into `Foo2` too.
+
+### Type ID
+
+All internal data types are expressed using unsigned ID `-64~-1`. Users can use `0~32703` for representing their types.
+At runtime, all type ids are added by `64`, represented and encoded as an unsigned int.
```
LiangliangSui, line 80: signed ID `-64 ~ -1` ?

LiangliangSui, line 80: Is it possible to use uint16 to represent unsigned int ?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,612 @@
...
+}
+```
+
+`intArray` has `int32_array` type. But both `objects` and `objectList` field in the serialize data have `list` data
+type. When deserializing, the implementation will create an `Object` array for `objects`, but create a `ArrayList`
+for `objectList` to populate it's elements. And the serialized data of `Foo` can be deserialized into `Foo2` too.
+
+### Type ID
+
+All internal data types are expressed using unsigned ID `-64~-1`. Users can use `0~32703` for representing their types.
```
LiangliangSui, line 79: > Users can use `0~32703` for representing their types.

Therefore, we need to indicate the ID range on the interface for users, and strictly check the input parameters internally.

chaokunyang: Yes, the ID will be checked

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Good catch, updated

docs/protocols/xlang_object_graph_spec.md
PragmaTwice: If we use `floatXX`, I think we should also use `intXX` instead of `iXX`

chaokunyang: Updated already in previous commit for update to `floatXX`

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,635 @@
...
+
+When process A received serialized `Foo2` from process B, here is how it deserialize the data:
+
+```c++
+Foo1 &foo1 = xxx;
+std::vector<fury::FieldInfo> &field_infos = type_meta.field_infos;
+for (auto &field_info : field_infos) {
+  switch (field_info.field_id) {
+    case 0:
+      foo1.v1 = buffer.read_varint32();
```
theweipeng, line 626: In this sample,  the field v1 in Foo1 assigned ` id 0` and it's type was int32, whereas in Foo2 , `id 0 ` refers to v0 which is a boolean. how did `case0:   foo1.v1 = buffer.read_varint32();` worked correctly  under these conditions?

chaokunyang: The id won't be written in the serialized data, the whole type meta will carry the field name, so when deserialization, fury can compare the field name and generate a different id for fields not match

theweipeng: Ok, i got it

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,635 @@
...
+Such information can be provided in other languages too:
+
+- cpp: use macro and template.
+- golang: use struct tag.
+- python: use typehint.
+- rust: use macro.
+
+### Type ID
+
+All internal data types are expressed using a ID in range `-64~-1`. Users can use `0~32703` for representing their
```
PragmaTwice, line 102:
```suggestion
All internal data types are expressed using an ID in range `-64~-1`. Users can use `0~32703` for representing their
```

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,635 @@
...
+types. At runtime, all type ids are added by `64`, and then encoded as an unsigned varint.
```
PragmaTwice, line 103: `unsigned varint`

So data type is different from encoding, right? There is no unsigned type, and also varint type.

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,635 @@
...
+
+- size: 1 byte
+- format: 0 for `false`, 1 for `true`
+
+#### Byte
+
+- size: 1 byte
+- format: write as pure byte.
+
+#### Short
```
PragmaTwice, line 323: How about unsigned short?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,635 @@
...
+      i.e. `b & 0x80 == 0x80`, then the next byte should be read until the first bit is unset.
+
+#### Signed long
+
+- size: 1~9 byte
+- Fury SLI(Small long as int) Encoding:
+    - If long is in [-1073741824, 1073741823], encode as 4 bytes int: `| little-endian: ((int) value) << 1 |`
+    - Otherwise write as 9 bytes: `| 0b1 | little-endian 8 bytes long |`
+- Fury PVL(Progressive Variable-length Long) Encoding:
+    - First convert the number into positive unsigned long by ` (v << 1) ^ (v >> 63)` ZigZag algorithm to reduce cost of
```
PragmaTwice, line 355: We don't use zigzag in signed int?

docs/protocols/xlang_object_graph_spec.md
PragmaTwice: And I'm a little worry about that maybe 64 is not enough in the future.

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,635 @@
...
+    - positive long format: first bit in every byte indicates whether to have the next byte. If first bit is set
+      i.e. `b & 0x80 == 0x80`, then the next byte should be read until the first bit is unset.
+
+#### Signed long
+
+- size: 1~9 byte
+- Fury SLI(Small long as int) Encoding:
+    - If long is in [-1073741824, 1073741823], encode as 4 bytes int: `| little-endian: ((int) value) << 1 |`
+    - Otherwise write as 9 bytes: `| 0b1 | little-endian 8 bytes long |`
+- Fury PVL(Progressive Variable-length Long) Encoding:
```
PragmaTwice, line 354: How to decide/indicate between these two encoding?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,635 @@
...
+  it as an unsigned int.
+
+#### Unsigned long
+
+- size: 1~9 byte
+- Fury PVL(Progressive Variable-length Long) Encoding:
+    - positive long format: first bit in every byte indicates whether to have the next byte. If first bit is set
+      i.e. `b & 0x80 == 0x80`, then the next byte should be read until the first bit is unset.
+
+#### Signed long
```
PragmaTwice, line 348: For performance, do we consider to provide a fixed length encoding rather than varint for integer types?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,635 @@
...
+  bool v0;
+  // id 1, but will have id 0 in process A
+  int32_t v1;
+  // id 2, but will have id 3 in process A
+  int64_t long_value;
+  // id 3, but will have id 1 in process A
+  std::string v2;
+  // id 4, but will have id 4 in process A
+  std::vector<std::string> list;
+}
```
PragmaTwice, line 614:
```suggestion
class Foo1 {
  int32_t v1; // id 0
  std::string v2; // id 1
};
// class Foo with version 2
class Foo2 {
  // id 0, but will have id 2 in process A
  bool v0;
  // id 1, but will have id 0 in process A
  int32_t v1;
  // id 2, but will have id 3 in process A
  int64_t long_value;
  // id 3, but will have id 1 in process A
  std::string v2;
  // id 4, but will have id 4 in process A
  std::vector<std::string> list;
};
```

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,635 @@
...
+    case 0:
+      foo1.v1 = buffer.read_varint32();
+      break;
+    case 1:
+      foo1.v2 = fury.read_string();
+      break;
+    default:
+      fury.skip_data(field_info);
+  }
+}
```
PragmaTwice, line 633:
```suggestion
Foo1 foo1 = ...;
const std::vector<fury::FieldInfo> &field_infos = type_meta.field_infos;
for (const auto &field_info : field_infos) {
  switch (field_info.field_id) {
    case 0:
      foo1.v1 = buffer.read_varint32();
      break;
    case 1:
      foo1.v2 = fury.read_string();
      break;
    default:
      fury.skip_data(field_info);
  }
}
```

docs/protocols/xlang_object_graph_spec.md
chaokunyang: If we extend it to 128, user registeded types will be encoded by 2 bytes, which will bloat the data.

I have two solutions:
- Preserve type id range `32000~32703` for extending
- Use `-96~1` for internal ID, but it only give 32 ids for users to encode type in one byte

docs/protocols/xlang_object_graph_spec.md
chaokunyang: We used zigzag in signed int too.

`varint` for long is computing-extensive, so we also provided `Fury SLI(Small long as int) Encoding` to reduce this cost

docs/protocols/xlang_object_graph_spec.md
chaokunyang: `SLI` is faster, but if most longs are very small, it will bloat the data a little. generally speaking, if users use long, probably the value is too large to hold as an int.

docs/protocols/xlang_object_graph_spec.md
chaokunyang: In java, we have a global config for it:
```java
  public FuryBuilder withIntCompressed(boolean intCompressed) {
    this.compressInt = intCompressed;
    return this;
  }

  /**
   * Use variable length encoding for long. Enabled by default, use {@link LongEncoding#SLI} (Small
   * long as int) for long encoding.
   */
  public FuryBuilder withLongCompressed(boolean longCompressed) {
    return withLongCompressed(longCompressed ? LongEncoding.SLI : LongEncoding.LE_RAW_BYTES);
  }

  /** Use variable length encoding for long. */
  public FuryBuilder withLongCompressed(LongEncoding longEncoding) {
    this.longEncoding = Objects.requireNonNull(longEncoding);
    return this;
  }
```

Fury will base this config to generate different code. But it seems we can't do this for c++/rust. I'm considering whether we can add encoding to the type systems. For example, we can have a `fint32/fint64`too.

docs/protocols/xlang_object_graph_spec.md
chaokunyang: @PragmaTwice I added `var_int32/fixed_int32/fixed_int64/var_int64/sli_int64` to type systems, user can use type alias ot annotation to mark such types:
```java
class Foo {
  List<@VarInt64 Integer> list;
}
```
```golang
type varint64 int64

type Foo struct {
  []varint64 list
}
```

chaokunyang: Unsigned types not added, since they are not supported in every language, take `int64` in java as an example.

docs/protocols/xlang_object_graph_spec.md
chaokunyang: I changed it to int16

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,657 @@
...
+## Meta share
+
+> This mode will forbid streaming writing since it needs to look back for update the start offset after the whole object
+> graph
+> writing and meta collecting is finished. Only in this way we can ensure deserialization failure doesn't lost shared
+> meta.
+> Meta streamline will be supported in the future for enclosed meta sharing which doesn't cross multiple serializations
+> of different objects.
+
+For Schema consistent mode, type will be encoded as an enumerated string by full type name. Here we mainly describe
```
theweipeng, line 207: In consistent mode, the type metadata is written as a varint of `type_id` .  However, in  `Meta share` mode, the full name of the type is written. I believe that  `Meta share` mode is used to eliminate the overhead of metadata transmission. But why is the full name necessary in  `Meta share` mode?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,657 @@
...
+If schema consistent mode is enabled globally or enabled for current type, type meta will be written as a fury unsigned
+varint of `type_id`.
+
+### Schema evolution
+
+If schema evolution mode is enabled globally or enabled for current type, type meta will be written as follows:
+
+- If meta share mode is not enabled, type meta will be written as schema consistent mode. Additionally, field meta such
+  as field type and name will be written with the field value using a key-value like layout.
+- If meta share mode is enabled, type meta will be written as a meta-share encoded binary if type hasn't been written
```
theweipeng, line 195: In `meta share` mode, will the `meta-share encode binary` be written at the offset pointed to by the `meta-share start offset`, or at the start of an object in the binary?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,657 @@
...
+### Schema consistent
+
+If schema consistent mode is enabled globally or enabled for current type, type meta will be written as a fury unsigned
+varint of `type_id`.
+
+### Schema evolution
+
+If schema evolution mode is enabled globally or enabled for current type, type meta will be written as follows:
+
+- If meta share mode is not enabled, type meta will be written as schema consistent mode. Additionally, field meta such
```
theweipeng, line 193: What is the meta layout when meta share mode is not enabled ?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,657 @@
...
+  as field type and name will be written with the field value using a key-value like layout.
+- If meta share mode is enabled, type meta will be written as a meta-share encoded binary if type hasn't been written
+  before, otherwise an unsigned varint id which references to previous written type meta will be written.
+
+## Meta share
```
theweipeng, line 198: We can explain the `Meta share` here，including its usage and the conditions under which it is used.

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,657 @@
...
+- In rust, Fury takes `Option:None` as a null value
+- In c++, Fury takes `std::nullopt` as a null value
+- In golang, Fury takes `null interface/pointer` as a null value
+
+If one want to deserialize in languages like `Java/Python/JavaScript`, he should mark the type with all fields
+not-null by default, or using schema-evolution mode to carry the not-null fields info in the data.
+
+## Type Meta
+
+For every type to be serialized, it must be registered with an optional ID first. The registered type will have a
```
theweipeng, line 177: If the tag is replaced by type_id, does it mean that the order of registration needs to be consistent?

docs/protocols/xlang_object_graph_spec.md
chaokunyang: The full name of the type name won't be written, I forget to update it

docs/protocols/xlang_object_graph_spec.md
chaokunyang: I updated the meta spec, could you review it again.

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Yes, users can specify an ID when registering, then the order doesn't have to be consistent. Otherwise, the order must be consistent.

docs/protocols/xlang_object_graph_spec.md
chaokunyang: If meta share mode is not enabled:
    - If type meta hasn't been written before, the data will be written as:
      ```
      | unsigned varint: 0b11111111 | type def |
      ```
    - If type meta has been written before, the data will be written as:
      ```
      | unsigned varint: written index << 1 | type def |
      ```

docs/protocols/xlang_object_graph_spec.md
chaokunyang: - If type meta hasn't been written before, add `type def` to `MetaContext`, then get
  index of the meta in `MetaContext`, write that index as an unsigned varint.
- The type meta in `MetaContext` will be carried into the serialized data when the type is first serialized in a
  batch. That what's `optional 4 bytes` used for. It records the start offset of serialized type defs which not in
  provided `MetaContext` in current object graph serialization. Type defs
  ```python
  buffer.write_var_uint32(len(writting_type_metas))
  for type_meta in writting_type_metas:
      type_meta.write_type_def(buffer)
  ```

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Specifing an ID is not that convenient compared to specify the tag, but it's faster and save space too.

chaokunyang: This do need more consideration, we must determine whether or not do this tradeoff

docs/protocols/xlang_object_graph_spec.md
PragmaTwice: Got it.

docs/protocols/xlang_object_graph_spec.md
PragmaTwice: I mean, what kind of types is allowed in key?

docs/protocols/xlang_object_graph_spec.md
PragmaTwice: If we allow then it's same as float64/float32, which is very efficient and take fixed storage.

If we have unlimited-precision decimal, should we also have bigint and unlimited-precision float?

docs/protocols/xlang_object_graph_spec.md
chaokunyang: list/map/set/array/tensor/arrow are not allowed, others are all allowed

docs/protocols/xlang_object_graph_spec.md
PragmaTwice: How about unsigned integer types?

docs/protocols/xlang_object_graph_spec.md
PragmaTwice: I see. Maybe give some notes on this doc.

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,671 @@
...
+
+```
+length(unsigned varint) | elements header | elements data
+```
+
+#### elements header
+
+In most cases, all elements are same type and not null, elements header will encode those homogeneous
+information to avoid the cost of writing it for every element. Specifically, there are four kinds of information
+which will be encoded by elements header, each use one bit:
```
LiangliangSui, line 428: Can `length+elements` also be concat as a long and encoded as an unsigned var long like string type?

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,671 @@
...
+- If the element types are different, use the 4rd bit `0b1000` header to flag it.
+
+By default, all bits are unset, which means all elements won't track ref, all elements are same type, not null and
+the actual element is the declared type in the custom type field.
+
+#### elements data
+
+Based on the elements header, the serialization of elements data may skip `ref flag`/`null flag`/`element type info`.
+
+`CollectionSerializer#write/read` can be taken as an example.
```
LiangliangSui, line 443: Could we add a hyperlink here?

docs/protocols/xlang_object_graph_spec.md
chaokunyang: The decimal will be serialized like:
```java
buffer.writeVarUint32(value.scale());
buffer.writeVarUint32(value.precision());
buffer.writeVarUint32(bytes.length);
buffer.writeBytes(value.unscaledValue().toByteArray());
```

Could you elaborate how decimal128/decimal256 save space or provide better performance?
One thing I can see is that we can skip write `buffer.writeVarUint32(bytes.length)`, we can take binary took up 128/256 bytes. but other data are still unavoidable.

And if the unscaledValue is small, take `new BigDecimal(BigInteger.valueOf(100), 200, MathContext.DECIMAL128)` for an example, it took only 6 bytes, but when using decimal128 format, it will take 131 bytes?

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Could you elaborate more? You mean group all unsigned types together?

chaokunyang: I added format for fixed int32/64 format, could you take a look at again?

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Good suggestion! It's a good optimization for small collection.

But elements header took 4 bits, only 3 bits in a varint of a byte can be used for encoding length, the max value for it is `7`. Then we will use two bytes for length encoding, and the byte saving with this trick will be ineffective for lsit with size larger than `7`

I don't know the common size for collection, so not sure whether we should add such optimization. Either those bits left empty for future extension, or for this optimization.

@PragmaTwice @theweipeng Do you have any suggestions?

docs/protocols/xlang_object_graph_spec.md
chaokunyang: updated

docs/protocols/xlang_object_graph_spec.md
PragmaTwice: I think you can add it to the spec.

chaokunyang: updated

docs/protocols/xlang_object_graph_spec.md
chaokunyang: I think we can concate length+elements header together

theweipeng: @chaokunyang  If it is larger than 7, will the 3 bits be empty?

chaokunyang:
One more thing is that do we need to add a int16 magic number at the header? JDK serialization use two bytes, avro use 4 bytes, kryo/protobuf/flatbuffer doesn't add magic number. @theweipeng @PragmaTwice @LiangliangSui @pjfanning

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,692 @@
...
+If schema evolution mode is enabled globally or enabled for current type, type meta will be written as follows:
+
+- If meta share mode is not enabled:
+    - If type meta hasn't been written before, the data will be written as:
+      ```
+      | unsigned varint: 0b11111111 | type def |
+      ```
+    - If type meta has been written before, the data will be written as:
+      ```
+      | unsigned varint: written index << 1 | type def |
```
theweipeng, line 200: `type def` is unnecessary when it had been written before.   Perhaps  `| unsigned varint: written index << 1 ` is correct

theweipeng:
> One more thing is that do we need to add a int16 magic number at the header? JDK serialization use two bytes, avro use 4 bytes, kryo/protobuf/flatbuffer doesn't add magic number. @theweipeng @PragmaTwice @LiangliangSui @pjfanning

I think it would be necessary,  we can use the magic number to indicate the protocol version

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Nope, it will be used, but next byte will have 3 bits empty

chaokunyang:
> > One more thing is that do we need to add a int16 magic number at the header? JDK serialization use two bytes, avro use 4 bytes, kryo/protobuf/flatbuffer doesn't add magic number. @theweipeng @PragmaTwice @LiangliangSui @pjfanning
> 
> I think it would be necessary, we can use the magic number to indicate the protocol version

There are 4 bits empty in fury header, which can be used to indicate the protocol version too.

docs/protocols/xlang_object_graph_spec.md
chaokunyang: Good catch, I missed it

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,692 @@
...
+      ```
+- If meta share mode is enabled:
```
theweipeng, line 202: The only difference I noticed between the two modes is that one mode sends metadata and the other mode does not.
If the share mode is not enabled, can we write the meta data at the `meta start offset` ?  When it is disabled, we leave the `meta start offset` empty. so we can more easily understand the difference between the two modes.

theweipeng:
> > > One more thing is that do we need to add a int16 magic number at the header? JDK serialization use two bytes, avro use 4 bytes, kryo/protobuf/flatbuffer doesn't add magic number. @theweipeng @PragmaTwice @LiangliangSui @pjfanning
> > 
> > 
> > I think it would be necessary, we can use the magic number to indicate the protocol version
> 
> There are 4 bits empty in fury header, which can be used to indicate the protocol version too.

I  have noticed that  sometimes, applications may encapsulate the data buffer using a proprietary protocol. On the receiving end, the use of a magic number is helpful for them to identify which protocol is being used.

docs/protocols/xlang_object_graph_spec.md
chaokunyang: update to :

- Normal mode(meta share not enabled):
    - If type meta hasn't been written before, add `type def`
      to `captured_type_defs`: `captured_type_defs[type def] = map size`
    - Get index of the meta in `captured_type_defs`, write that index as an unsigned varint.
    - The meta will be written as:
      ```
      | unsigned varint: written index |
      ```
    - After finished the serialization of the object graph, fury will start to write `captured_type_defs`:
        - Firstly, set current to `meta start offset` of fury header
        - Then write `captured_type_defs` one by one:
          ```python
          buffer.write_var_uint32(len(writting_type_defs))
          for type_meta in writting_type_defs:
          type_meta.write_type_def(buffer)
          ```
- Meta share mode: the writing steps are same as the normal mode, but `captured_type_defs` will be shared across
  multiple serializations of different objects. For example, suppose we have a batch to serialize:
    ```python
    captured_type_defs = {}
    stream = ...
    # add `Type1` to `captured_type_defs` and write `Type1`
    fury.serialize(stream, [Type1()])
    # add `Type2` to `captured_type_defs` and write `Type2`, `Type1` is written before.
    fury.serialize(stream, [Type1(), Type2()])
    # `Type1` and `Type2` are written before, no need to write meta.
    fury.serialize(stream, [Type1(), Type2()])
    ```

- Streaming mode(streaming mode doesn't support meta share):
    - If type meta hasn't been written before, the data will be written as:
      ```
      | unsigned varint: 0b11111111 | type def |
      ```
    - If type meta has been written before, the data will be written as:
      ```
      | unsigned varint: written index << 1 |
      ```
      `written index` is the id in `captured_type_defs`.
    - With this mode, `meta start offset` can be omitted.

chaokunyang:
> > > > One more thing is that do we need to add a int16 magic number at the header? JDK serialization use two bytes, avro use 4 bytes, kryo/protobuf/flatbuffer doesn't add magic number. @theweipeng @PragmaTwice @LiangliangSui @pjfanning
> > > 
> > > 
> > > I think it would be necessary, we can use the magic number to indicate the protocol version
> > 
> > 
> > There are 4 bits empty in fury header, which can be used to indicate the protocol version too.
> 
> I have noticed that sometimes, applications may encapsulate the data buffer using a proprietary protocol. On the receiving end, the use of a magic number is helpful for them to identify which protocol is being used.

Added two bytes header to the spec

LiangliangSui:
> > > > > One more thing is that do we need to add a int16 magic number at the header? JDK serialization use two bytes, avro use 4 bytes, kryo/protobuf/flatbuffer doesn't add magic number. @theweipeng @PragmaTwice @LiangliangSui @pjfanning
> > > > 
> > > > 
> > > > I think it would be necessary, we can use the magic number to indicate the protocol version
> > > 
> > > 
> > > There are 4 bits empty in fury header, which can be used to indicate the protocol version too.
> > 
> > 
> > I have noticed that sometimes, applications may encapsulate the data buffer using a proprietary protocol. On the receiving end, the use of a magic number is helpful for them to identify which protocol is being used.
> 
> Added two bytes header to the spec

Use two bytes to store the magic number, and use two bytes to store the protocol version, just like [JDK serialization](https://docs.oracle.com/javase/8/docs/api/java/io/ObjectStreamConstants.html#STREAM_MAGIC) (maybe we can also use one byte to store the protocol version), we still keep 4 empty bits in the header(after all, the number of version numbers that can be represented by 4 bits is relatively small.) , leaving room for future expansion.

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,747 @@
...
+    stream = ...
+    # add `Type1` to `captured_type_defs` and write `Type1`
+    fury.serialize(stream, [Type1()])
+    # add `Type2` to `captured_type_defs` and write `Type2`, `Type1` is written before.
+    fury.serialize(stream, [Type1(), Type2()])
+    # `Type1` and `Type2` are written before, no need to write meta.
+    fury.serialize(stream, [Type1(), Type2()])
+    ```
+
+- Streaming mode(streaming mode doesn't support meta share):
```
theweipeng, line 226: How can we detect which mode is used? I noticed that there is no flag in the header to indicate the mode

docs/protocols/xlang_object_graph_spec.md
```diff
@@ -0,0 +1,747 @@
...
+For every type to be serialized, it must be registered with an optional ID first. The registered type will have a
+user-provided or an auto-growing unsigned int i.e. `type_id`. The registration can be used for security check and type
+identification. The id of user registered type will be added by `64` to make space for Fury internal data types.
+
+Depending on whether meta share mode and registration is enabled for current type, Fury will write type meta
+differently.
+
+### Schema consistent
+
+If schema consistent mode is enabled globally or enabled for current type, type meta will be written as a fury unsigned
```
theweipeng, line 190: How to determine whether an object is of `Schema consistent` or `Schema evolution` mode  during deserialization. There are no flags in the `type meta` header to indicate this

docs/protocols/xlang_object_graph_spec.md
chaokunyang: This is configured globally when creating fury. All the fury instance across multiple languages must have same config

docs/protocols/xlang_object_graph_spec.md
chaokunyang: I updated the spec:

- If schema consistent mode is enabled globally, type meta will be written as a fury unsigned varint of `type_id`.
  Schema evolution related meta will be ignored.
- If schema evolution mode is enabled globally and current class is configured to use schema consistent mode like struct
  vs table in flatbuffers:
    - Type meta will be add to `captured_type_defs`: `captured_type_defs[type def stub] = map size` ahead when
      registering type.
    - Get index of the meta in `captured_type_defs`, write that index as `| unsigned varint: index |`.

chaokunyang:
> > > > > > One more thing is that do we need to add a int16 magic number at the header? JDK serialization use two bytes, avro use 4 bytes, kryo/protobuf/flatbuffer doesn't add magic number. @theweipeng @PragmaTwice @LiangliangSui @pjfanning
> > > > > 
> > > > > 
> > > > > I think it would be necessary, we can use the magic number to indicate the protocol version
> > > > 
> > > > 
> > > > There are 4 bits empty in fury header, which can be used to indicate the protocol version too.
> > > 
> > > 
> > > I have noticed that sometimes, applications may encapsulate the data buffer using a proprietary protocol. On the receiving end, the use of a magic number is helpful for them to identify which protocol is being used.
> > 
> > 
> > Added two bytes header to the spec
> 
> Use two bytes to store the magic number, and use two bytes to store the protocol version, just like [JDK serialization](https://docs.oracle.com/javase/8/docs/api/java/io/ObjectStreamConstants.html#STREAM_MAGIC) (maybe we can also use one byte to store the protocol version), we still keep 4 empty bits in the header(after all, the number of version numbers that can be represented by 4 bits is relatively small.) , leaving room for future expansion.

@LiangliangSui I think 2 bytes is OK, the protocol will be backward/forward compatible in the future, we barely bump protocol version. Only things like proto2 -> proto3 need to bump protocol version. But that can be expressed by bump magic number

theweipeng (approved):
Great!

chaokunyang:
@PragmaTwice @pjfanning @theweipeng @LiangliangSui If there are no further comments anymore, I think we can merge it now.

PragmaTwice (approved):

docs/protocols/xlang_object_graph_spec.md
chaokunyang: I added this optimization to the spec since empty list and small list is common. thanks @LiangliangSui for this suggestion.

LiangliangSui (approved):
Great!, I don’t have any comments anymore, thank you for your hard work.

PR #1417, opened 2024-03-21
refactor(java): Streamline ClassResolver responsibilities

LiangliangSui:
To reduce the complexity of the `ClassResolver` class, we mainly do the following two things:

1. Move ClassId related responsibilities into ClassIdAllocator class.

2. Move `registeredId2ClassInfo` into `ExtRegistry` static inner class since they are content related.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -21,6 +21,7 @@ package org.apache.fury.resolver;
 
 import static org.apache.fury.codegen.Expression.Invoke.inlineInvoke;
 import static org.apache.fury.codegen.ExpressionUtils.eq;
+import static org.apache.fury.resolver.ClassIdAllocator.BuiltinClassId;
 import static org.apache.fury.serializer.CodegenSerializer.loadCodegenSerializer;
 import static org.apache.fury.serializer.CodegenSerializer.loadCompatibleCodegenSerializer;
 import static org.apache.fury.serializer.CodegenSerializer.supportCodegenForJavaSerialization;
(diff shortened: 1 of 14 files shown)


java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -156,55 +157,14 @@ public class ClassResolver {
...
   // use a lower load factor to minimize hash collision
   private static final float loadFactor = 0.25f;
   private static final float furyMapLoadFactor = 0.25f;
   private static final String META_SHARE_FIELDS_INFO_KEY = "shareFieldsInfo";
   private static final ClassInfo NIL_CLASS_INFO =
-      new ClassInfo(null, null, null, null, false, null, null, ClassResolver.NO_CLASS_ID);
-
+      new ClassInfo(null, null, null, null, false, null, null, BuiltinClassId.NO_CLASS_ID);
   private final Fury fury;
-  private ClassInfo[] registeredId2ClassInfo = new ClassInfo[] {};
```
chaokunyang, line 207: `registeredId2ClassInfo` is in the critical path, we'd better to keep it here instead of `extRegistry`, which incur another pointer indirection.

chaokunyang:
Hi @LiangliangSui , thanks for this work. We're standardlizing Fury xlang serialization protocol in #1413 , which may change class resolver too, especially the class id part. Can we hold this until the xlang serialization protocol are merged, so we don't have to refactor `ClassResolver` again.

The type id is lift to the first-class concept in #1413, we do need to strealine it from class resolver.

LiangliangSui:
Ok, we keep this PR until the xlang serialization protocol is merged.

PR #1420, opened 2024-03-23
feat(java): carry read objects when deserialization fail for better trouble shooting

### LINKED ISSUE #1419: [Java] print read objects when deserialization fail for better trouble shooting
Currently when deserialization fail, it's hard to locate which object deserialization failed. We must debug locally line by line to find the issue.

We can carry with read objects in exception, so when this exception  is printed, we can know which objects are serialized correctly. Then we more close to the root cause.

chaokunyang:
This PR carry read objects when deserialization fail for better trouble shooting.

Closes #1419


theweipeng (approved):
LGTM

PR #1422, opened 2024-03-24
feat(java): implement define_class insteadof using javaassist

chaokunyang:
This PR implements `defineClass` insteadof using javaassist to reduce the dependency


theweipeng (approved):
LGTM

PR #1423, opened 2024-03-24
feat(java): skip print exception when class not exist

chaokunyang:
skip print exception when `deserializeUnexistedClass` option enabled and class not exist


theweipeng (approved):
LGTM

PR #1424, opened 2024-03-24
refactor(java): remove registerWithCheck function

LiangliangSui:
Remove `registerWithCheck` function and register function enables duplicate registration check by default. When Class is repeatedly registered, an `IllegalArgumentException` is thrown.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -372,8 +372,6 @@ public class ClassResolver {
     register(ConcurrentHashMap.class);
     register(ArrayBlockingQueue.class);
     register(LinkedBlockingQueue.class);
-    register(Boolean[].class, Byte[].class, Short[].class, Character[].class);
-    register(Integer[].class, Float[].class, Long[].class, Double[].class);
     register(AtomicBoolean.class);
     register(AtomicInteger.class);
     register(AtomicLong.class);
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -263,46 +263,42 @@ public class ClassResolver {
     register(LambdaSerializer.ReplaceStub.class, LAMBDA_STUB_ID);
     register(JdkProxySerializer.ReplaceStub.class, JDK_PROXY_STUB_ID);
     register(ReplaceResolveSerializer.ReplaceStub.class, REPLACE_STUB_ID);
-    registerWithCheck(void.class, PRIMITIVE_VOID_CLASS_ID);
-    registerWithCheck(boolean.class, PRIMITIVE_BOOLEAN_CLASS_ID);
-    registerWithCheck(byte.class, PRIMITIVE_BYTE_CLASS_ID);
-    registerWithCheck(char.class, PRIMITIVE_CHAR_CLASS_ID);
-    registerWithCheck(short.class, PRIMITIVE_SHORT_CLASS_ID);
-    registerWithCheck(int.class, PRIMITIVE_INT_CLASS_ID);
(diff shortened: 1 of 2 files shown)


java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -356,20 +352,12 @@ private void addDefaultSerializer(Class type, Serializer serializer) {
     register(type);
   }
 
-  /** Register common class ahead to get smaller class id for serialization. */
-  private void registerCommonUsedClasses() {
```
chaokunyang, line 360: Those classes shouldn't be removed, otherwise they will be allocated a bigger ID in `addDefaultSerializers`, which will be serialized using 2 bytes. 

The principle is to ensure frequently used classes get a smaller ID to make them be serialized in a byte, which is the range: `0~63`

chaokunyang: Maybe we should add a comment for this

LiangliangSui: These classes have not been deleted. Since there are some duplicate registered classes in `registerCommonUsedClasses` and `initialize()`, deduplication has been carried out in `registerCommonUsedClasses` (For example, `Class.class` is registered once in `initialize()`, and then registered again in `registerCommonUsedClasses()`, so the registration of `Class.class` is deleted in `registerCommonUsedClasses()`).

Then, I moved all the classes registered in the cleaned `registerCommonUsedClasses()` to `registerDefaultClasses()` (because currently There are a lot of places to register classes), and then called `registerDefaultClasses()` before calling `addDefaultSerializers()` in `initialize()`

This still ensures that these classes are registered before executing `addDefaultSerializers()`.

chaokunyang: I see, thanks

chaokunyang: Another issue to note is that there are some classes registered in `ClassResolver#addDefaultSerializers` which may be more frequent than classes in `registerDefaultClasses`:
```java
    register(Boolean[].class, Byte[].class, Short[].class, Character[].class);
    register(Integer[].class, Float[].class, Long[].class, Double[].class);
    register(AtomicBoolean.class);
    register(AtomicInteger.class);
    register(AtomicLong.class);
    register(AtomicReference.class);
```
Could we register such frequent classes in `registerDefaultClasses` ahead to avoid the less frequent classes such as `AtomicReference` to use an id in range `0~63`?

chaokunyang: I mean Move frequently used classes in common user scenarios to the front of `registerDefaultClasses`

chaokunyang: This may be a little complex, we can merge this PR first, and I can help submit a pr to fix it later

LiangliangSui: > I mean Move frequently used classes in common user scenarios to the front of `registerDefaultClasses`

I got it, in this way, classes that users frequently use can be assigned smaller ClassIds.

We can keep the `registerCommonUsedClasses` function and then move the frequently used classes by users in `registerDefaultClasses` to `registerCommonUsedClasses` and then call `registerCommonUsedClasses` before `registerDefaultClasses` (For example, move `Boolean[].class, Byte[].class, Short[].class, Character[].class`, etc. to `registerCommonUsedClasses` for registration.), is that so?

chaokunyang: Yes, that's what I mean.

LiangliangSui: Ok, I will do this work in this pr.

LiangliangSui: Updated.

Issue #1389, opened 2024-02-28
[Release] Tracking issue for Fury v0.5.0 release, the first release under ASF


chaokunyang:
## Is your feature request related to a problem? Please describe.
This is a tracking issue for Fury 0.5.0 release, our first release since Fury joined ASF incubator.

Release plan:
| Component | Old version | New version|
|-------------|-------------|------------|
| Java|  0.4.1 |  0.5.0|
| Python |  0.4.1 |  0.5.0 | 0.5.0 |
| Javascript |  0.4.1 |  not released |
| Golang | 0.4.1 |  not released |
| C++ | not released | not released |
| Rust| not released | not released |

## Work items
- [ ] #1421

PR #1425, opened 2024-03-25
fix: fix license issue for release

chaokunyang:
- set license to Apache 2
- add disclaimer: according to https://incubator.apache.org/guides/distribution.html, the release of incubating project should carry with a disclaimer.

 Releated issue: #1389

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -37,6 +37,18 @@
   <name>Fury Project Parent POM</name>
   <description>
     Apache Fury™ is a blazing fast multi-language serialization framework powered by jit and zero-copy.
+
+    Apache Fury (incubating) is an effort undergoing incubation at the Apache
+    Software Foundation (ASF), sponsored by the Apache Incubator PMC.
+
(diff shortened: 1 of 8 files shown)


java/pom.xml
```diff
@@ -37,6 +37,17 @@
   <name>Fury Project Parent POM</name>
   <description>
     Apache Fury™ is a blazing fast multi-language serialization framework powered by jit and zero-copy.
+
+    Apache Fury (incubating) is an effort undergoing incubation at the Apache
```
tisonkun, line 41:
```suggestion
    Fury is an effort undergoing incubation at the Apache
```

Already "Apache Fury™" above, let's keep a consistent tidy format.

tisonkun: Oh. It's about Javadoc footer? Than let's keep it as is.

tisonkun (approved):

java/pom.xml
chaokunyang: It's not about Javadoc footer, I will submit a PR later to fix this

PR #1427, opened 2024-03-27
fix: update CI actions

pjfanning:
avoid actions that use node16

https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

also - only publish if running on main fork - not on user forks


chaokunyang (approved):
LGTM

PR #1428, opened 2024-03-27
fix: use blazingly instead of blazing

pjfanning:
As a native English speaker, the existing text doesn't read properly for me.

Blazing is being used as an adverb so should become Blazingly.

https://en.wiktionary.org/wiki/blazingly

For instance, if you wanted to use `incredible` - you would say `incredibly fast`, not `incredible fast`.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -19,7 +19,7 @@
 
 github:
   description: >-
-    A blazing fast multi-language serialization framework powered by JIT and zero-copy.
+    A blazingly fast multi-language serialization framework powered by JIT and zero-copy.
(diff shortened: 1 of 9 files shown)


chaokunyang (approved):
LGTM, thanks. Rust is using `blazingly` too.

chaokunyang:
Is there a way to change this too?
image

pjfanning:
The .asf.yaml file is used to update GitHub settings like the project description. I think the value is updated now because I included a change in .asf.yaml in this PR.

chaokunyang:
Great, thanks

PR #1434, opened 2024-03-28
chore(java): Reuse unsafePutPositiveVarInt in unsafeWritePositiveVarInt

LiangliangSui:
N/A

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
@@ -1233,49 +1233,9 @@ public final class MemoryBuffer {
    * to avoid using two memory operations.
    */
   public int unsafeWritePositiveVarInt(int v) {
-    // The encoding algorithm are based on kryo UnsafeMemoryOutput.writeVarInt
-    // varint are written using little endian byte order.
-    // This version should have better performance since it remove an index update.
-    long value = v;
-    final int writerIndex = this.writerIndex;
-    long varInt = (value & 0x7F);
-    value >>>= 7;
-    if (value == 0) {
-      UNSAFE.putByte(heapMemory, address + writerIndex, (byte) varInt);
-      this.writerIndex = writerIndex + 1;
-      return 1;
-    }
-    // bit 8 `set` indicates have next data bytes.
-    varInt |= 0x80;
-    varInt |= ((value & 0x7F) << 8);
-    value >>>= 7;
-    if (value == 0) {
-      unsafePutInt(writerIndex, (int) varInt);
-      this.writerIndex = writerIndex + 2;
-      return 2;
-    }
-    varInt |= (0x80 << 8);
-    varInt |= ((value & 0x7F) << 16);
-    value >>>= 7;
-    if (value == 0) {
-      unsafePutInt(writerIndex, (int) varInt);
-      this.writerIndex = writerIndex + 3;
-      return 3;
-    }
-    varInt |= (0x80 << 16);
-    varInt |= ((value & 0x7F) << 24);
-    value >>>= 7;
-    if (value == 0) {
-      unsafePutInt(writerIndex, (int) varInt);
-      this.writerIndex = writerIndex + 4;
-      return 4;
-    }
-    varInt |= (0x80L << 24);
-    varInt |= ((value & 0x7F) << 32);
-    varInt &= 0xFFFFFFFFFL;
-    unsafePutLong(writerIndex, varInt);
-    this.writerIndex = writerIndex + 5;
-    return 5;
+    int varintBytes = unsafePutPositiveVarInt(writerIndex, v);
+    writerIndex += varintBytes;
+    return varintBytes;
   }
 
   /**


chaokunyang (approved):
LGTM

java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
```diff
@@ -1233,49 +1233,9 @@ public int readVarInt() {
...
-      this.writerIndex = writerIndex + 4;
-      return 4;
-    }
-    varInt |= (0x80L << 24);
-    varInt |= ((value & 0x7F) << 32);
-    varInt &= 0xFFFFFFFFFL;
-    unsafePutLong(writerIndex, varInt);
-    this.writerIndex = writerIndex + 5;
-    return 5;
+    int varintBytes = unsafePutPositiveVarInt(writerIndex, v);
```
chaokunyang, line 1236: JVM JIT inline has max depth limit, which is 9 by default. Reuse this method may disabled some jit inline. Could we benchmark `org.apache.fury.benchmark.UserTypeDeserializeSuite` to check this code reuse doesn't introduce a preformance regression?

```java
  public static void main(String[] args) throws IOException {
    if (args.length == 0) {
      String commandLine =
          "io.*UserTypeDeserializeSuite.fury* -f 0 -wi 3 -i 3 -t 1 -w 2s -r 2s -rf csv";
      System.out.println(commandLine);
      args = commandLine.split(" ");
    }
    Main.main(args);
  }
```
This will run fury serialization only

LiangliangSui: Okay, I will do this check later.

LiangliangSui: After passing the benchmark, it can still be inlined and there is no performance regression.

image

The difference between the two is 155266, which should be within the error.

In addition, I saw in the log that `unsafePutPositiveVarInt` can be inlined in the benchmark process by using `-XX:+PrintCompilation -XX:+UnlockDiagnosticVMOptions -XX:+PrintInlinin`

```
@ 152   java.lang.Integer::intValue (5 bytes)   accessor
@ 155   org.apache.fury.memory.MemoryBuffer::writeVarInt (17 bytes)   inline (hot)
  @ 8   org.apache.fury.memory.MemoryBuffer::ensure (37 bytes)   inline (hot)
  @ 13   org.apache.fury.memory.MemoryBuffer::unsafeWriteVarInt (15 bytes)   inline (hot)
    @ 11   org.apache.fury.memory.MemoryBuffer::unsafeWritePositiveVarInt (22 bytes)   inline (hot)
      @ 6   org.apache.fury.memory.MemoryBuffer::unsafePutPositiveVarInt (208 bytes)   inline (hot)
        @ 39   sun.misc.Unsafe::putByte (11 bytes)   force inline by annotation
          @ 7   jdk.internal.misc.Unsafe::putByte (0 bytes)   (intrinsic)
        @ 81   org.apache.fury.memory.MemoryBuffer::unsafePutInt (45 bytes)   inline (hot)
          @ 23   sun.misc.Unsafe::putInt (11 bytes)   force inline by annotation
            @ 7   jdk.internal.misc.Unsafe::putInt (0 bytes)   (intrinsic)
```

chaokunyang: This looks good, does this have an influence in a end-to-end benchmark? JVM inline has max depth limit, the current case shouldn't be influenced by this optimization. Could you run fury benchmark for MediaContent object?

LiangliangSui: with this pr
```
Benchmark                                         (bufferType)   (objectType)  (references)   Mode  Cnt        Score        Error  Units
UserTypeSerializeSuite.fury_serialize_compatible         array  MEDIA_CONTENT         false  thrpt    9  4587893.508 ±  42116.854  ops/s
UserTypeSerializeSuite.fury_serialize_compatible         array  MEDIA_CONTENT          true  thrpt    9  3177202.586 ±  14736.954  ops/s
UserTypeSerializeSuite.fury_serialize_compatible  directBuffer  MEDIA_CONTENT         false  thrpt    9  4562769.304 ± 340601.623  ops/s
UserTypeSerializeSuite.fury_serialize_compatible  directBuffer  MEDIA_CONTENT          true  thrpt    9  3079333.765 ±  61135.629  ops/s
```

before this pr
```
Benchmark                                         (bufferType)   (objectType)  (references)   Mode  Cnt        Score        Error  Units
UserTypeSerializeSuite.fury_serialize_compatible         array  MEDIA_CONTENT         false  thrpt    9  4602232.298 ± 107310.721  ops/s
UserTypeSerializeSuite.fury_serialize_compatible         array  MEDIA_CONTENT          true  thrpt    9  3141775.420 ±  59485.987  ops/s
UserTypeSerializeSuite.fury_serialize_compatible  directBuffer  MEDIA_CONTENT         false  thrpt    9  4755918.332 ±  65920.575  ops/s
UserTypeSerializeSuite.fury_serialize_compatible  directBuffer  MEDIA_CONTENT          true  thrpt    9  3118601.861 ± 140368.292  ops/s
```

I run fury benchmark for MediaContent object, there should be no degradation in performance, within the error range.

chaokunyang: Great! Thanks for taking time to benchmark this.

Issue #1433, opened 2024-03-28
Execute an error in the openj9 environment


mintonzhang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-fury/issues) and found no similar issues.

### Version

fury-core: 0.4.0
openjdk version "17.0.10" 2024-01-16
IBM Semeru Runtime Open Edition 17.0.10.0 (build 17.0.10+7)
Eclipse OpenJ9 VM 17.0.10.0 (build openj9-0.43.0, JRE 17 Mac OS X amd64-64-Bit Compressed References 20240116_636 (JIT enabled, AOT enabled)
OpenJ9   - 2c3d78b48
OMR      - ea8124dbc
JCL      - 2aad089841f based on jdk-17.0.10+7)

### Component(s)

Java

### Minimal reproduce step

```java

    @Getter
    @Setter
    public static class UserCacheInfoTest {

        private String extraUserId;

        private String unionId;

        private String name;

        private String jobNumber;

        private String email;

        private Long userId;

        private String mobile;

        private String avatar;

        private String position;

        private String deptName;
    }

    public static void main(String[] args) {
        UserCacheInfoTest userCacheInfoTest = new UserCacheInfoTest();
        userCacheInfoTest.setExtraUserId("1");
        userCacheInfoTest.setUnionId("1");
        userCacheInfoTest.setName("1");
        userCacheInfoTest.setJobNumber("1");
        userCacheInfoTest.setEmail("1");
        userCacheInfoTest.setUserId(0L);
        userCacheInfoTest.setMobile("1");
        userCacheInfoTest.setAvatar("1");
        userCacheInfoTest.setPosition("1");
        userCacheInfoTest.setDeptName("1");
        //ThreadLocalFury fury = new ThreadLocalFury(classLoader -> );

/
        Fury fury = Fury.builder()
                .withLanguage(Language.JAVA)
                .withCodegen(true)
                //.registerGuavaTypes(true)
                .withCompatibleMode(CompatibleMode.COMPATIBLE)
                .requireClassRegistration(false)
                .suppressClassRegistrationWarnings(true)
                .build();



        byte[] serialize = fury.serialize(userCacheInfoTest);

        UserCacheInfoTest deserialize = (UserCacheInfoTest) fury.deserialize(serialize);

        System.out.println();
    }
```

### What did you expect to see?

Normal serialization and deserialization

### What did you see instead?

```
[...]

PR #1437, opened 2024-03-31
feat(java): add openj9 ci

chaokunyang:
This PR add openj9 ci, since we have openj9 users, see #1433 . Class GC in linux distribution are not supported in this PR.


theweipeng (approved):

java/fury-core/src/main/java/org/apache/fury/util/unsafe/_JDKAccess.java
```diff
@@ -65,7 +65,7 @@ public class _JDKAccess {
       property = property.substring(2);
     }
     String jmvName = System.getProperty("java.vm.name", "");
```
LiangliangSui, line 67: Nit: `jmvName` should be renamed to `jvmName`.

chaokunyang: Nice catch, fixed in https://github.com/apache/incubator-fury/pull/1439

PR #1450, opened 2024-04-01
chore(java): add comments for loop unrolling.

LiangliangSui:
N/A

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
@@ -1239,6 +1239,9 @@ public final class MemoryBuffer {
   }
 
   private int readPositiveVarIntSlow() {
+    // Note:
+    //  Loop are not used here to improve performance.
+    //  We manually unroll the loop for better performance.
     int b = readByte();
     int result = b & 0x7F;
     if ((b & 0x80) != 0) {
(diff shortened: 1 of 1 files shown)


java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
```diff
@@ -1316,24 +1316,15 @@ public int readPositiveVarInt() {
   }
 
   private int readPositiveVarIntSlow() {
-    int b = readByte();
-    int result = b & 0x7F;
-    if ((b & 0x80) != 0) {
+    int shift = 0;
```
chaokunyang, line 1319: Could we benchmark the performance of those two implementation using JMH? The previous one are loop unrolled for performance

chaokunyang: The implementation in this class are kind of messy, we manually inline code and unrolled some loop for performance

chaokunyang: Yes, could you add some comments in this pr? I think we'd better add comments to similar methods

LiangliangSui: Updated.

BTW, do we need to add the benchmark code (sample code above) that uses loop and does not use loop in the `MemoryBuffer#readPositiveVarIntSlow` method to the benchmark directory?

chaokunyang: I don't think it's necessary, only rolling code will be used in the release

chaokunyang:
@LiangliangSui There are some conflict between your branch and main branch, could you resolve the conflict?

LiangliangSui:
> @LiangliangSui There are some conflict between your branch and main branch, could you resolve the conflict?

Updated, ptal.

chaokunyang (approved):
LGTM

PR #1467, opened 2024-04-04
feat(java): extract public Fury methods to BaseFury

Munoon:
Extract public Fury method to the `BaseFury` interface and implement them in `ThreadPoolFury` and `ThreadLocalFury`, so end users can avoid using `execute()` and other utility methods.

Off topic: further steps might be changing `BaseFury` interface name to `Fury` while existing class renaming to something like `FuryImpl`. `FuryBuilder` should return the new `Fury` interface, so we will hide system methods like `xwriteRef` and others. WDYT?

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/BaseFury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/BaseFury.java
@@ -19,8 +19,12 @@
 
 package org.apache.fury;
 
+import java.io.InputStream;
+import java.io.OutputStream;
 import org.apache.fury.memory.MemoryBuffer;
+import org.apache.fury.serializer.BufferCallback;
 import org.apache.fury.serializer.Serializer;
+import org.apache.fury.serializer.SerializerFactory;
 import org.apache.fury.serializer.Serializers;
 
 /** All Fury’s basic interface, including Fury’s basic methods. */
@@ -55,6 +59,9 @@ public interface BaseFury {
    */
   void register(Class<?> cls, Short id, boolean createSerializer);
 
+  /** register class with given type tag which will be used for cross-language serialization. */
+  void register(Class<?> cls, String typeTag);
+
   /**
    * Register a Serializer.
    *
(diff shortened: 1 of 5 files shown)


java/fury-core/src/main/java/org/apache/fury/BaseFury.java
```diff
@@ -64,9 +71,16 @@ public interface BaseFury {
    */
   <T> void registerSerializer(Class<T> type, Class<? extends Serializer> serializerClass);
 
+  void registerSerializer(Class<?> type, Serializer<?> serializer);
```
chaokunyang, line 74: The serializer is bind to a Fury instance, we may not be able to create a serializer object for registration unless we change the serializer interfere. Could we change this to register(Class type, Class serializer class)?

java/fury-core/src/main/java/org/apache/fury/BaseFury.java
```diff
@@ -55,6 +59,9 @@ public interface BaseFury {
    */
   void register(Class<?> cls, Short id, boolean createSerializer);
 
+  /** register class with given type tag which will be used for cross-language serialization. */
+  void register(Class<?> cls, String typeTag);
```
chaokunyang, line 63: This API is used for clang serialization, we removed it in our new xlang serialization spec in #1413 . Could we remove this API? We can add  new API when new protocol are implemented

java/fury-core/src/main/java/org/apache/fury/BaseFury.java
Munoon: There might be a singleton serializer. I've personally create some of them in my project. In this case they will be applied to each existing and each new Fury instances in `ThreadLocalFury` and `ThreadPoolFury`.
Maybe this is something that worth mentioning in the documentation.

However, if you didn't like it, we can remove this method and make user register singleton serializers using `execute` method. In this case, there is more chances that user will know what he is doing.

chaokunyang: I see, thanks. This is reasonable to add to BaseFury

chaokunyang (approved):
LGTM

PR #1468, opened 2024-04-06
feat(spec): add xlang type mapping

chaokunyang:
This PR add xlang serialization type mapping to complete the xlang spec in #1413 .

The changes include:
- Xlang type mapping
- Remove fixed_int32/fixed_int64 type since it's duplicate with `int32`/`int64`


chaokunyang:
BYW, I was thinking whether we should make all type id starts from 0, and let users to register type id from 64. In this way, we don't have to add ID by 64, and there won't be two kinds of ID to understand. One is the registed ID, the other is internal ID. But the drawback is that users must understand the concept of 64 and the pre-registered ID.

theweipeng (approved):

docs/guide/xlang_type_mapping.md
```diff
@@ -4,4 +4,75 @@ sidebar_position: 3
...
+| bool_array         | 21      | bool[]          | ndarray(np.bool_)    | /               | `bool[x]`                      | `[x]bool/[]~`    | `Vec<bool>`      |
+| int16_array        | 22      | short[]         | ndarray(int16)       | /               | `int16_t[x]/vector<~>`         | `[x]int16/[]~`   | `Vec<i16>`       |
+| int32_array        | 23      | int[]           | ndarray(int32)       | /               | `int32_t[x]/vector<~>`         | `[x]int32/[]~`   | `Vec<i32>`       |
+| int64_array        | 24      | long[]          | ndarray(int64)       | /               | `int64_t[x]/vector<~>`         | `[x]int64/[]~`   | `Vec<i64>`       |
+| float16_array      | 25      | float[]         | ndarray(float16)     | /               | `fury::float16_t[x]/vector<~>` | `[x]float16/[]~` | `Vec<fury::f16>` |
+| float32_array      | 26      | float[]         | ndarray(float32)     | /               | `float[x]/vector<~>`           | `[x]float32/[]~` | `Vec<f32>`       |
+| float64_array      | 27      | float[]         | ndarray(float64)     | /               | `double[x]/vector<~>`          | `[x]float64/[]~` | `Vec<f64>`       |
```
LiangliangSui, line 43: `float[]` should be `double[]`

docs/guide/xlang_type_mapping.md
```diff
@@ -4,4 +4,75 @@ sidebar_position: 3
...
+# Type annotation(not implemented)
+
+Due to differences between type systems of languages, those types can't be mapped one-to-one between languages. Users
+can provide meta hints for fields of a type, or for the whole type. Such information can be provided in other languages
+too:
+
+- java: use annotation.
+- cpp: use macro and template.
+- golang: use struct tag.
+- python: use typehint.
```
LiangliangSui, line 58: Can python also support using decorator to achieve this function?

docs/guide/xlang_type_mapping.md
```diff
@@ -4,4 +4,75 @@ sidebar_position: 3
...
+    ```java
+    class Foo {
+      @Int32Type(varint = true)
+      int f1;
+      List<@Int32Type(varint = true) Integer> f2;
+    }
+    ```
+- Python:
+
+```python
```
LiangliangSui, line 73: Nit: should be indented

chaokunyang:
@LiangliangSui @theweipeng @PragmaTwice I will revert this PR first, and submit a new PR for review.

PR #1469, opened 2024-04-06
feat(java): reduce readInt/readVarInt code size for for jvm jit inline

chaokunyang:
This PR optimizes readVarInt/readInt performance by reduce code size of readVarInt for better jvm jit


chaokunyang:
Before:
image
After:
image

theweipeng (approved):

PR #1451, opened 2024-04-01
feat(java): native streaming mode deserialization

chaokunyang:
## What this PR do?
This PR implements streaming mode deserialization, which make the protocol between `Fury#serialize(buffer, o)` and `Fury#serialize(stream, o)` consistent, and make the deserialization support real streaming mode to reduce latency, and facilitate the pipeline between deserialization and transfer.

The changes inlucde:
- Introduce a `FuryStreamReader` to allow buffered streaming reading.
- Make `MemoryBuffer` support  streaming reading without introduce cost on critical path.
- Implements `InputStream` based `FuryStreamReader` for Fury.

`java.nio.Channel` based zero-copy is not implemented in this PR.

## Related issues
Closes #1435

## Media Benchmark
with this PR:
```
Benchmark                                  (bufferType)   (objectType)  (references)   Mode  Cnt        Score       Error  Units
UserTypeDeserializeSuite.fury_deserialize         array  MEDIA_CONTENT         false  thrpt   50  2867721.514 ± 55102.830  ops/s
UserTypeDeserializeSuite.fury_deserialize         array  MEDIA_CONTENT          true  thrpt   50  1841586.657 ± 46192.190  ops/s
UserTypeDeserializeSuite.fury_deserialize  directBuffer  MEDIA_CONTENT         false  thrpt   50  2542987.512 ± 49835.129  ops/s
UserTypeDeserializeSuite.fury_deserialize  directBuffer  MEDIA_CONTENT          true  thrpt   50  1712919.362 ± 26500.068  ops/s
```
Before this PR: 4c60eb9780ec51ffe1aac3ee2da594c0762799e0
```
Benchmark                                  (bufferType)   (objectType)  (references)   Mode  Cnt        Score       Error  Units
UserTypeDeserializeSuite.fury_deserialize         array  MEDIA_CONTENT         false  thrpt   50  2971081.559 ± 52348.975  ops/s
UserTypeDeserializeSuite.fury_deserialize         array  MEDIA_CONTENT          true  thrpt   50  1938252.031 ± 25077.333  ops/s
UserTypeDeserializeSuite.fury_deserialize  directBuffer  MEDIA_CONTENT         false  thrpt   50  2511729.964 ± 39278.437  ops/s
UserTypeDeserializeSuite.fury_deserialize  directBuffer  MEDIA_CONTENT          true  thrpt   50  1714610.534 ± 29778.339  ops/s
```

Baseline(7bb21d06):
```
[...]

PR #1472, opened 2024-04-06
feat(java): optimize read float/double for jvm jit inline

chaokunyang:
- optimize read float jvm jit inline by separate little/big endian methods
- optimize read double jvm jit inline by separate little/big endian methods
- generate unsafe get float code online
- generate unsafe get double code online

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
@@ -522,12 +522,6 @@ public final class MemoryBuffer {
     }
   }
 
-  public short getShortN(int index) {
-    final long pos = address + index;
-    checkPosition(index, pos, 2);
-    return UNSAFE.getShort(heapMemory, pos);
-  }
-
   /** Get short in big endian order from provided buffer. */
   public static short getShortB(byte[] b, int off) {
     return (short) ((b[off + 1] & 0xFF) + (b[off] << 8));
@@ -2150,6 +2144,36 @@ public final class MemoryBuffer {
       return Short.reverseBytes(UNSAFE.getShort(heapMemory, pos));
     }
   }
+
+
+  // Reduce method body for better inline in the caller.
+  @CodegenInvoke
+  public short readShortOnLE() {
+    int readerIdx = readerIndex;
+    // use subtract to avoid overflow
+    int remaining = size - readerIdx;
+    if (remaining < 2) {
+      throw new IndexOutOfBoundsException(
+        String.format(
+          "readerIndex(%d) + length(%d) exceeds size(%d): %s", readerIdx, 2, size, this));    }
+    readerIndex = readerIdx + 2;
+    return UNSAFE.getShort(heapMemory, address + readerIdx);
+  }
+
+  // Reduce method body for better inline in the caller.
+  @CodegenInvoke
+  public short readShortOnBE() {
+    int readerIdx = readerIndex;
+    // use subtract to avoid overflow
+    int remaining = size - readerIdx;
+    if (remaining < 2) {
+      throw new IndexOutOfBoundsException(
+        String.format(
+          "readerIndex(%d) + length(%d) exceeds size(%d): %s", readerIdx, 2, size, this));    }
+    readerIndex = readerIdx + 2;
+    return Short.reverseBytes(UNSAFE.getShort(heapMemory, address + readerIdx));
+  }
+
 
   public int readInt() {
     int readerIdx = readerIndex;
--- a/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
@@ -2131,6 +2131,36 @@ public final class MemoryBuffer {
     }
   }
 
+  // Reduce method body for better inline in the caller.
+  @CodegenInvoke
+  public char readCharOnLE() {
+    int readerIdx = readerIndex;
+    // use subtract to avoid overflow
+    int remaining = size - readerIdx;
+    if (remaining < 2) {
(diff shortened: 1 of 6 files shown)


theweipeng (approved):

java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
```diff
@@ -2318,6 +2205,37 @@ public float readFloat() {
...
   }
 
+  // Reduce method body for better inline in the caller.
+  @CodegenInvoke
+  public float readFloatOnLE() {
+    int readerIdx = readerIndex;
+    // use subtract to avoid overflow
+    int remaining = size - readerIdx;
+    if (remaining < 4) {
+      throw new IndexOutOfBoundsException(
```
LiangliangSui, line 2215: Nit: we can reuse [`throwIndexOutOfBoundsException`](https://github.com/apache/incubator-fury/blob/main/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java#L1853-L1857). `throwIndexOutOfBoundsException` is not executed frequently and should have no impact on inlining.

java/fury-core/src/test/java/org/apache/fury/serializer/PrimitiveSerializersTest.java
```diff
@@ -54,4 +59,62 @@ public void testUint16Serializer() {
     assertThrows(IllegalArgumentException.class, () -> serializer.xwrite(buffer, -1));
     assertThrows(IllegalArgumentException.class, () -> serializer.xwrite(buffer, 65536));
   }
+
+  @Data
+  @AllArgsConstructor
+  public static class PrimitiveStruct {
```
LiangliangSui, line 65: Should be private.
```java
private static class PrimitiveStruct
```

java/fury-core/src/test/java/org/apache/fury/serializer/PrimitiveSerializersTest.java
```diff
@@ -54,4 +59,62 @@ public void testUint16Serializer() {
...
+            Integer.MIN_VALUE,
+            Integer.MIN_VALUE,
+            Long.MIN_VALUE,
+            Long.MIN_VALUE,
+            Float.MIN_VALUE,
+            Float.MIN_VALUE,
+            Double.MIN_VALUE,
+            Double.MIN_VALUE);
+    if (compressNumber) {
+      FuryBuilder builder =
```
LiangliangSui, line 101: Nit: We can move the common part of FuryBuilder before the if branch, so that if and else can share this part

LiangliangSui:
Hi @chaokunyang , I have a suggestion. For this kind of optimization PR, we should force the submitter to provide a benchmark data in the PR, so that we can ensure that the optimization is beneficial to the codebase.

Then for the PR of the optimization category, we should force the PR title to indicate that it is an optimization PR (standardizing the Fury community). We can add a new category `opt` in [pr-lint.yml](https://github.com/apache/incubator-fury/blob/main/.github/workflows/pr-lint.yml), and I will submit a PR to do this.

java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
chaokunyang: This is temprary , it will be replaced by `streamReader.fillBuffer` in #1451. The exception will be trown in `DoundChecker`

java/fury-core/src/test/java/org/apache/fury/serializer/PrimitiveSerializersTest.java
chaokunyang: Private will work here too, but the generate code would be different. Codegen work best for public classes

chaokunyang:
> Hi @chaokunyang , I have a suggestion. For this kind of optimization PR, we should force the submitter to provide a benchmark data in the PR, so that we can ensure that the optimization is beneficial to the codebase.
> 
> Then for the PR of the optimization category, we should force the PR title to indicate that it is an optimization PR (standardizing the Fury community). We can add a new category `opt` in [pr-lint.yml](https://github.com/apache/incubator-fury/blob/main/.github/workflows/pr-lint.yml), and I will submit a PR to do this.

How abouting add a `perf` category, as described in https://platform.uno/docs/articles/uno-development/git-conventional-commits.html.  `opt` can represent many things, like `refactor`

chaokunyang:
Here is the benchmark result:
```
With This PR:
Benchmark                                  (bufferType)   (objectType)  (references)   Mode  Cnt        Score        Error  Units
UserTypeDeserializeSuite.fury_deserialize         array  Struct         false  thrpt   10  4466936.113 ± 386617.618  ops/s
UserTypeDeserializeSuite.fury_deserialize         array  Struct          true  thrpt   10  3963160.097 ± 178393.730  ops/s
UserTypeDeserializeSuite.fury_deserialize  directBuffer  Struct         false  thrpt   10  4879409.124 ±  79998.330  ops/s
UserTypeDeserializeSuite.fury_deserialize  directBuffer  Struct          true  thrpt   10  4140782.389 ±  95492.701  ops/s

Before this PR:
Benchmark                                  (bufferType)   (objectType)  (references)   Mode  Cnt        Score        Error  Units
UserTypeDeserializeSuite.fury_deserialize         array  Struct         false  thrpt   10  4135698.223 ± 344998.039  ops/s
UserTypeDeserializeSuite.fury_deserialize         array  Struct          true  thrpt   10  4062397.644 ±  94063.642  ops/s
UserTypeDeserializeSuite.fury_deserialize  directBuffer  Struct         false  thrpt   10  4311519.019 ± 251129.086  ops/s
UserTypeDeserializeSuite.fury_deserialize  directBuffer  Struct          true  thrpt   10  4056600.395 ± 414134.048  ops/s
```

PR #1474, opened 2024-04-07
docs: Use perf type when pr has performance impact.

LiangliangSui:
When PR has an impact on Fury performance, it is recommended to use perf type and provide benchmark data in the PR description.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,10 @@ For example, here are good PR titles:
 - fix(c++): blablabla
 - chore(python): remove useless yyy file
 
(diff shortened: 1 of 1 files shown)


chaokunyang (approved):
LGTM, We may need to update our pr template too

LiangliangSui:
> LGTM, We may need to update our pr template too

The pr template already contains [perf type](https://github.com/apache/incubator-fury/blob/main/.github/workflows/pr-lint.yml#L48).

chaokunyang:
I mean a pr template like: https://github.com/apache/arrow/blob/main/.github/pull_request_template.md

LiangliangSui:
Okay, I got it, I will update later.

PR #1476, opened 2024-04-07
feat(doc): add type mapping for xlang serialization

chaokunyang:
add type mapping for xlang serialization


docs/guide/xlang_type_mapping.md
```diff
@@ -4,4 +4,76 @@ sidebar_position: 3
...
+      List<@Int32Type(varint = true) Integer> f2;
+    }
+    ```
+- Python:
+
+```python
+class Foo:
+    f1: Int32Type(varint=True)
+    f2: List[Int32Type(varint=True)]
+```
```
LiangliangSui, line 78:
```suggestion
- Python:

    ```python
    class Foo:
        f1: Int32Type(varint=True)
        f2: List[Int32Type(varint=True)]
    ```
```

LiangliangSui (changes requested):

docs/guide/xlang_type_mapping.md
```diff
@@ -4,4 +4,76 @@ sidebar_position: 3
...
+# Type annotation(not implemented)
+
+Due to differences between type systems of languages, those types can't be mapped one-to-one between languages. Users
+can provide meta hints for fields of a type, or for the whole type. Such information can be provided in other languages
+too:
+
+- java: use annotation.
+- cpp: use macro and template.
+- golang: use struct tag.
+- python: use typehint.
```
LiangliangSui, line 59: I have a confusion here, whether decorator can also achieve similar functions.

chaokunyang: decorator can only be used to decorate method or class, I don't think it can be used here. Descriptor may be suitable, see https://docs.djangoproject.com/en/5.0/ref/models/fields/

LiangliangSui: Great! Thanks for replying to this question.

LiangliangSui (approved):
LGTM, I have no other comments.

docs/guide/xlang_type_mapping.md
```diff
@@ -4,4 +4,76 @@ sidebar_position: 3
...
+| set                | 15      | Set             | set                  | /               | set                            | fury.Set         | Set              |
+| map                | 16      | Map             | dict                 | Map             | unordered_map                  | map              | HashMap          |
+| duration           | 17      | Duration        | timedelta            | Number          | duration                       | Duration         | Duration         |
+| timestamp          | 18      | Instant         | datetime             | Number          | std::chrono::nanoseconds       | Time             | DateTime         |
+| decimal            | 19      | BigDecimal      | Decimal              | bigint          | /                              | /                | /                |
+| binary             | 20      | byte[]          | bytes                | /               | `int8_t[n]/vector<T>`          | `[n]int8/[]T`    | `Vec<i8>`        |
+| array              | 20      | array           | np.ndarray           | /               | /                              | array/slice      | Vec              |
```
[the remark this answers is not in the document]
chaokunyang: There are many languages which have a dedicated type for binary instead of array. For example, python uses `bytes` for binary type. Binary us not used as an array in many cases. I suggest we keep it. There is also a binary type in Arrow.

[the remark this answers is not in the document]
chaokunyang: Thrift also has a binary type. And we don't provide uint8 type for now. The int8 array can be a different type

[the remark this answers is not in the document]
theweipeng: We can serialize a bytes to int8 array and deserialize int8 array to bytes. But it would introduce some confusion. since it is so, we keep it. but should assign a new Typeid to binary to differentiate it from the array.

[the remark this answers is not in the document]
theweipeng: > Thrift also has a binary type. And we don't provide uint8 type for now. The int8 array can be a different type

Yes I have noticed that we can not distinguish 255 from -1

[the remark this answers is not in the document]
chaokunyang: > We can serialize a bytes to int8 array and deserialize int8 array to bytes. But it would introduce some confusion. since it is so, we keep it. but should assign a new Typeid to binary to differentiate it from the array.
Yes  , I forgot to update the type Id. I will update it later

docs/guide/xlang_type_mapping.md
```diff
@@ -4,4 +4,90 @@ sidebar_position: 3
...
-Coming soon.
+Note:
+
+- For type definition, see [Type Systems in Spec](../specification/xlang_serialization_spec.md#type-systems)
+- `int16_t[n]/vector<T>` indicates `int16_t[n]/vector<int16_t>`
+- The cross-language serialization is not stable, do not use it in your production environment.
+
+# Type Mapping
+
+| Type               | Type ID | Java            | Python               | Javascript      | C++                            | Golang           | Rust             |
```
LiangliangSui, line 15:
```suggestion
| Fury Type          | Fury Type ID | Java            | Python               | Javascript      | C++                            | Golang           | Rust             |
```

docs/guide/xlang_type_mapping.md
```diff
@@ -4,4 +4,90 @@ sidebar_position: 3
...
+| tensor             | 30      | /               | /                    | /               | /                              | /                | /                |
+| sparse tensor      | 31      | /               | /                    | /               | /                              | /                | /                |
+| arrow record batch | 32      | /               | /                    | /               | /                              | /                | /                |
+| arrow table        | 33      | /               | /                    | /               | /                              | /                | /                |
+
+# Type info(not implemented currently)
+
+Due to differences between type systems of languages, those types can't be mapped one-to-one between languages.
+
+If a user sees one type has multiple fury types, for example, `long` in java has type `int64/varint64/sliint64`, it
```
LiangliangSui, line 55: I suggest changing it to the following:

If the user notices that one language type corresponds to multiple Fury types

chaokunyang: thanks, updated

theweipeng (approved):
LGTM

PR #1480, opened 2024-04-08
chore(java): rename MemoryBuffer read/write/put/getType with read/write/put/getTypeNumber 

chaokunyang:
This PR refactor `MemoryBuffer` by:
- Rename all methods in name for unsigned varint/varlong write/read methods from write/readPositiveXXX.
- Rename all methods of get/put/read/writeType to type with number. For example, rename `readInt` to `readInt32`
- Remove rebundant code in `MemoryBuffer`
- Rename unsafe method as `_unsafeXXX`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
           java-version: 8
           distribution: 'temurin'
       - name: Install fury java
-        run: cd java && mvn -T10 clean install -DskipTests && cd -
+        run: cd java && mvn -T10 --no-transfer-progress clean install -DskipTests && cd -
       - name: Test
         run: |
           # Avoid sbt download jackson-databind error
--- a/NOTICE
+++ b/NOTICE
@@ -28,10 +28,6 @@ The text of each license is the standard Apache 2.0 license.
       java/fury-core/src/main/java/org/apache/fury/util/Platform.java
       java/fury-format/src/main/java/org/apache/fury/format/vectorized/ArrowWriter.java
 
-* flink (https://github.com/apache/flink)
-    Files:
-      java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
-
 * commons-io (https://github.com/apache/commons-io)
     Files:
       java/fury-core/src/main/java/org/apache/fury/io/ClassLoaderObjectInputStream.java
--- a/cpp/fury/util/buffer.h
+++ b/cpp/fury/util/buffer.h
@@ -133,7 +133,7 @@ public:
 
   inline double GetDouble(uint32_t offset) { return Get<double>(offset); }
 
-  inline uint32_t PutPositiveVarInt32(uint32_t offset, int32_t value) {
+  inline uint32_t PutVarUint32(uint32_t offset, int32_t value) {
     if (value >> 7 == 0) {
       data_[offset] = (int8_t)value;
       return 1;
@@ -164,7 +164,7 @@ public:
     return 5;
   }
 
-  inline int32_t GetPositiveVarInt32(uint32_t offset,
+  inline int32_t GetVarUint32(uint32_t offset,
                                      uint32_t *readBytesLength) {
     uint32_t position = offset;
     int b = data_[position++];
--- a/cpp/fury/util/buffer.h
+++ b/cpp/fury/util/buffer.h
@@ -164,8 +164,7 @@ public:
     return 5;
   }
 
-  inline int32_t GetVarUint32(uint32_t offset,
-                                     uint32_t *readBytesLength) {
+  inline int32_t GetVarUint32(uint32_t offset, uint32_t *readBytesLength) {
     uint32_t position = offset;
     int b = data_[position++];
     int result = b & 0x7F;
--- a/cpp/fury/util/buffer_test.cc
+++ b/cpp/fury/util/buffer_test.cc
@@ -39,7 +39,7 @@ TEST(Buffer, ToString) {
 }
 
 void checkVarUint32(int32_t startOffset, std::shared_ptr<Buffer> buffer,
-                         int32_t value, uint32_t bytesWritten) {
+                    int32_t value, uint32_t bytesWritten) {
   uint32_t actualBytesWritten = buffer->PutVarUint32(startOffset, value);
   EXPECT_EQ(actualBytesWritten, bytesWritten);
   uint32_t readBytesLength;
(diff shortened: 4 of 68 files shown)


chaokunyang:
@theweipeng @PragmaTwice @LiangliangSui Please take a look, this is a breaking API changes

chaokunyang:
@theweipeng @PragmaTwice @LiangliangSui Do we need to change all naming to names like `readInt16/readInt32/readInt64/readVarInt32/readVarInt64/readFloat16/readFloat32/readFloat64`/`writeInt16/writeInt32/...`?

chaokunyang:
There are so many `get/put/read/write` `short/int/long/float/double`, change it all to `int16/32/64`/`float16/32/64` will be a huge change.

One solution is that we can keep old naming intact, and add new names too. i.e. `readInt32/readInt` exists at the same time

theweipeng:
> @theweipeng @PragmaTwice @LiangliangSui Do we need to change all naming to names like `readInt16/readInt32/readInt64/readVarInt32/readVarInt64/readFloat16/readFloat32/readFloat64`/`writeInt16/writeInt32/...`?

I think it is very necessary, as it is good for a clear understanding of the data to be process.

theweipeng:
> There are so many `get/put/read/write` `short/int/long/float/double`, change it all to `int16/32/64`/`float16/32/64` will be a huge change.
> 
> 
> 
> One solution is that we can keep old naming intact, and add new names too. i.e. `readInt32/readInt` exists at the same time

There are all built-in APIs, which  user is unaccessable to it. I don't think we need to keep compatibility.

chaokunyang:
> > There are so many `get/put/read/write` `short/int/long/float/double`, change it all to `int16/32/64`/`float16/32/64` will be a huge change.
> > One solution is that we can keep old naming intact, and add new names too. i.e. `readInt32/readInt` exists at the same time
> 
> There are all built-in APIs, which user is unaccessable to it. I don't think we need to keep compatibility.

Users will use this APi to implement their own serializers. Take this library for an example: https://github.com/tommyettinger/tantrum/blob/main/tantrum-jdkgdxds/src/main/java/com/github/tommyettinger/tantrum/jdkgdxds/helpers/Support.java

LiangliangSui:
> @theweipeng @PragmaTwice @LiangliangSui Do we need to change all naming to names like `readInt16/readInt32/readInt64/readVarInt32/readVarInt64/readFloat16/readFloat32/readFloat64`/`writeInt16/writeInt32/...`?

+1, I also think this is very necessary. It’s a cool thing to normalize the naming of data types in various languages.

LiangliangSui:
> There are so many `get/put/read/write` `short/int/long/float/double`, change it all to `int16/32/64`/`float16/32/64` will be a huge change.
> 
> One solution is that we can keep old naming intact, and add new names too. i.e. `readInt32/readInt` exists at the same time

I prefer not to exist simultaneously, as existing at the same time will lead to more confusion of concepts and bloated code base.

The current PR has modified the common interface in [MemoryBuffer(e.g. writePositiveVarInt -> writeVarUint32)](https://github.com/apache/incubator-fury/pull/1480/files#diff-f321dd2324a9ae0cd5808f0281fd200090a9175830937aa3fb48cf309f7e36a1R1136-R1142), which has had an impact on some custom serializers. Now that this compatibility has been broken, we can complete all the renaming work before releasing pkg under `org.apache.fury`. We won't be able to make such more dangerous modifications in the future.

chaokunyang:
The thing is that java `java.nio.ByteBuffer` use `getInt/getLong/putInt/putLong`. If we use a different name here, it may intruduce confusition to users. `org.apache.arrow.memory.ArrowBuf` also use `setInt/setLong/putInt/putLong`

chaokunyang:
Protobuf does use `int32/int64`, see `com.google.protobuf.CodedInputStream#readRawVarint32()`

LiangliangSui:
> The thing is that java `java.nio.ByteBuffer` use `getInt/getLong/putInt/putLong`. If we use a different name here, it may intruduce confusition to users. `org.apache.arrow.memory.ArrowBuf` also use `setInt/setLong/putInt/putLong`

I still prefer `int32/int64`. We can emphasize the compatibility issues of ByteBuffer in the document. We can ask other members of the community for suggestions.

chaokunyang:
> > The thing is that java `java.nio.ByteBuffer` use `getInt/getLong/putInt/putLong`. If we use a different name here, it may intruduce confusition to users. `org.apache.arrow.memory.ArrowBuf` also use `setInt/setLong/putInt/putLong`
> 
> I still prefer `int32/int64`. We can emphasize the compatibility issues of ByteBuffer in the document. We can ask other members of the community for suggestions.

Seems `int32/64` style is better, I will post this to the dev@fury apache.org for more feedbacks

PragmaTwice:
Actually if it's C++/rust I will recommend to use template/generics instead of put type into function name.

But not sure if it's a good way in Java.

PragmaTwice (approved):

chaokunyang:
> Actually if it's C++/rust I will recommend to use template/generics instead of put type into function name.
> 
> But not sure if it's a good way in Java.

Java generics it erased, it's not feasible to invoke it using generics to dispatch

chaokunyang:
Hi @tommyettinger, do you have any suggestions? You used `MemoryBuffer` widely inhttps://github.com/tommyettinger/tantrum.

Will it be better if we rename all  `read/write/get/put Int/Long/Float/Double` in methods of `MemoryBuffer` to `read/write/get/put Int32/Int64/Float32/Float64`

tommyettinger:
I have no problem using Float32 and Int64 etc. as the terms, I think they're language-neutral and are almost completely self-explanatory. I imagine anyone writing code that uses MemoryBuffer won't have any problem knowing what Float32 vs Float64 mean. In C, I know `long` has some sort of compiler- and architecture-dependent meaning, and also that there's `long long` (which would not be good to see in a method name in other languages, since it looks like a pair of `long`s). So moving towards the types of numbers that at least C++ seems to encourage, like how it has int64_t, should be a good action.

chaokunyang:
> I have no problem using Float32 and Int64 etc. as the terms, I think they're language-neutral and are almost completely self-explanatory. I imagine anyone writing code that uses MemoryBuffer won't have any problem knowing what Float32 vs Float64 mean. In C, I know `long` has some sort of compiler- and architecture-dependent meaning, and also that there's `long long` (which would not be good to see in a method name in other languages, since it looks like a pair of `long`s). So moving towards the types of numbers that at least C++ seems to encourage, like how it has int64_t, should be a good action.

Hi @tommyettinger , thanks for your reply and suggestion. The type with number looks better to me now. I will udpate this PR to use this name style

theweipeng (approved):
LGTM

PR #1483, opened 2024-04-08
feat(java): channel stream reader

Munoon:


java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
```diff
@@ -20,30 +20,143 @@
...
     this(channel, ByteBuffer.allocate(4096));
   }
 
-  private FuryReadableChannel(ReadableByteChannel channel, ByteBuffer directBuffer) {
+  public FuryReadableChannel(ReadableByteChannel channel, ByteBuffer directBuffer) {
     this.channel = channel;
     this.byteBuffer = directBuffer;
-    this.buffer = MemoryBuffer.fromByteBuffer(directBuffer);
+
+    int offset = directBuffer.arrayOffset() + directBuffer.position();
```
chaokunyang, line 44: We need an offheap buffer here. Otherwise wr can just warp channel into a fury input stream

java/fury-core/src/test/java/org/apache/fury/StreamTest.java
```diff
@@ -204,4 +207,38 @@ public void testJavaOutputStream() throws IOException {
       assertEquals(newObj, beanA);
     }
   }
+
+  @Test
+  public void testReadableChannel() throws IOException {
+    Fury fury = Fury.builder().requireClassRegistration(false).build();
+    BeanA beanA = BeanA.createBeanA(2);
+    {
+      ByteArrayOutputStream bas = new ByteArrayOutputStream();
```
chaokunyang, line 216: How about using try-with for auto resource release

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
```diff
@@ -20,30 +20,130 @@
...
+  }
+
+  @Override
+  public void readToByteBuffer(ByteBuffer dst, int length) {
+    readToByteBuffer0(dst, length);
+  }
+
+  @Override
+  public int readToByteBuffer(ByteBuffer dst) {
+    return readToByteBuffer0(dst, dst.remaining());
```
chaokunyang, line 136: It's OK for this method to return data less than `dst.remaining()` since we returned a number how much data we read

Munoon: Do you think we should change the implementation of this method?
As I see, the `FuryInputStream` is implemented in the similar way: it checks the `dst.remaining()` and read from stream if he didn't have enough data. I think this method should be consistent over implementation and writing `dst.remaining()` is a correct desired logic (since you can control `dst.remaining()` by specifying `dst.limit`).

chaokunyang: `FuryInputStream ` only read `stream.available()` instead of read until `dst.remaining()`. If we read until `dst.remaining()`, it may block the current thread, which is unpected if users want to invoke this method
```java
      try {
        buf.read(dst, remaining);
        int available = stream.available();
        if (available > 0) {
          fillBuffer(available);
          int newRemaining = buf.remaining();
          buf.read(dst, newRemaining);
          return newRemaining + remaining;
        } else {
          return remaining;
        }
      } catch (IOException e) {
      }
```

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
```diff
@@ -20,30 +20,130 @@
...
+    if (remaining >= dstRemaining) {
+      byte[] bytes = buf.readBytes(dstRemaining);
+      dst.put(bytes);
+      return dstRemaining;
+    } else {
+      int filledSize = fillBuffer(dstRemaining - remaining);
+      int length = remaining + filledSize;
+      byte[] bytes = buf.readBytes(length);
+      dst.put(bytes);
+      return length;
```
chaokunyang, line 96:
```suggestion
    MemoryBuffer buf = buffer;
    int remaining = buf.remaining();
    int len = dst.remaining();
    if (remaining >= len) {
      buf.read(dst, len);
      return len;
    } else {
      try {
        buf.read(dst, remaining);
        return channel.read(dst) + remaining;
      } catch (IOException e) {
        throw new RuntimeException(e);
      }
    }
```

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
```diff
@@ -20,30 +20,130 @@
...
+    Platform.copyMemory(null, address, target, targetPointer, numBytes);
+    buf.increaseReaderIndex(numBytes);
+  }
+
+  @Override
+  public void readToByteBuffer(ByteBuffer dst, int length) {
+    MemoryBuffer buf = memoryBuffer;
+    int remaining = buf.remaining();
+    if (remaining < length) {
+      remaining += fillBuffer(length - remaining);
```
chaokunyang, line 134: This introduce an extra copy, the data doesn't have to copy into `memoryBuffer` to read into `dst`

java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
```diff
@@ -156,8 +156,17 @@ private MemoryBuffer(long offHeapAddress, int size, ByteBuffer offHeapBuffer) {
    *     the memory being released.
    * @param streamReader a reader for reading from a stream.
    */
-  private MemoryBuffer(
+  public MemoryBuffer(
```
chaokunyang, line 159: We'd better keep it private, the MemoryBuffer should only be created by factory method

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
```diff
@@ -128,21 +128,21 @@ public void readToUnsafe(Object target, long targetPointer, int numBytes) {
...
-    return readToByteBuffer0(dst, dst.remaining());
-  }
-
-  private int readToByteBuffer0(ByteBuffer dst, int length) {
     MemoryBuffer buf = memoryBuffer;
     int remaining = buf.remaining();
-    if (remaining < length) {
-      remaining += fillBuffer(length - remaining);
+    if (remaining > 0) {
+      buf.read(dst, remaining);
```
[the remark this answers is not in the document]
Munoon: Did you mean `dst.remaining()`?

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
Munoon: I thought we need it, to have the ability to *go back*.
If we don't, to avoid using some transfer buffer we'll be to set a limit to the destination buffer. Is this acceptable?

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
Munoon: Hmm, looks like we can also rewrite it using `readToByteBuffer(ByteBuffer, int)` method. I'll use this method in `read(ByteBuffer)` method at least until we decide the final implementation of `readToByteBuffer(ByteBuffer, int)`.

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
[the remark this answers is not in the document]
chaokunyang: If there are no remaining in `memoryBuffer`, and `channel` has extra data, the invocation will stop read anything here.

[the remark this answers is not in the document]
chaokunyang: > Did you mean `dst.remaining()`?

`Channel` doesn't have an `available` method, we can just invoke `Channle.read` instead

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
```diff
@@ -20,30 +20,114 @@
...
+  }
+
+  @Override
+  public int readToByteBuffer(ByteBuffer dst) {
+    MemoryBuffer buf = memoryBuffer;
+    int remaining = buf.remaining();
+    if (remaining > 0) {
+      buf.read(dst, remaining);
+    }
+    return remaining;
```
chaokunyang, line 120:
```suggestion
    return buf.read(dst, remaining) + channel.read(dst);
```

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
chaokunyang: Actually I'm considering only provide streaming interface to customized serializer to forbidden users to use API that can go back

java/fury-core/src/main/java/org/apache/fury/io/FuryReadableChannel.java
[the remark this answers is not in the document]
Munoon: Fine, so we are expecting the the `ByteBuffer dst` have already set up the correct limit before invoking this method.

chaokunyang (approved):
LGTM, thanks @Munoon

Issue #1466, opened 2024-04-04
[Java] optimize MemoryBuffer.unsafePutPositiveVarInt bytecode size


chaokunyang:
## Is your feature request related to a problem? Please describe.
The bytecode size of `MemoryBuffer.unsafePutPositiveVarInt` is 466 bytes, which is too big for inline.

![image](https://github.com/apache/incubator-fury/assets/12445254/5720f452-9d84-4a47-ad72-84825f01e08d)

## Describe the solution you'd like
We can rewrite it using algorithms like I used in #1462

PR #1489, opened 2024-04-11
fix(java): optimize unsafePutPositiveVarInt bytecode size

pandalee99:
## optimize unsafePutPositiveVarInt bytecode size
i think,the code can be reduce something memory
the  bytecode reduced from 208 to 192
image

#1462 has been reduced from three variables to two variables, and #1466 is already very few variables, I don't think it can be reduced any more

Variable-length coding requires at least one variable determination and one variable storage

## Related issues
#1466 

## Benchmark
pre 
image
after
image


chaokunyang (approved):
Great, thanks for this optimization

LiangliangSui:
Hi @pandalee99 , thanks for contributing to Fury.

We can further simplify some intermediate steps to reduce the generated bytecode. I submitted a PR to do this. https://github.com/apache/incubator-fury/pull/1490

pandalee99:
is so cool

Issue #1465, opened 2024-04-04
[Java] optimize MemoryBuffer.unsafeWritePositiveVarLong bytecode size


chaokunyang:
## Is your feature request related to a problem? Please describe.
The bytecode size of `MemoryBuffer.unsafeWritePositiveVarLong` is 466 bytes, which is too big for jvm jit inline.

![image](https://github.com/apache/incubator-fury/assets/12445254/9204e795-b33a-46e8-af8c-b2ffb141f5a7)

## Describe the solution you'd like
We can rewrite it using algorithms like I used in #1463

wangjie-fourth:
@chaokunyang it is done. maybe need close this issue?

PR #1491, opened 2024-04-11
perf(java): Reduce unsafeWritePositiveVarLong bytecode size.

pandalee99:
## Reduce unsafeWritePositiveVarLong bytecode size
Reduce bytecode size from 431 to 413.

## Related issues
Through the calculation in advance to reduce memory
#1465 

## Benchmark
pre
image

after
image

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
@@ -1697,64 +1697,56 @@ public final class MemoryBuffer {
     final int writerIndex = this.writerIndex;
     int varInt;
     varInt = (int) (value & 0x7F);
-    value >>>= 7;
-    if (value == 0) {
+    if (value >>> 7 == 0) {
       UNSAFE.putByte(heapMemory, address + writerIndex, (byte) varInt);
       this.writerIndex = writerIndex + 1;
       return 1;
     }
-    varInt |= (int) ((value & 0x7F) << 8) | 0x80;
-    value >>>= 7;
-    if (value == 0) {
+    varInt |= (((value & 0x3f80) << 1) | 0x80);
+    if (value >>> 14 == 0) {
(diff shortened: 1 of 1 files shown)


java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
```diff
@@ -1697,64 +1697,56 @@ public int unsafeWritePositiveVarLong(long value) {
...
+    varLong |= ((value & 0x1fc0000000000L) << 6) | 0x800000000000L;
+    if (value >>> 49 == 0) {
       unsafePutLong(writerIndex, varLong);
       this.writerIndex = writerIndex + 7;
       return 7;
     }
-    varLong |= ((value & 0x7F) << 56) | 0x80000000000000L;
-    value >>>= 7;
+    varLong |= ((value & 0xfe000000000000L) << 7) | 0x80000000000000L;
+    value >>>= 56;
```
LiangliangSui, line 1743: `value` does not need to be reassigned here

LiangliangSui (changes requested):

java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
pandalee99: sure，I already tested it.
I originally thought
```java
    if (value >>> 56 == 0) {
      unsafePutLong(writerIndex, varLong);
      this.writerIndex = writerIndex + 8;
      return 8;
    }
    unsafePutLong(writerIndex, varLong | 0x8000000000000000L);
    UNSAFE.putByte(heapMemory, address + writerIndex + 8, (byte) (value >>> 56 & 0xFF));
    this.writerIndex = writerIndex + 9;
```
but, this bytecode size will be from 413 to 414.

chaokunyang (approved):
LGTM, I didn't find a better way to optimize it

PR #1492, opened 2024-04-11
chore(java): FuryLogger adds level control.

LiangliangSui:
## What does this PR do?

FuryLooger adds default log level control.


java/fury-core/src/main/java/org/apache/fury/logging/FuryLogger.java
```diff
@@ -23,88 +23,139 @@
...
 public class FuryLogger implements Logger {
+
+  private static final int ERROR_LEVEL = 0;
+
+  private static final int WARN_LEVEL = 1;
+
+  private static final int INFO_LEVEL = 2;
+
+  // Default is INFO level.
+  private static final int LEVEL = INFO_LEVEL;
```
chaokunyang, line 34: how aboubt make it as a final field of `FuryLogger`, and pass it from `LoggerFactory` to allow users to custimize level

LiangliangSui: Pass in `Level` when calling `LoggerFactory#getLogger`, and then pass in `Level` through the `FuryLogger` constructor?

chaokunyang: We can add a static method in LoggerFactory named `setLoggerLevel`, when creating logger, we can pass that level to FuryLogger constructor

java/fury-core/src/main/java/org/apache/fury/logging/Slf4jLogger.java
```diff
@@ -25,10 +25,13 @@
 public class Slf4jLogger implements Logger {
   private static final String FQCN = Slf4jLogger.class.getName();
 
+  private final boolean isLocationAwareLogger;
+
   private final org.slf4j.Logger logger;
 
   public Slf4jLogger(Class<?> cls) {
     this.logger = org.slf4j.LoggerFactory.getLogger(cls);
+    this.isLocationAwareLogger = logger instanceof LocationAwareLogger;
```
chaokunyang, line 34: Good optimization!

chaokunyang (approved):
LGTM, thanks

java/fury-core/src/main/java/org/apache/fury/logging/LoggerFactory.java
```diff
@@ -27,6 +27,7 @@
 public class LoggerFactory {
   private static volatile boolean disableLogging;
   private static volatile boolean useSlf4jlogger;
+  private static volatile int logLevel;
```
chaokunyang, line 30: The default value 0 will disable all fury logging

LiangliangSui: Yes, this is a potential bug, I see you have fixed it in https://github.com/apache/incubator-fury/pull/1503, thanks!

PR #1502, opened 2024-04-13
chore(java): deprecate  serializeJavaObjectAndClass/deserializeJavaObjectAndClass API

chaokunyang:
## What does this PR do?
This PR deprecates serializeJavaObjectAndClass/deserializeJavaObjectAndClass API.
This APi is vargue and kind of duplicate with `Fury#serialize`. 
It doesn't provide `BufferCallback` support too.

We should make our API minimal and does not bloat it

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/BaseFury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/BaseFury.java
@@ -149,17 +149,31 @@ public interface BaseFury {
 
   <T> T deserializeJavaObject(FuryReadableChannel channel, Class<T> cls);
 
+  /** This method is deprecated, please use {@link #serialize} instead. */
+  @Deprecated
   byte[] serializeJavaObjectAndClass(Object obj);
 
+  /** This method is deprecated, please use {@link #serialize} instead. */
+  @Deprecated
   void serializeJavaObjectAndClass(MemoryBuffer buffer, Object obj);
 
+  /** This method is deprecated, please use {@link #serialize} instead. */
+  @Deprecated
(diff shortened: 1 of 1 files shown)


chaokunyang:
@theweipeng @PragmaTwice @LiangliangSui Could you please take a look?

LiangliangSui:
Hi @chaokunyang , I prefer to use the `@Deprecated` annotation to retain several versions rather than directly deleting these user interfaces, because there are already many users using Fury, maybe they are using these interfaces, and it is very risky for us to do so. After we release a few versions under org.apache.fury, we will remove them. This might be better.

chaokunyang:
> Hi @chaokunyang , I prefer to use the `@Deprecated` annotation to retain several versions rather than directly deleting these user interfaces, because there are already many users using Fury, maybe they are using these interfaces, and it is very risky for us to do so. After we release a few versions under org.apache.fury, we will remove them. This might be better.

We haven't make a release under `org.apache.fury`, all exsiting users are using class under `io.fury` package, there will still be breaking changes here. So maybe we can remove it directly.

LiangliangSui:
Although users are still using `io.fury`, if we delete these interfaces, users will still have to do some extra work to adapt to `org.apache.fury`(we have currently broken part of the first version of org.apache.fury). I still recommend retaining these interfaces and marking them as `@Deprecated`, because they are not for a particularly urgent modification, we can officially remove them after a few versions, which can reduce the user's first `org.apache.fury` adaptation workload.

chaokunyang:
> Deprecated

I see, I'll mark it as `@Deprecated`

LiangliangSui (approved):
LGTM, thanks!

theweipeng (approved):

PR #1503, opened 2024-04-13
perf(java): fastpath for read/write small varint in range [0,127]

chaokunyang:
## What does this PR do?
This PR add fastpath for read/write small varint in range `[0,127]`.

This PR also fix wrong log level setting in #1492

## Benchmark
With this PR:
image

Before this PR:
image

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/logging/LoggerFactory.java
+++ b/java/fury-core/src/main/java/org/apache/fury/logging/LoggerFactory.java
@@ -27,7 +27,7 @@ import org.apache.fury.util.GraalvmSupport;
 public class LoggerFactory {
   private static volatile boolean disableLogging;
   private static volatile boolean useSlf4jlogger;
-  private static volatile int logLevel;
+  private static volatile int logLevel = 2;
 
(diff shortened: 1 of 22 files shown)


theweipeng (approved):

java/fury-core/src/main/java/org/apache/fury/logging/LoggerFactory.java
```diff
@@ -47,7 +47,7 @@ public static void useSlf4jLogging(boolean useSlf4jLogging) {
 
   public static Logger getLogger(Class<?> clazz) {
     if (disableLogging) {
-      return new NilLogger();
+      return new FuryLogger(clazz, logLevel);
```
LiangliangSui, line 50: I have a confusion, why we disableLogging, but still use `FuryLogger` to output Log.

chaokunyang: The logging is disabled in https://github.com/apache/incubator-fury/pull/1492 disabled logging, I changed this for debug but forget to revert it. I will fix it in another PR

PR #1505, opened 2024-04-13
chore(javaScript):  enhancement code readability

theweipeng:
## What does this PR do? 
**Reduce the use of closures to enhance code readability.**   
We used closures extensively in the past, which could easily trigger Inline-Cache in `V8`, but this practice was detrimental to code readability. Now that the infrastructure is stable, we can replace them with classes to enhance readability without affecting performance.

There are no user-facing change.

## Benchmark
| (index) | serialize | deserialize |
|----------|------------|--------------|
| fury     | 151        | 120          |
| protobuf | 29         | 39           |
| json     | 28         | 35           |


theweipeng:
@LiangliangSui @chaokunyang  Hi, could you help review this?

javascript/packages/fury/lib/reader/index.ts
```diff
@@ -22,175 +22,184 @@ import { isNodeEnv } from "../util";
...
+    this.cursor += len;
     return result;
   }
 
-  function varUInt32() {
+  varUInt32() {
     // Reduce memory reads as much as possible. Reading a uint32 at once is far faster than reading four uint8s separately.
-    if (byteLength - cursor >= 5) {
-      const u32 = dataView.getUint32(cursor++, true);
+    if (this.byteLength - this.cursor >= 5) {
```
LiangliangSui, line 167: Could we optimize this code using something similar to the [Java implementation](https://github.com/apache/incubator-fury/blob/main/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java#L1777-L1810)? So that we can reduce some bit move and intermediate variables.

theweipeng: Good idea. I will change it.

chaokunyang (approved):
LGTM

PR #1508, opened 2024-04-13
feat(spec): add doc for why map use chunk by chunk spec

chaokunyang:
## What does this PR do?
This pr add doc for why map use chunk by chunk spec

## Related issues
#1413 

### DIFF
--- a/docs/specification/xlang_serialization_spec.md
+++ b/docs/specification/xlang_serialization_spec.md
@@ -559,7 +559,7 @@ Map iteration is too expensive, Fury won't compute the header like for list sinc
 Users can use `MapFieldInfo` annotation to provide the header in advance. Otherwise Fury will use first key-value pair
 to predict header optimistically, and update the chunk header if the prediction failed at some pair.
 
-Fury will serialize the map chunk by chunk, every chunk has 127 pairs at most.
+Fury will serialize the map chunk by chunk, every chunk has 255 pairs at most.
 
 ```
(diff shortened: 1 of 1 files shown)


docs/specification/xlang_serialization_spec.md
```diff
@@ -592,6 +592,21 @@ format will be:
...
+
+When fury will use first key-value pair to predict header optimistically, it can't know how many pairs have same
+meta(tracking kef ref, key has null and so on). If we don't write chunk by chunk with max chunk size, we must write at
+least `X` bytes to take up a place for later to update the number which has same elements, `X` is the num_bytes for
+encoding varint encoding of map size.
+
+And most map size are smaller than 255, if all pairs have same data, the chunk will be 1. This is common in golang/rust,
+which object are not reference by default.
+
+Also, if only one or two keys have different meta, we can make it into a different chunk, so that most pairs can share
```
theweipeng, line 605: Am I understanding this correctly as follows?
If the type prediction fails, we might initiate a new chunk. This implies that when streaming mode is disabled, all elements within the same chunk are of the same type.  In this scenario, both the third bit and the corresponding third bit in the header will be set to zero.

chaokunyang: Elements in a chunk can be different types too. For example, we can profile a a mapand if types are different but no null values, we can still use one chunk nut encode types for every item

theweipeng (approved):

PR #1511, opened 2024-04-14
perf(java): optimize fury creation speed

### LINKED ISSUE #1510: [Java] Optimize Fury creation speed
Current fury creation takes about 2 millseconds, which is slow, we need to optimize it.

A profiler shows that Fury creation is slow at ClassResolver:
image
image

Avoid String generation, and reduce map resize cost by creating a map with bigger size

## Additional context
#1482

chaokunyang:
## Related issues

Closes #1510 

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -427,16 +427,18 @@ public class ClassResolver {
   public void register(Class<?> cls, int classId) {
     // class id must be less than Integer.MAX_VALUE/2 since we use bit 0 as class id flag.
     Preconditions.checkArgument(classId >= 0 && classId < Short.MAX_VALUE);
-    Preconditions.checkArgument(
-        !extRegistry.registeredClassIdMap.containsKey(cls),
-        "Class %s already registered with id %s.",
-        cls,
-        extRegistry.registeredClassIdMap.get(cls));
-    Preconditions.checkArgument(
-        !extRegistry.registeredClasses.containsKey(cls.getName()),
-        "Class %s with name %s has been registered, registering class with same name are not allowed.",
-        extRegistry.registeredClasses.get(cls.getName()),
-        cls.getName());
+    if (extRegistry.registeredClassIdMap.containsKey(cls)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Class %s already registered with id %s.",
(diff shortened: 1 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -431,15 +429,14 @@ public void register(Class<?> cls, int classId) {
     Preconditions.checkArgument(classId >= 0 && classId < Short.MAX_VALUE);
     Preconditions.checkArgument(
         !extRegistry.registeredClassIdMap.containsKey(cls),
```
LiangliangSui, line 431: We can use `if` instead of `Preconditions#checkArgument`, thereby reducing the number of `extRegistry.registeredClassIdMap.get`.

chaokunyang: Could you elabrate more? I don't see any difference here

LiangliangSui:
```java
    Preconditions.checkArgument(
        !extRegistry.registeredClassIdMap.containsKey(cls),
        "Class %s already registered with id %s.",
        cls,
        extRegistry.registeredClassIdMap.get(cls));
    Preconditions.checkArgument(
        !extRegistry.registeredClasses.containsKey(cls.getName()),
        "Class %s with name %s has been registered, registering class with same name are not allowed.",
        extRegistry.registeredClasses.get(cls.getName()),
        cls.getName());
```

```java
    if (extRegistry.registeredClassIdMap.containsKey(cls)) {
      throw new IllegalArgumentException(
          String.format(
              "Class %s already registered with id %s.",
              cls, extRegistry.registeredClassIdMap.get(cls)));
    }

    if (extRegistry.registeredClasses.containsKey(cls.getName())) {
      throw new IllegalArgumentException(
          String.format(
              "Class %s with name %s has been registered, registering class with same name are not allowed.",
              extRegistry.registeredClasses.get(cls.getName()), cls.getName()));
    }
```

The former `extRegistry.registeredClassIdMap.get` will definitely be executed, while the latter will only be executed when there is an exception. We can save the cost of `extRegistry.registeredClassIdMap.get`.

chaokunyang: Good suggestion, updated

LiangliangSui (approved):
Great!

theweipeng (approved):
Great!

PR #1514, opened 2024-04-14
feat(java): meta string encoding algorithm in java

chaokunyang:
## What does this PR do?
This PR implements meta string encoding described in [fury java serialization spec](https://fury.apache.org/docs/specification/fury_java_serialization_spec#meta-string) and [xlang serialization spec](https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#meta-string)

We have `3/8` space saveing for most string:
```java
    // utf8 use 30 bytes, we use only 19 bytes
    assertEquals(encoder.encode("org.apache.fury.benchmark.data").getBytes().length, 19);
    // utf8 use 12 bytes, we use only 9 bytes.
    assertEquals(encoder.encode("MediaContent").getBytes().length, 9);
```

The integration with ClassResolver is left in another PR.

## Related issues
#1240 
#1413 

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
+++ b/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
@@ -40,6 +40,9 @@ public class MetaStringEncoder {
   }
 
   public MetaString encode(String input, Encoding encoding) {
+    if (input.isEmpty()) {
+      return new MetaString(Encoding.LOWER_SPECIAL, new byte[0], 0);
+    }
     int length = input.length();
     switch (encoding) {
       case LOWER_SPECIAL:
@@ -160,7 +163,7 @@ public class MetaStringEncoder {
 
   public byte[] encodeFirstToLowerSpecial(char[] chars) {
     chars[0] = Character.toLowerCase(chars[0]);
-    return encodeGeneric(chars, 6, false);
+    return encodeGeneric(chars, 5, false);
   }
 
   public byte[] encodeAllToLowerSpecial(char[] chars) {
--- a/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
+++ b/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
@@ -52,7 +52,7 @@ public class MetaStringEncoder {
         char[] chars = input.toCharArray();
         int upperCount = countUppers(chars);
         return new MetaString(
-            encoding, encodeRepAllToLowerSpecial(chars), (upperCount + length) * 5);
+            encoding, encodeRepAllToLowerSpecial(chars, upperCount), (upperCount + length) * 5);
       default:
         byte[] bytes = input.getBytes(StandardCharsets.UTF_8);
         return new MetaString(Encoding.UTF_8, bytes, bytes.length * 8);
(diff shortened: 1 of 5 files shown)


chaokunyang:
@theweipeng @PragmaTwice @LiangliangSui This PR is ready for review, could you help review it?

theweipeng (approved):
LGTM

java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  public Encoding computeEncoding(String input) {
+    if (input.isEmpty()) {
+      return Encoding.LOWER_SPECIAL;
+    }
+    char[] chars = input.toCharArray();
+    if (canBeLowerSpecialEncoded(chars)) {
+      return Encoding.LOWER_SPECIAL;
+    } else if (canBeLowerUpperDigitSpecialEncoded(chars)) {
```
LiangliangSui, line 114: `canBeLowerUpperDigitSpecialEncoded`, `countDigits` and `countUppers` traverse `chars` three times. Could we traverse the `chars` only once in `canBeLowerUpperDigitSpecialEncoded`, calculate the digits of numbers, the number of uppercase letters and return a bean(the number of characters of both types that should be included in the bean).

java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
```diff
@@ -0,0 +1,288 @@
...
+    int upperCount = 0;
+    for (char c : chars) {
+      if (Character.isUpperCase(c)) {
+        upperCount++;
+      }
+    }
+    return upperCount;
+  }
+
+  private boolean canBeLowerUpperDigitSpecialEncoded(String input) {
```
LiangliangSui, line 170: This function is not used, we should remove it.

java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
```diff
@@ -0,0 +1,288 @@
...
+          return Encoding.ALL_TO_LOWER_SPECIAL;
+        } else {
+          return Encoding.LOWER_UPPER_DIGIT_SPECIAL;
+        }
+      }
+    }
+    return Encoding.UTF_8;
+  }
+
+  private boolean canBeLowerSpecialEncoded(String input) {
```
LiangliangSui, line 132: This function is not used, we should remove it.

java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
chaokunyang: updated

java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
chaokunyang: removed

java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
```diff
@@ -0,0 +1,283 @@
...
+    return encodeGeneric(input, 6, false);
+  }
+
+  public byte[] encodeFirstToLowerSpecial(String input) {
+    return encodeFirstToLowerSpecial(input.toCharArray());
+  }
+
+  public byte[] encodeFirstToLowerSpecial(char[] chars) {
+    chars[0] = Character.toLowerCase(chars[0]);
+    return encodeGeneric(chars, 5, false);
```
LiangliangSui, line 207:
```suggestion
    return encodeGeneric(chars, 5, true);
```

chaokunyang: removed  `lowerSpecial` param

LiangliangSui (approved):
LGTM

PR #1525, opened 2024-04-16
fix(java): ThreadPoolFury and ThreadLocalFury concurrency security issue

LiangliangSui:
## What does this PR do?

Fix potential thread safety issues with ThreadPoolFury and ThreadLocalFury.

## Related issues

https://github.com/apache/incubator-fury/issues/1524

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/pool/ClassLoaderFuryPooled.java
+++ b/java/fury-core/src/main/java/org/apache/fury/pool/ClassLoaderFuryPooled.java
@@ -29,11 +29,13 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import javax.annotation.concurrent.ThreadSafe;
 import org.apache.fury.Fury;
 import org.apache.fury.logging.Logger;
 import org.apache.fury.logging.LoggerFactory;
 
 /** A thread-safe object pool of {@link Fury}. */
+@ThreadSafe
 public class ClassLoaderFuryPooled {
 
   private static final Logger LOG = LoggerFactory.getLogger(ClassLoaderFuryPooled.class);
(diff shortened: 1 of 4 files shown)


java/fury-core/src/main/java/org/apache/fury/pool/ClassLoaderFuryPooled.java
```diff
@@ -122,6 +124,20 @@ private void addFury() {
   }
 
   void setFactoryCallback(Consumer<Fury> factoryCallback) {
```
chaokunyang, line 126: This method should not be called concurently, users should invoke regitster and other fury config methods when creating ThreadSafeFury. It's not necessary to add a lock

LiangliangSui: In fact, all our register and config functions should be called when `ThreadSafeFury` is being created, but the purpose of adding the `processCallback` function should be contrary to this design. The `processCallback` interface allows register or config operations to be performed after created `ThreadSafeFury`.

LiangliangSui:
```java
ThreadSafeFury safeFury = Fury.builder()
    .requireClassRegistration(true)
    .buildThreadSafeFury();
safeFury.register(A.class);
safeFury.register(B.class);

// Thread A
new Thread(() -> {
  safeFury.serialize(new A());
}).start();

// Thread B
new Thread(() -> {
  // processCallback(fury -> fury.register(clz));
  safeFury.register(C.class);
  safeFury.serialize(new C());
}).start();
```

As shown in the above Demo, how can we avoid `safeFury.register(C.class)` being called in Thread B.

chaokunyang: Then we should add code to avoid this, such as capture Thread, and check thread in later call

LiangliangSui: I think this is a feasible approach, but I think the methods that implement the ThreadSafeFury interface should be thread-safe, and all methods in ThreadSafeFury should be allowed to be called in any thread.

PR #1533, opened 2024-04-17
refactor(java): Detach MemoryBuffer from Fury

LiangliangSui:
## What does this PR do?

Separate MemoryBuffer management tasks from Fury and reduce Fury class responsibilities.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/Fury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/Fury.java
@@ -42,6 +42,7 @@ import org.apache.fury.io.FuryReadableChannel;
 import org.apache.fury.logging.Logger;
 import org.apache.fury.logging.LoggerFactory;
 import org.apache.fury.memory.MemoryBuffer;
+import org.apache.fury.memory.MemoryBufferManager;
 import org.apache.fury.memory.MemoryUtils;
 import org.apache.fury.resolver.ClassInfo;
 import org.apache.fury.resolver.ClassInfoHolder;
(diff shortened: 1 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -117,6 +116,7 @@ public final class Fury implements BaseFury {
   private Iterator<MemoryBuffer> outOfBandBuffers;
   private boolean peerOutOfBandEnabled;
   private int depth;
+  private final MemoryBufferManager bufferManager;
```
chaokunyang, line 119: The buffer logic is pretty simple, I don't think it deserves a separate abstraction.  The abstraction is not always a good thing.

LiangliangSui: Could we apply `execute` method in `Fury`?

chaokunyang: I prefer not, we should reduce callstack and avoid use lambda here to reduce performance cost. There are alreay several nested call before the real serialization logic run

LiangliangSui: Okay, I got it, I will close this PR later.

LiangliangSui:
What can I do for Fury currently? Implement the design in [xlang_serialization_spec.md](https://github.com/apache/incubator-fury/blob/main/docs/specification/xlang_serialization_spec.md)?

chaokunyang:
> What can I do for Fury currently? Implement the design in [xlang_serialization_spec.md](https://github.com/apache/incubator-fury/blob/main/docs/specification/xlang_serialization_spec.md)?

Yep, we basically redesigned our xlang serialization protocol, the implementation in java/python/golang/javascript are not alighed to the new proposal. Maybe we can take over some features development in the new protocol

LiangliangSui:
> > What can I do for Fury currently? Implement the design in [xlang_serialization_spec.md](https://github.com/apache/incubator-fury/blob/main/docs/specification/xlang_serialization_spec.md)?
> 
> Yep, we basically redesigned our xlang serialization protocol, the implementation in java/python/golang/javascript are not alighed to the new proposal. Maybe we can take over some features development in the new protocol

Okay, I'll take some of the features and implement it.

PR #1549, opened 2024-04-19
feat(javascript): Implement Xlang map

theweipeng:
## What does this PR do?
1. Implement the map protocol.
2. Cache the primitive type serializer, guess serializer by data type every time is unnecessary.

In this PR , the code generator is simple call the AnySerializer which is much more slower then jit code, I will implement the code generator next PR.

No user-facing change.

## Benchmark
Before:
| (index)    | Values |
|-------------|--------|
| serialize  | 365    |
| deserialize | 193    |

After:

| (index)    | Values |
|-------------|--------|
| serialize  | 519    |
| deserialize | 238    |


theweipeng:
@chaokunyang @LiangliangSui  Hi，Can you help take a review？

javascript/packages/fury/lib/classResolver.ts
```diff
@@ -114,8 +115,26 @@ export default class SerializerResolver {
     this.registerSerializer(fury, Type.float16Array());
     this.registerSerializer(fury, Type.float32Array());
     this.registerSerializer(fury, Type.float64Array());
+
+    this.numberSerializer = this.getSerializerById(SerializerResolver.getTypeIdByInternalSerializerType(InternalSerializerType.FLOAT64));
+    this.bigintSerializer = this.getSerializerById(SerializerResolver.getTypeIdByInternalSerializerType(InternalSerializerType.INT64));
```
chaokunyang, line 120: Maybe not in this PR, could we rename bigint to int64? Normally bigint means an unlimited integer which can represent number of any bits

javascript/packages/fury/lib/gen/map.ts
```diff
@@ -21,77 +21,237 @@ import { MapTypeDescription, TypeDescription } from "../description";
...
+  }
+
+  read(fromRef: boolean): any {
+    let count = this.fury.binaryReader.varInt32();
+    const result = new Map();
+    if (fromRef) {
+      this.fury.referenceResolver.reference(result);
+    }
+    while (count > 0) {
+      const header = this.fury.binaryReader.uint16();
```
chaokunyang, line 201: Do we need to rename it as `read/writeXXX` in future?

chaokunyang (approved):
LGTM

javascript/packages/fury/lib/gen/map.ts
```diff
@@ -21,77 +21,237 @@ import { MapTypeDescription, TypeDescription } from "../description";
...
+    }
+    flag <<= 4;
+    if (valueFlag & 0b10) {
+      flag |= MapFlags.HAS_NULL;
+    }
+    if (valueFlag & 0b01) {
+      flag |= MapFlags.TRACKING_REF;
+    }
+    return flag;
+  }
```
LiangliangSui, line 70:
```suggestion
  private getHead(keyFlag: number, valueFlag: number) {
    let flag = 0;
    if (keyFlag & MapFlags.HAS_NULL) {
      flag |= MapFlags.HAS_NULL;
    }
    if (keyFlag & MapFlags.TRACKING_REF) {
      flag |= MapFlags.TRACKING_REF;
    }
    flag <<= 4;
    if (valueFlag & MapFlags.HAS_NULL) {
      flag |= MapFlags.HAS_NULL;
    }
    if (valueFlag & MapFlags.TRACKING_REF) {
      flag |= MapFlags.TRACKING_REF;
    }
    return flag;
  }
```

javascript/packages/fury/lib/gen/map.ts
theweipeng: I think it is uncessary, unlike java the binaryWriter and binaryReader  are separate classes.  In `v8` , if the count of a object's property greater than 30, the object will degenerate into hashtable, which is very very shlow. So I  split the class `MemoryBuffer` into two  separate class.

PR #1557, opened 2024-04-23
feat(java): support adjust logger level dynamically

### LINKED ISSUE #1554: [Java] Cannot disable logging with ThreadSafeFury in native-image
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-fury/issues) and found no similar issues.

### Version

Version: 0.5.0 (built from source)
OS: macOS
JDK: 22

After following the [GraalVM guide](https://github.com/apache/incubator-fury/blob/main/docs/guide/graalvm_guide.md) by using a static class field for `native-image` compilation, I was able to disable logging with `LoggerFactory.disableLogging()` if a using a `Fury`. However, logging is not disabled if a `ThreadSafeFury` is used instead.

### Component(s)

Java

### Minimal reproduce step

```
import org.apache.fury.Fury;
import org.apache.fury.ThreadSafeFury;
import org.apache.fury.logging.LoggerFactory;
public class Main {
    static ThreadSafeFury fury;
    static {
        LoggerFactory.disableLogging();
        fury = new ThreadLocalFury(classLoader -> {
            return Fury.builder().build();
        });
        // Also tried: fury = Fury.builder().buildThreadSafeFury();
    }
        fury.serialize(Main.class);
    }
}
```

No output

### What did you see instead?

`2024-04-21 05:33:31 INFO  Fury:144 [main] - Created new fury org.apache.fury.Fury@50c4ee77`

### Anything Else?

Correct behaviour is observed in the JVM with a `ThreadSafeFury`, i.e. when not compiled with `native-image`. Similarly, correct behaviour is observed in both the JVM and `native-image` when using a `Fury` instead of a `ThreadSafeFury`.

chaokunyang:
## What does this PR do?
This PR supports adjust logger level dynamically even Fury related classes are initialized.

## Related issues

Closes #1554


java/fury-core/src/main/java/org/apache/fury/logging/LoggerFactory.java
```diff
@@ -32,6 +32,7 @@ public class LoggerFactory {
   /** Disable Logger, there will be no log output. */
   public static void disableLogging() {
     disableLogging = true;
+    logLevel = LogLevel.ERROR_LEVEL - 1;
```
LiangliangSui, line 35: Why do we need to set the log level when `disableLogging/enableLogging`? This is an implicit operation.

Currently, even if `FuryLogger` has been instantiated, log output can be disabled through `disableLogging`.

chaokunyang: This is to keep the consistency. If logging is disabled, logLevel should reflect such config too

chaokunyang: You can take disableLogging as a convenience method for set logging level

LiangliangSui: Could we use `logLevel` instead of `disableLogging` field? 

Because currently `disableLogging` field is equivalent to `logLevel == LogLevel.ERROR_LEVEL - 1`, so we can remove the judgment in `setLogLevel`.

chaokunyang: updated

LiangliangSui (approved):
Looks good to me, but there is code format ci error that needs to be processed.

PR #1560, opened 2024-04-24
chore(java): clean code for Fury ArrayAsList

chaokunyang:
## What does this PR do?

This PR clean code for Fury ArrayAsList, and make it as a internal class of `FuryArrayAsListSerializer`


chaokunyang:
Hi @justinmclean, I didn't find the similarities between Fury ArrayAsList and other projects. But in order to reduce confusion, I  cleaned the code in  Fury ArrayAsList. Could you help review it?

java/fury-core/src/main/java/org/apache/fury/serializer/collection/FuryArrayAsListSerializer.java
```diff
@@ -43,4 +47,91 @@ public Collection newCollection(MemoryBuffer buffer) {
...
+
+    @Override
+    public int indexOf(Object o) {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean contains(Object o) {
+      throw new UnsupportedOperationException();
+    }
```
tisonkun, line 112: This seems to be the effective changes beyond moving the code from a standalone file to a nested class.

But such methods are quite common and I suppose every Java developer can write:

```java
    @Override
    public boolean add(E e) {
      array[size++] = e;
      return true;
    }

    @Override
    public E get(int index) {
      return array[index];
    }
```

If such code can have compliance issues, where is the unique origin for all the code in such pattern?

chaokunyang: I just removed the implementation of methods which won't be invoked by Fury currently.

But to be honest, I think Java developer  will write similar code if he want to implement a List based on an Array.

PR #1562, opened 2024-04-24
fix(license): remove benchmark license when build

chaokunyang:
## What does this PR do?
Benchmark code are excluded from source release, we should remove license mention in the final LICENSE file.

The NOTICE fof those files are empty, so no need to change NOTICE file

## Related issues

#1389 


tisonkun (approved):
Thank you!

Issue #1540, opened 2024-04-19
[Go] Implement meta string encoding algorithm for golang


chaokunyang:
## Is your feature request related to a problem? Please describe.
We've implemented  meta string encoding algorithm in https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#meta-string for java in #1514   , it's time to implement it in golang.

## Describe the solution you'd like
Java implementation in #1514 can be taken as a reference. But note that the meta string encoding algorithm is used for encode field name only, so the special charater can't be `.` or `$`, thus the implementation will be simpler

## Additional context
#1413

qingoba:
Could you assign it to me? This is my first try of open source and I'm very interested in this task. Thanks.

chaokunyang:
Great, thanks for the willingness to contribute to Fury

chaokunyang:
hmm, this is a bug, UTF-8 is barely used in meta string. Acutally, most chars are ascii chars. So it's not covered in Fury serialization tests. We need to fix it and add some unit tests.

Thanks for pointing out this bug @qingoba

chaokunyang:
I have a new idea, we can add a bit to incidate whether strip last char in encoded meta string if the encoding is not UTF-8. In this way, we don't have to store num bits and num chars in MetaString

qingoba:
I don't understand what you mean, could you give an example?
Actually, I think `numChars` can be omitted, because it can be computed by `numBits` when needed.

qingoba:
In this way, the Decoder does not need to accept `numBits` arguments.

chaokunyang:
> I have a new idea, we can add a bit to incidate whether strip last char in encoded meta string if the encoding is not UTF-8. In this way, we don't have to store num bits and num chars in MetaString

Hi @qingoba , I added stip last char flag to spec in #1565 . I believe this will make the implementation simpler

PR #1565, opened 2024-04-24
feat(spec/java): add strip flag in meta string encoding spec

chaokunyang:
## What does this PR do?

add strip flag in meta string encoding spec

## Related issues

#1540

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringDecoder.java
+++ b/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringDecoder.java
@@ -45,19 +45,18 @@ public class MetaStringDecoder {
    *
    * @param encodedData encoded data using passed <code>encoding</code>.
    * @param encoding encoding the passed data.
-   * @param numBits total bits for encoded data.
    * @return Decoded string.
    */
-  public String decode(byte[] encodedData, Encoding encoding, int numBits) {
+  public String decode(byte[] encodedData, Encoding encoding) {
     switch (encoding) {
       case LOWER_SPECIAL:
-        return decodeLowerSpecial(encodedData, numBits);
+        return decodeLowerSpecial(encodedData);
       case LOWER_UPPER_DIGIT_SPECIAL:
-        return decodeLowerUpperDigitSpecial(encodedData, numBits);
+        return decodeLowerUpperDigitSpecial(encodedData);
       case FIRST_TO_LOWER_SPECIAL:
-        return decodeRepFirstLowerSpecial(encodedData, numBits);
+        return decodeRepFirstLowerSpecial(encodedData);
       case ALL_TO_LOWER_SPECIAL:
-        return decodeRepAllToLowerSpecial(encodedData, numBits);
+        return decodeRepAllToLowerSpecial(encodedData);
       case UTF_8:
         return new String(encodedData, StandardCharsets.UTF_8);
       default:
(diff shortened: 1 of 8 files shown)


theweipeng (approved):
LGTM

java/fury-core/src/main/java/org/apache/fury/meta/MetaStringDecoder.java
```diff
@@ -66,30 +65,36 @@ public String decode(byte[] encodedData, Encoding encoding, int numBits) {
...
+  private String decodeLowerSpecial(byte[] data) {
     StringBuilder decoded = new StringBuilder();
-    int bitIndex = 0;
-    int bitMask = 0b11111; // 5 bits for mask
-    while (bitIndex + 5 <= numBits) {
+    int totalBits = data.length * 8; // Total number of bits in the data
+    boolean stripLastChar = (data[0] & 0x80) != 0; // Check the first bit of the first byte
+    int bitMask = 0b11111; // 5 bits for the mask
+    int bitIndex = 1; // Start from the second bit
+    while (bitIndex + 5 <= totalBits) {
```
LiangliangSui, line 74: When `stripLastChar` is true, the while loop will be executed one more time, and an extra redundant char will be added to `decoded`, and then we will delete it.

Through this logic, we can delete the `numChars` and `numBits` fields in `MetaString`, because we can infer these two values ​​through `stripLastChar`. Can I understand this problem this way?

chaokunyang: Yes, since one char is encoded using `5/6` chars, the last byte may have `1~7` rebundant bits, which can hold another char, and we must delete such char

qingoba:
In `MetaStringEncoder.java`, we use `0x08`:
```java
boolean stripLastChar = bytes.length * 8 >= totalBits + bitsPerChar;
if (stripLastChar) {
bytes[0] = (byte) (bytes[0] | 0x80);
}
```

In `MetaString.java`, we use `0b1`:
```java
if (encoding != Encoding.UTF_8) {
  Preconditions.checkArgument(bytes.length > 0);
  this.stripLastChar = (bytes[0] & 0b1) != 0;
} else {
  this.stripLastChar = false;
}
```
These two are not consistent.

LiangliangSui:
> In `MetaStringEncoder.java`, we use `0x08`:
> 
> ```java
> boolean stripLastChar = bytes.length * 8 >= totalBits + bitsPerChar;
> if (stripLastChar) {
> bytes[0] = (byte) (bytes[0] | 0x80);
> }
> ```
> 
> In `MetaString.java`, we use `0b1`:
> 
> ```java
> if (encoding != Encoding.UTF_8) {
>   Preconditions.checkArgument(bytes.length > 0);
>   this.stripLastChar = (bytes[0] & 0b1) != 0;
> } else {
>   this.stripLastChar = false;
> }
> ```
> 
> These two are not consistent.

Thanks for pointing out this issue, this is indeed a bug and our unit tests don't cover it.

`0x80` should also be used in `MetaString`.
```java
if (encoding != Encoding.UTF_8) {
  Preconditions.checkArgument(bytes.length > 0);
  this.stripLastChar = (bytes[0] & 0x80) != 0;
} else {
  this.stripLastChar = false;
}
```

Do you want to submit a PR to fix this issue and add unit tests? @qingoba

qingoba:
Sure, I will do this after I finish #1566

chaokunyang:
> In `MetaStringEncoder.java`, we use `0x08`:
> 
> ```java
> boolean stripLastChar = bytes.length * 8 >= totalBits + bitsPerChar;
> if (stripLastChar) {
> bytes[0] = (byte) (bytes[0] | 0x80);
> }
> ```
> 
> In `MetaString.java`, we use `0b1`:
> 
> ```java
> if (encoding != Encoding.UTF_8) {
>   Preconditions.checkArgument(bytes.length > 0);
>   this.stripLastChar = (bytes[0] & 0b1) != 0;
> } else {
>   this.stripLastChar = false;
> }
> ```
> 
> These two are not consistent.

The MetaStringDecoder parse the flag by itself, and didn't use this flag. Currently the `org.apache.fury.meta.MetaString#stripLastChar` is not called by other classes. So the fury serializaiton ut still succceed.

qingoba:
There is another question.
If input string is empty, endcode bytes is also empty:
```java
public MetaString encode(String input, Encoding encoding) {
  Preconditions.checkArgument(
      input.length() < Short.MAX_VALUE, "Long meta string than 32767 is not allowed");
  if (input.isEmpty()) {
    return new MetaString(input, Encoding.UTF_8, specialChar1, specialChar2, new byte[0]);
  }
}
```
However, decode procedure doesn't check empty. It will access `bytes[0]` directly:
```java
private String decodeLowerSpecial(byte[] data) {
  StringBuilder decoded = new StringBuilder();
  int totalBits = data.length * 8; // Total number of bits in the data
  boolean stripLastChar = (data[0] & 0x80) != 0; // Check the first bit of the first byte
}
```
So decode procedure should check empty.

qingoba:
I can fix this now.

PR #1568, opened 2024-04-24
perf(java): Optimize MetaStringDecoder

LiangliangSui:
## What does this PR do?

When `MetaStringDecoder` is decoding, it will read two bytes at a time from the `data` and then perform the operation. This PR determines whether to skip the byte value. If it does not skip the byte decoding, it will only read one byte from the data.


chaokunyang (approved):
Great!

PR #1573, opened 2024-04-25
fix(java): Prevent exception in ObjectArray.clearObjectArray()

tommyettinger:
## What does this PR do?

Without this change, if `ObjectArray.clearObjectArray()` is called with start > 0 and size < 128 (which is `COPY_THRESHOLD`), then an IllegalArgumentException is thrown. This happens because `Arrays.fill()` takes a from index and a to index, in contrast with `clearObjectArray()`, which takes a start index and a size. The exception looks like this:

```
Exception in thread "main" java.lang.IllegalArgumentException: fromIndex(11) > toIndex(5)
	at java.base/java.util.Arrays.rangeCheck(Arrays.java:718)
	at java.base/java.util.Arrays.fill(Arrays.java:3453)
```

This wasn't thrown in Fury code, because Fury itself never calls clearObjectArray() with any start value other than 0. The method is part of a public API, though, so user code can call it.

I added an extra check to ObjectArrayTest that calls clearObjectArray() with 1 as a start value, instead of only testing with a start value of 0. I can run the tests one at a time and the only test I touched passes, but (maybe because of the commit I checked out) running the command in CONTRIBUTING.md (`mvn -T10 clean test`) fails on `org.apache.fury.serializer.MetaSharedCompatibleTest`. I'm not sure if there's anything I can do about that; the issue may already be fixed.

Spotless and Checkstyle have both been run and found no issues.

## Related issues

None.

No.

## Benchmark

I don't see any way this could have a significant effect on performance. It only adds one addition performed once during some calls to an O(n) function, and clearing an array can't be done multiple times in sequence with any difference from doing it once.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/collection/ObjectArray.java
+++ b/java/fury-core/src/main/java/org/apache/fury/collection/ObjectArray.java
@@ -95,7 +95,7 @@ public final class ObjectArray {
    */
   public static void clearObjectArray(Object[] objects, int start, int size) {
     if (size < COPY_THRESHOLD) {
-      Arrays.fill(objects, start, size, null);
+      Arrays.fill(objects, start, start + size, null);
     } else {
       if (size < NIL_ARRAY_SIZE) {
         System.arraycopy(NIL_ARRAY, 0, objects, start, size);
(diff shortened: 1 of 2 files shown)


chaokunyang (approved):
Looks great, thanks for this fix

PR #1579, opened 2024-04-26
Update pom.xml

stardustman:
fix using high version jdk compile to low version such as jdk8 using --release instead of -target
solve the https://github.com/apache/incubator-fury/issues/1577

> Starting JDK 9, the javac executable can accept the --release option to specify against which Java SE release you want to build the project. For example, you have JDK 11 installed and used by Maven, but you want to build the project against Java 8. The --release option ensures that the code is compiled following the rules of the programming language of the specified release, and that generated classes target the release as well as the public API of that release. This means that, unlike the [-source and -target options](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html), the compiler will detect and generate an error when using APIs that don't exist in previous releases of Java SE.
https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

### DIFF
--- a/java/fury-core/pom.xml
+++ b/java/fury-core/pom.xml
@@ -48,7 +48,7 @@
 
   <properties>
     <maven.compiler.source>8</maven.compiler.source>
-    <maven.compiler.target>8</maven.compiler.target>
+    <maven.compiler.release>8</maven.compiler.release>
     <fury.java.rootdir>${basedir}/..</fury.java.rootdir>
   </properties>
 


chaokunyang:
Hi @stardustman , thanks for your contribution. Could you add a ByteBuffer ut for this ?

chaokunyang:
> Hi @stardustman , thanks for your contribution. Could you add a ByteBuffer ut for this ?

Seems this is not easy to add a test, we must compile use one java version, and run it with another java version.

java/fury-core/pom.xml
```diff
@@ -127,6 +127,11 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
```
chaokunyang, line 130: I did a test, seems not work? Could you test it locally, you can compile it with java9+, then select to use java8 to run. And do we need to add this config to parent pom?

stardustman:
this 0.4.1 version in [mvnrepository](https://mvnrepository.com/artifact/org.furyio/fury-core/0.4.1) compile with jdk17? or other version?

chaokunyang:
> this 0.4.1 version in [mvnrepository](https://mvnrepository.com/artifact/org.furyio/fury-core/0.4.1) compile with jdk17? or other version?

The major in class file is 52, so it should be java8 already

stardustman:
> > this 0.4.1 version in [mvnrepository](https://mvnrepository.com/artifact/org.furyio/fury-core/0.4.1) compile with jdk17? or other version?
> 
> The major in class file is 52, so it should be java8 already
if this jar was compiled using jdk8's javac to get class file, this should not report error when execute with jdk8.
but I get error when I serialize the ByteBuffer, example code like this:

```java
import io.fury.Fury;
import io.fury.ThreadLocalFury;
import io.fury.ThreadSafeFury;
import io.fury.config.Language;

import java.nio.ByteBuffer;

public class FuryTest {


    // Note that Fury instances should be reused between
    // multiple serializations of different objects.
    private static final ThreadSafeFury fury = new ThreadLocalFury(classLoader -> {
        Fury f = Fury.builder()
                .withLanguage(Language.JAVA)
                .withClassLoader(classLoader)
                .build();
        return f;
    });

    public static byte[] encoder(Object object) {
        return fury.serialize(object);
    }

    public static <T> T decoder(byte[] bytes) {
        return (T) fury.deserialize(bytes);
    }

        byte len = 10;
        ByteBuffer byteBuffer = ByteBuffer.allocate(len);
        for (int i = 0; i < len; i++) {
            byteBuffer.put((byte) i);
        }
        byte[] bytes = FuryTest.encoder(byteBuffer);
        System.out.println(bytes.length);

    }

}

```

chaokunyang:
I forgot which javac version I'm using.

I use following steps to guess javac version:
- download jar from https://mvnrepository.com/artifact/org.furyio/fury-core/0.4.1
- Extract classfile: `jar xf fury-core-0.4.1.jar io/fury/Fury.class`
- Get major version: `javap -verbose -cp fury-core-0.4.1.jar io/fury/Fury.class | grep "major"`

stardustman:
I use the branch `releases-0.4.1` and `jdk1.8.0_241` then  execute the `mvn package` in the IDE to get the final jar file, and test using this jar, this time the test code can execute, but using the jar download from mvn central repo  report error.
I also try to use jdk11 to execute the `mvn package` but get the  `error: package sun.misc does not exist`.

chaokunyang:
Hi @stardustman , does latest fury 0.8.0 works for you on JDK8?

Issue #1576, opened 2024-04-26
[Java] Memory issue about FuryInputStream


yihwang6:
In #1551, I noticed that a growBuffer method was added to FuryInputStream, which continuously increases the size of the buffer. This operation may potentially give rise to memory-related concerns due to the large buffer size. It might be better to consider implementing the buffered streaming reading with a fixed-size buffer.

chaokunyang:
`org.apache.fury.io.FuryInputStream#shrinkBuffer` will be invoked after the deserialziation finished

yihwang6:
Ok, but there may still be a large buffer size when deserializing a large object.

chaokunyang:
Yes, that is possible. We don't release buffer before we finished the deserialization. One case I can imagine is that when buffer close to int max size, we can't multiple 1.5.

chaokunyang:
Would you like to help optimize this?

PR #1582, opened 2024-04-27
fix(java): grow buffer to interger.max

chaokunyang:
## What does this PR do?

grow buffer to interger.max

## Related issues

#1576 

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
@@ -1210,7 +1210,8 @@ public final class MemoryBuffer {
   /** For off-heap buffer, this will make a heap buffer internally. */
   public void ensure(int length) {
     if (length > size) {
-      byte[] data = new byte[length * 2];
+      int newSize = length < 104857600 ? length << 2 : (int) Math.min(length * 1.5d, Integer.MAX_VALUE);
+      byte[] data = new byte[newSize];
       copyToUnsafe(0, data, Platform.BYTE_ARRAY_OFFSET, size());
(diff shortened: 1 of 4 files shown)


java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
```diff
@@ -1204,18 +1204,27 @@ public void writePrimitiveArray(Object arr, int offset, int numBytes) {
...
-      byte[] data = new byte[length * 2];
-      copyToUnsafe(0, data, Platform.BYTE_ARRAY_OFFSET, size());
-      initHeapBuffer(data, 0, data.length);
+      growBuffer(length);
     }
   }
 
+  private void growBuffer(int length) {
+    int newSize =
+        length < 104857600 ? length << 2 : (int) Math.min(length * 1.5d, Integer.MAX_VALUE);
```
LiangliangSui, line 1222: How is the value `104857600` calculated?

chaokunyang: It' just 100 * 1024 * 1024

LiangliangSui: Move `org.apache.fury.io.FuryStreamReader#BUFFER_GROW_STEP_THRESHOLD` to `MemoryBuffer` and reuse it?

chaokunyang: good suggestion, updated

LiangliangSui (approved):
LGTM, thanks!

PR #1590, opened 2024-04-27
feat(spec): remove list/map header from type meta spec

chaokunyang:
## What does this PR do?

This PR removes list/map header from type meta spec. 

Such header may be computed ahead sometimes. But it may need to compute based the data at runtime. If we write header into type meta, we may still need to compute and write header when serialization. This will introduce extra space cost.

Instead, we can write all such header at runtime, and when creating/generating serializer, we can compute part of such header ahead, and let remaing parts of header computed at serialization.

When deserialization, the deserializer can generate different deserializer based on read header, which will be more efficient.

## Related issues
#1413


chaokunyang:
cc @theweipeng @PragmaTwice @LiangliangSui

LiangliangSui (approved):
LGTM

theweipeng (approved):
LGTM

PR #1591, opened 2024-04-28
perf(java): Reduce performance regression caused by deleteCharAt

LiangliangSui:
## What does this PR do?

This PR reduces the performance regression caused by StringBuilder#deleteCharAt, avoids using deleteCharAt, and makes early judgments in while.

## Benchmark

before this PR
image

with this pr
image

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringDecoder.java
+++ b/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringDecoder.java
@@ -74,7 +74,7 @@ public class MetaStringDecoder {
     boolean stripLastChar = (data[0] & 0x80) != 0; // Check the first bit of the first byte
     int bitMask = 0b11111; // 5 bits for the mask
     int bitIndex = 1; // Start from the second bit
-    while (bitIndex + 5 <= totalBits) {
+    while (bitIndex + 5 <= totalBits && !(stripLastChar && (bitIndex + 2 * 5 > totalBits))) {
       int byteIndex = bitIndex / 8;
       int intraByteIndex = bitIndex % 8;
       // Extract the 5-bit character value across byte boundaries if needed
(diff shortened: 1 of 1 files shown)


chaokunyang:
MetaString is not int critical path, maybe we don't need to optimize this?

LiangliangSui:
Through this optimization we can also reduce the operations of appending and deleting to StringBuilder. Although it is not on the critical path, it is still profitable.

java/fury-core/src/main/java/org/apache/fury/meta/MetaStringDecoder.java
```diff
@@ -74,7 +74,7 @@ private String decodeLowerSpecial(byte[] data) {
     boolean stripLastChar = (data[0] & 0x80) != 0; // Check the first bit of the first byte
     int bitMask = 0b11111; // 5 bits for the mask
     int bitIndex = 1; // Start from the second bit
-    while (bitIndex + 5 <= totalBits) {
+    while (bitIndex + 5 <= totalBits && !(stripLastChar && (bitIndex + 2 * 5 > totalBits))) {
```
chaokunyang, line 77: Could you explain what `!(stripLastChar && (bitIndex + 2 * 5 > totalBits))` mean? It seems not intutive to me.

LiangliangSui: `(bitIndex + 2 * 5 > totalBits)` indicates that the current character is the last character.

`!(stripLastChar && (bitIndex + 2 * 5 > totalBits))` means that if the current character is the last character and the last character needs to be deleted, the while loop will not be entered.

This can reduce the last redundant `StringBuilder#append -> StringBuilder#deleteCharAt`

chaokunyang: Got it, thanks

chaokunyang (approved):
LGTM

Issue #1595, opened 2024-04-29
[Java] Support userContext for user custom global data


MrChang0:
## Is your feature request related to a problem? Please describe.

when I want compress some `Map<String, String> map`,  it have lots of same string key, I want make user serializer contain a map key dict suche as `Map<String, Integer> dict` and I can just serializer key as int, then write dict data after metaContext.
but fury cant costum metaContext to save user global data.

btw, `stringRefIgnored` open will take slow serilizer time.

## Describe the solution you'd like

maybe we can support user context with open interface like interface `userContext` with method `wirteData(MemoryBuffer buffer)` and `write(MemoryBuffer buffer)` and `constructr(Fury fury)`. we can take it just like `fury.registerUserContext("name", CompressStringDictMapContext.class)`.  we can take userContext like `fury.getUserContext("name")`
when after wirte metaContext, wirte all register userContext data by register order, same to read.

## Additional context

I make a simple compress map version in my local project, so I think a userContext can take better flexibility

thanks

chaokunyang:
Hi, Fury has a `SerializationContext` class. Could we use this class to implement this feature?

MrChang0:
sure, I will take a simple code for this feature.

PR #1596, opened 2024-04-29
feat(java): support user context for serialize global data(#1595)

MrChang0:
## What does this PR do?

support user context for serialize user global context in some special sence.
the changes in this pr are too simple and without design, just an example for the issue.

## Related issues

- #1595


MrChang0:
maybe it could be. I use Fury for RPC and have a big trouble is fury option can't change when code is going on. if we have more 
flexible context so that Fury could transmit data like ' fury options'/'user data'.

chaokunyang:
> maybe it could be. I use Fury for RPC and have a big trouble is fury option can't change when code is going on. if we have more
> flexible context so that Fury could transmit data like ' fury options'/'user data'.

Could you share more details why you want to change fury options at runtime. It is designed to be immutable

MrChang0:
> > maybe it could be. I use Fury for RPC and have a big trouble is fury option can't change when code is going on. if we have more
> > flexible context so that Fury could transmit data like ' fury options'/'user data'.
> 
> Could you share more details why you want to change fury options at runtime. It is designed to be immutable

hoho,  just image a scence for this feature. maybe the deserializer can deserialize any Fury data even options is different. for now I dont need it. I just need a user context to save custom data.

java/fury-core/src/main/java/org/apache/fury/resolver/UserContextResolver.java
```diff
@@ -0,0 +1,40 @@
...
+ * under the License.
+ */
+
+package org.apache.fury.resolver;
+
+import org.apache.fury.Fury;
+import org.apache.fury.memory.MemoryBuffer;
+
+/** write/read user custom global data after metaContext data. */
+public abstract class UserContextResolver {
```
chaokunyang, line 26: How about naming it as `UserContext` just like `MetaContext`
```suggestion
public abstract class UserContext {
```

java/fury-core/src/main/java/org/apache/fury/resolver/UserContextResolver.java
```diff
@@ -0,0 +1,40 @@
...
+  public UserContextResolver(Fury fury) {
+    this.fury = fury;
+  }
+
+  public abstract void write(MemoryBuffer buffer);
+
+  public abstract void read(MemoryBuffer buffer);
+
+  /** write/read end should clear user data. */
+  public abstract void reset();
```
chaokunyang, line 39: In meta share mode, we can write this context in a channel only once. If we reset it every time, does it means that we will write this every time?

MrChang0: I would like context used without state, context is global in a fury and reset every write/read. if we do it like meta context, we need store contexts and fury. so I think context design should be same and I wander your advice.

java/fury-core/src/main/java/org/apache/fury/resolver/UserContextResolver.java
MrChang0: ok

java/fury-core/src/main/java/org/apache/fury/resolver/UserContextResolver.java
chaokunyang: If it's stateless, we can implement this like :
```java
  @Override
  public void write(MemoryBuffer buffer, Object value) {
    Map<String, Integer> o = (Map<String, Integer>) fury.getSerializationContext().get("dict");
    if (o == null) {
      o = xxx;
      fury.getSerializationContext().add("dict", o);
      fury.writeNonRef(buffer, o);
      buffer.writeBoolean(true);
    } else {
      buffer.writeBoolean(false);
    }
    
  }

  @Override
    Map<String, Integer> o;
    if (buffer.readBoolean()) {
      o = (Map<String, Integer>) fury.readNonRef(buffer);
      fury.getSerializationContext().add("dict", o);
    } else {
      o = (Map<String, Integer>) fury.getSerializationContext().get("dict");
    }
  }
```

But it introduce one extra byte

MrChang0: with `UserContext#reset()` just like `SerializationContext#reset()`, take user a choice  whether clean their data after fury serialize/deserialize. we can save tmp data in userContext and not in `SerializationContext#objects`.

PR #1598, opened 2024-05-03
chore(java): make enum serializer as an upper level class

chaokunyang:
## What does this PR do?

Enum is a core concept in fury serialization, and we will have more specific optimization on this type.

This make enum serializer as an upper level class, so we can make all enum releated code more cohesive


LiangliangSui (approved):
LGTM

PR #1603, opened 2024-05-06
feat(sepc): update type meta field info spec

chaokunyang:
## What does this PR do?

Update type meta field info spec:
```
- field info:
    - header(8
      bits): `3 bits size + 2 bits field name encoding + polymorphism flag + nullability flag + ref tracking flag`.
      Users can use annotation to provide those info.
        - 2 bits field name encoding:
            - encoding: `UTF8/ALL_TO_LOWER_SPECIAL/LOWER_UPPER_DIGIT_SPECIAL/TAG_ID`
            - If tag id is used, i.e. field name is written by an unsigned varint tag id. 2 bits encoding will be `11`.
        - size of field name:
            - The `3 bits size: 0~7`  will be used to indicate length `1~7`, the value `7` indicates to read more bytes,
              the encoding will encode `size - 7` as a varint next.
            - If encoding is `TAG_ID`, then num_bytes of field name will be used to store tag id.
        - ref tracking: when set to 1, ref tracking will be enabled for this field.
        - nullability: when set to 1, this field can be null.
        - polymorphism: when set to 1, the actual type of field will be the declared field type even the type if
          not `final`.
    - field name: If tag id is set, tag id will be used instead. Otherwise meta string encoding `[length]` and data will
      be written instead.
```

## Related issues
#1556 

### DIFF
--- a/docs/specification/xlang_serialization_spec.md
+++ b/docs/specification/xlang_serialization_spec.md
@@ -288,23 +288,27 @@ Meta header is a 64 bits number value encoded in little endian order.
       fields, then use fields info in meta for deserializing compatible fields.
 - type id: the registered id for the current type, which will be written as an unsigned varint.
 - field info:
-    - Header(8 bits):
-        - Format:
-            - `reserved 1 bit + 3 bits field name encoding + polymorphism flag + nullability flag + ref tracking flag + tag id flag`.
-        - Users can use annotation to provide that info.
-            - tag id: when set to 1, the field name will be written by an unsigned varint tag id.
-            - ref tracking: when set to 0, ref tracking will be disabled for this field.
-            - nullability: when set to 0, this field won't be null.
-            - polymorphism: when set to 1, the actual type of field will be the declared field type even the type if
-              not `final`.
-            - 3 bits field name encoding will be set to meta string encoding flags when tag id is not set.
-    - Type id:
+    - header(8
+      bits): `3 bits size + 2 bits field name encoding + polymorphism flag + nullability flag + ref tracking flag`.
+      Users can use annotation to provide those info.
+        - 2 bits field name encoding:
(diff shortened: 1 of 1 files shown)


docs/specification/xlang_serialization_spec.md
```diff
@@ -288,23 +288,27 @@ Meta header is a 64 bits number value encoded in little endian order.
...
-    - Type id:
+    - header(8
+      bits): `3 bits size + 2 bits field name encoding + polymorphism flag + nullability flag + ref tracking flag`.
+      Users can use annotation to provide those info.
+        - 2 bits field name encoding:
+            - encoding: `UTF8/ALL_TO_LOWER_SPECIAL/LOWER_UPPER_DIGIT_SPECIAL/TAG_ID`
+            - If tag id is used, i.e. field name is written by an unsigned varint tag id. 2 bits encoding will be `11`.
+        - size of field name:
+            - The `3 bits size: 0~7`  will be used to indicate length `1~7`, the value `7` indicates to read more bytes,
+              the encoding will encode `size - 7` as a varint next.
```
LiangliangSui, line 299: > The `3 bits size: 0~7` will be used to indicate length `1~7`, the value `7` indicates to read more bytes,

Do we give up using 0? I don't understand this paragraph very thoroughly. Could you help explain it in depth?

LiangliangSui: So `0(0b000)~6(0b110)` means `1~7`, and `7 (0b111)` means there are more bytes. Can I understand it this way?

chaokunyang: Yes, that's what it means

LiangliangSui: Got it, thanks!

LiangliangSui (approved):
LGTM

PR #1613, opened 2024-05-08
feat(java): Support CopyOnWriteArrayListSerializer

MrChang0:
## What does this PR do?

Support CopyOnWriteArrayListSerializer

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
@@ -259,36 +259,21 @@ public class CollectionSerializers {
   public static class CopyOnWriteArrayListSerializer
       extends CollectionSerializer<CopyOnWriteArrayList> {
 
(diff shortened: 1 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
```diff
@@ -255,6 +256,42 @@ public List<?> xread(MemoryBuffer buffer) {
...
   }
 
+  public static class CopyOnWriteArrayListSerializer
+      extends CollectionSerializer<CopyOnWriteArrayList> {
+
+    private static final long ARRAY_OFFSET;
+
+    static {
+      try {
+        final Field field = CopyOnWriteArrayList.class.getDeclaredField("array");
```
chaokunyang, line 266: Could we implement this serializer like ImmutableListSerializer. The array field here won't be referenced by other objects. So we can use collection serialization protocol

chaokunyang (approved):
LGTM

PR #1614, opened 2024-05-08
refactor(java): Remove Guava's Concurrency utils usages

Munoon:
## What does this PR do?

Removes Guava's Concurrency utils usages

## Related issues

#1113

## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

PR #1616, opened 2024-05-09
feat(java): add set serializer for concurrent set

MrChang0:
## What does this PR do?

we use `Sets.newConcurrentHashSet()`  to create set, with lastest guava version it use `ConcurrentHashMapKeySetView` and old version it use `Collections.newSetFromMap(map)`. 

remove `CopyOnWriteArrayListSerializer` from native-image.properties.  #1614 forget delete.

### DIFF
--- a/integration_tests/graalvm_tests/src/main/java/org/apache/fury/graalvm/Collection.java
+++ b/integration_tests/graalvm_tests/src/main/java/org/apache/fury/graalvm/Collection.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fury.graalvm;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import org.apache.fury.Fury;
+import org.apache.fury.util.Preconditions;
+
+public class Collection {
+  static Fury fury;
+
+  static {
+    fury = Fury.builder().requireClassRegistration(true).build();
+  }
+
+  static void test(Fury fury) {
+    final Map<String, String> unmodifiableMap = Map.of("k1", "v1", "k2", "v2");
+    Preconditions.checkArgument(
+        unmodifiableMap.equals(fury.deserialize(fury.serialize(unmodifiableMap))));
+    System.out.println(unmodifiableMap);
+    final List<Integer> arrayasList = Arrays.asList(1, 2, 3);
+    Preconditions.checkArgument(arrayasList.equals(fury.deserialize(fury.serialize(arrayasList))));
+    System.out.println(arrayasList);
+    final Set<String> setFromMap = Collections.newSetFromMap(new ConcurrentHashMap<>());
+    setFromMap.add("a");
+    setFromMap.add("b");
+    setFromMap.add("c");
+    Preconditions.checkArgument(setFromMap.equals(fury.deserialize(fury.serialize(setFromMap))));
+    System.out.println(setFromMap);
+  }
+
+  public static void main(String[] args) {
+    test(fury);
+    System.out.println("Collection succeed");
+  }
+}
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
@@ -49,7 +49,6 @@ import org.apache.fury.exception.FuryException;
 import org.apache.fury.memory.MemoryBuffer;
 import org.apache.fury.memory.Platform;
 import org.apache.fury.reflect.ReflectionUtils;
-import org.apache.fury.resolver.ClassInfo;
 import org.apache.fury.resolver.ClassResolver;
 import org.apache.fury.serializer.ReplaceResolveSerializer;
 import org.apache.fury.serializer.Serializer;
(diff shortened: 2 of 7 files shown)


java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
```diff
@@ -411,6 +415,66 @@ public ConcurrentSkipListSet newCollection(MemoryBuffer buffer) {
...
+    public SetFromMapSerializer(Fury fury, Class<Set<?>> type) {
+      super(fury, type, false);
+    }
+
+    @Override
+    public Collection newCollection(MemoryBuffer buffer) {
+      int numElements = buffer.readVarUint32Small7();
+      setNumElements(numElements);
+      final ClassInfo mapClassInfo = fury.getClassResolver().readClassInfo(buffer);
+      final MethodHandle methodHandle = ReflectionUtils.getCtrHandle(mapClassInfo.getCls(), true);
```
chaokunyang, line 440: This won't work for graalvm. It's too late, we should move it to constructor, so we can generate code at graalvm  build time

MrChang0: ok, I change it. I test this version in graalvm and it works, I wandering why.

chaokunyang: Could you add a test in graalvm_tests module?

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
```diff
@@ -411,6 +414,60 @@ public ConcurrentSkipListSet newCollection(MemoryBuffer buffer) {
...
+    @Override
+    public void write(MemoryBuffer buffer, Set value) {
+      final Map<?, Boolean> map = (Map<?, Boolean>) Platform.getObject(value, MAP_FIELD_OFFSET);
+      fury.writeRef(buffer, map);
+    }
+
+    @Override
+    public Set<?> read(MemoryBuffer buffer) {
+      final Map<?, Boolean> map = (Map<?, Boolean>) fury.readRef(buffer);
+      final Set<?> set = Collections.newSetFromMap(Collections.emptyMap());
```
chaokunyang, line 447: Why not just invoke `Collections.newSetFromMap(map)`

integration_tests/graalvm_tests/src/main/java/org/apache/fury/graalvm/Collection.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import org.apache.fury.Fury;
+import org.apache.fury.util.Preconditions;
+
+public class Collection {
```
chaokunyang, line 31: This may be confusing with java.util.Collection. How about naming it as `CollectionTest`

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
```diff
@@ -411,6 +415,66 @@ public ConcurrentSkipListSet newCollection(MemoryBuffer buffer) {
...
+        throw new RuntimeException(e);
+      }
+    }
+
+    public SetFromMapSerializer(Fury fury, Class<Set<?>> type) {
+      super(fury, type, false);
+    }
+
+    @Override
+    public Collection newCollection(MemoryBuffer buffer) {
```
chaokunyang, line 436: I just found that this method should not be used if collection `codegen` is false. 
We can just implement serialization like this:
```java

    @Override
    public void write(MemoryBuffer buffer, Collection value) {
      Object fieldValue = Platform.getObject(value, offset);
      fury.writeRef(buffer, fieldValue);
    }

    @Override
    public Collection read(MemoryBuffer buffer) {
      final Map map = (Map) fury.readRef(buffer);
      return (Collection) Collections.newSetFromMap(map);
    }
```

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
```diff
@@ -433,28 +436,18 @@ public SetFromMapSerializer(Fury fury, Class<Set<?>> type) {
...
 
     @Override
-    public Collection onCollectionWrite(MemoryBuffer buffer, Set<?> value) {
-      buffer.writeVarUint32Small7(value.size());
-      final Map<?, Boolean> map = (Map<?, Boolean>) Platform.getObject(value, MAP_FIELD_OFFSET);
-      fury.getClassResolver().writeClassAndUpdateCache(buffer, map.getClass());
-      return value;
+    public Set<?> read(MemoryBuffer buffer) {
+      final Map map = (Map) fury.readRef(buffer);
+      final Set set = Collections.newSetFromMap(Collections.emptyMap());
```
chaokunyang, line 447: How about:
```java
final Map map = (Map) fury.readRef(buffer);
return Collections.newSetFromMap(map);
```

MrChang0:
what fuck [discussions](https://github.com/orgs/community/discussions/10369)

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
chaokunyang: I see, thanks. If only allow users to create set first, then add values.

If so, the serialization should be simplified.

Here is the serialization code:
```java
public boolean add(E e) { return m.put(e, Boolean.TRUE) == null; }
```

All map values should not be serialized. And map will be empty, so there is not shared reference for this map. 

Your previous implementation seems better, only need to set codegen to true. So fury will generate serialization code for this set. In `onCollectonWrite`, you write map class, and `newCollection`, you read class and create map from it. You can get serializer for that map, and it must be a AbstractMapSerializer, then you can use ` AbstractMapSerializer#newMap` to create the empty map, and build the set from it.

Current implementation is OK for me too. The default implementation in Fury will serializer whole map too. 

We can merge this PR first, and create another PR to optimize it if you like .

chaokunyang (approved):
LGTM

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
MrChang0: I prefer to save map classInfo, It looks likes `set serializer`  and less reflection operate. anyway tomorrow is workday, I will finish it.

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
```diff
@@ -418,36 +419,39 @@ public static final class SetFromMapSerializer extends CollectionSerializer<Set<
...
     @Override
-    public Set<?> read(MemoryBuffer buffer) {
-      final Map map = (Map) fury.readRef(buffer);
-      final Set set = Collections.newSetFromMap(Collections.emptyMap());
-      Platform.putObject(set, MAP_FIELD_OFFSET, map);
-      Platform.putObject(set, KEY_SET_FIELD_OFFSET, map.keySet());
-      return (Set<?>) set;
+    public Collection onCollectionWrite(MemoryBuffer buffer, Set<?> value) {
+      final Map<?, Boolean> map = (Map<?, Boolean>) Platform.getObject(value, MAP_FIELD_OFFSET);
+      fury.getClassResolver().writeClassAndUpdateCache(buffer, map.getClass());
```
chaokunyang, line 451: UpdateCache is unnecessary, next written item will be different types. Other part looks great

MrChang0: fine, I careless.

PR #1620, opened 2024-05-10
feat(go/java): Add ASCII check before meta string encoding

jasonmokk:
## What does this PR do?

This PR introduces a validation method to ensure that all input strings to the `MetaString` encoder are ASCII.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
+++ b/java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
@@ -49,6 +49,9 @@ public class MetaStringEncoder {
    * @return A MetaString object representing the encoded string.
    */
   public MetaString encode(String input) {
+    if (!isASCII(input)) {
+      throw new IllegalArgumentException("Non-ASCII characters in meta string are not allowed");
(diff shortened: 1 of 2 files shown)


java/fury-core/src/main/java/org/apache/fury/meta/MetaStringEncoder.java
```diff
@@ -70,6 +70,9 @@ public MetaString encode(String input, Encoding[] encodings) {
   public MetaString encode(String input, Encoding encoding) {
     Preconditions.checkArgument(
         input.length() < Short.MAX_VALUE, "Long meta string than 32767 is not allowed");
+    if (encoding != Encoding.UTF_8 && !isASCII(input)) {
```
chaokunyang, line 73: Just found that we can use `org.apache.fury.serializer.StringSerializer#isLatin` in java instead. `isLatin` is OK too here

jasonmokk: Got it, PR has been updated. Thanks for the review!

chaokunyang (approved):
LGTM, thanks

PR #1622, opened 2024-05-10
feat(java): register old version guava collect

MrChang0:
## What does this PR do?

before guava 19.0, of() return EmptyImmutableSet/EmptyImmutableBiMap/EmptyImmutableSortedMap/EmptyImmutableSortedSet. check class and register serializer for them.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/collection/GuavaCollectionSerializers.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/collection/GuavaCollectionSerializers.java
@@ -421,7 +421,7 @@ public class GuavaCollectionSerializers {
       class GuavaEmptyBiMap {}
 
       cls = GuavaEmptyBiMap.class;
-      fury.registerSerializer(cls, new ImmutableSetSerializer(fury, cls));
+      fury.registerSerializer(cls, new ImmutableMapSerializer(fury, cls));
     }
     if (checkClassExist(pkg + ".EmptyImmutableSortedSet")) {
       cls = loadClass(pkg + ".EmptyImmutableSortedSet", ImmutableSortedSet.of().getClass());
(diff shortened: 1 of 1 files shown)


java/fury-core/src/main/java/org/apache/fury/serializer/collection/GuavaCollectionSerializers.java
```diff
@@ -399,6 +400,26 @@ public static void registerDefaultSerializers(Fury fury) {
     fury.registerSerializer(cls, new ImmutableSortedSetSerializer<>(fury, cls));
     cls = loadClass(pkg + ".ImmutableSortedMap", ImmutableSortedMap.of(1, 2).getClass());
     fury.registerSerializer(cls, new ImmutableSortedMapSerializer<>(fury, cls));
+
+    // Guava version before 19.0, of() return
+    // EmptyImmutableSet/EmptyImmutableBiMap/EmptyImmutableSortedMap/EmptyImmutableSortedSet
+    // we register if class exist
+    if (checkClassExist(pkg + ".EmptyImmutableSet")) {
```
chaokunyang, line 407: If caller and server have different guava version, this method seems introduce deserialization error

chaokunyang: Could we make classes before and after guava 19.0 mapped to same class id. In this way, teh different guava version across processes won't introduce deserialization error

MrChang0: > Could we make classes before and after guava 19.0 mapped to same class id. In this way, teh different guava version across processes won't introduce deserialization error

I perfer it more, with different guava version, it can take `of()` to deserialize.

chaokunyang: If classid are different, the deserialization will laod wrong class and deserializer. We met such errors before. it's very hard to locate it

MrChang0: > If classid are different, the deserialization will laod wrong class and deserializer. We met such errors before. it's very hard to locate it

I see comments, and I take same classId to deserialize low version guava.

chaokunyang (approved):
Looks great

PR #1626, opened 2024-05-13
fix(java): subclass without fields will encode superclass

MrChang0:
## What does this PR do?

when subclass without fields. `org.apache.fury.meta.ClassDefEncoder#getClassFields` will skip subclass so className will be superclass. however we should save class even fields is empty.


chaokunyang:
Thanks for this fix. Could we also add a test in `MetaSharedCompatibleTest`?

chaokunyang (approved):
Looks great, thanks

PR #1628, opened 2024-05-13
fix(java): ThreadPoolFury#factoryCallback don't work when create new classLoaderFuryPooled

MrChang0:
## What does this PR do?

ThreadPoolFury#factoryCallback don't work when create new classLoaderFuryPooled.


chaokunyang (approved):
Nice catch, I missed classloader change case

PR #1631, opened 2024-05-14
fix(go/java): Enhance ASCII check in meta string encoding

jasonmokk:
## What does this PR do?

This PR enhances the current ASCII check (before meta string encoding) I implemented in #1620 to return a UTF-8 encoded `MetaString` early if the input is non-ASCII. This improves efficiency and saves time on `computeEncoding` and `encode`. Unit tests are also added.

## Related issues
https://github.com/apache/incubator-fury/issues/1619


LiangliangSui (approved):
Looks good to me, thanks!

PR #1648, opened 2024-05-24
feat: add a script to automatically generate doc

LiangliangSui:
## What does this PR do?

Add a script to automatically generate doc, Implement the first stage goals mentioned in https://github.com/apache/incubator-fury/issues/1647.

## Related issues

https://github.com/apache/incubator-fury/issues/1647


chaokunyang:
Could we use parse javadoc html instead?

LiangliangSui:
> Could we use parse javadoc html instead?

Good suggestion!

It can be implemented by parsing javadoc html. Even if it is implemented by parsing html, we need to clearly point out which config fields are in the comments. 

In addition, we need to clearly point out the defaultValue in the comments of each filed, because it will not be included in javadoc. The true default value of the field (for example, `false` in `boolean enable = false` will not appear in the javadoc html)

Our config field comments can be designed in the following format?

```
/**
 * xxxxxxxxxxxxx
 *
 * @defaultValue: xxxx
 */
```

LiangliangSui:
Do you have any other suggestions for generating Tables based on Fields and writing them to `java_serialization_guide.md`? @chaokunyang

LiangliangSui:
Demo

image

LiangliangSui:
Hi @chaokunyang , I implemented this function based on parsing javadoc. Could you help review this?

tools/gen_fury_builder_doc.py
```diff
@@ -0,0 +1,224 @@
...
+    )
+    with open(fury_build_src) as f:
+        content = f.read()
+    shutil.rmtree(javadoc_dir)
+
+    html = etree.HTML(content)
+    # There is only one `Field Detail`
+    field_detail = html.xpath('//div[@class="details"]//section[1]')[0]
+    if field_detail is None:
+        return 1
```
chaokunyang, line 216: how about raise exception with detailed message instead of exit by error code

LiangliangSui: Update

tools/gen_fury_builder_doc.py
```diff
@@ -0,0 +1,225 @@
...
+                    language to <a href="Language.html#JAVA"><code>Language.JAVA</code></a>, which will have much better performance.
+                </div>
+                <dl>
+                    <dt><span class="simpleTagLabel">defaultValue</span></dt>
+                    <dd>Language.JAVA</dd>
+                </dl>
+            </li>
+        </ul>
+        """
+        tag_labels = field.xpath('li/dl/dt/span[@class="simpleTagLabel"]/text()')
```
chaokunyang, line 72: This seems coupled with javadoc structure. WIll javadoc generated html change between JDK vendor and different JDK versions?

LiangliangSui: Yes, we may be able to automatically install the fixed version and vendor in the script before generating javadoc.

Or do you have any other good ideas?

PR #1650, opened 2024-05-27
perf(java): Lower the threshold for multi-byte reading

LiangliangSui:
## What does this PR do?

We only need to ensure that `_unsafeGetInt32` and `_unsafeGetInt64` can take out values, which can reduce the performance loss caused by multiple readBytes.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
@@ -1625,7 +1625,7 @@ public final class MemoryBuffer {
     // noinspection Duplicates
     int readIdx = readerIndex;
     int result;
-    if (size - readIdx < 5) {
+    if (size - readIdx < 4) {
       result = (int) readVarUint36Slow();
     } else {
       long address = this.address;
@@ -1668,7 +1668,7 @@ public final class MemoryBuffer {
     // noinspection Duplicates
     int readIdx = readerIndex;
     int result;
-    if (size - readIdx < 5) {
+    if (size - readIdx < 4) {
       result = (int) readVarUint36Slow();
     } else {
       long address = this.address;
@@ -1706,7 +1706,7 @@ public final class MemoryBuffer {
     // Duplicate and manual inline for performance.
     // noinspection Duplicates
     int readIdx = readerIndex;
-    if (size - readIdx >= 9) {
+    if (size - readIdx >= 8) {
       long bulkValue = _unsafeGetInt64(readIdx++);
       // noinspection Duplicates
       long result = bulkValue & 0x7F;
@@ -1774,7 +1774,7 @@ public final class MemoryBuffer {
   /** Reads the 1-5 byte int part of a non-negative varint. */
   public int readVarUint32() {
     int readIdx = readerIndex;
-    if (size - readIdx < 5) {
+    if (size - readIdx < 4) {
       return (int) readVarUint36Slow();
     }
     // | 1bit + 7bits | 1bit + 7bits | 1bit + 7bits | 1bit + 7bits |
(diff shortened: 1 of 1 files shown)


chaokunyang:
Hi @LiangliangSui , could you elabrate more why we need this optimization? I don't see the performance loss before this PR

LiangliangSui:
> Hi @LiangliangSui , could you elabrate more why we need this optimization? I don't see the performance loss before this PR

Let's take `_readVarInt32OnLE` as an example. When `size - readIdx >= 4`, we can use `UNSAFE.getInt` to read out 4 bytes at a time. We do not need `size - readIdx >= 5` to read out 4 bytes at a time.

If `size - readIdx == 4`, according to the previous logic, we need to execute `readByte()` 4 times (read from memory 4 times), and the optimized logic only requires one getInt (operate the memory once).

The reason why we name some reading methods `readVarXXXSlow` should be that these methods will loop through `readByte()` to read data. Is my understanding correct?

java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
```diff
@@ -1774,7 +1774,7 @@ private long readVarUint36Slow() {
   /** Reads the 1-5 byte int part of a non-negative varint. */
   public int readVarUint32() {
     int readIdx = readerIndex;
-    if (size - readIdx < 5) {
+    if (size - readIdx < 4) {
```
chaokunyang, line 1777: later `result |= (UNSAFE.getByte(heapMemory, address + readIdx++) & 0x7F) << 28;` wqill read one more byte without check. If we lower threshold, we may get crash in such cases.

chaokunyang: Could you add a test to cover such cases?

LiangliangSui: > later result |= (UNSAFE.getByte(heapMemory, address + readIdx++) & 0x7F) << 28; wqill read one more byte without check. If we lower threshold, we may get crash in such cases.

The premise for executing `result |= (UNSAFE.getByte(heapMemory, address + readIdx++) & 0x7F) << 28` is that the fourth byte of `fourByteValue` is `0b1xxxxxxxx` (that is, there are more bytes to be parsed).

In `readVarUint36Slow()`, it is also possible to read 5 bytes continuously (provided that `(b & 0x80) != 0`).

Therefore, the premise of executing `result |= (UNSAFE.getByte(heapMemory, address + readIdx++) & 0x7F) << 28` is that we still have content that can be read in the Buffer. If there is no content to read, it means that an error occurred during writing.

chaokunyang: As you can see, we have `address + readIdx++`. It will read the 5rd bytes without range check. This will crash jvm if we don't have such big buffer or data is maliciously constructed

LiangliangSui: So the five consecutive `getByte` calls in `readVarUint36Slow` are no different from `UNSAFE.getInt + UNSAFE.getByte`. They all have read checks for `(fourByteValue & 0x80000000) != 0` and `(b & 0x80) != 0`.

We only need to ensure that `size - readIdx >= 4` to use `UNSAFE.getInt` to read the content normally.

chaokunyang: You read five bytes, but you just check 4 bytes here

chaokunyang: The readVarUint36Slow is very rare, only last five bytes of the whole binary will go to this path. So we try to reduce the check in critical path

LiangliangSui: As discussed in WeChat, this solution only benefits the last 4 bytes, but it will introduce extra code, causing the bytecode of this method to become larger. In most cases, this optimization will not take effect, so close this PR.

PR #1661, opened 2024-05-30
feat(java): Implement classId related logic

LiangliangSui:
## What does this PR do?

Implementing xlang_spec typeId.

## Related issues

https://github.com/apache/incubator-fury/pull/1413

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
@@ -160,45 +160,45 @@ public class ClassResolver {
   // bit 0 unset indicates class is written as an id.
   public static final byte USE_CLASS_VALUE_FLAG = 0b1;
   // preserve 0 as flag for class id not set in ClassInfo`
-  public static final short NO_CLASS_ID = (short) 0;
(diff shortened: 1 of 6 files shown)


LiangliangSui:
Hi @chaokunyang , could you help me check if there is any problem with the implementation logic? If there is no problem, I will handle CI related errors.

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -160,45 +160,45 @@ public class ClassResolver {
   // bit 0 unset indicates class is written as an id.
   public static final byte USE_CLASS_VALUE_FLAG = 0b1;
   // preserve 0 as flag for class id not set in ClassInfo`
-  public static final short NO_CLASS_ID = (short) 0;
-  public static final short LAMBDA_STUB_ID = 1;
```
chaokunyang, line 164: Different languages have different type systems. The type id for java native serialization and xlang serialization may be different.

LiangliangSui: Ok, I see, so we need to define a new set of TypeId in java for xlang spec?

PR #1662, opened 2024-05-30
chore(rust): resolve deprecate warnings

waynexia:
## What does this PR do?

- Resolve deprecate warnings from `chrono` dep
- Remove patch version in Cargo.toml

## Related issues

N/A

No

### DIFF
--- a/rust/fury-derive/Cargo.toml
+++ b/rust/fury-derive/Cargo.toml
@@ -25,7 +25,12 @@ rust-version.workspace = true
 proc-macro = true
 
 [dependencies]
-proc-macro2 = { default-features = false, version = "1.0.66" }
-syn = { default-features = false, version = "2.0.26", features = ["parsing", "proc-macro", "derive", "printing"] }
-quote = { default-features = false, version = "1.0.31" }
-thiserror = { default-features = false, version = "1.0.43" }
+proc-macro2 = { default-features = false, version = "1.0" }
+syn = { default-features = false, version = "2.0", features = [
+    "parsing",
+    "proc-macro",
+    "derive",
+    "printing",
+] }
+quote = { default-features = false, version = "1.0" }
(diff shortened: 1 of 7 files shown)


rust/fury-derive/Cargo.toml
```diff
@@ -25,7 +25,12 @@ rust-version.workspace = true
 proc-macro = true
 
 [dependencies]
-proc-macro2 = { default-features = false, version = "1.0.66" }
```
waynexia, line 28: Cargo will try to resolve the "latest" package version that meets the semver rule. Hence the last patch version in Cargo.toml doesn't take efforts, unless use a strict match like `version = "=1.0.66"`

rust/tests/tests/test_complex_struct.rs
```diff
@@ -60,7 +60,7 @@ fn complex_struct() {
         op: Some("option".to_string()),
         op2: None,
         date: NaiveDate::from_ymd_opt(2025, 12, 12).unwrap(),
-        time: NaiveDateTime::from_timestamp_opt(1689912359, 0).unwrap(),
+        time: DateTime::from_timestamp(1689912359, 0).unwrap().naive_utc(),
```
waynexia, line 63: The semantic doesn't changes: https://docs.rs/chrono/latest/chrono/struct.NaiveDateTime.html#method.from_timestamp_opt

The previous call `NaiveDateTime::from_timestamp_opt` also returns UTC datetime

>Makes a new NaiveDateTime corresponding to a UTC date and time, from the number of non-leap seconds since the midnight UTC on January 1, 1970 (aka “UNIX timestamp”) and the number of nanoseconds since the last whole non-leap second.

chaokunyang (approved):
Looks great! Thanks for contributing to fury @waynexia

PR #1660, opened 2024-05-29
feat(java): scoped meta share mode for type forward/backward compaibility

chaokunyang:
## What does this PR do?

This PR implements scoped meta share mode for type forward/backward compaibility

## Related issues

#202 

## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

## Benchmark
Perf increased from `1900102.586` to `2430410.064`
```
Before:

Benchmark                                                       (bufferType)   (objectType)  (references)   Mode  Cnt        Score        Error  Units
fury_deserialize                              array  MEDIA_CONTENT         false  thrpt   10  2734151.212 ± 253921.628  ops/s
fury_deserialize_compatible                   array  MEDIA_CONTENT         false  thrpt   10  1900102.586 ±  62176.872  ops/s
furymetashared_deserialize_compatible         array  MEDIA_CONTENT         false  thrpt   10  3011439.327 ± 260518.752  ops/s

After:

Benchmark                                                       (bufferType)   (objectType)  (references)   Mode  Cnt        Score        Error  Units
fury_deserialize                              array  MEDIA_CONTENT         false  thrpt   10  2661186.814 ± 279377.198  ops/s
fury_deserialize_compatible                   array  MEDIA_CONTENT         false  thrpt   10  2430410.064 ± 164165.865  ops/s
furymetashared_deserialize_compatible         array  MEDIA_CONTENT         false  thrpt   10  3098083.064 ± 259391.053  ops/s
```

Size decreased from **732 to 577**:
```
Before
2024-05-30 01:00:49 INFO  FuryState:157 [fury_deserialize_compatible-jmh-worker-1] - ======> Fury | MEDIA_CONTENT | false | array | 732 |

After
2024-05-30 12:57:00 INFO  FuryState:157 [fury_deserialize_compatible-jmh-worker-1] - ======> Fury | MEDIA_CONTENT | false | array | 577 |
```

The

PR #1663, opened 2024-05-30
feat(java): support meta compression by Deflater

chaokunyang:
## What does this PR do?

This PR support meta compression and add Deflater as default compressor.

In our test, it can compress meta by reduce size of **243**  without introducing any performance cost:

```
before:
Fury | STRUCT | false | array | 1227 |

after
STRUCT | false | array | 984 |

```

## Related issues
#1660 


java/fury-core/src/main/java/org/apache/fury/meta/ClassDefEncoder.java
```diff
@@ -146,36 +135,63 @@ static MemoryBuffer encodeClassDef(
...
-        writeTypeName(buffer, encoded.f1);
+        writePkgName(classDefBuf, encoded.f0);
+        writeTypeName(classDefBuf, encoded.f1);
       }
-      writeFieldsInfo(buffer, fields);
+      writeFieldsInfo(classDefBuf, fields);
+    }
+    byte[] compressed =
+        classResolver
+            .getFury()
```
LiangliangSui, line 151: I had an idea when I was looking at the code recently. 

Could `org.apache.fury.config.Config` be set to `org.apache.fury.resolver.SerializationContext`?
I think it is more reasonable to get `Config` from `SerializationContext` during runtime. It feels a bit weird to get it from `Fury` object.

This will reduce the burden on the `Fury` class.

java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
```diff
@@ -251,6 +254,37 @@ public FuryBuilder withScopedMetaShare(boolean scoped) {
...
+                    + "may be incorrect if different compressor instance of same type "
+                    + "indicates different compressor.",
+                metaCompressor);
+            metaCompressor = MetaCompressor.typeEqualMetaCompressor(metaCompressor);
+          }
+          break;
+        } catch (NoSuchMethodException e) {
+          clz = clz.getSuperclass();
+        }
+      }
```
LiangliangSui, line 282: Could this part of the logic be moved to `MetaCompressor`? This is strongly related to `MetaCompressor` rather than `FuryBuilder`.

java/fury-core/src/main/java/org/apache/fury/meta/ClassDefEncoder.java
chaokunyang: Good suggestion, I like the idea to get config from context. But other get context, user must get fury first, then get config from context. It will introduce an extra indirection

java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
chaokunyang: good suggestion, updated

java/fury-core/src/main/java/org/apache/fury/meta/ClassDefEncoder.java
chaokunyang: I think it's not related to this PR, we can discuss it in another issue

LiangliangSui (approved):
LGTM

PR #1666, opened 2024-05-31
chore(doc): fix doc example code

wangjie-fourth:
## What does this PR do?
according to comments,  should set `requireClassRegistration` to false

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,8 +159,6 @@ public class Example {
     // multiple serializations of different objects.
     {
       Fury fury = Fury.builder().withLanguage(Language.JAVA)
-        // Allow to deserialize objects unknown types, more flexible
-        // but may be insecure if the classes contains malicious code.
         .requireClassRegistration(true)
         .build();
       // Registering types can reduce class name serialization overhead, but not mandatory.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -173,11 +173,8 @@ public class Example {
       ThreadSafeFury fury = Fury.builder().withLanguage(Language.JAVA)
         // Allow to deserialize objects unknown types, more flexible
         // but may be insecure if the classes contains malicious code.
-        .requireClassRegistration(true)
+        .requireClassRegistration(false)
```
chaokunyang, line 176: This is a dangerous option, how about removing this option from code in readme?

chaokunyang: If we set it to false, it may mislead users think that it's OK to set it to false. But it's not actually. Disable check may introduce remote code execution

wangjie-fourth: your are right. But this looks like comments and code are opposite. How about keep `requireClassRegistration` to true, and delete comments?

chaokunyang: Looks good, could you update it in this PR

wangjie-fourth: done. please help review again

docs/guide/java_serialization_guide.md
```diff
@@ -83,7 +83,7 @@ public class Example {
   private static final ThreadSafeFury fury = Fury.builder()
     // Allow to deserialize objects unknown types, more flexible
     // but may be insecure if the classes contains malicious code.
-    .requireClassRegistration(true)
+    .requireClassRegistration(false)
```
chaokunyang, line 86: Could we delete comments and keep it as true? And invoke `fury.register(SomeClass)`  at the initialization

wangjie-fourth: done. please help review again

chaokunyang (approved):
Looks great

PR #1667, opened 2024-05-31
perf(java): Add ClassInfo ClassBytes generation conditions.

LiangliangSui:
## What does this PR do?

`ClassInfo#classNameBytes` and `ClassInfo#packageNameBytes` are only used when `classInfo.classId == NO_CLASS_ID && !metaContextShareEnabled`

```java
public void writeClass(MemoryBuffer buffer, ClassInfo classInfo) {
  if (classInfo.classId == NO_CLASS_ID) { // no class id provided.
    // use classname
    if (metaContextShareEnabled) {
      buffer.writeByte(USE_CLASS_VALUE_FLAG);
      // FIXME(chaokunyang) Register class but not register serializer can't be used with
      //  meta share mode, because no class def are sent to peer.
      writeClassWithMetaShare(buffer, classInfo);
    } else {
      // if it's null, it's a bug.
      assert classInfo.packageNameBytes != null;
      metaStringResolver.writeMetaStringBytesWithFlag(buffer, classInfo.packageNameBytes);
      assert classInfo.classNameBytes != null;
      metaStringResolver.writeMetaStringBytes(buffer, classInfo.classNameBytes);
    }
  } else {
    // use classId
    buffer.writeVarUint32(classInfo.classId << 1);
  }
}
```

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/meta/Encoders.java
+++ b/java/fury-core/src/main/java/org/apache/fury/meta/Encoders.java
@@ -36,10 +36,10 @@ import org.apache.fury.util.StringUtils;
 
 /** A class used to encode package/class/field name. */
 public class Encoders {
-  public static final MetaStringEncoder PACKAGE_ENCODER = new MetaStringEncoder('.', '_');
-  public static final MetaStringDecoder PACKAGE_DECODER = new MetaStringDecoder('.', '_');
-  public static final MetaStringEncoder GENERIC_ENCODER = PACKAGE_ENCODER;
-  public static final MetaStringDecoder GENERIC_DECODER = PACKAGE_DECODER;
+  public static final MetaStringEncoder GENERIC_ENCODER = new MetaStringEncoder('.', '_');
+  public static final MetaStringDecoder GENERIC_DECODER = new MetaStringDecoder('.', '_');
+  public static final MetaStringEncoder PACKAGE_ENCODER = GENERIC_ENCODER;
+  public static final MetaStringDecoder PACKAGE_DECODER = GENERIC_DECODER;
   public static final MetaStringEncoder TYPE_NAME_ENCODER = new MetaStringEncoder('$', '_');
   public static final MetaStringDecoder TYPE_NAME_DECODER = new MetaStringDecoder('$', '_');
   public static final String ARRAY_PREFIX = "1";


java/fury-core/src/main/java/org/apache/fury/resolver/ClassInfo.java
```diff
@@ -87,13 +86,15 @@ public class ClassInfo {
       this.fullClassNameBytes = null;
     }
     if (cls != null
-        && (classId == ClassResolver.NO_CLASS_ID || classId == ClassResolver.REPLACE_STUB_ID)) {
+        && ((classId == ClassResolver.NO_CLASS_ID
+                && !classResolver.getFury().getConfig().isMetaShareEnabled())
+            || classId == ClassResolver.REPLACE_STUB_ID)) {
```
LiangliangSui, line 91: In addition, there is another question, that is, why do we need to create `packageNameBytes` and `classNameBytes` when classId is `ClassResolver.REPLACE_STUB_ID`? I see `packageNameBytes` and `classNameBytes` are only used when `classInfo.classId == NO_CLASS_ID && !metaContextShareEnabled`

chaokunyang: Sometimes only classes are serialized, not the instance. If we serialize such class only, we need to write classname bytes. Could you document it?

java/fury-core/src/main/java/org/apache/fury/resolver/ClassInfo.java
```diff
@@ -86,14 +85,19 @@ public class ClassInfo {
...
       // REPLACE_STUB_ID for write replace class in `ClassSerializer`.
       Tuple2<String, String> tuple2 = Encoders.encodePkgAndClass(cls);
       this.packageNameBytes =
-          metaStringResolver.getOrCreateMetaStringBytes(PACKAGE_ENCODER.encode(tuple2.f0));
+          metaStringResolver.getOrCreateMetaStringBytes(Encoders.encodePackage(tuple2.f0));
       this.classNameBytes =
-          metaStringResolver.getOrCreateMetaStringBytes(TYPE_NAME_ENCODER.encode(tuple2.f1));
+          metaStringResolver.getOrCreateMetaStringBytes(Encoders.encodeTypeName(tuple2.f1));
```
chaokunyang, line 100: Could you check whether we need to update MetaStringBytes.decode

LiangliangSui: Updated.

`MetaStringBytes.decode` can also reuse existing `MetaStringDecoder` objects

LiangliangSui: We can also create a `Decoders` class, similar to `org.apache.fury.meta.Encoders`, which can be used to cache `ConcurrentMap<MetaString, String>`, merge this PR first, and then I will do this thing.

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -1701,7 +1701,7 @@ public Class<?> xreadClass(MemoryBuffer buffer) {
     Class<?> cls = classNameBytes2Class.get(byteString);
     if (cls == null) {
       Preconditions.checkNotNull(byteString);
-      String className = byteString.decode('.', '_');
+      String className = byteString.decode(Encoders.PACKAGE_DECODER);
```
chaokunyang, line 1704: Should not be PACKAGE_DECODER here for classname decode

java/fury-core/src/main/java/org/apache/fury/resolver/ClassInfo.java
chaokunyang: I prefer put it into Encoders instead

chaokunyang: > Updated.
> 
> `MetaStringBytes.decode` can also reuse existing `MetaStringDecoder` objects

You should update MetaStringBytes#decode too:
```java
  public String decode(MetaStringDecoder decoder) {
    int header = (int) (hashCode & 0xff);
    int encodingFlags = header & 0b111;
    MetaString.Encoding encoding = MetaString.Encoding.values()[encodingFlags];
    return decoder.decode(bytes, encoding);
  }
```
Currently it use the whole encoding set, but `Encoders.encodePackage` only use part of the encodings. The decode error will happen if a classname with unicode chars occur.

Could you add a unit test to cover such cases? It should raise error in your current change.

And the key here is that MetaStringBytes can't distinguish whether the data is  package or classname, so the decode need use the whole encoding set

LiangliangSui: I think we should not update `MetaStringBytes#decode` for the following reasons:

`Encoders.encodePackage` supports UTF-8 encoding by default, so even if the classname contains unicode characters, it can be encoded successfully.

```java
static final Encoding[] pkgEncodings =
      new Encoding[] {UTF_8, ALL_TO_LOWER_SPECIAL, LOWER_UPPER_DIGIT_SPECIAL};

static final Encoding[] typeNameEncodings =
    new Encoding[] {
      UTF_8, LOWER_UPPER_DIGIT_SPECIAL, FIRST_TO_LOWER_SPECIAL, ALL_TO_LOWER_SPECIAL
    };
```

I coded a unit test, as follows, which pass
```java
@Test
public void testEncoding() {
    String str = "你好，世界";
    MetaString metaString = Encoders.encodePackage(str);
    MetaStringBytes metaStringBytes = new MetaStringBytes(metaString);
    String decode = metaStringBytes.decode(Encoders.PACKAGE_DECODER);
    Assert.assertEquals(str, decode);
}
```

> And the key here is that MetaStringBytes can't distinguish whether the data is package or classname, so the decode need use the whole encoding set

`MetaStringBytes#decode` uses the encoding stored in `MetaStringBytes#hashCode & 0x111`, which means that if the encoding is successful, the decoding will be successful. It should have nothing to do with the encoding set.

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
LiangliangSui: I saw that the [ClassInfo constructor](https://github.com/apache/incubator-fury/blob/main/java/fury-core/src/main/java/org/apache/fury/resolver/ClassInfo.java#L84-L85) also uses PACKAGE_ENCODER encoding `Class#getName()`

java/fury-core/src/main/java/org/apache/fury/resolver/ClassInfo.java
LiangliangSui: And `MetaStringEncoder` supports UTF8 by default, even if my `EncodingSet` does not specify UTF8

chaokunyang: The encoding set in `MetaStringBytes#hashCode & 0x111` use whole `org.apache.fury.meta.MetaString.Encoding.values()`. But the `Encoders.encodePackage` only use `UTF_8, ALL_TO_LOWER_SPECIAL, LOWER_UPPER_DIGIT_SPECIAL`. The inconsistency may introduce some error

LiangliangSui: `MetaStringBytes#hashCode & 0b111` only controls the range of the encoding set, not the encoding set to be used.

> The inconsistency may introduce some error

Could you elaborate on this? I'm not sure what will cause the inconsistency.

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
LiangliangSui: Done

java/fury-core/src/main/java/org/apache/fury/resolver/ClassInfo.java
chaokunyang: I misunderstood it. `Encoders.encodePkgName` use `UTF_8, ALL_TO_LOWER_SPECIAL, LOWER_UPPER_DIGIT_SPECIAL` encoding. The `ClassDefEncoder` use index in `UTF_8, ALL_TO_LOWER_SPECIAL, LOWER_UPPER_DIGIT_SPECIAL` to store encoding. But `getOrCreateMetaStringBytes` uses the encoding in returned `MetaString` instead, so it will use `metaString.getEncoding().getValue()` as the index, and there won't be inconsistency when decoding the meta string

java/fury-core/src/main/java/org/apache/fury/meta/Encoders.java
```diff
@@ -38,6 +38,8 @@
 public class Encoders {
   public static final MetaStringEncoder PACKAGE_ENCODER = new MetaStringEncoder('.', '_');
   public static final MetaStringDecoder PACKAGE_DECODER = new MetaStringDecoder('.', '_');
+  public static final MetaStringEncoder GENERIC_ENCODER = PACKAGE_ENCODER;
```
chaokunyang, line 41: How about move GENERIC_ENCODER front and make `PACKAGE_ENCODER = GENERIC_ENCODER`

chaokunyang (approved):
LGTM overall, I left a minor comement

java/fury-core/src/main/java/org/apache/fury/meta/Encoders.java
LiangliangSui: Done

Issue #1542, opened 2024-04-19
[JavaScript] Implement meta string encoding algorithm for JavaScript


chaokunyang:
## Is your feature request related to a problem? Please describe.
We've implemented  meta string encoding algorithm in https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#meta-string for java in #1514   , it's time to implement it in javascript.

## Describe the solution you'd like
Java implementation in #1514 can be taken as a reference. But note that the meta string encoding algorithm is used for encode field name only, so the special charater can't be `.` or `$`, thus the implementation will be simpler

## Additional context
#1413

Forchapeatl:
Hello @chaokunyang  , I am Forcha Pearl from Cameroon . I am very good  bit manipulation, endianess, encoding DataStructures and Algorithm. I also experienced with Java, python and javascript.  I am interested in participting in OSPP 2024 under Apache furry this year. Please  has this issue ben reserved for the OSPP  contribution period.

chaokunyang:
> Hello @chaokunyang , I am Forcha Pearl from Cameroon . I am very good bit manipulation, endianess, encoding DataStructures and Algorithm. I also experienced with Java, python and javascript. I am interested in participting in OSPP 2024 under Apache furry this year. Please has this issue ben reserved for the OSPP contribution period.

Hi @Forchapeatl , this is not an issue reserved for the OSPP. If it's not taken by someone else in this month, we will take it over. The issues for OSPP can be found at #1670

theweipeng:
> Hello @chaokunyang , I am Forcha Pearl from Cameroon . I am very good bit manipulation, endianess, encoding DataStructures and Algorithm. I also experienced with Java, python and javascript. I am interested in participting in OSPP 2024 under Apache furry this year. Please has this issue ben reserved for the OSPP contribution period.

@Forchapeatl 
Hi, I am the mentor for the OSPP Fury Javascript project. Your participation is welcome. However, I noticed that the OSPP application period is about to end on June 4th at 18:00. You should submit your application on OSPP as soon as possible.

Forchapeatl:
Thank you @theweipeng . @chaokunyang , Please  assign this issue to me. I would love to work on this

chaokunyang:
> Thank you @theweipeng . @chaokunyang , Please assign this issue to me. I would love to work on this

Great, assigned to you

barrerajuanjose:
[...]

PR #1675, opened 2024-06-05
feat(javascript): Meta string encoding algorithm for JavaScript

Forchapeatl:
## What does this PR do?
Meta string encoding algorithm for JavaScript

#1542


Forchapeatl:
Hello @theweipeng ,  A metastring encoding and decoding algorithm has been implemented in javascript but I don't know hat else to do. Please have a look at my draft.

theweipeng:
> Hello @theweipeng ,  A metastring encoding and decoding algorithm has been implemented in javascript but I don't know hat else to do. Please have a look at my draft. 

Great work! Since we are using TypeScript in this project, could you please rename the file to have a .ts extension and resolve the TSLint issues?

chaokunyang:
Do we need to merge those three classes into one file named `metastring.ts`?

Forchapeatl:
I went through the [https://www.conventionalcommits.org/en/v1.0.0/](https://www.conventionalcommits.org/en/v1.0.0/) but I still manage to fail the lint test.  
>Do we need to merge those three classes into one file named metastring.ts?

I am still studying the structure of the apache furry code base.  I don't really know much but I think it's best to use the same file structure as the ones in the `java`  code base , so that it will be easier for new developers to navigate through.

Forchapeatl:
Hi, @chaokunyang , please can you  tell me how to pass this test [https://github.com/apache/incubator-fury/actions/runs/9441081022/job/26001294763?pr=1675](https://github.com/apache/incubator-fury/actions/runs/9441081022/job/26001294763?pr=1675)

chaokunyang:
CI failed with:
image

@theweipeng could you take a look at it

Forchapeatl:
Hi  i did the `npm run test`  , but my logs are different @chaokunyang  , please can you tell me the command to typed in to get those logs ? 
image

theweipeng:
> Hi i did the `npm run test` , but my logs are different @chaokunyang , please can you tell me the command to typed in to get those logs ? image

If you use VSCode to edit the TypeScript code, you will see type errors underlined. Alternatively, you can navigate to the `packages/fury`  directory and run `npm run build`; typically, you can see compilation errors there.

I cloned your branch, and after executing npm install and npm run test, I was able to reproduce the errors on my computer.

2024-06-13 23-23-43屏幕截图

chaokunyang:
Code style check faileld:
image

Seems we have not add code lint document for javascript, @theweipeng could you update the lint document?

chaokunyang:
> I went through the https://www.conventionalcommits.org/en/v1.0.0/ but I still manage to fail the lint test.
> 
> > Do we need to merge those three classes into one file named metastring.ts?
> 
> I am still studying the structure of the apache furry code base. I don't really know much but I think it's best to use the same file structure as the ones in the `java` code base , so that it will be easier for new developers to navigate through.

I prefer keeping javascript idioms. In fury python, all those three class will be put int one file named `metastring.py`. We won't get a fully consistent file layout across language.

@theweipeng How javascript organize such classes ?

theweipeng:
> > I went through the https://www.conventionalcommits.org/en/v1.0.0/ but I still manage to fail the lint test.
> 
> > 
> 
> > > Do we need to merge those three classes into one file named metastring.ts?
> 
> > 
> 
> > I am still studying the structure of the apache furry code base. I don't really know much but I think it's best to use the same file structure as the ones in the `java` code base , so that it will be easier for new developers to navigate through.
> 
> 
> 
> I prefer keeping javascript idioms. In fury python, all those three class will be put int one file named `metastring.py`. We won't get a fully consistent file layout across language.
> 
> 
> 
> @theweipeng How javascript organize such classes ?

We can put it in one file in fury typescript

theweipeng:
> Hello @theweipeng , I tried `npm run lint --fix`, but I can't seem to pass [this  test](https://github.com/apache/fury/actions/runs/9526992061/job/26263101471?pr=1675) . Please help me.

Could you try npm run lint-fix ? This command is declare in package.json. 'npm run lint --fix' means pass the param '--fix' to npm, but not to lint

javascript/packages/fury/lib/meta/MetaString.ts
```diff
@@ -0,0 +1,148 @@
...
+      chars.push(bitsPerChar === MetaString.LOWER_SPECIAL
+        ? MetaString.valueToCharLowerSpecial(value)
+        : MetaString.valueToCharLowerUpperDigitSpecial(value));
+    }
+
+    return chars.join("");
+  }
+
+  // Convert a character to its value for LOWER_SPECIAL encoding
+  static charToValueLowerSpecial(char: string): number {
```
theweipeng, line 78: It should be declared as private

javascript/packages/fury/lib/meta/MetaString.ts
```diff
@@ -0,0 +1,148 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+class MetaString {
+  static LOWER_SPECIAL = 5;
+  static LOWER_UPPER_DIGIT_SPECIAL = 6;
+
+  // Encode function based on the specified bits per character
+  static encode(str: string, bitsPerChar: number): Uint8Array {
```
theweipeng, line 25: I noticed that in Java, the Encoding flag is detected automatically:

```Java
  public MetaString encode(String input, Encoding[] encodings) {
    if (input.isEmpty()) {
      return new MetaString(input, Encoding.UTF_8, specialChar1, specialChar2, new byte[0]);
    }
    if (!StringSerializer.isLatin(input.toCharArray())) {
      return new MetaString(
          input,
          Encoding.UTF_8,
          specialChar1,
          specialChar2,
          input.getBytes(StandardCharsets.UTF_8));
    }
    Encoding encoding = computeEncoding(input, encodings);
    return encode(input, encoding);
  }
```
The bitsPerChar value is determined by the Encoding used. Perhaps, for consistency with Java's approach, we should handle this detail within our method. This could prevent confusion for users calling the encode method, who might otherwise be unsure about how to set the bitsPerChar argument.

javascript/packages/fury/lib/meta/MetaString.ts
```diff
@@ -0,0 +1,148 @@
...
+    const totalBits = str.length * bitsPerChar + 1;
```
theweipeng, line 26: In Javascript, The length of emoji is greater than 1.
```Javascript
("😁").length === 2
```
We always need to be careful, it often causes bugs.

javascript/packages/fury/lib/meta/MetaString.ts
```diff
@@ -0,0 +1,148 @@
...
+    const byteLength = Math.ceil(totalBits / 8);
+    const bytes = new Uint8Array(byteLength);
+    let currentBit = 1;
+
+    for (const char of str) {
```
theweipeng, line 31: Although the emoji length greater than 1, but it will only be iterated once.

javascript/packages/fury/lib/meta/MetaString.ts
Forchapeatl: This  is the new  encode function which automatically detects , encoding and handles Emojis .  Please , are there any other test cases  ?
```js
  // Encode function with automatic encoding detection
  static encode(str: string): Uint8Array {
    const bitsPerChar = MetaString.detectEncoding(str);
    const totalBits = str.length * bitsPerChar + 1;
    const byteLength = Math.ceil(totalBits / 8);
    const bytes = new Uint8Array(byteLength);
    let currentBit = 1;

    for (const char of [...str]) { // Using spread operator to handle emojis correctly
      const value = bitsPerChar === MetaString.LOWER_SPECIAL
        ? MetaString.charToValueLowerSpecial(char)
        : MetaString.charToValueLowerUpperDigitSpecial(char);

      for (let i = bitsPerChar - 1; i >= 0; i--) {
        if ((value & (1 << i)) !== 0) {
          const bytePos = Math.floor(currentBit / 8);
          const bitPos = currentBit % 8;
          bytes[bytePos] |= (1 << (7 - bitPos));
        }
        currentBit++;
      }
    }

    const stripLastChar = bytes.length * 8 >= totalBits + bitsPerChar;
    if (stripLastChar) {
      bytes[0] |= 0x80;
    }
    return bytes;
  }
```

Forchapeatl: can i use [TextEncoder
](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) instead ?

theweipeng: > can i use [TextEncoder ](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) instead ?

Yes, the TextEncoder is implemented by both Node.js and browsers.

Forchapeatl:
Here is the test file MetaString.test.ts
```ts
import {describe, expect, test} from '@jest/globals';
import MetaString from './MetaString'; // Adjust the import path as necessary

describe('MetaString', () => {
  test('should encode and decode a simple string', () => {
    const input = "helloWorld_123";
    const encoded = MetaString.encode(input);
    const decoded = MetaString.decode(encoded);
    expect(decoded).toEqual(input);
  });

  test('should encode and decode a string with emojis', () => {
    const input = "helloWorld_123 💯";
    const encoded = MetaString.encode(input);
    const decoded = MetaString.decode(encoded);
    expect(decoded).toEqual(input);
  });

  test('should throw error for encoding string with .', () => {
    const input = "hello.World_123";
    expect(() => {
      MetaString.encode(input);
    }).toThrow("Unsupported character for encoding: .");
  });

  test('should throw error for encoding string with $', () => {
    const input = "hello$World_123";
    expect(() => {
      MetaString.encode(input);
    }).toThrow("Unsupported character for encoding: $");
  });

  test('should throw error for decoding string with .', () => {
    const input = "helloWorld_123";
    const encoded = MetaString.encode(input);
    encoded[0] = 0x2E; // Add a `.` character code point in the encoded data
    expect(() => {
      MetaString.decode(encoded);
    }).toThrow("Unsupported character for decoding: .");
  });

  test('should throw error for decoding string with $', () => {
    const input = "helloWorld_123";
    const encoded = MetaString.encode(input);
    encoded[0] = 0x24; // Add a `$` character code point in the encoded data
    expect(() => {
      MetaString.decode(encoded);
    }).toThrow("Unsupported character for decoding: $");
  });

  test('should encode and decode a string with various unicode characters', () => {
    const input = "helloWorld_123 💯 你好 мир";
    const encoded = MetaString.encode(input);
    const decoded = MetaString.decode(encoded);
    expect(decoded).toEqual(input);
  });
});

```

Please how do I initialize fury ?

theweipeng:
> Here is the test file MetaString.test.ts
> 
> ```ts
> 
> import {describe, expect, test} from '@jest/globals';
> 
> import MetaString from './MetaString'; // Adjust the import path as necessary
> 
> 
> 
> describe('MetaString', () => {
> 
>   test('should encode and decode a simple string', () => {
> 
>     const input = "helloWorld_123";
> 
>     const encoded = MetaString.encode(input);
> 
>     const decoded = MetaString.decode(encoded);
> 
>     expect(decoded).toEqual(input);
> 
>   });
> 
> 
> 
>   test('should encode and decode a string with emojis', () => {
> 
>     const input = "helloWorld_123 💯";
> 
>     const encoded = MetaString.encode(input);
> 
>     const decoded = MetaString.decode(encoded);
> 
>     expect(decoded).toEqual(input);
> 
>   });
> 
> 
> 
>   test('should throw error for encoding string with .', () => {
> 
>     const input = "hello.World_123";
> 
>     expect(() => {
> 
>       MetaString.encode(input);
> 
>     }).toThrow("Unsupported character for encoding: .");
> 
>   });
> 
> 
> 
>   test('should throw error for encoding string with $', () => {
> 
>     const input = "hello$World_123";
> 
>     expect(() => {
> 
>       MetaString.encode(input);
> 
>     }).toThrow("Unsupported character for encoding: $");
> 
>   });
> 
> 
> 
>   test('should throw error for decoding string with .', () => {
> 
>     const input = "helloWorld_123";
> 
>     const encoded = MetaString.encode(input);
> 
>     encoded[0] = 0x2E; // Add a `.` character code point in the encoded data
> 
>     expect(() => {
> 
>       MetaString.decode(encoded);
> 
>     }).toThrow("Unsupported character for decoding: .");
> 
>   });
> 
> 
> 
>   test('should throw error for decoding string with $', () => {
> 
>     const input = "helloWorld_123";
> 
>     const encoded = MetaString.encode(input);
> 
>     encoded[0] = 0x24; // Add a `$` character code point in the encoded data
> 
>     expect(() => {
> 
>       MetaString.decode(encoded);
> 
>     }).toThrow("Unsupported character for decoding: $");
> 
>   });
> 
> 
> 
>   test('should encode and decode a string with various unicode characters', () => {
> 
>     const input = "helloWorld_123 💯 你好 мир";
> 
>     const encoded = MetaString.encode(input);
> 
>     const decoded = MetaString.decode(encoded);
> 
>     expect(decoded).toEqual(input);
> 
>   });
> 
> });
> 
> 
> 
> ```
> 
> 
> 
> Please how do I initialize fury ?

Great work! You can now initialize a Fury instance by invoking 'new Fury()'. For examples of its usage, please search within the project. Note that the Compatible Mode is currently not supported, meaning that the MateString encoding and decoding functions are not being called anywhere in the system.

Forchapeatl:
Please , how may I extend the `BaseSerializerGenerator`

this is my new `MetaString.ts`

```ts
import { TypeDescription } from "../description";
import { CodecBuilder } from "./builder";
import { BaseSerializerGenerator } from "./serializer";
import { CodegenRegistry } from "./router";
import { InternalSerializerType } from "../type";
import { Scope } from "./scope";

class MetaString extends BaseSerializerGenerator {
  description: TypeDescription;

  constructor(description: TypeDescription, builder: CodecBuilder, scope: Scope) {
    super(description, builder, scope);
    this.description = description;
  }

  // Encode function that handles all Unicode characters
  static writeStmt(accessor: string): Uint8Array {
    const encoder = new TextEncoder();
    return encoder.encode(accessor);
  }

  // Decoding function that handles all Unicode characters
  static readStmt(accessor: Uint8Array): string {
    const decoder = new TextDecoder();
    return decoder.decode(accessor);
  }
}

CodegenRegistry.register(InternalSerializerType.METASTRING, MetaString);
```

but  I get this error 
image

theweipeng:
> Please , how may I extend the `BaseSerializerGenerator`
> 
> this is my new `MetaString.ts`
> 
> ```ts
> import { TypeDescription } from "../description";
> import { CodecBuilder } from "./builder";
> import { BaseSerializerGenerator } from "./serializer";
> import { CodegenRegistry } from "./router";
> import { InternalSerializerType } from "../type";
> import { Scope } from "./scope";
> 
> class MetaString extends BaseSerializerGenerator {
>   description: TypeDescription;
> 
>   constructor(description: TypeDescription, builder: CodecBuilder, scope: Scope) {
>     super(description, builder, scope);
>     this.description = description;
>   }
> 
>   // Encode function that handles all Unicode characters
>   static writeStmt(accessor: string): Uint8Array {
>     const encoder = new TextEncoder();
>     return encoder.encode(accessor);
>   }
> 
>   // Decoding function that handles all Unicode characters
>   static readStmt(accessor: Uint8Array): string {
>     const decoder = new TextDecoder();
>     return decoder.decode(accessor);
>   }
> }
> 
> CodegenRegistry.register(InternalSerializerType.METASTRING, MetaString);
> ```
> 
> but I get this error image

`BaseSerializerGenerator` is used to generate code for types that users register using `Type.xxxx()`. MetaString is used to compute the meta information when we generate code for `Type.Object`. Therefore, you can invoke your function in `gen/object.ts`. Take a look at the other function `computeStructHash`, which is used to compute the hash of the type.`MetaString` works similarly to `computeStructHash`. `MetaString` only needs to be called once for each type during code generation, meaning that it is unnecessary to generate JavaScript code every time.

Forchapeatl:
Hello @theweipeng , I have exported the `metastring`  class and called it  objects.ts , please have a look at my PR. Am I in the right direction ?

theweipeng:
> Hello @theweipeng , I have exported the `metastring`  class and called it  objects.ts , please have a look at my PR. Am I in the right direction ?

Yep, It is in the right direction.

Forchapeatl:
@theweipeng  , please help me with a few hints on passing this job [https://github.com/apache/fury/actions/runs/9747569516/job/26900584218?pr=1675](https://github.com/apache/fury/actions/runs/9747569516/job/26900584218?pr=1675)

Forchapeatl:
Hi, @theweipeng , Please have a look at my PR. I have been trying to adjust the object.test.ts files so as to pass the pipline tests  but i really don't know the `writeStmt` and `readStmt` functions  are tested in the objects.test.ts files. Please help me with a few hints.

theweipeng:
> Hi, @theweipeng , Please have a look at my PR. I have been trying to adjust the object.test.ts files so as to pass the pipline tests but i really don't know the `writeStmt` and `readStmt` functions are tested in the objects.test.ts files. Please help me with a few hints.

First, let me explain why we need to generate code, which is necessary for a general understanding.

Without generating code for each object type, the JavaScript would look like this:
```JavaScript
 function writeObject(input, writer, description) {
     description.fields.forEach(field => {
         const binary = metaString.encode(field.name);
         writer.write(binary);
     });
     description.fields.forEach(field => {
         writeXxxx(input[field.name], writer, field.description);
     });
 }
 ```
In this pseudo-code, the function writeObject calls metaString.encode every time and iterating over description.fields is unnecessary. Performance is very important for Fury.

If we generate code for each object type, the JavaScript will look like this:
```JavaScript
 const metaBinary = description.fields.map(field => metaString.encode(field.name)).join(/* code to join the binary data */);
 function writeObject(input, writer) {
     writer.write(metaBinary);
     writeString(input.b);
     writeBoolean(input.c);
 }
 ```
In this pseudo-code, metaBinary is called only once when writeObject is created, and we can avoid all overhead except for the necessary write buffer.

writeStmt is used to generate a code fragment that will be embedded by the function toSerializer, declared in BaseSerializerGenerator.

For example, a description like the one below will generate code when registerSerializer is called:
```JavaScript
 const description = Type.object('example.foo', {
     b: Type.string(),
     c: Type.bool()
 });
 
 const serializer = fury.registerSerializer(description).serializer;
```
The call chain will be generateSerializer(gen/index.ts) -> generate(gen/index.ts) -> generator.toSerializer(gen/index.ts) -> toSerializer(gen/serializer.ts) -> writeStmt(gen/object.ts). `funcString` declared in generate(gen/index.ts) store the final code. 

writeStmt iterates through the properties defined in the description (in the example, they are b and c) and generates the code based on the property's description.

In this example, prop b will call writeStmt declared in gen/string.ts.

You can add a script for debugging like this:

https://github.com/apache/fury/assets/16490211/2c1f7b34-a4bd-4774-bc26-556b80a89c02

And click Debug:

https://github.com/apache/fury/assets/16490211/77d67ae8-042a-4ae1-b878-9de39777fb49

The test will pause at the breakpoint, allowing you to trace the call chain by debugging the code generator.

In your case, you should create a write instance for meta inside a closure, thus allowing direct writing of the meta binary.

For instance:
```JavaScript
const metaBinary = this.scope.declare("metaBinary", `${metaString.encode(JSON.stringify(description))}`); // This line declares a variable inside a closure.
return `
${this.builder.writer.binary(metaBinary)};
`;
```
tagWriter is in a similar situation; this should be helpful to you:

https://github.com/apache/fury/assets/16490211/401e9f36-09c6-4204-9102-b115fac9631a

Forchapeatl:
thank you for your guidance sir , the explanations saved me , please have a look at my PR.

javascript/packages/fury/lib/meta/MetaString.ts
```diff
@@ -0,0 +1,32 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+export class MetaString {
+  // Encode function that handles all Unicode characters
+  static encode(str: string): Uint8Array {
+    const encoder = new TextEncoder();
```
theweipeng, line 23: You can not replace the `encoding algorithm` which was implemented on the previous PRs to TextEncoder. TextEncoder is used for encoding the string to a utf8 buffer. The encoding algorithm in our protocol (https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#encoding-algorithms) is used for compressing the string to a minimum byte size.  Please revert this file to the last pr

Forchapeatl: The TextEncoder/Decoder native functions have been removed.

theweipeng: > The TextEncoder/Decoder native functions have been removed.
The main issue is not whether to use TextEncoder or not, but that you need to implement our protocol. 
Neither `TextEncoder` nor `atob` can meet our requirements because our protocol is private and not a simple conversion from char to numbers. ()
The code you submitted in (https://github.com/apache/fury/pull/1675/commits/e5a9c3acc120e73e403b6860f81c630480141094) is right.

PR #1684, opened 2024-06-12
feat(java): decoder add cache

LiangliangSui:
## What does this PR do?

`org.apache.fury.meta.Encoders` adds cache to decoder and uses `hashCode` as key.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/meta/Encoders.java
+++ b/java/fury-core/src/main/java/org/apache/fury/meta/Encoders.java
@@ -36,14 +36,14 @@ import org.apache.fury.util.StringUtils;
 
 /** A class used to encode package/class/field name. */
 public class Encoders {
+  public static final String ARRAY_PREFIX = "1";
+  public static final String ENUM_PREFIX = "2";
   public static final MetaStringEncoder GENERIC_ENCODER = new MetaStringEncoder('.', '_');
   public static final MetaStringDecoder GENERIC_DECODER = new MetaStringDecoder('.', '_');
   public static final MetaStringEncoder PACKAGE_ENCODER = GENERIC_ENCODER;
   public static final MetaStringDecoder PACKAGE_DECODER = GENERIC_DECODER;
   public static final MetaStringEncoder TYPE_NAME_ENCODER = new MetaStringEncoder('$', '_');
   public static final MetaStringDecoder TYPE_NAME_DECODER = new MetaStringDecoder('$', '_');
-  public static final String ARRAY_PREFIX = "1";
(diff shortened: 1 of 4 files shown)


java/fury-core/src/main/java/org/apache/fury/meta/Encoders.java
```diff
@@ -139,4 +134,32 @@ public static MetaString encodeFieldName(String fieldName) {
...
+  public static MetaString encodePackage(String pkg) {
+    return pgkMetaStringCache.computeIfAbsent(pkg, k -> PACKAGE_ENCODER.encode(pkg, pkgEncodings));
+  }
+
+  public static MetaString encodeTypeName(String typeName) {
+    return typeMetaStringCache.computeIfAbsent(
+        typeName, k -> TYPE_NAME_ENCODER.encode(typeName, typeNameEncodings));
+  }
+
+  public static String decodeFieldName(long hashCode, byte[] bytes, MetaString.Encoding encoding) {
```
chaokunyang, line 147: Hashcode is an internal optimization, maybe we shouldn't make it pubic as an API for meta string encoder

chaokunyang:
Hi @LiangliangSui , meta string encoding is not in critical path, we don't need to cache it.

LiangliangSui:
> Hi @LiangliangSui , meta string encoding is not in critical path, we don't need to cache it. And we've alreadly have cache in ClassResolver

The meta string encoding cache is already implemented in `org.apache.fury.meta.Encoders`. I just moved them a bit. Then I implemented the decode cache based on the meta string encoding cache.

LiangliangSui:
Why is meta string encoding not in the critical path? I see that encoding is cached in Encoders, maybe decoding needs similar optimization?

chaokunyang:
> Why is meta string encoding not in the critical path? I see that encoding is cached in Encoders, maybe decoding needs similar optimization?

It's encoded only once and cached in ClassResolver

chaokunyang:
Is there a way to add cache for decode and didn't introduce API like hashcode stuff?

LiangliangSui:
> Is there a way to add cache for decode and didn't introduce API like hashcode stuff?

I also thought about what you said, we can only establish an association with String through `(byte[] encodedData, Encoding encoding)`, but the best way to integrate `(byte[] encodedData, Encoding encoding)` is to use hashcode, and this can also avoid the secondary calculation of `(byte[] encodedData, Encoding encoding)`

LiangliangSui:
Perhaps we can use `(byte[] encodedData, Encoding encoding)` to construct an incomplete `MetaString`, and then use` MetaString` as the key and `String` as the value.

PR #1686, opened 2024-06-13
chore(java): Disallow writing meta classdef when obj is null

LiangliangSui:
## What does this PR do?

When obj is null, if shareMeta is enabled, no need to write meta classdef

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/Fury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/Fury.java
@@ -156,6 +156,7 @@ public final class Fury implements BaseFury {
     arrayListSerializer = new ArrayListSerializer(this);
     hashMapSerializer = new HashMapSerializer(this);
     originToCopyMap = new IdentityMap<>();
+    classDefEndOffset = -1;
     LOG.info("Created new fury {}", this);
   }
 
@@ -795,7 +796,7 @@ public final class Fury implements BaseFury {
     } catch (Throwable t) {
       throw ExceptionUtils.handleReadFailed(this, t);
     } finally {
-      if (shareMeta) {
+      if (shareMeta && classDefEndOffset != -1) {
         buffer.readerIndex(classDefEndOffset);
       }
       resetRead();
@@ -1122,7 +1123,7 @@ public final class Fury implements BaseFury {
     } catch (Throwable t) {
       throw ExceptionUtils.handleReadFailed(this, t);
     } finally {
-      if (shareMeta) {
+      if (shareMeta && classDefEndOffset != -1) {
         buffer.readerIndex(classDefEndOffset);
       }
       resetRead();
@@ -1232,7 +1233,7 @@ public final class Fury implements BaseFury {
     } catch (Throwable t) {
       throw ExceptionUtils.handleReadFailed(this, t);
     } finally {
-      if (shareMeta) {
+      if (shareMeta && classDefEndOffset != -1) {
         buffer.readerIndex(classDefEndOffset);
       }
       resetRead();
@@ -1444,7 +1445,6 @@ public final class Fury implements BaseFury {
   private void readClassDefs(MemoryBuffer buffer) {
     int relativeClassDefOffset = buffer.readInt32();
     if (relativeClassDefOffset == -1) {
-      classDefEndOffset = buffer.readerIndex();
       return;
     }
     int readerIndex = buffer.readerIndex();
@@ -1484,6 +1484,7 @@ public final class Fury implements BaseFury {
     nativeObjects.clear();
     peerOutOfBandEnabled = false;
     depth = 0;
+    classDefEndOffset = -1;
   }
 
   public void resetCopy() {
--- a/java/fury-core/src/main/java/org/apache/fury/Fury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/Fury.java
@@ -1122,7 +1122,7 @@ public final class Fury implements BaseFury {
     } catch (Throwable t) {
       throw ExceptionUtils.handleReadFailed(this, t);
     } finally {
-      if (shareMeta && classDefEndOffset != -1) {
+      if (classDefEndOffset != -1) {
         buffer.readerIndex(classDefEndOffset);
       }
       resetRead();
--- a/java/fury-core/src/main/java/org/apache/fury/Fury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/Fury.java
@@ -48,6 +48,7 @@ import org.apache.fury.resolver.ClassInfo;
 import org.apache.fury.resolver.ClassInfoHolder;
 import org.apache.fury.resolver.ClassResolver;
 import org.apache.fury.resolver.MapRefResolver;
+import org.apache.fury.resolver.MetaContext;
 import org.apache.fury.resolver.MetaStringResolver;
 import org.apache.fury.resolver.NoRefResolver;
 import org.apache.fury.resolver.RefResolver;
@@ -339,12 +340,14 @@ public final class Fury implements BaseFury {
       buffer.writeInt32(-1); // preserve 4-byte for nativeObjects start offsets.
     }
     // reduce caller stack
-    if (!refResolver.writeRefOrNull(buffer, obj)) {
+    boolean isNull = refResolver.writeRefOrNull(buffer, obj);
+    if (!isNull) {
       ClassInfo classInfo = classResolver.getOrUpdateClassInfo(obj.getClass());
       classResolver.writeClass(buffer, classInfo);
       writeData(buffer, classInfo, obj);
     }
-    if (shareMeta) {
+    MetaContext metaContext = getSerializationContext().getMetaContext();
+    if (shareMeta && !isNull && metaContext != null && !metaContext.writingClassDefs.isEmpty()) {
       buffer.putInt32(startOffset, buffer.writerIndex() - startOffset - 4);
       classResolver.writeClassDefs(buffer);
     }
@@ -1063,9 +1066,12 @@ public final class Fury implements BaseFury {
         if (!refResolver.writeRefOrNull(buffer, obj)) {
           ClassInfo classInfo = classResolver.getOrUpdateClassInfo(obj.getClass());
           writeData(buffer, classInfo, obj);
+          MetaContext metaContext = getSerializationContext().getMetaContext();
+          if (metaContext != null && !metaContext.writingClassDefs.isEmpty()) {
+            buffer.putInt32(startOffset, buffer.writerIndex() - startOffset - 4);
+            classResolver.writeClassDefs(buffer);
+          }
         }
-        buffer.putInt32(startOffset, buffer.writerIndex() - startOffset - 4);
-        classResolver.writeClassDefs(buffer);
       } else {
         if (!refResolver.writeRefOrNull(buffer, obj)) {
           ClassInfo classInfo = classResolver.getOrUpdateClassInfo(obj.getClass());
@@ -1437,6 +1443,9 @@ public final class Fury implements BaseFury {
 
   private void readClassDefs(MemoryBuffer buffer) {
     int relativeClassDefOffset = buffer.readInt32();
+    if (relativeClassDefOffset == -1) {
+      return;
+    }
     int readerIndex = buffer.readerIndex();
     buffer.readerIndex(readerIndex + relativeClassDefOffset);
     classResolver.readClassDefs(buffer);
--- a/java/fury-core/src/main/java/org/apache/fury/Fury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/Fury.java
@@ -341,14 +341,13 @@ public final class Fury implements BaseFury {
       buffer.writeInt32(-1); // preserve 4-byte for nativeObjects start offsets.
     }
     // reduce caller stack
-    boolean isNull = refResolver.writeRefOrNull(buffer, obj);
-    if (!isNull) {
+    if (!refResolver.writeRefOrNull(buffer, obj)) {
       ClassInfo classInfo = classResolver.getOrUpdateClassInfo(obj.getClass());
       classResolver.writeClass(buffer, classInfo);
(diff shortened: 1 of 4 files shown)


java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -310,12 +310,13 @@ private void write(MemoryBuffer buffer, Object obj) {
...
     // reduce caller stack
-    if (!refResolver.writeRefOrNull(buffer, obj)) {
+    boolean isNull = refResolver.writeRefOrNull(buffer, obj);
+    if (!isNull) {
       ClassInfo classInfo = classResolver.getOrUpdateClassInfo(obj.getClass());
       classResolver.writeClass(buffer, classInfo);
       writeData(buffer, classInfo, obj);
     }
-    if (shareMeta) {
+    if (shareMeta && !isNull) {
```
chaokunyang, line 319: The deserialization will still read class meta, but the offset will be -1, then the read will fail. Could you add an UT for such case?

LiangliangSui: Done.

CI failure is not related to the current changes, fix it in https://github.com/apache/fury/pull/1691.

chaokunyang:
The code looks good to me. But do we really to optimize this? When value is null, the ClassDef list is empty, only one byte are written in such cases

LiangliangSui:
> The code looks good to me. But do we really to optimize this? When value is null, the ClassDef list is empty, only one byte are written in such cases

There will be two more write overheads (putInt32, writeVarUint32Small7), which is also a small problem I found while reading the code. If there is no negative impact, I think it can be merged.

LiangliangSui:
Hi @chaokunyang , could you reply to this PR?

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -307,23 +309,26 @@ private void write(MemoryBuffer buffer, Object obj) {
...
       buffer.putInt32(startOffset, buffer.writerIndex());
       classResolver.writeClassDefs(buffer);
     }
   }
 
   private void xserializeInternal(MemoryBuffer buffer, Object obj) {
     int startOffset = buffer.writerIndex();
-    buffer.writeInt32(-1); // preserve 4-byte for nativeObjects start offsets.
+    buffer.writeInt32(
+        Constants.INVALID_OFFSITE); // preserve 4-byte for nativeObjects start offsets.
```
chaokunyang, line 331: This is an internal, could we keep it inside this class

LiangliangSui: Also used in ClassResolver https://github.com/apache/fury/pull/1686/files#diff-849c034dbc915d01d33d09221ec8c78b290ac53a845e3197ab648ec05bea480aR1427-R1430

chaokunyang:
How about checking meta list size and remove this part of code from class resolver into Fury

LiangliangSui:
I don't think it's necessary. Reading `classDefOffset` and reading `numClassDefs` should be what `readClassDefs` should do. We shouldn't distribute what `readClassDefs` should do to other functions.

Why would we do this? Just to move `Constants.INVALID_OFFSITE` to `Fury`?

chaokunyang:
> I don't think it's necessary. Reading `classDefOffset` and reading `numClassDefs` should be what `readClassDefs` should do. We shouldn't distribute what `readClassDefs` should do to other functions.
> 
> Why would we do this? Just to move `Constants.INVALID_OFFSITE` to `Fury`?

When object is not null, but the type is internal types such pritimitive type, string or ArrayList/HashMap, the meta list will be empty too

LiangliangSui:
> but the type is internal types such pritimitive type

Not only internal types, but also the meta list of registered classes(`classInfo.classId != NO_CLASS_ID`) is empty.

> How about checking meta list size and remove this part of code from class resolver into Fury

```java
  private void write(MemoryBuffer buffer, Object obj) {
    int startOffset = buffer.writerIndex();
    boolean shareMeta = config.isMetaShareEnabled();
    if (shareMeta) {
      buffer.writeInt32(INVALID_OFFSITE); // preserve 4-byte for nativeObjects start offsets.
    }
    // reduce caller stack
    boolean isNull = refResolver.writeRefOrNull(buffer, obj);
    if (!isNull) {
      ClassInfo classInfo = classResolver.getOrUpdateClassInfo(obj.getClass());
      classResolver.writeClass(buffer, classInfo);
      writeData(buffer, classInfo, obj);
    }
    // **Just determine the meta list size.**
    if (shareMeta && !isNull && !serializationContext.getMetaContext().writingClassDefs.isEmpty()) {
      buffer.putInt32(startOffset, buffer.writerIndex());
      classResolver.writeClassDefs(buffer);
    }
  }
```
Is it possible to determine the meta list size at the end?

LiangliangSui:
> > but the type is internal types such pritimitive type
> 
> Not only internal types, but also the meta list of registered classes(`classInfo.classId != NO_CLASS_ID`) is empty.
> 
> > How about checking meta list size and remove this part of code from class resolver into Fury
> 
> ```java
>   private void write(MemoryBuffer buffer, Object obj) {
>     int startOffset = buffer.writerIndex();
>     boolean shareMeta = config.isMetaShareEnabled();
>     if (shareMeta) {
>       buffer.writeInt32(INVALID_OFFSITE); // preserve 4-byte for nativeObjects start offsets.
>     }
>     // reduce caller stack
>     boolean isNull = refResolver.writeRefOrNull(buffer, obj);
>     if (!isNull) {
>       ClassInfo classInfo = classResolver.getOrUpdateClassInfo(obj.getClass());
>       classResolver.writeClass(buffer, classInfo);
>       writeData(buffer, classInfo, obj);
>     }
>     // **Just determine the meta list size.**
>     if (shareMeta && !isNull && !serializationContext.getMetaContext().writingClassDefs.isEmpty()) {
>       buffer.putInt32(startOffset, buffer.writerIndex());
>       classResolver.writeClassDefs(buffer);
>     }
>   }
> ```
> 
> Is it possible to determine the meta list size at the end?

@chaokunyang Any other suggestions? If so, please let me know.

chaokunyang:
> the meta list of registered classes(`classInfo.classId != NO_CLASS_ID`) is empty.

For the meta list of registered classes(classInfo.classId != NO_CLASS_ID), we still need to write meta, this needs being done before we set scopde meta share as the default compatible mode

chaokunyang:
We can check meta size before writing meta

LiangliangSui:
> > the meta list of registered classes(`classInfo.classId != NO_CLASS_ID`) is empty.
> 
> For the meta list of registered classes(classInfo.classId != NO_CLASS_ID), we still need to write meta, this needs being done before we set scopde meta share as the default compatible mode

I mean if the current class has been registered, the information of the current class will not be written back to meta.

```java
public void writeClass(MemoryBuffer buffer, ClassInfo classInfo) {
  if (classInfo.classId == NO_CLASS_ID) { // no class id provided.
    // use classname
    if (metaContextShareEnabled) {
      buffer.writeByte(USE_CLASS_VALUE_FLAG);
      // FIXME(chaokunyang) Register class but not register serializer can't be used with
      //  meta share mode, because no class def are sent to peer.
      writeClassWithMetaShare(buffer, classInfo);
    } else {
      // if it's null, it's a bug.
      assert classInfo.packageNameBytes != null;
      metaStringResolver.writeMetaStringBytesWithFlag(buffer, classInfo.packageNameBytes);
      assert classInfo.classNameBytes != null;
      metaStringResolver.writeMetaStringBytes(buffer, classInfo.classNameBytes);
    }
  } else {
    // use classId
    buffer.writeVarUint32(classInfo.classId << 1);
  }
}
```

LiangliangSui:
> We can check meta size before writing meta

Done

java/fury-core/src/main/java/org/apache/fury/type/Constants.java
```diff
@@ -0,0 +1,24 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fury.type;
+
+public class Constants {
+  public static final int INVALID_OFFSITE = -1;
```
chaokunyang, line 23: I perfer not expose this variable. It's an internal implementation, we'd better keep it private

LiangliangSui:
Hi @chaokunyang , could you please take a look?

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -339,12 +341,14 @@ private void write(MemoryBuffer buffer, Object obj) {
...
-    if (!refResolver.writeRefOrNull(buffer, obj)) {
+    boolean isNull = refResolver.writeRefOrNull(buffer, obj);
+    if (!isNull) {
       ClassInfo classInfo = classResolver.getOrUpdateClassInfo(obj.getClass());
       classResolver.writeClass(buffer, classInfo);
       writeData(buffer, classInfo, obj);
     }
-    if (shareMeta) {
+    MetaContext metaContext = getSerializationContext().getMetaContext();
+    if (shareMeta && !isNull && metaContext != null && !metaContext.writingClassDefs.isEmpty()) {
```
chaokunyang, line 351:
```suggestion
    if (shareMeta && !metaContext.writingClassDefs.isEmpty()) {
```

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -339,12 +341,14 @@ private void write(MemoryBuffer buffer, Object obj) {
       buffer.writeInt32(-1); // preserve 4-byte for nativeObjects start offsets.
     }
     // reduce caller stack
-    if (!refResolver.writeRefOrNull(buffer, obj)) {
+    boolean isNull = refResolver.writeRefOrNull(buffer, obj);
+    if (!isNull) {
```
chaokunyang, line 345:
```suggestion
    if (!refResolver.writeRefOrNull(buffer, obj)) {
```

java/fury-core/src/main/java/org/apache/fury/Fury.java
LiangliangSui: When obj is null, can we also not execute `classResolver.writeClassDefs(buffer)`?

chaokunyang: When obj is null, the writingClassDefs is empty

LiangliangSui: got, `metaContext.writingClassDefs` may cause NPE risk, maybe we should check `assert metaContext != null` before using `metaContext.writingClassDefs`

chaokunyang: then we can just write `metaContext != null && !metaContext.writingClassDefs.isEmpty()`

LiangliangSui: updated

LiangliangSui:
The above comments has been completed, please help review again.

LiangliangSui:
@chaokunyang please help review this

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -344,7 +346,8 @@ private void write(MemoryBuffer buffer, Object obj) {
       classResolver.writeClass(buffer, classInfo);
       writeData(buffer, classInfo, obj);
     }
-    if (shareMeta) {
+    MetaContext metaContext = getSerializationContext().getMetaContext();
```
chaokunyang, line 349:
```suggestion
    MetaContext metaContext = serializationContext.getMetaContext();
```

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1226,7 +1232,7 @@ public Object deserializeJavaObjectAndClass(MemoryBuffer buffer) {
     } catch (Throwable t) {
       throw ExceptionUtils.handleReadFailed(this, t);
     } finally {
-      if (shareMeta) {
+      if (shareMeta && classDefEndOffset != -1) {
```
chaokunyang, line 1235:
```suggestion
      if (classDefEndOffset != -1) {
```

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1116,7 +1122,7 @@ public <T> T deserializeJavaObject(MemoryBuffer buffer, Class<T> cls) {
...
```
chaokunyang, line 1125:
```suggestion
      if (classDefEndOffset != -1) {
```

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1063,9 +1066,12 @@ public void serializeJavaObject(MemoryBuffer buffer, Object obj) {
         if (!refResolver.writeRefOrNull(buffer, obj)) {
           ClassInfo classInfo = classResolver.getOrUpdateClassInfo(obj.getClass());
           writeData(buffer, classInfo, obj);
+          MetaContext metaContext = getSerializationContext().getMetaContext();
```
chaokunyang, line 1069:
```suggestion
          MetaContext metaContext = serializationContext.getMetaContext();
```

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -792,7 +795,7 @@ public Object deserialize(MemoryBuffer buffer, Iterable<MemoryBuffer> outOfBandB
     } catch (Throwable t) {
       throw ExceptionUtils.handleReadFailed(this, t);
     } finally {
-      if (shareMeta) {
+      if (shareMeta && classDefEndOffset != -1) {
```
chaokunyang, line 798:
```suggestion
      if (classDefEndOffset != -1) {
```

chaokunyang (approved):
LGTM

PR #1690, opened 2024-06-16
feat(java/python): new xlang type system spec implementation

chaokunyang:
## What does this PR do?

This PR implements a new [type system](https://fury.apache.org/docs/specification/fury_xlang_serialization_spec/#type-systems)  for xlang serialization between java and python.

The changes includes:
- Refine type system spec: added new types:
  - named_enum: an enum whose value will be serialized as the registered name.
  - struct: a morphic(final) type serialized by Fury Struct serializer.
  - polymorphic_struct: a type which is not morphic(not final). i.e. it don't have subclasses. Suppose we're deserializing
    `List<SomeClass>`, we can save dynamic serializer dispatch if `SomeClass` is morphic(final).
  - compatible_struct: a morphic(final) type serialized by Fury compatible Struct serializer.
  - polymorphic_compatible_struct: a non-morphic(non-final) type serialized by Fury compatible Struct serializer.
  - named_struct: a `struct` whose type mapping will be encoded as a name.
  - named_polymorphic_struct: a `polymorphic_struct` whose type mapping will be encoded as a name.
  - named_compatible_struct: a `compatible_struct` whose type mapping will be encoded as a name.
  - named_polymorphic_compatible_struct: a `polymorphic_compatible_struct` whose type mapping will be encoded as a name.
  - ext: a type which will be serialized by a customized serializer.
  - polymorphic_ext: an `ext` type which is not morphic(not final).
  - named_ext: an `ext` type whose type mapping will be encoded as a name.
  - named_polymorphic_ext: an `polymorphic_ext` type whose type mapping will be encoded as a name.
- Added a new XtypeResolver in java to resolve xlang types
- Support register class mapping by id. Before this PR, we only support register class by name, which is more expensive at space/performance cost.
- Support pass type into to resolve type ambiguation such as `ArrayList/Object[]` in java. Users can `serialize(List.of(1, 2, ,3))` and deserialize it into array by `deserialize(bytes, Integer[].class)`
- Refactor pyfury serialization by moving type resolver into python code from cython, this will make debug more easy and reduce code duplciation, it also speed serialization performance.
- golang xtype serialization  test are disabled, it will be reenabled after new type system is implemented in golang


cpp/fury/type/type.h
```diff
@@ -0,0 +1,68 @@
+#include <cstdint> // For fixed-width integer types
```
pandalee99, line 1: This pr is relatively large, can we briefly describe what this pr does?

pandalee99 (commented):
This pr is relatively large, can we briefly describe what this pr does?

chaokunyang:
> This pr is relatively large, can we briefly describe what this pr does?

I haven't finish it, this PR mainly reimplement xlang serialization with new protocol. Detailed description will be added in next days.

chaokunyang:
Hi @theweipeng @pandalee99 @penguin-wwy @PragmaTwice  @jiacai2050 , this PR is ready for review, could you help review it?

python/pyfury/_serialization.pyx
```diff
@@ -1570,10 +1203,20 @@ cdef class DoubleSerializer(CrossLanguageCompatibleSerializer):
 
 
 @cython.final
-cdef class StringSerializer(CrossLanguageCompatibleSerializer):
-    cpdef inline int16_t get_xtype_id(self):
-        return FuryType.STRING.value
+cdef class DynamicFloatSerializer(CrossLanguageCompatibleSerializer):
+    cpdef inline xwrite(self, Buffer buffer, value):
+        # TOTO(chaokunyang) check value range and write type and value
```
penguin-wwy, line 1208:
```suggestion
        # TODO(chaokunyang) check value range and write type and value
```

python/pyfury/_registry.py
```diff
@@ -0,0 +1,627 @@
...
+        cross-language serialization."""
+        if serializer is not None and not isinstance(serializer, Serializer):
+            try:
+                serializer = serializer(self.fury, cls)
+            except BaseException:
+                try:
+                    serializer = serializer(self.fury)
+                except BaseException:
+                    serializer = serializer()
+        n_params = len({typename, type_id, None}) - 1
```
penguin-wwy, line 348:
```suggestion
        n_params = len((typename, type_id, None)) - 1
```
In CPython, both `tuple` and `list` use a freelists to speed up allocation and deallocation, whereas `set` and `dict` do not have this. Therefore, using a tuple or list is more appropriate if you need a temporary structure.

python/pyfury/_serialization.pyx
```diff
@@ -247,608 +238,364 @@ cdef int32_t NOT_NULL_STRING_FLAG = NOT_NULL_VALUE_FLAG & 0b11111111 | \
...
+        cdef ClassInfo class_info
+        if classinfo_ptr != NULL:
+            class_info = <object> classinfo_ptr
+            if class_info.serializer is not None:
+                return class_info
+            else:
+                class_info.serializer = self._resolver._create_serializer(cls)
+                return class_info
+        elif not create:
+            return
```
penguin-wwy, line 499:
```suggestion
            return None
```

python/pyfury/_registry.py
chaokunyang: We need set here, because we are deduplicating the arguments to check `n_params`. And this is not in critical path, the performance is not important

java/fury-core/src/main/java/org/apache/fury/resolver/XtypeResolver.java
```diff
@@ -0,0 +1,485 @@
...
+
+  public ClassInfo writeClassInfo(MemoryBuffer buffer, Object obj) {
+    ClassInfo classInfo = getClassInfo(obj.getClass(), classInfoCache);
+    int xtypeId = classInfo.getXtypeId();
+    byte internalTypeId = (byte) xtypeId;
+    buffer.writeVarUint32Small7(xtypeId);
+    switch (internalTypeId) {
+      case Types.NAMED_ENUM:
+      case Types.NAMED_STRUCT:
+      case Types.NAMED_EXT:
```
Handsome-cong, line 363: Lack of support for:
NAMED_POLYMORPHIC_STRUCT
NAMED_COMPATIBLE_STRUCT
NAMED_POLYMORPHIC_COMPATIBLE_STRUCT
NAMED_POLYMORPHIC_EXT

java/fury-core/src/main/java/org/apache/fury/resolver/XtypeResolver.java
```diff
@@ -0,0 +1,485 @@
...
+  }
+
+  public ClassInfo readClassInfo(MemoryBuffer buffer) {
+    long xtypeId = buffer.readVarUint32Small14();
+    byte internalTypeId = (byte) xtypeId;
+    switch (internalTypeId) {
+      case Types.NAMED_ENUM:
+      case Types.NAMED_STRUCT:
+      case Types.NAMED_COMPATIBLE_STRUCT:
+      case Types.NAMED_EXT:
```
Handsome-cong, line 382: Lack of support for:
NAMED_POLYMORPHIC_STRUCT
NAMED_POLYMORPHIC_COMPATIBLE_STRUCT
NAMED_POLYMORPHIC_EXT

java/fury-core/src/main/java/org/apache/fury/resolver/XtypeResolver.java
chaokunyang: good catch, fixed

python/pyfury/_registry.py
penguin-wwy: lgtm

pandalee99 (approved):
I think this PR is basically okay. 
It supports new types and improves the specifications of the type system. For instance, the `XtypeResolver` class has established a rather intricate yet fully functional mechanism for type resolution, registration, and the processing of related information.This PR has added a lot of functions and optimized the performance.

penguin-wwy (approved):

PR #1691, opened 2024-06-18
ci: fix ci error

LiangliangSui:
## What does this PR do?
The current CI keeps failing.

[Error](https://github.com/apache/fury/actions/runs/9560557264/job/26352935638):
```
  A module that was compiled using NumPy 1.x cannot be run in
  NumPy 2.0.0 as it may crash. To support both 1.x and 2.x
  versions of NumPy, modules must be compiled with NumPy 2.0.
  Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.
```

`numpy` was updated to 2.0.0 on Jun 16, 2024, which caused compatibility issues with CI, so `numpy<2.0.0` was set.

image


chaokunyang (approved):
LGTM

PR #1697, opened 2024-06-21
fix(java): return fury to pooled which get from

MrChang0:
## What does this PR do?

incase of pooledCache expired from cache, and fury will return to another pooledCache.
so we save pooledCache instead of get from cache.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/pool/ClassLoaderFuryPooled.java
+++ b/java/fury-core/src/main/java/org/apache/fury/pool/ClassLoaderFuryPooled.java
@@ -109,6 +109,7 @@ public class ClassLoaderFuryPooled {
       furyCondition.signalAll();
     } catch (Exception e) {
       LOG.error(e.getMessage(), e);
+      throw new RuntimeException(e);
     } finally {
       lock.unlock();
     }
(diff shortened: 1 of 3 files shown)


chaokunyang (approved):
LGTM

juliojgd:
This PR os introducing a breaking change in the API, right? By rethrowing exceptions that were not rethrown before.

chaokunyang:
Hi, @juliojgd , could you elaborate more? I don't see API changes

juliojgd:
Sure. This new throw sentence, not present before:
image

Makes that (1) is a new code point where a RuntimeException can be thrown (before this change it was not) as it's inside a finally and I can't see a try-catch surrounding it.

In addition in (2) there is another new point in code where a RuntimeException can be thrown (before this change the catched exception were swallowed silently) and I can't see a try-catch surrounding it.

image

For me adding a new point throwing a new RuntimeException that before did not throw it makes part of de API contract, and could lead to problems in clients using the library.

I am new to Fury code, maybe I'm wrong.

Regards

chaokunyang:
I see, thanks for you detailed explation. This indeed introduce an API change. But it does need to throw exception here, the previous implementation is not right

juliojgd:
Maybe it should be noted in the release notes, highlighting that here is such API change.

chaokunyang:
> Maybe it should be noted in the release notes, highlighting that here is such API change.

Great suggestion, we should have an API change section in our release notes

PR #1699, opened 2024-06-25
feat: use TextDecoder to decode buffer

bytemain:
## What does this PR do?

Use the browser builtin decode utils: TextDecoder

https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder

This can improve a lot of performance. see the diff of the benchmark.

and the original implementation will cause many `minor gc` in javascript, the longer the string, the longer the GC time.

## Benchmark

CleanShot 2024-06-25 at 17 04 45@2x


chaokunyang (approved):
Awesome!

theweipeng (approved):

Issue #1679, opened 2024-06-11
[Java] fast object copy natively in fury java


chaokunyang:
## Is your feature request related to a problem? Please describe.
Deep copy is common in programing to avoid modify existing data, which may introduce side effect and subtle bugs.

Currently fury doesn't support copy natively, users need to serialized data into binary, then deserialize it into object. This is costly, it introduce extra copy, and for immutable objects such as string, time objects, boxed objects, there is no need to copy.

## Describe the solution you'd like

We should support copy natively in fury:
- Design a copy interface
- Make`Serializer` implement the copy interface, throw UnsupportedException by default.
- For non-jit serializer, override the copy interface to implement copy
- For immutable object such as `String`, `java.time`, just return itself
- For mutable object, create new object and set all attributes
- For pojo/bean/record object, implement the copy in a separate class, and forward the copy to that class to reuse the copy implementation in `ObjectSerializer/CompatibleObjectSerializer`

PR #1701, opened 2024-06-26
feat(java): fast object copy framework in fury java

zhaommmmomo:
## What does this PR do?

Object deep copy framework in fury java. This PR is only for non-jit serializer

- For immutable object such as `String`、`java.time`、`int`、`byte`... return itself.
- For mutable object, create new object and set all attributes.
- For `pojo` / `bean` object, Use `ObjectSerializer.copy(obj)` to create a new object
- For `Arrays`、`Collection`、`Map` object, create new object, traverse the elements, call `fory.copy(obj)` to generate new elements and set them to the new object. Some serializers have special `copy` methods.

## Related issues

[[Java] fast object copy framework in fury java #1679](https://github.com/apache/fury/issues/1679)

1. Provide `fury.copy(obj)` interface, which can deep copy the object without serialize / deserialize. For example:

```java
User user = new User();
user.setName("a");

// If object contain circular references, please enable copy ref tracking by FuryBuilder#withCopyRefTracking(true) 
Fury fury = Fury.builder().withLanguage(Language.JAVA).withCopyRefTracking(true).build();
User newUser = fury.copy(user);
```

2. Provide `FuryCopyable<T>`  interface, which can customize the copy method of object. For example: 

```java
public class User implements Serializable, FuryCopyable<User> {
  
  @Override
  public User copy(Fury fury) {
    // do something
    System.out.println("object custom copy method");
    return newUser;
  }
}
```

windows 10、12 cores、24G

> JDK

java version "1.8.0_202"
Java(TM) SE Runtime Environment (build 1.8.0_202-b08)
Java HotSpot(TM) 64-Bit Server VM (build 25.202-b08, mixed mode)

java version "17.0.8" 2023-07-18 LTS
Java(TM) SE Runtime Environment (build 17.0.8+9-LTS-211)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.8+9-LTS-211, mixed mode, sharing)

> Test data

see [benchmark code](https://github.com/zhaommmmomo/fury/tree/copy_benchmark/java/fury-benchmark/src/main/java/org/apache/fury/benchmark)
```java
int size = 128;
int[] intArr = new int[size];
List<Object> list = new ArrayList<>(size);
Map<Object, Object> map = new ConcurrentHashMap<>();
BeanA beanA = BeanA.createBeanA(size); // org.apache.fury.test.bean.BeanA

for (int i = 0; i < size; i++) {
  intArr[i] = i;
  list.add(i);
  map.put(i, UUID.randomUUID().toString());
}
```

> Test

**JMH**: benchmarkMode({Mode.Throughput})、fork(1)、threads(1)、warmup(iterations = 3, time = 1)、measurement(iterations = 5, time = 5)、outputTimeUnit(TimeUnit.SECONDS)
*JDK8*

*JDK17*

**JMH**: benchmarkMode({Mode.AverageTime})、fork(1)、threads(1)、warmup(iterations = 3, time = 1)、measurement(iterations = 5, time = 5)、outputTimeUnit(TimeUnit.MILLISECONDS)
**Each benchmark method will be looped 10,000 times**
*JDK8*

*JDK17*


zhaommmmomo:
@chaokunyang hi, if there is anything I need to provide or modify, please let me know.

chaokunyang:
This is awesome! Very cool. Thanks for taking time to implement this. I will take a review tomorrow and give you may feedbacks. Thanks for your contribution @zhaommmmomo

java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
```diff
@@ -262,6 +272,13 @@ public void write(MemoryBuffer buffer, boolean[] value) {
       }
     }
 
+    @Override
+    public boolean[] copy(boolean[] originArray) {
+      boolean[] newArray = new boolean[originArray.length];
+      System.arraycopy(originArray, 0, newArray, 0, originArray.length);
```
chaokunyang, line 278: Could we use `Arrays.copyOf` instead for primitive types? Some JDK JIT compiler has special optimization for `Arrays.copyOf`

java/fury-core/src/main/java/org/apache/fury/serializer/Serializer.java
```diff
@@ -38,11 +38,20 @@ public abstract class Serializer<T> {
   protected final Class<T> type;
   protected final boolean isJava;
   protected final boolean needToWriteRef;
+  protected boolean immutable;
```
chaokunyang, line 41: Could we make this field final?

java/fury-core/src/main/java/org/apache/fury/serializer/collection/ImmutableCollectionSerializers.java
```diff
@@ -105,6 +105,7 @@ class MapNStub {}
   public static class ImmutableListSerializer extends CollectionSerializer {
     public ImmutableListSerializer(Fury fury, Class cls) {
       super(fury, cls, true);
+      setImmutable(true);
```
chaokunyang, line 108: If value is not immutable, we should not mark this list as immutable. Maybe we need to check value types in copy implementation

java/fury-core/src/main/java/org/apache/fury/serializer/collection/UnmodifiableSerializers.java
```diff
@@ -85,6 +85,7 @@ public UnmodifiableCollectionSerializer(Fury fury, Class cls, Function factory,
       super(fury, cls, false);
       this.factory = factory;
       this.offset = offset;
+      setImmutable(true);
```
chaokunyang, line 88: ditto as ImmutableList

java/fury-core/src/main/java/org/apache/fury/serializer/collection/UnmodifiableSerializers.java
```diff
@@ -109,6 +110,7 @@ public UnmodifiableMapSerializer(Fury fury, Class cls, Function factory, long of
...
```
chaokunyang, line 113: ditto as ImmutableList

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
```diff
@@ -228,6 +230,40 @@ public void write(MemoryBuffer buffer, T value) {
...
   }
 
+  @Override
+  public T copy(T originObj) {
+    T newObj = newBean(constructor, type);
+    List<Field> fieldsList = getAllFieldCache();
+    for (Field field : fieldsList) {
+      try {
+        Object value = field.get(originObj);
+        field.set(newObj, fury.copy(value));
```
chaokunyang, line 240: This use reflection, which is pretty slow.
Could we use Plafform util here? Just like:

```java
for(FieldInfo info: fieldClassInfoList) {
long offset = info.offset;
swithc(info.classId) {
  ....
 case PRIMITIVE_INT: Platform.putInt(newObj,offset, Platform.getInt(originObj, info.offset); break;
 case PRIMITIVE_DOULE: Platform.putDouble(newObj, offset, Platform.getDouble(originObj, info.offset); break;
 case STRING, other immutable types ids:  Platform.putObject(newObj, offset, Platform.getObject(originObj, info.offset); break;
default: Platform.putObject(newObj, offset, fury.copy(Platform.getObject(originObj, info.offset))); break;
}
}
```

chaokunyang:
Another thing I found is that we didn't handle circular reference here. If we have  circular reference, current implementation will raise recurision error

java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
zhaommmmomo: No problem, I will use `Arrays.copy` to implement the copy method of primitive types.

Also, can you provide articles or links about special optimizations? I didn't find any relevant descriptions. thanks

java/fury-core/src/main/java/org/apache/fury/serializer/Serializer.java
zhaommmmomo: okay

java/fury-core/src/main/java/org/apache/fury/serializer/collection/ImmutableCollectionSerializers.java
zhaommmmomo: Yes, you are right, value maybe mutable. We should iterate over the elements in `collection / map`, call `fury.copy()` method and then add them to the new object.

java/fury-core/src/main/java/org/apache/fury/serializer/collection/UnmodifiableSerializers.java
zhaommmmomo: okay

java/fury-core/src/main/java/org/apache/fury/serializer/collection/UnmodifiableSerializers.java
zhaommmmomo: okay

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
zhaommmmomo: No problem, I will try to use `Plafform util` to implement it and compare the execution efficiency with some common implementations

zhaommmmomo:
Hi @chaokunyang, I will complete the following in the next few days:
1. For `primitive types` array, use `Arrays.copy()` to implement
2. Mark the immutable field as final
3. Unmark the immutable mark of Immutable `Collection` and `Map`, and traverse and call the `fury.copy()` method for the elements
4. Optimize the performance of `ObjectSerializer.copy()`
5. Solve the circular reference problem

Others:
1. Output a complete FuryCopyTest use case (including: `primitive type`, `collection`, `map`, `some common objects`, `complex objects`, `circular reference`, `ThreadPoolFury.copy()`, `ThreadPoolFury.copy()`...)

Thank you for your code review and suggestions

chaokunyang:
For collection and map, you can use the genetics for elements to reduce call to fury.copy. This is important for performance. Ditto for array.

 But It's ok to left in another pr to reduce the complexity in this pr

zhaommmmomo:
> 对于集合和映射，您可以使用元素的遗传来减少对 fury.copy 的调用。这对于性能很重要。数组也是如此。
> 
> 但是可以将其留在另一个 pr 中以降低此 pr 的复杂性

zhaommmmomo:
> For collection and map, you can use the genetics for elements to reduce call to fury.copy. This is important for performance. Ditto for array.
> 
> But It's ok to left in another pr to reduce the complexity in this pr

OK, I will try to use these to optimize copy next

zhaommmmomo:
> Hi @chaokunyang, I will complete the following in the next few days:
> 
> 1. For `primitive types` array, use `Arrays.copy()` to implement
> 2. Mark the immutable field as final
> 3. Unmark the immutable mark of Immutable `Collection` and `Map`, traverse and call the `fury.copy()` method for the elements
> 4. Optimize the performance of `ObjectSerializer.copy()`
> 5. Solve the circular reference problem
> 
> Others:
> 
> 1. Output a complete FuryCopyTest use case (including: `primitive type`, `collection`, `map`, `some common objects`, `complex objects`, `circular reference`, `ThreadPoolFury.copy()`, `ThreadPoolFury.copy()`...)
> 
> Thank you for your code review and suggestions

hi @chaokunyang, i have already dealt with the issues raised by code review. If you have any other questions, please let me know.  thanks

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1221,15 +1240,55 @@ public Object deserializeJavaObjectAndClass(FuryReadableChannel channel) {
...
+      }
+      Object newObj = originToCopyMap.get(obj);
+      if (Objects.nonNull(newObj)) {
+        return (T) newObj;
+      }
+      if (copyRefTracking) {
+        originObjRef = obj;
+      }
+      T copy;
+      if (obj instanceof FuryCopyable) {
```
chaokunyang, line 1262: this check is expensive, we may better put it into a Serializer#copy

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1221,15 +1240,55 @@ public Object deserializeJavaObjectAndClass(FuryReadableChannel channel) {
...
+  /**
+   * Record the mapping between the object currently being copied and the newly generated object
+   * after copying.
+   *
+   * @param obj the new object instance
+   */
+  @SuppressWarnings("unchecked")
+  public void copyReference(Object obj) {
+    if (copyDepth > 0 && Objects.nonNull(originObjRef) && Objects.nonNull(obj)) {
+      originToCopyMap.put(originObjRef, obj);
```
chaokunyang, line 1290: Could we add a new method in `RefResolver` and reuse the IndentifyMap in this ref resolver?

java/fury-core/src/main/java/org/apache/fury/config/Config.java
```diff
@@ -38,6 +38,7 @@ public class Config implements Serializable {
   private final boolean basicTypesRefIgnored;
   private final boolean stringRefIgnored;
   private final boolean timeRefIgnored;
+  private final boolean copyTrackingRef;
```
chaokunyang, line 41: If we reuse ref resolver, this option can be omitted

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1221,15 +1240,55 @@ public Object deserializeJavaObjectAndClass(FuryReadableChannel channel) {
...
-    if (obj instanceof FuryCopyable) {
-      return (T) ((FuryCopyable) obj).copy(this);
-    } else {
-      return (T) classResolver.getOrUpdateClassInfo(obj.getClass()).getSerializer().copy(obj);
+    try {
+      copyDepth++;
+      if (Objects.isNull(originToCopyMap)) {
+        originToCopyMap = new IdentityMap();
+      }
+      Object newObj = originToCopyMap.get(obj);
```
chaokunyang, line 1254: For objects like String, Integer, there is no need to track ref, the map put/get cost is expensive for small objects copy

chaokunyang: You can take Serializer#needToTrackRef as an example

java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
chaokunyang: I forgot the original articles. You can run a jmh benchmark, and `Arrays.copy` is 2X faster than `System.arraycopy`.  
This is what we do in MemoryBuffer:
```java
  public byte[] getBytes(int index, int length) {
    if (index == 0 && heapMemory != null && heapOffset == 0) {
      // Arrays.copyOf is an intrinsics, which is faster
      return Arrays.copyOf(heapMemory, length);
    }
    if (index + length > size) {
      throwIndexOOBExceptionForRead(length);
    }
    byte[] data = new byte[length];
    copyToUnsafe(index, data, Platform.BYTE_ARRAY_OFFSET, length);
    return data;
  }

```

java/fury-core/src/main/java/org/apache/fury/Fury.java
zhaommmmomo: Okay, I'll refer to the circular dependency processing of serialize/deserialize and try to implement it based on MapRefResolver

java/fury-core/src/main/java/org/apache/fury/config/Config.java
zhaommmmomo: I am considering whether it is necessary to handle the circular references of `copy` and `serialize/deserialize` separately

zhaommmmomo:
hi @chaokunyang, the latest code has been adjusted
1. Add Serializer#needToCopyRef to indicate whether serializer needs to track ref
2. For immutable object, no need to track ref, just return. such as `string`, `time`, `primitive type`...
3. For mutable object, use Serializer#needToCopyRef to determine whether track ref is needed, and use IndentifyMap to record the mapping between source instance and new instance

zhaommmmomo:
> 你好@chaokunyang，最新代码已调整
> 
> 1. 添加 Serializer#needToCopyRef 来指示序列化器是否需要跟踪 ref
> 2. 对于不可变对象，不需要跟踪引用，只需返回。例如`string`，，...`time``primitive type`
> 3. 对于可变对象，使用 Serializer#needToCopyRef 判断是否需要 track ref，使用 IndentifyMap 记录源实例与新实例的映射

@chaokunyang hi, if there is anything I need to provide or modify, please let me know.

chaokunyang:
image
Hi @zhaommmmomo , could you fix compilation error?

zhaommmmomo:
> image Hi @zhaommmmomo , could you fix compilation error?

sorry, I didn't notice that the lower version of jdk does not support the corresponding method. solved

java/fury-core/src/main/java/org/apache/fury/Fury.java
chaokunyang: We'd better to create a new Serializer like FuryCopyableSerializer for this instead of using isntanceof here. It's very slow.

The code could looks like in getSerializer:
```java
serializer = XXX;
if (FuryCopyable.class.isAssignableFrom(cls)) {
  serializer = new FuryCopyableSerializer(serializer);
}

class FuryCopyableSerializer {
 private Serializer serializer;

write(...) {
serializer.write(...);
}

read(...) {
serializer.read(...);
}

T copy(T x) {
        return (T) ((FuryCopyable<?>) obj).copy(x);
}

}
```

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1220,6 +1240,47 @@ public Object deserializeJavaObjectAndClass(FuryReadableChannel channel) {
...
+
+  /**
+   * Record the mapping between the object currently being copied and the newly generated object
+   * after copying.
+   *
+   * @param originObj the origin object instance
+   * @param newObj the new object instance
+   */
+  public void copyReference(Object originObj, Object newObj) {
+    if (Objects.nonNull(originObj) && Objects.nonNull(newObj)) {
```
chaokunyang, line 1272: Why not just write `originObj != null && newObj != null`? And for ref tracking, I think checking originObj is enough

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1220,6 +1240,47 @@ public Object deserializeJavaObjectAndClass(FuryReadableChannel channel) {
...
+      originToCopyMap.put(originObj, newObj);
+    }
+  }
+
+  public Object getCopyObject(Object originObj) {
+    if (Objects.isNull(originObj)) {
```
chaokunyang, line 1278: We'd better check null directly, we need to reduce call stack for inline

java/fury-core/src/main/java/org/apache/fury/config/Config.java
```diff
@@ -99,6 +101,10 @@ public boolean trackingRef() {
     return trackingRef;
   }
 
+  public boolean copyTrackingRef() {
```
chaokunyang, line 104: How about naming it to `copyRef`

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
```diff
@@ -228,6 +229,65 @@ public void write(MemoryBuffer buffer, T value) {
...
+    if (!needToCopyRef) {
+      newObj = newBean(constructor, type);
+      copyFields(originObj, newObj);
+      return newObj;
+    }
+    newObj = (T) fury.getCopyObject(originObj);
+    if (Objects.nonNull(newObj)) {
+      return newObj;
+    }
+    fury.incCopyDepth(1);
```
chaokunyang, line 247: why increase depth here

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
```diff
@@ -228,6 +229,65 @@ public void write(MemoryBuffer buffer, T value) {
     writeContainerFields(buffer, value, fury, refResolver, classResolver);
   }
 
+  @Override
+  public T copy(T originObj) {
+    if (isRecord) {
+      return originObj;
```
chaokunyang, line 235: fields of Record may be mutable, we should not skip its update unless we can ensure all fields are immutable types. And if all fields value are immutable, and fields are final, we can take it as immutable even if it's not record

chaokunyang, line 235: fields of Record may be mutable, we should not skip its update unless we can ensure all fields are immutable types. And if all fields value are immutable, and fields are final, we can take it as immutable even if it's not record

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
```diff
@@ -228,6 +229,65 @@ public void write(MemoryBuffer buffer, T value) {
...
+    }
+    T newObj;
+    if (!needToCopyRef) {
+      newObj = newBean(constructor, type);
```
chaokunyang, line 239: this line can be put front

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializer.java
```diff
@@ -45,6 +51,35 @@ public T onCollectionRead(Collection collection) {
...
+      copyElements(originCollection, newCollection);
+      return (T) newCollection;
+    }
+    newCollection = (Collection) fury.getCopyObject(originCollection);
+    if (Objects.nonNull(newCollection)) {
+      return (T) newCollection;
+    }
+    fury.incCopyDepth(1);
+    newCollection = newCollection(originCollection);
+    fury.copyReference(originCollection, newCollection);
```
chaokunyang, line 71: `copyReference` seems just track ref for copy, how about name it as `reference(T o1, T o2)`, it's an overload method, if we give it a detailed javadoc, it won't introduce confusion

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
```diff
@@ -101,6 +102,33 @@ public ArraysAsListSerializer(Fury fury, Class<List<?>> cls) {
...
+      if (!needToCopyRef) {
+        copyElements(originCollection, elements);
+        return Arrays.asList(elements);
+      }
+      List<?> newCollection;
+      newCollection = (List<?>) fury.getCopyObject(originCollection);
+      if (Objects.nonNull(newCollection)) {
+        return newCollection;
+      }
+      fury.incCopyDepth(1);
```
chaokunyang, line 117: We can udpate depth in Fury#copy

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
```diff
@@ -221,6 +249,22 @@ public T newCollection(MemoryBuffer buffer) {
       fury.getRefResolver().reference(collection);
       return collection;
     }
+
+    @Override
+    public Collection newCollection(Collection originCollection) {
+      Collection collection;
+      Comparator comparator = ((SortedSet) originCollection).comparator();
+      if (Objects.equals(type, TreeSet.class)) {
```
chaokunyang, line 257: I noticed we used `Objects` in many places, could we remove such usage, we need to reduce callstack and the size of method bytecode

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
```diff
@@ -221,6 +249,22 @@ public T newCollection(MemoryBuffer buffer) {
...
+        collection = new TreeSet(comparator);
```
chaokunyang, line 258: comparator needs copy to sometimes, we should copy them too

java/fury-core/src/main/java/org/apache/fury/serializer/collection/CollectionSerializers.java
chaokunyang, line 117: We can udpate depth in Fury#copy

java/fury-core/src/main/java/org/apache/fury/Fury.java
zhaommmmomo: This looks like changing `instanceof` to `Class#isAssignableFrom`, maybe we can do this:
```java
Class<?> cls = obj.getClass();
if (FuryCopyable.class.isAssignableFrom(cls)) {
    return (T) ((FuryCopyable<?>) obj).copy(this);
} else {
    return (T) classResolver.getOrUpdateClassInfo(cls).getSerializer().copy(obj);
}
```

The benefits of doing this are:
1. No need to instantiate an additional `FuryCopyableSerializer` object
2. For `write/read`, do not call `FuryCopyableSerializer`, but use the original `Serializer`

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
```diff
@@ -228,6 +229,65 @@ public void write(MemoryBuffer buffer, T value) {
     writeContainerFields(buffer, value, fury, refResolver, classResolver);
   }
 
+  @Override
+  public T copy(T originObj) {
+    if (isRecord) {
```
chaokunyang, line 234: We need to implement copy for org.apache.fury.builder.Generated.GeneratedSerializer, CompatibleSerializer and MetaSharedSerializer too. Perhaps we need a new base class for those serializer

zhaommmmomo:
I have summarized the above areas that need to be adjusted. Thank you for your code review and suggestions. @chaokunyang 

Adjustment Points:
1. Create a `FuryCopyableSerializer` to handle the custom `copy` method, and convert to this type in `createSerializer` to avoid using `instanceof` checks every time a copy is made.
2. Adjust `Fury#copyReference(Object o1, Object o2)` to `Fury#reference(T o1, T o2)`, which is only used to track the references of o1 and o2.
3. Create a base class `BaseObjectSerializer` for all `pojo / record / bean`, which implements the common `copy()` method. `ObjectSerializer`, `GeneratedSerializer`, `CompatibleSerializer`, and `MetaSharedSerializer` will inherit from this class.
4. `record` is no longer the default immutable object.
5. The `comparator` also needs to `Fury#copy()`.
6. Code style:
   - Reduce calls like `Objects.isNull(obj)`, and directly check if `obj == null`.
   - Place `inc copy depth` inside `Fury#copy(T obj)`.

java/fury-core/src/main/java/org/apache/fury/serializer/collection/AbstractMapSerializer.java
```diff
@@ -416,6 +416,10 @@ public T xread(MemoryBuffer buffer) {
     return onMapRead(map);
   }
 
+  protected void copyEntry(Map originMap, Map newMap) {
+    originMap.forEach((k, v) -> newMap.put(fury.copy(k), fury.copy(v)));
```
chaokunyang, line 420: Could we avoid lambda in data plane, it's a little slow

java/fury-core/src/main/java/org/apache/fury/serializer/collection/SynchronizedSerializers.java
```diff
@@ -105,11 +106,23 @@ public Collection read(MemoryBuffer buffer) {
...
 
     public SynchronizedMapSerializer(Fury fury, Class cls, Function factory, long offset) {
       super(fury, cls, false);
       this.factory = factory;
       this.offset = offset;
+      this.isSortedMap = SortedMap.class.isAssignableFrom(cls);
+    }
+
+    @Override
+    public Map newMap(Map map) {
```
chaokunyang, line 119: Why not just:
```java
public Map copy(Map object) {
      Object unwrapped = Platform.getObject(object, offset);
      synchronized (object) {
        return factory.apply(fury.copy(unwrapped));
      }
}
```

java/fury-core/src/main/java/org/apache/fury/serializer/collection/SynchronizedSerializers.java
```diff
@@ -105,11 +106,23 @@ public Collection read(MemoryBuffer buffer) {
...
+      if (isSortedMap) {
+        Comparator comparator = fury.copy(((SortedMap) map).comparator());
+        return new TreeMap(comparator);
```
chaokunyang, line 122: The internal value may not be TreeMap type

java/fury-core/src/main/java/org/apache/fury/serializer/collection/UnmodifiableSerializers.java
```diff
@@ -104,11 +105,23 @@ public Collection read(MemoryBuffer buffer) {
...
 
     public UnmodifiableMapSerializer(Fury fury, Class cls, Function factory, long offset) {
       super(fury, cls, false);
       this.factory = factory;
       this.offset = offset;
+      this.isSortedMap = SortedMap.class.isAssignableFrom(cls);
+    }
+
+    @Override
+    public Map newMap(Map map) {
```
chaokunyang, line 118: ditto

chaokunyang: We may need a test to cover those types

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1220,6 +1239,43 @@ public Object deserializeJavaObjectAndClass(FuryReadableChannel channel) {
...
   }
 
+  @Override
+  public <T> T copy(T obj) {
+    if (obj == null) {
+      return null;
+    }
+    copyDepth++;
+    try {
+      return (T) classResolver.getOrUpdateClassInfo(obj.getClass()).getSerializer().copy(obj);
```
chaokunyang, line 1249: Would be better to use a switch here to create a fastpath for common types, which can avoid the virtual method call cost

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1220,6 +1239,43 @@ public Object deserializeJavaObjectAndClass(FuryReadableChannel channel) {
...
+    } catch (StackOverflowError e) {
```
chaokunyang, line 1250: The try catch is expensive, the nested copy should not go to here. Maybe we should add a new copy method like `copyObject` which is used in nested copy, and replace all internal invoke of this `copy(T obj)` with `copyObject(T obj)`

chaokunyang:
Could we benchmark this copy framework with `serialize/deserialize` in fury benchmark? So we can know where we should go in next steps

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -0,0 +1,146 @@
...
+      return originObj;
+    }
+    if (isRecord) {
+      Object[] fieldValues = copyFields(originObj);
+      try {
+        return (T) constructor.invokeWithArguments(fieldValues);
+      } catch (Throwable e) {
+        Platform.throwException(e);
+      }
+      return originObj;
```
chaokunyang, line 72: We can throw IllgalStateException("unreachable")  after `Platform.throwException(e);` instead

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -0,0 +1,146 @@
...
+        return copyObject;
+      }
+      fury.reference(originObj, newObj);
+    }
+    copyFields(originObj, newObj);
+    return newObj;
+  }
+
+  private Object[] copyFields(T originObj) {
+    return classResolver.getFieldResolver(type).getAllFieldsList().stream()
```
chaokunyang, line 87: this is expensive, we can `buildFieldInfos(fury, descriptorGrouper)` to cache a `InternalFieldInfo[]` instead.

Like:

```
if (fieldsInfo == null) {
  fieldsInfo = xxx;
}

Object[] values = new Object[fieldsInfo.length];
for (FieldInfo info : fieldsInfo) {
 switch(info.classId) {

}
}
```

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -0,0 +1,146 @@
...
+              if (classResolver.isPrimitive(fieldInfo.getEmbeddedClassId())) {
+                return fieldAccessor.get(originObj);
+              }
+              return fury.copyObject(fieldAccessor.get(originObj));
+            })
+        .toArray();
+  }
+
+  private void copyFields(T originObj, T newObj) {
+    List<FieldInfo> fieldsList = classResolver.getFieldResolver(type).getAllFieldsList();
```
chaokunyang, line 100: use cached fieldsInfo instead

chaokunyang:
ScalaMapSerializer needs to implements copy too

chaokunyang (approved):
LGTM

PR #1702, opened 2024-06-26
feat(Python): meta string encoding algorithm in Python

### LINKED ISSUE #1541: [Python] Implement meta string encoding algorithm for python
We've implemented  meta string encoding algorithm in https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#meta-string for java in #1514   , it's time to implement it in python.

Java implementation in #1514 can be taken as a reference. But note that the meta string encoding algorithm is used for encode field name only, so the special charater can't be `.` or `$`, thus the implementation will be simpler

## Additional context
#1413

pandalee99:
## What does this PR do?
The Python algorithm for meta string is implemented and has passed all 13 tests
image

I am not familiar with Python, there may be a few problems, but the overall code is relatively complete

How to have a better implementation, I would also like to improve it, I like this projectBut I think this test environment is relatively bad, and I feel that it can be further built

## Related issues
Closes #1541 


python/pyfury/lib/meta/meta_string.py
```diff
@@ -0,0 +1,365 @@
...
+        if not input_string:
+            return MetaString(
+                input_string,
+                Encoding.UTF_8,
+                self.special_char1,
+                self.special_char2,
+                np.array([], np.uint8),
+                0,
+            )
+
```
LiangliangSui, line 75: Maybe we should check the length of `input_string` and limit it to `Short.MAX_VALUE`.

chaokunyang:
Could we move meta into an upper level directory under pyfury? And we need to add a `__init__.py` file under meta to mark it as a package. Also, I prefer name `meta_string.py` as `metastring.py`, it's like `datetime.py`. Many python module omit underscore like this

python/pyfury/lib/meta/meta_string.py
```diff
@@ -0,0 +1,365 @@
...
+            char_value = (
+                (data[byte_index] << 8)
+                | (data[byte_index + 1] if byte_index + 1 < len(data) else 0)
+            ) >> (11 - intra_byte_index) & bit_mask
+            bit_index += 5
+            decoded.append(self.decode_lower_special_char(char_value))
+
+        return "".join(decoded)
+
+    def decode_lower_upper_digit_special(self, data):
```
chaokunyang, line 298:
```suggestion
    def _decode_lower_upper_digit_special(self, data):
```
private method should starts with `_`

python/pyfury/meta/preconditions.py
LiangliangSui, line 1: We can create a util folder under pyfury and move preconditions.py to the util folder, because it is not only meta, it should be used later.

chaokunyang:
Could ee merge encoder and decoder into metastring.py? Python module is like Java package

python/pyfury/meta/encoding.py
```diff
@@ -0,0 +1,31 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from enum import Enum
+
+
+# Defines the types of supported encodings for MetaStrings.
+class Encoding(Enum):
```
chaokunyang, line 22: Could we merge this class into metastring.py?

python/pyfury/tests/test_meta_string.py
```diff
@@ -0,0 +1,186 @@
...
+# under the License.
+
+import unittest
+import numpy as np
+
+from pyfury.meta.encoding import Encoding
+from pyfury.meta.metastring import MetaStringEncoder, MetaStringDecoder
+
+
+class TestMetaString(unittest.TestCase):
```
chaokunyang, line 25: fury use pytest instead. Could you change to pytest

python/pyfury/utils/preconditions.py
```diff
@@ -0,0 +1,56 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
+# Check utils
+class Preconditions:
```
chaokunyang, line 20: We can just use python assert instead

python/pyfury/tests/test_meta_string.py
```diff
@@ -0,0 +1,199 @@
...
+        elif n < 62:
+            chars.append(chr(ord("0") + (n - 52)))
+        elif n == 62:
+            chars.append(special_char1)
+        else:
+            chars.append(special_char2)
+    return "".join(chars)
+
+
+def test_meta_string():
```
chaokunyang, line 78: Could we rename this as `test_metastring`? And rename `test_meta_string.py` as `test_metastring.py`

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,386 @@
...
+        self.special_char1 = special_char1
+        self.special_char2 = special_char2
+
+    def decode(self, encoded_data, encoding):
+        if len(encoded_data) == 0:
+            return ""
+        return self.decode_with_encoding(encoded_data, encoding)
+
+    def decode_with_encoding(self, encoded_data, encoding):
+
```
chaokunyang, line 65: delete this line

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+class Encoding(Enum):
+    UTF_8 = 0x00
+    LOWER_SPECIAL = 0x01
+    LOWER_UPPER_DIGIT_SPECIAL = 0x02
+    FIRST_TO_LOWER_SPECIAL = 0x03
+    ALL_TO_LOWER_SPECIAL = 0x04
+
+
+# SHORT_MAX_VALUE is used to check whether the length of the value is valid.
+Short_MAX_VALUE = 32767
```
chaokunyang, line 34: How about name it as _METASTRING_NUM_CHARS_LIMIT? Short is not an native type in Python

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+            return MetaString(
+                input_string, encoding, encoded_data, (upper_count + length) * 5
+            )
+        else:
+            encoded_data = np.frombuffer(input_string.encode("utf-8"), dtype=np.uint8)
+            return MetaString(
+                input_string, Encoding.UTF_8, encoded_data, len(encoded_data) * 8
+            )
+
+    def compute_encoding(self, input_string):
```
chaokunyang, line 211: Could you add type hints for all related methods?

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+        self.encoding = encoding
+        self.encoded_data = encoded_data
+        self.length = length
+        if self.encoding != Encoding.UTF_8:
+            self.strip_last_char = (encoded_data[0] & 0x80) != 0
+        else:
+            self.strip_last_char = False
+
+
+# Decodes MetaString objects back into their original plain text form.
```
chaokunyang, line 49: Could we make this comment as the docstring for this method?

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+            # Extract the 6-bit character value across byte boundaries if needed
+            char_value = (
+                (data[byte_index] << 8)
+                | (data[byte_index + 1] if byte_index + 1 < len(data) else 0)
+            ) >> (10 - intra_byte_index) & bit_mask
+            bit_index += 6
+            decoded.append(self._decode_lower_upper_digit_special_char(char_value))
+        return "".join(decoded)
+
+    # Decoding special char for LOWER_SPECIAL based on encoding mapping.
```
chaokunyang, line 111: Ditto

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+        elif char_value == 27:
+            return "_"
+        elif char_value == 28:
+            return "$"
+        elif char_value == 29:
+            return "|"
+        else:
+            raise ValueError(f"Invalid character value for LOWER_SPECIAL: {char_value}")
+
+    # Decoding special char for LOWER_UPPER_DIGIT_SPECIAL based on encoding mapping.
```
chaokunyang, line 126: Ditto

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+# Encodes plain text strings into MetaString objects with specified encoding mechanisms.
+class MetaStringEncoder:
+    def encode(self, input_string):
+
+        # Long meta string than 32767 is not allowed.
+        assert (
+            len(input_string) < _METASTRING_NUM_CHARS_LIMIT
+        ), "Long meta string than 32767 is not allowed."
```
chaokunyang, line 170: Use variable instead of magic number

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+
+        if not input_string:
+            return MetaString(input_string, Encoding.UTF_8, np.array([], np.uint8), 0)
```
chaokunyang, line 173: I think `bytes` type are enough here

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+
+        encoding = self.compute_encoding(input_string)
+        return self.encode_with_encoding(input_string, encoding)
+
+    def encode_with_encoding(self, input_string, encoding):
+
+        # Long meta string than 32767 is not allowed.
+        assert (
+            len(input_string) < _METASTRING_NUM_CHARS_LIMIT
+        ), "Long meta string than 32767 is not allowed."
```
chaokunyang, line 183: Do not use magic number. You can use format string

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+            return MetaString(
+                input_string, Encoding.UTF_8, encoded_data, len(encoded_data) * 8
+            )
+
+    def compute_encoding(self, input_string):
+        if not input_string:
+            return Encoding.LOWER_SPECIAL
+
+        chars = list(input_string)
+        statistics = self._compute_statistics(chars)
```
chaokunyang, line 216: Could you return a named tuple instead of a dict. And we'd better use encoding enum instead of string

python/pyfury/meta/metastring.py
```diff
@@ -0,0 +1,324 @@
...
+        return self._encode_generic(input_string, 6)
+
+    def _encode_first_to_lower_special(self, input_string):
+        chars = list(input_string)
+        chars[0] = chars[0].lower()
+        return self._encode_generic(chars, 5)
+
+    def _encode_all_to_lower_special(self, chars, upper_count):
+        new_chars = []
+        for c in chars:
```
chaokunyang, line 268: We only need do this if upper count is greater than 0

PR #1706, opened 2024-06-27
fix(doc): codestyle version in python

pandalee99:
## What does this PR do?
In fact, in Python's code style, here's the hint
```bash
pip install black==22.1.0 flake8==3.9.1 flake8-quotes flake8-bugbear
```
But there is no version of click
black actually depends on click, by default, our click is the latest, so there is a problem of version inconsistency, and then the developer may not know what version is needed, and may upgrade black. Unfortunately, the ci/cd monitoring results are not the same for different versions of black

I think it's the right thing to do to avoid ambiguity
```bash
pip install black==22.1.0 flake8==3.9.1 flake8-quotes flake8-bugbear click==8.0.2
```

## Related issues
https://github.com/apache/fury/blob/main/CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,7 +96,7 @@ mvn checkstyle:check
 ```bash
 cd python
 # install dependencies fro styling
-pip install black==22.1.0 flake8==3.9.1 flake8-quotes flake8-bugbear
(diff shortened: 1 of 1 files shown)


LiangliangSui (approved):
LGTM

Issue #1544, opened 2024-04-19
[Rust] Implement meta string encoding algorithm for rust


chaokunyang:
## Is your feature request related to a problem? Please describe.
We've implemented  meta string encoding algorithm in https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#meta-string for java in #1514   , it's time to implement it in rust.

## Describe the solution you'd like
Java implementation in #1514 can be taken as a reference. But note that the meta string encoding algorithm is used for encode field name only, so the special charater can't be `.` or `$`, thus the implementation will be simpler

## Additional context
#1413

Kelvinyu1117:
I would like to try.

chaokunyang:
> #1514

Great, thanks for the willingness to contribute to Apache Fury. Note that Java implementation in https://github.com/apache/incubator-fury/pull/1514 has some changes, please see latest code in main branch

urlyy:
Hi！I saw this issue in the Alibaba Cloud Tianchi TONGYI Lingma competition. I would like to try implementing it.

chaokunyang:
> Hi！I saw this issue in the Alibaba Cloud Tianchi TONGYI Lingma competition. I would like to try implementing it.

Great! thanks for contributing to Apache Fury. Seems @Kelvinyu1117 hasn't been working on this, you could take it over

PR #1712, opened 2024-06-29
feat(Rust): add meta string encoding algorithm in Rust

urlyy:
## What does this PR do?
Implementing meta string encoding algorithm in Rust and has passed all 13 tests. The code is in `rust/fury/src/meta/*` and `rust/tests/tests/test_meta.rs` 
Snipaste_2024-06-29_17-07-41
I am not familiar with Rust, I implemented this feature by referring to the Java implementation and utilizing TONGYI Lingma. Therefore, there might be some problems.

Please don't review it immediately. As this version is essentially the same as the Java version, I have a question for `But note that the meta string encoding algorithm is used for encode field name only, so the special charater can't be . or $, thus the implementation will be simpler`. Does this imply I should remove the checks in the Rust version for whether a character is equal to `.` or `$`? In that case, I also need to modify the test cases.

## Related issues

[#1544](https://github.com/apache/fury/issues/1544)

## 本改动借助了通义灵码进行辅助编程
image


chaokunyang:
Could we merge multiple files decoder, encoder, meta_string together? The meta directory will contain other files in the future. Also, test_meta.rs could be renamed to `test_meta_string.rs`

urlyy:
> Could we merge multiple files decoder, encoder, meta_string together? The meta directory will contain other files in the future. Also, test_meta.rs could be renamed to `test_meta_string.rs`

Like this?
image

chaokunyang:
The `$` is a keyword in rust, it won't be included in field name, so we can ignore it to simplify the implementation

chaokunyang:
Could we merge error and encoding file into meta_string.rs?

rust/fury/src/meta/meta_string.rs
```diff
@@ -0,0 +1,488 @@
...
+        }
+    }
+
+    fn decode_lower_upper_digit_special_char(&self, char_value: u8) -> Result<char, Error> {
+        match char_value {
+            0..=25 => Ok((b'a' + char_value) as char),
+            26..=51 => Ok((b'A' + char_value - 26) as char),
+            52..=61 => Ok((b'0' + char_value - 52) as char),
+            62 => Ok(self.special_char1),
+            63 => Ok(self.special_char2),
```
theweipeng, line 449: `special_char1 and special_char2 `  are only used for Java. Java requires three specific characters:  `._$`, but only  62 and 63 are available. In Xlang, this logic is unnecessary because you can't determine the meanings of the 62 and 63 when decoding binary.

theweipeng: This will be helpful for you.  `https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#encoding-algorithms`

theweipeng:
MetaString is used for compressing strings, not only for encoding field names but also supporting UTF-8. Therefore, you don't need to perform any checking on the encoded input. If you wish to verify the field names, the function `derive_fury_meta` might be the right place.

rust/fury/src/meta/meta_string.rs
urlyy: My fault. So I will remove these two variables.
However, I have a question. Besides Java, I also referenced the Python PR [#1702](https://github.com/apache/fury/pull/1702), but that python source code still contains `special_char1` and `special_char2`. This is why I kept them in Rust as well. But Python also doesn't support `$` in field name, right?

urlyy:
> MetaString is used for compressing strings, not only for encoding field names but also supporting UTF-8. Therefore, you don't need to perform any checking on the encoded input. If you wish to verify the field names, the function `derive_fury_meta` might be the right place.

Sorry, I'm a bit unclear about what you mean and what should I do. Do you mean that in the `MetaStringDecoder`, in the following situation, we don't need to check for other characters and return an `Error`?
```rust
 fn decode_lower_special_char(&self, char_value: u8) -> Result<char, Error> {
        match char_value {
            0..=25 => Ok((b'a' + char_value) as char),
            26 => Ok('.'),
            27 => Ok('_'),
            29 => Ok('|'),
            // don't need to return Error?
            _ => Err(Error::InvalidLowerSpecialValue { value: char_value }),
        }
    }
```

rust/fury/src/meta/meta_string.rs
theweipeng: > My fault. So I will remove these two variables. However, I have a question. Besides Java, I also referenced the Python PR [#1702](https://github.com/apache/fury/pull/1702), but that python source code still contains `special_char1` and `special_char2`. This is why I kept them in Rust as well. But Python also doesn't support `$` in field name, right?



> My fault. So I will remove these two variables. However, I have a question. Besides Java, I also referenced the Python PR [#1702](https://github.com/apache/fury/pull/1702), but that python source code still contains `special_char1` and `special_char2`. This is why I kept them in Rust as well. But Python also doesn't support `$` in field name, right?

The  `Encoding Flag` is determined by the function `computeStatistics`.  
The `Encoding Flag` for a  string that containing `$`  would be set to `UTF8` or  `LOWER_SPECIAL`.

I noticed that `special_char1` and `special_char2` were deleted in the commit `https://github.com/apache/fury/pull/1702/commits/266e4e8031bd38b5263b134d5d8a73b78c401a43`

theweipeng:
> > MetaString is used for compressing strings, not only for encoding field names but also supporting UTF-8. Therefore, you don't need to perform any checking on the encoded input. If you wish to verify the field names, the function `derive_fury_meta` might be the right place.
> 
> Sorry, I'm a bit unclear about what you mean and what should I do. Do you mean that in the `MetaStringDecoder`, in the following situation, we don't need to check for other characters and return an `Error`?
> 
> ```rust
>  fn decode_lower_special_char(&self, char_value: u8) -> Result<char, Error> {
>         match char_value {
>             0..=25 => Ok((b'a' + char_value) as char),
>             26 => Ok('.'),
>             27 => Ok('_'),
>             29 => Ok('|'),
>             // don't need to return Error?
>             _ => Err(Error::InvalidLowerSpecialValue { value: char_value }),
>         }
>     }
> ```

Sorry for the confusion. I mean that `MetaString Encode` is used for more than just compressing field names. You don't need to worry about whether to support $; just focus on implementing the compression algorithms.

urlyy:
> Sorry for the confusion. I mean that `MetaString Encode` is used for more than just compressing field names. You don't need to worry about whether to support $; just focus on implementing the compression algorithms.

I appreciate your patience and explanation！😄

rust/tests/tests/test_meta_string.rs
```diff
@@ -0,0 +1,244 @@
...
+    let bytes1 = encoder.encode_lower_special("abc_def").unwrap();
+    assert_eq!(bytes1.len(), 5);
+    let bytes2 = encoder
+        .encode("org.apache.fury.benchmark.data")
+        .unwrap()
+        .bytes;
+    assert_eq!(bytes2.len(), 19);
+    let bytes3 = encoder.encode("MediaContent").unwrap().bytes;
+    assert_eq!(bytes3.len(), 9);
+    // 验证解码
```
chaokunyang, line 34: Could you use English for all comments and documents?

urlyy: sorry about that

theweipeng (approved):
LGTM

PR #1720, opened 2024-07-03
feat(C++): String detection is performed using SIMD techniques

### LINKED ISSUE #313: [C++] String encoding detection by SIMD
We've supported string ascii check using jdk8 unsafe superword in #90. But the simd string encoding check is not implemented in c++.

Add simd string encoding check by SIMD by using [simdutf](https://github.com/simdutf/simdutf)

## Additional context
https://github.com/simdjson/simdjson and it's paper https://arxiv.org/pdf/1902.08318.pdf use SIMD to speed up parsing.

pandalee99:
## What does this PR do?
ref: https://arxiv.org/pdf/1902.08318.pdf
ref: https://github.com/simdutf/simdutf
I learned about the related simd technology, as well as this paper and project implementation.
Using SIMD technique for string detection.
First, I need to implement the logic and complete the latin character detection
``` c++
// Baseline implementation
bool isLatin_Baseline(const std::string& str) {
    for (char c : str) {
        if (static_cast<unsigned char>(c) >= 128) {
            return false;
        }
    }
    return true;
}
```
image
Then, I tried to use SSE2 to speed it up, which is obviously a little bit faster, the logic is to read multiple characters at once and then do the bit arithmetic
Obviously, there was a speed boost, but I didn't think it was enough, so I tried it again with AVX2
image
I think in terms of efficiency, it's already much faster than before. 
But how do you prove that it's also logically true?
I added test samples to verify

``` C++
TEST(StringUtilTest, TestIsLatinLogic)
```

Finally, I ran the test
image
done.

## Related issues
Closes #313 

### DIFF
--- a/cpp/fury/util/string_util.h
+++ b/cpp/fury/util/string_util.h
@@ -26,7 +26,6 @@ namespace fury {
     bool isLatin_Baseline(const std::string& str);
     bool isLatin_AVX2(const std::string& str);
     bool isLatin_SSE2(const std::string& str);
-    bool isLatin_AVX2_vpshufb(const std::string& str);
     std::string generateRandomString(size_t length);
 
 }
--- a/cpp/fury/util/string_util.h
+++ b/cpp/fury/util/string_util.h
@@ -1,0 +1,130 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "string_util.h"
+#include <immintrin.h> // AVX2 header
+#include <emmintrin.h> // SSE2 header
+#include <random>
+
+namespace fury {
+
+    bool isLatin_AVX2(const std::string& str) {
+        const char* data = str.data();
+        size_t len = str.size();
+
+        // Process 32 characters at a time with AVX2
+        size_t i = 0;
+        __m256i latin_mask = _mm256_set1_epi8(0x80);
+        for (; i + 32 <= len; i += 32) {
+            __m256i chars = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(data + i));
+            __m256i result = _mm256_and_si256(chars, latin_mask);
+            if (!_mm256_testz_si256(result, result)) {
+                return false;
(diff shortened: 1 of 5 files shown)


pandalee99:
@chaokunyang can u help me review this pr?

pandalee99:
@chaokunyang can u help me see this problem? 
image
this seems to be a CI error

chaokunyang:
image
Seems this test has an compilation error

cpp/fury/util/string_util_test.cc
```diff
@@ -0,0 +1,95 @@
...
+                           end_time - start_time)
+                           .count();
+  std::cout << "SSE2 Running Time: " << duration_sse2 << " ns" << std::endl;
+  EXPECT_TRUE(result_sse2);
+}
+
+// TEST Logic
+TEST(StringUtilTest, TestIsLatinLogic) {
+
+  EXPECT_TRUE(isLatin_AVX2("Fury"));
```
chaokunyang, line 66: AVX2 doesn't exists for ARM based CPU, could we make this work on ARM by using some macro check

pandalee99: thanks!

cpp/fury/util/string_util.h
```diff
@@ -0,0 +1,33 @@
...
+#include <string>
+
+namespace fury {
+
+bool isLatin_Baseline(const std::string &str);
+bool isLatin_AVX2(const std::string &str);
+bool isLatin_SSE2(const std::string &str);
+bool isLatin_NEON(const std::string &str);
+bool isLatin_RISCV(const std::string &str);
+std::string generateRandomString(size_t length);
```
chaokunyang, line 31: Could we add a method which select implementation based the macro. Otherwise, how do know which method should be invoked?

cpp/fury/util/string_util_test.cc
```diff
@@ -0,0 +1,84 @@
...
+
+#include "string_util.h"
+#include <chrono>
+#include <gtest/gtest.h>
+#include <iostream>
+
+namespace fury {
+
+#if defined(__x86_64__) || defined(_M_X64)
+bool isLatin(const std::string &str) { return isLatin_AVX2(str); }
```
chaokunyang, line 28: this should be moved into string_util.h

PragmaTwice:
I'm wondering if we can add a benchmark into the codebase via https://github.com/google/benchmark.

cpp/fury/util/string_util.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <string>
+
+namespace fury {
+
+bool isLatin_Baseline(const std::string &str);
+bool isLatin_AVX2(const std::string &str);
+bool isLatin_SSE2(const std::string &str);
+bool isLatin_NEON(const std::string &str);
+bool isLatin_RISCV(const std::string &str);
```
PragmaTwice, line 30: Could we detect the platform automatically and provide a `isLatin` that can select the suitable implementation?

pandalee99: I get it

cpp/fury/util/string_util.h
pandalee99: I get it

pandalee99:
Of course, we can add benchmark, but the change of this pr is relatively small. If we need to add benchmark, we can build the basic framework in another pr, how about that?
For this pr I added Fury log to express the increased efficiency.
image

chaokunyang (approved):
LGTM

PR #1725, opened 2024-07-08
build: update scala build

pjfanning:
* sbt 1.10.0
* scalatest 3.2.19
* scala 2.13.14
* add Apache Snapshot Repo resolver so you don't need to build the Java jars, the build can get them from the snapshot repo


chaokunyang (approved):
LGTM

Issue #1547, opened 2024-04-19
[Rust] Support convert utf16 encoded string to utf8 string


chaokunyang:
## Is your feature request related to a problem? Please describe.
Currently Fury xlang serialization use utf8 for string encoding, which is not performance efficient in many languages. 

We introduced utf16 in https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#string . But rust native string doesn't support utf16, it's is utf-8 encoded. 

We should support to transcode  utf16 encoded string to utf8 string in fury rust deserialization.

## Describe the solution you'd like

Implement utf16 to utf8 convertion in fury rust. The implementation should use SIMD to provide faster speed.

## Additional context
#1413

urlyy:
Hi, I've created a basic demo. Since Rust's string encoding is UTF-8, we can directly use the String's API to convert UTF-16 encoded data into a string. However, this method doesn't utilize SIMD. So, I'm wondering what else needs to be done on top of this. 
```rust
fn main() {
    let s = "Hello";
    // initial utf-16 encoded data
    let utf16_encoded: Vec<u16> = s.encode_utf16().collect();
    // convert utf-16 to utf-8
    match String::from_utf16(&utf16_encoded) {
        Ok(string) => {
            // let utf8_data = string.into_bytes();
            println!("{}", string);
        }
        Err(e) => {
            println!("Error converting UTF-16 to UTF-8: {:?}", e);
        }
    }
}
```
I'm not familiar with high-performance computing and I've only found the [std::simd](https://doc.rust-lang.org/beta/std/simd/index.html) library, however, it is a **nightly-only experimental** API.

urlyy:
Hi, I've created a basic demo. Since Rust's string encoding is UTF-8, we can directly use the String's API to convert UTF-16 encoded data into a string. However, this method doesn't utilize SIMD. So, I'm wondering what else needs to be done on top of this. 
```rust
let bytes = [
        0b01101000, // 'h'
        0b00000000, 
        0b01100101, // 'e'
        0b00000000, 
        0b01101100, // 'l'
        0b00000000, 
        0b01101100, // 'l'
        0b00000000, 
        0b01101111, // 'o'
        0b00000000, 
        0b00010110, // '世' in UTF-16 little-endian
        0b01001110, 
        0b01001100, // '界' in UTF-16 little-endian
        0b01110101, 
    ];
    let utf16_vec: Vec<u16> = bytes
        .chunks_exact(2)
[...]

PR #1730, opened 2024-07-10
feat(Rust): Implement utf16 to utf8 conversion algorithm

urlyy:
## What does this PR do?
Implement an util function converting utf16 encoded string to utf8 string in Rust.

## Related issues
#1547 

## Benchmark

### DIFF
--- a/rust/fury/src/util.rs
+++ b/rust/fury/src/util.rs
@@ -1,0 +1,93 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#[derive(thiserror::Error, Debug)]
+pub enum UtilError {
+    #[error("Invalid UTF-16 string: missing surrogate pair")]
+    MissingSurrogatePair,
+}
+
+// Swapping the high 8 bits and the low 8 bits of a 16-bit value
+fn swap_endian(value: u16) -> u16 {
+    ((value & 0xff) << 8) | ((value & 0xff00) >> 8)
+}
+
+pub fn to_utf8(utf16: &[u16], is_little_endian: bool) -> Result<Vec<u8>, UtilError> {
+    // Pre-allocating approximate capacity to minimize dynamic resizing
+    let mut utf8_bytes = Vec::with_capacity(utf16.len() * 2);
+    let mut iter = utf16.iter();
+    while let Some(&wc) = iter.next() {
+        let wc = if is_little_endian {
+            swap_endian(wc)
+        } else {
+            wc
+        };
+        match wc {
+            // 1-byte UTF-8
+            code_point if code_point < 0x80 => {
+                utf8_bytes.push(code_point as u8);
+            }
+            // 2-byte UTF-8
+            code_point if code_point < 0x800 => {
+                // 110????? 10??????
+                // Need 11 bit suffix of wc
+                let second = ((code_point & 0b111111) as u8) | 0b10000000;
+                let first = ((code_point >> 6 & 0b11111) as u8) | 0b11000000;
+                utf8_bytes.push(first);
+                utf8_bytes.push(second);
+            }
+            // Surrogate pair (4-byte UTF-8)
+            wc1 if (0xd800..=0xdbff).contains(&wc1) => {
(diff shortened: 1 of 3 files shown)


rust/fury/src/util.rs
```diff
@@ -0,0 +1,93 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#[derive(thiserror::Error, Debug)]
+pub enum UtilError {
```
chaokunyang, line 19: Could we use standard error instead? UtilError looks vague to me

chaokunyang:
Looks good to me overall, only some minor error handle issues. @theweipeng do you have some suggestion?

rust/fury/src/util.rs
```diff
@@ -0,0 +1,90 @@
...
+// under the License.
+
+// Swapping the high 8 bits and the low 8 bits of a 16-bit value
+fn swap_endian(value: u16) -> u16 {
+    (value << 8) | (value >> 8)
+}
+
+pub fn to_utf8(utf16: &[u16], is_little_endian: bool) -> Result<Vec<u8>, String> {
+    // Pre-allocating approximate capacity to minimize dynamic resizing
+    let mut utf8_bytes = Vec::with_capacity(utf16.len() * 2);
```
theweipeng, line 25: A UTF16 char can be encoded into up to 4 UTF8 bytes. Perhaps we can  change it to `Vec::with_capacity(utf16.len() * 4);`

rust/fury/src/util.rs
```diff
@@ -0,0 +1,90 @@
...
+                    // utf16 to unicode
+                    let code_point =
+                        ((((wc1 as u32) - 0xd800) << 10) | ((wc2 as u32) - 0xdc00)) + 0x10000;
+                    // 11110??? 10?????? 10?????? 10??????
+                    // Need 21 bit suffix of code_point
+                    let fourth = ((code_point & 0b11_1111) as u8) | 0b1000_0000;
+                    let third = ((code_point >> 6 & 0b11_1111) as u8) | 0b1000_0000;
+                    let second = ((code_point >> 12 & 0b11_1111) as u8) | 0b1000_0000;
+                    let first = ((code_point >> 18 & 0b111) as u8) | 0b1111_0000;
+                    utf8_bytes.push(first);
```
theweipeng, line 68: `push` will check the memory boundary, since we initialize utf8_bytes with a fixed capacity, we can avoid this overhead by using a more efficient way. Merging four u8 values into a single u32 to reduce additional memory writes.

theweipeng (commented):
Some suggestion

rust/fury/src/util.rs
urlyy: I think the maximum length should be `utf16.len()*3`, because `four u8` uses `two u16`

rust/fury/src/util.rs
urlyy: Thanks for your advice. I have replaced `push` with `ptr.add(offset).write(value)` inside an `unsafe` block. However, I can't write a `u32` into `Vec<u8>` at once, so I'm still using multiple write operations.

urlyy:
I also find a dataset of character set: https://github.com/lemire/unicode_lipsum, hope this will be useful for future benchmarks.

rust/fury/src/util.rs
theweipeng: You can cast u8 ptr to u32 ptr.  for instance  `buf.as_mut_ptr() as *mut i32;`

urlyy: I found that write u32 to [u8:4] will be affected by processor's endianness.
```rust
fn main() {
    let mut vec = vec![0u8; 4];
    unsafe {
        let ptr = vec.as_mut_ptr() as *mut u32;
        ptr.write(0xa1b2c3d4);
        vec.set_len(4);
    }
    for i in 0..vec.len() {
        print!("0x{:02x} ", vec[i]);
    }
    // print: 0xd4 0xc3 0xb2 0xa1
}
````
So I use `#[cfg(target_endian = "little | big")]` to handle this.
```rust
#[cfg(target_endian = "little")]
ptr.write(0xd4c3b2a1);

#[cfg(target_endian = "big")]
ptr.write(0xa1b2c3d4);
```

rust/fury/src/util.rs
```diff
@@ -0,0 +1,122 @@
...
+                unsafe {
+                    ptr.add(offset).write(code_point as u8);
+                }
+                offset += 1;
+            }
+            code_point if code_point < 0x800 => {
+                // 2-byte UTF-8
+                // [0000|0bbb|bbcc|cccc] => [110|bbbbb], [10|cccccc]
+                // write a u16 to [u8;2] is affected by the processor's endianness.
+                #[cfg(target_endian = "little")]
```
theweipeng, line 51: We can use copy_from_slice 
```
let sl = utf8_bytes.as_mut_slice()
sl.copy_from_slice(&[u1,u2,u3,u4]);
```
Or copy_nonoverlapping
```
ptr::copy_nonoverlapping(src.as_ptr(), self.as_mut_ptr(), self.len());
```

```
to avoid the multiple times memory write overhead.  Meanwhile, enhance the readability of the code

theweipeng (approved):
LGTM

Issue #1735, opened 2024-07-17
FR: Only serialize annotated fields


Frotty:
## Is your feature request related to a problem? Please describe.
I want to move an existing project to fury, but right now data fields in classes are marked with an annotation. Fury only supports not de-/serializing transient fields, but it's not convenient to mark a majority of fields as transient.

## Describe the solution you'd like
I would like there to be an annotation that allows me to mark only the fields of a class that should be de-/serialized.

## Additional context
This was discussed @chaokunyang in https://www.reddit.com/r/java/comments/1cll68f/comment/ld3suns/?context=3 and he is generally supportive. However, there seems to be no real guide for me to start implementing this. Perhaps someone else can do it in a shorter amount of time.

urlyy:
@chaokunyang I'm interested in this issue. However , as it's my first time delving into the Java version of Fury, I'm not quite sure if my approach is correct. By using DEBUG, I've traced the code and found a method which obtains the list of fields of the object being serialized, this `descriptorMap` ultimately also participates in serializer codegen. Here is already a check on the `Ignore` annotation; I need to modify the logic.
Do you have some advice? To be honest I don't know if it's enough to modify this method , maybe I should explore more.

![image](https://github.com/user-attachments/assets/e466c7f8-dbf5-4fbc-8622-37bf4d507f14)

Frotty:
Cool, would be nice to see this implemented. I suppose there must be a configuration or mode, or annotation on the class itself, to switch the fury behavior to only serialize marked fields. Otherwise, might be quite straightforward. Currently, the annotation is `@Exposed` but I don't really care about the name.

chaokunyang:
The implementation should not be complex, but the naming is an issue. This is a public API, how do we name this annotation?

Pil0tXia:
Good job. Might we consider this issue as completed?

Frotty:
I wasn't able to test it yet, when will it be available in a release?

chaokunyang:
> I wasn't able to test it yet, when will it be available in a release?

You can use snapshot jar for test. A release jar will be released by the end of this week

PR #1751, opened 2024-07-22
feat(java): Add 'Expose' annotation to support "only de/serialize annotated fields"

urlyy:
## What does this PR do?
Add `Expose` annotation to support "only de/serialize annotated fields"

## Related issues
#1735 

## Benchmark
N/A

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/annotation/Expose.java
+++ b/java/fury-core/src/main/java/org/apache/fury/annotation/Expose.java
@@ -27,4 +27,4 @@ import java.lang.annotation.Target;
 /** Only fields with this annotation are permitted for de/serialization. */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
-public @interface Exposed {}
+public @interface Expose {}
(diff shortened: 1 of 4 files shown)


urlyy:
@chaokunyang I have already modified the code style according to the output of `mvn checkstyle:check`, but it still did not pass the GitHub Actions check. Can you help me identify the reason?

chaokunyang:
image

You need to format code by `mvn spotless:apply`

chaokunyang:
The code looks good, but I have concerns about the annotation name. This is the hard one. We have many choices such as `Expose`/`SerializeField`/`FuryField`/`Include`/`FuryInclude`. I'm not sure which is better

chaokunyang:
I think we can name it as `Expose`, and in future we can add more advanced annotation like:
```java
@interface Include {
  String[] fieldsNames();
}

@interface Exclude {
  String[] fieldsNames();
}

@Include(fieldsNames={"f1"})
class Foo {
  inf f1;
  bool f2;
}

@Exclude(fieldsNames={"f2"})
class Foo {
  inf f1;
  bool f2;
}
```

chaokunyang (approved):
LGTM

PR #1768, opened 2024-07-27
fix(java): Fix memory leak in `StructSerializer.xread()` caused by re-pushing instead of popping `GenericType`.

komamitsu:
## What does this PR do?

Fix memory leak in `StructSerializer.xread()` caused by re-pushing instead of popping `GenericType`.

## Related issues

None, probably.

## Benchmark

N/A


chaokunyang (approved):
Good catch, thanks

PR #1775, opened 2024-07-27
feat(rust): support simd convert utf16 to utf8

urlyy:
## What does this PR do?
For the conversion from UTF-16 to UTF-8, a SIMD method based on AVX/SSE/NEON instruction sets was added on the basis of #1730 , and benchmarks were written.

referencing 
- https://github.com/simdutf/simdutf/blob/master/src/westmere/sse_convert_utf16_to_utf8.cpp
- https://github.com/simdutf/simdutf/blob/master/src/haswell/avx2_convert_utf16_to_utf8.cpp
- https://github.com/simdutf/simdutf/blob/5c1a86887010cd2b4d648049c4d73de81a026341/src/arm64/arm_convert_utf16_to_utf8.cpp
- https://github.com/simdutf/simdutf/blob/master/src/tables/utf16_to_utf8_tables.h

## Related issues
- #1547 
- #1730 

## Benchmark
In local machine benchmark , simd approach seems to be Two to Three times faster than normal approach. But when handling with surrogate pair, algorithm will use fall_back (normal) way, in this case simd approach is worse than normal way.

### DIFF
--- a/rust/fury/Cargo.toml
+++ b/rust/fury/Cargo.toml
@@ -36,6 +36,9 @@ thiserror = { default-features = false, version = "1.0" }
 name = "simd_bench"
 harness = false
 
+[[bench]]
+name = "simd_utf16_to_utf8"
+harness = false
 
 [dev-dependencies]
 criterion = "0.5.1"
--- a/rust/fury/benches/simd_utf16_to_utf8.rs
+++ b/rust/fury/benches/simd_utf16_to_utf8.rs
@@ -1,0 +1,178 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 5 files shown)


urlyy:
I learned the technique used in https://github.com/simdutf/simdutf  , and  implemented in fury.
Here are some questions:
1. Sorry for I can only test on x86 platform. I rely on GitHub CI to debug the logic of the AArch64 version.😢 I can't give the aarch64 benchmark data.
2. There may be a lot of comments. If they are not desired , I will remove them.
3.  I have seen another SIMD code in rust , `string_util.rs`, should I merge it with my code like what has done in `cpp`?
4. As it is helpful to use the precomputing table, I have put it in my code , but takes approximately 1600 lines. Do you have some advice on this? 
5. For better readability, I have used nested mod and divided the various steps into different functions. If this approach is not satisfactory, I am willing to make adjustments。
6. To be honest, I am uncertain whether introducing such an extensive amount of code is justified merely for the purpose of converting UTF to UTF-8. However, I have observed optimizations, particularly for the conversion of 1 UTF-16 to 1, 2, or 3 UTF-8 bytes. This has added complexity。

PR #1780, opened 2024-07-31
feat(java): ThreadSafeFury add getClassResolver method

funky-eyes:
## What does this PR do?

Because using ThreadSafeFury cannot get ClassResolver, so I submitted this pr to increase some checklistener and other behavior


java/fury-core/src/test/java/org/apache/fury/FuryCopyTest.java
```diff
@@ -137,8 +137,10 @@ public void threadLocalCopyTest() {
     BeanA beanA = BeanA.createBeanA(2);
     ExecutorService executor = Executors.newSingleThreadExecutor();
     AtomicReference<Throwable> ex = new AtomicReference<>();
-    ThreadLocalFury threadLocalFury =
-        builder().withCodegen(false).withRefCopy(true).buildThreadLocalFury();
+    ThreadLocalFury threadLocalFury = builder().withCodegen(false).withRefCopy(true).buildThreadLocalFury();
+    threadLocalFury.getClassResolver().setClassChecker((classResolver, className1) -> true);
```
chaokunyang, line 141: This won't set set class checker for other fury instance. We could set Class checker when building threadsafefury by creating threadlocalfury directly

funky-eyes: > This won't set set class checker for other fury instance. We could set Class checker when building threadsafefury by creating threadlocalfury directly

I believe that in some cases, adding check listeners and serializers involves dynamically adding or modifying them at runtime, which cannot be predicted with absolute certainty

chaokunyang: If so, maybe we need to add `setClassChecker` to BaseFury/ThreadSafeFur directly.

funky-eyes: > If so, maybe we need to add `setClassChecker` to BaseFury/ThreadSafeFur directly.

That's a good idea. I will make modifications based on this suggestion

funky-eyes: done

LiangliangSui:
Hi @funky-eyes , the rust ci failure is not related to your changes, I will fix this soon, track the ci failure issue at https://github.com/apache/fury/issues/1781

funky-eyes:
> Hi @funky-eyes , the rust ci failure is not related to your changes, I will fix this soon, track the ci failure issue at #1781

thx

java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
```diff
@@ -266,8 +268,18 @@ public ClassLoader getClassLoader() {
   }
 
   @Override
-  public ClassResolver getClassResolver() {
-    return bindingThreadLocal.get().get().getClassResolver();
+  public void setClassChecker(ClassChecker classChecker) {
+    bindingThreadLocal.get().get().getClassResolver().setClassChecker(classChecker);
```
chaokunyang, line 272: We need to use processCallback

funky-eyes: > We need to use processCallback

I don't quite understand what you mean. Could you give me an example?

chaokunyang: You can take `setSerializerFactory` as an example:
```java
  public void setSerializerFactory(SerializerFactory serializerFactory) {
    processCallback(fury -> fury.setSerializerFactory(serializerFactory));
  }
```

chaokunyang: And maybe we can add following API for extending:
```java

public void configureClassResolver(Consumer<ClassResolver> apply);

```

funky-eyes: > You can take `setSerializerFactory` as an example:
> 
> ```java
>   public void setSerializerFactory(SerializerFactory serializerFactory) {
>     processCallback(fury -> fury.setSerializerFactory(serializerFactory));
>   }
> ```

I feel doing it this way might be simpler.

java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
```diff
@@ -265,6 +267,16 @@ public ClassLoader getClassLoader() {
     return bindingThreadLocal.get().getClassLoader();
   }
 
+  @Override
+  public ClassChecker getClassChecker() {
+    return bindingThreadLocal.get().get().getClassResolver().getClassChecker();
+  }
+
+  @Override
+  public SerializerFactory getSerializerFactory() {
```
chaokunyang, line 276: We should not provide this api. It may return different object when classloader changes or we invoke It from another thread

java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
```diff
@@ -265,6 +267,16 @@ public ClassLoader getClassLoader() {
     return bindingThreadLocal.get().getClassLoader();
   }
 
+  @Override
+  public ClassChecker getClassChecker() {
```
chaokunyang, line 271: Please implement this in AbstractThreadSafeFury

java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
funky-eyes: You are correct; I will proceed with the corrections

java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
funky-eyes: ok

funky-eyes: > Please implement this in AbstractThreadSafeFury

AbstractThreadSafeFury does not have bindingThreadLocal, so it seems unable to implement this interface.

chaokunyang: You can invoke processCallback instead

funky-eyes: > You can invoke processCallback instead

processCallback ReturnType is void

chaokunyang: I think you can create ClassChecker first and set it to Fury. Then you don't need to add any API to Fury. We won't add such API to get Classchecker or ClassResolver/SerializerFactory

funky-eyes: > We can only allow set ClassChecker

ok

chaokunyang (approved):
LGTM

PR #1789, opened 2024-08-04
feat(Rust): Implement TypeMeta part of the compatibility features

theweipeng:
### Refine code struct and implement some compatibility features
1. Split fury_meta file into individual files, as the file is too large to be easily read.  There are a lot of features should be implemented by macro, split it is necessary.
2. Implement compatibility by the macro, which will now generate a pattern match expression.
3. Write type meta to binary when serializing objects.
4. Read type meta for the pattern match which is generated at compile time when deserializing.
5. Replace lazy_static by Once for it has been stabled in stdlib

### Some unimplemented features
1. Add a v-table for looking up generic methods by type_id, which is used for support the Any type.
2. Support draining unused binary data when receiving an unregistered type.


chaokunyang:
Do we need to rename `internal` to `serializer` and rename `resolvers` to `resolver` to keep consistent with other languages?

chaokunyang (approved):
LGTM

PR #1791, opened 2024-08-05
feat(java): Add EnumNameSerializer for Enum

lionel-xie:
## What does this PR do?

Add EnumNameSerializer for Enum, can config it by org.apache.fury.config.Config#useEnumNameSerializer, default value is false


java/fury-core/pom.xml
```diff
@@ -25,7 +25,7 @@
   <parent>
     <groupId>org.apache.fury</groupId>
     <artifactId>fury-parent</artifactId>
-    <version>0.7.0</version>
+    <version>0.7.1-SNAPSHOT</version>
```
chaokunyang, line 28:
```suggestion
    <version>0.8.0-SNAPSHOT</version>
```

java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
```diff
@@ -312,6 +313,11 @@ public FuryBuilder withAsyncCompilation(boolean asyncCompilation) {
     return this;
   }
 
+  public FuryBuilder withUseEnumNameSerializer(boolean useEnumNameSerializer) {
```
chaokunyang, line 316: How about naming to `withSerializeEnumByName`?

java/fury-core/src/main/java/org/apache/fury/config/Config.java
```diff
@@ -248,6 +251,10 @@ public boolean isScalaOptimizationEnabled() {
     return scalaOptimizationEnabled;
   }
 
+  public boolean getUseEnumNameSerializer() {
```
chaokunyang, line 254: How about `isSerializeEnumByName`

java/fury-core/src/main/java/org/apache/fury/resolver/EnumStringResolver.java
```diff
@@ -0,0 +1,46 @@
+package org.apache.fury.resolver;
+
+import org.apache.fury.memory.MemoryBuffer;
+import org.apache.fury.meta.MetaString;
+
+import static org.apache.fury.meta.Encoders.GENERIC_DECODER;
+import static org.apache.fury.meta.Encoders.GENERIC_ENCODER;
+
+public class EnumStringResolver {
```
chaokunyang, line 9: Why not use MetaStringResolver directly

java/fury-core/src/main/java/org/apache/fury/serializer/EnumNameSerializer.java
```diff
@@ -0,0 +1,80 @@
...
+      Preconditions.checkNotNull(enclosingClass);
+      Preconditions.checkArgument(enclosingClass.isEnum());
+      enumConstantsMap = Arrays.stream(enclosingClass.getEnumConstants())
+              .collect(Collectors.toMap(Enum::name, Function.identity()));
+    }
+  }
+
+  @Override
+  public void write(MemoryBuffer buffer, Enum value) {
+    enumStringResolver.writeMetaStringBytes(buffer, value);
```
chaokunyang, line 60: We'd better use `org.apache.fury.resolver.MetaStringResolver#writeMetaStringBytes` instead. We can create `MetaStringBytes[]` when building `EnumNameSerializer`

java/fury-core/src/main/java/org/apache/fury/serializer/EnumNameSerializer.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Override
+  public Enum read(MemoryBuffer buffer) {
+    String key = enumStringResolver.readMetaString(buffer);
```
chaokunyang, line 65: We do not need to read a string, we can read `MetaStringBytes` using `org.apache.fury.resolver.MetaStringResolver#readMetaStringBytes(org.apache.fury.memory.MemoryBuffer)` instead

chaokunyang:
Hi @xxd819240366 , thanks for contribuing to Apache Fury. The code looks to me overall, I left some minor comments

java/fury-format/pom.xml
```diff
@@ -25,7 +25,7 @@
   <parent>
     <groupId>org.apache.fury</groupId>
     <artifactId>fury-parent</artifactId>
-    <version>0.7.0</version>
+    <version>0.7.1-SNAPSHOT</version>
```
chaokunyang, line 28:
```suggestion
    <version>0.8.0-SNAPSHOT</version>
```

java/fury-test-core/pom.xml
```diff
@@ -25,7 +25,7 @@
   <parent>
     <artifactId>fury-parent</artifactId>
     <groupId>org.apache.fury</groupId>
-    <version>0.7.0</version>
+    <version>0.7.1-SNAPSHOT</version>
```
chaokunyang, line 28:
```suggestion
    <version>0.8.0-SNAPSHOT</version>
```

java/fury-testsuite/pom.xml
```diff
@@ -25,7 +25,7 @@
   <parent>
     <artifactId>fury-parent</artifactId>
     <groupId>org.apache.fury</groupId>
-    <version>0.7.0</version>
+    <version>0.7.1-SNAPSHOT</version>
```
chaokunyang, line 28:
```suggestion
    <version>0.8.0-SNAPSHOT</version>
```

java/pom.xml
```diff
@@ -33,7 +33,7 @@
   <groupId>org.apache.fury</groupId>
   <artifactId>fury-parent</artifactId>
   <packaging>pom</packaging>
-  <version>0.7.0</version>
+  <version>0.7.1-SNAPSHOT</version>
```
chaokunyang, line 36:
```suggestion
  <version>0.8.0-SNAPSHOT</version>
```

java/fury-core/src/main/java/org/apache/fury/config/Config.java
xxd819240366: no problem

java/fury-core/src/main/java/org/apache/fury/resolver/EnumStringResolver.java
xxd819240366: because org.apache.fury.resolver.MetaStringResolver#getOrCreateMetaStringBytes is not public and MetaStringBytes annotated by @Internal. I feel like I shouldn't break the existing design

chaokunyang: You can use it in enum serialization, `@internal` just means users' serializer should not use this API

xxd819240366:
> 

Ok，I'll update it again

java/fury-core/src/main/java/org/apache/fury/resolver/EnumStringResolver.java
lionel-xie: Ok, so i can add public before org.apache.fury.resolver.MetaStringResolver#getOrCreateMetaStringBytes？

lionel-xie: And I also need to add public before MetaStringBytes class, are you sure I can do this? 😂

lionel-xie:
By the way, I can pass all test on my local development environment with java 8，but can't pass test on Java CI(8) . There is a error in org.apache.fury.serializer.EnumSerializerTest.testEnumSubclassFieldCompatible(org.apache.fury.serializer.EnumSerializerTest) only happen on github CI pipline

lionel-xie:
@chaokunyang

chaokunyang:
Seems some your jar version is 0.7.0 but others are 0.8.0-SNAPSHOT. Could you update all mavem module to 0.8.0-SNAPSHOT?

lionel-xie:
@chaokunyang  Seems all maven module version already updated to 0.8.0-SNAPSHOT
1723965789422_paste_img

java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
```diff
@@ -241,13 +252,40 @@ public Object read(MemoryBuffer buffer) {
...
+                        return new AbstractMap.SimpleEntry<>(stringBytes, e);
+                    })
+                    .collect(Collectors.toMap(AbstractMap.SimpleEntry::getKey, AbstractMap.SimpleEntry::getValue));
+        }
 
+        @Override
+        public Enum read(MemoryBuffer buffer) {
+            MetaStringBytes metaStringBytes = metaStringResolver.readMetaStringBytes(buffer);
+            if (!enumMap.containsKey(metaStringBytes)) {
+                Optional<NonexistentEnum> first = enumMap.values().stream().findFirst();
```
chaokunyang, line 276: if `deserializeNonexistentEnumValueAsNull` is enabled, we should return null here, otherwise we should throw Exception

java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
```diff
@@ -241,13 +252,40 @@ public Object read(MemoryBuffer buffer) {
...
+                return first.orElse(null);
+            }
+            return enumMap.get(metaStringBytes);
```
chaokunyang, line 279: This introduce a second map loopkup, we should avoid it. we can `value = enumMap.get(xxx);` and check whether `value` is null instead of check by `containsKey`

java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
chaokunyang: Or instead of returning null, we could return `NonexistentEnum.UNKNOWN` instead

chaokunyang:
image

chaokunyang:
@lionel-xie Could you merge fury main branch first, your base branch is 0.7.0, not the developer main branch

chaokunyang:
done in #1892

PR #1792, opened 2024-08-05
fix(java): fix classloader get npe

### LINKED ISSUE #1763: NullPointer from ThreadPoolFury.serialize
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

0.6.0

### Component(s)

Java

### Minimal reproduce step

I have not been able to produce a minimal code example. It is very difficult to debug, since I am just getting: `java.lang.NullPointerException: null` as the exception. This is occurring when trying to serialize values.

I am also seeing some errors when trying to deserialize values (far fewer then the serialize error):
```
java.lang.NullPointerException: Cannot invoke "String.length()" because "msg" is null
	at org.apache.fury.logging.FuryLogger.log(FuryLogger.java:145)
	at org.apache.fury.logging.FuryLogger.error(FuryLogger.java:133)
	at org.apache.fury.pool.FuryPooledObjectFactory.getPooledCache(FuryPooledObjectFactory.java:95)
	at org.apache.fury.pool.ThreadPoolFury.execute(ThreadPoolFury.java:77)
	at org.apache.fury.pool.ThreadPoolFury.deserialize(ThreadPoolFury.java:135)
```

No null pointers when serializing and deserializing data.

### What did you see instead?

j.l.NullPointerException: null

### Anything Else?

I am using Scala 2.13.12, my application runs in k8s (debian 11), with JDK 17. In production there are around 200 pods, and 20 cpus per pod.

I configure Fury in a scala Object like so:

```scala
object FuryUtils {
  lazy val fury = Fury
    .builder()
    .withLanguage(Language.JAVA)
    .withScalaOptimizationEnabled(true)
    .withRefTracking(false)
    .buildThreadSafeFuryPool(Runtime.getRuntime.availableProcessors() * 2, Runtime.getRuntime.availableProcessors() * 2)
}
```

and i serialize/deserialze my values like:
```scala
fury.serialize(value)
fury.deserialize(bytes).asInstanceOf[MyData]
```

I am calling this function ~20,000 times per second. And it is failing ~30 times per second.

chaokunyang:
## What does this PR do?

 fix classloader get npe

## Related issues

Closes #1763

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
@@ -257,6 +257,9 @@ public class ThreadLocalFury extends AbstractThreadSafeFury {
 
   @Override
   public void setClassLoader(ClassLoader classLoader, StagingType stagingType) {
+    if (classLoader == null) {
+      classLoader = Fury.class.getClassLoader();
+    }
     bindingThreadLocal.get().setClassLoader(classLoader, stagingType);
   }
 
--- a/java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
+++ b/java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
@@ -257,9 +257,6 @@ public class ThreadLocalFury extends AbstractThreadSafeFury {
 
   @Override
   public void setClassLoader(ClassLoader classLoader, StagingType stagingType) {
-    if (classLoader == null) {
-      classLoader = Fury.class.getClassLoader();
-    }
     bindingThreadLocal.get().setClassLoader(classLoader, stagingType);
   }
 
--- a/java/fury-core/src/main/java/org/apache/fury/pool/FuryPooledObjectFactory.java
+++ b/java/fury-core/src/main/java/org/apache/fury/pool/FuryPooledObjectFactory.java
@@ -21,8 +21,6 @@ package org.apache.fury.pool;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
-
-import java.net.URLClassLoader;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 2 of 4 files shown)


java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
```diff
@@ -257,6 +257,9 @@ public void setClassLoader(ClassLoader classLoader) {
 
   @Override
   public void setClassLoader(ClassLoader classLoader, StagingType stagingType) {
+    if (classLoader == null) {
```
LiangliangSui, line 260: This should be redundant, there is already such a way in `org.apache.fury.util.LoaderBinding#setClassLoader`

java/fury-core/src/main/java/org/apache/fury/pool/ThreadPoolFury.java
```diff
@@ -279,6 +279,9 @@ public void setClassLoader(ClassLoader classLoader) {
 
   @Override
   public void setClassLoader(ClassLoader classLoader, LoaderBinding.StagingType stagingType) {
+    if (classLoader == null) {
```
LiangliangSui, line 282: `org.apache.fury.pool.FuryPooledObjectFactory#setClassLoader`

java/fury-core/src/main/java/org/apache/fury/pool/FuryPooledObjectFactory.java
```diff
@@ -50,7 +52,13 @@ public class FuryPooledObjectFactory {
 
   /** ThreadLocal: ClassLoader. */
   private final ThreadLocal<ClassLoader> classLoaderLocal =
-      ThreadLocal.withInitial(() -> Thread.currentThread().getContextClassLoader());
+      ThreadLocal.withInitial(() -> {
+        ClassLoader loader = Thread.currentThread().getContextClassLoader();
+        if (loader == null) {
+          loader = Fury.class.getClassLoader();
+        }
```
LiangliangSui, line 59: Could we move this into `org.apache.fury.config.FuryBuilder#finish` ?

java/fury-core/src/main/java/org/apache/fury/ThreadLocalFury.java
chaokunyang: Good catch

LiangliangSui:
Do we need to add the same logic in https://github.com/apache/fury/blob/main/java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java#L323?

chaokunyang:
> Do we need to add the same logic in https://github.com/apache/fury/blob/main/java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java#L323?

Updated to:
```java
    if (classLoader == null) {
      classLoader = Thread.currentThread().getContextClassLoader();
      if (classLoader == null) {
        classLoader = Fury.class.getClassLoader();
      }
    }
```

LiangliangSui (approved):
LGTM

PR #1795, opened 2024-08-07
feat(Rust): Support polymorphism

theweipeng:
## What does this PR do?
1. Support polymorphism
Add a class resolver which is used for managing the relationship between type_id and serializer. The serializer is generated by a generic function which will be called by dynamic dispatch. When a struct uses `Box<dyn Any>` as a field type and does not specify the specific type, we will use the class resolver to load the handler by type_id. In this situation, there is some performance overhead due to hash lookup, but it greatly improves convenience.
Use as follow:
```Rust
#[test]
fn any() {
    #[derive(Fury, Debug)]
    struct Animal {
        f3: String,
    }

    #[derive(Fury, Debug)]
    struct Person {
        f1: Box<dyn Any>,
    }

    let person = Person {
        f1: Box::new(Animal {
            f3: String::from("hello"),
        }),
    };

    let mut fury = Fury::default();
    fury.register::<Animal>(999);
    fury.register::<Person>(1000);
    let bin = fury.serialize(&person);
    let obj: Person = fury.deserialize(&bin).expect("");
    assert_eq!(true, obj.f1.is::<Animal>())
}

```
2. Add a register  function for user to register Struct and id
3. Remove tag and hash which were generate by macro before and were removed in our protocol now.

## TODO
1. Internal types like String、Set and Map should be registered by fury by default and lookup by pattern match to avoid hash overhead.
2. More unit testcases.
3. Support `Box<dyn customTrait> `

### DIFF
--- a/rust/fury-core/src/serializer/string.rs
+++ b/rust/fury-core/src/serializer/string.rs
@@ -38,7 +38,7 @@ impl Serializer for String {
         Ok(context.reader.string(len as usize))
     }
 
-    fn ty(_fury: &Fury) -> i16 {
+    fn get_type_id(_fury: &Fury) -> i16 {
         FieldType::STRING.into()
     }
 }
--- a/rust/fury-core/src/serializer/string.rs
+++ b/rust/fury-core/src/serializer/string.rs
@@ -16,6 +16,7 @@
 // under the License.
 
 use crate::error::Error;
+use crate::fury::Fury;
 use crate::resolver::context::ReadContext;
 use crate::resolver::context::WriteContext;
 use crate::serializer::Serializer;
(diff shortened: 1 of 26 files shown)


rust/fury-core/src/serializer/string.rs
```diff
@@ -37,8 +38,8 @@ impl Serializer for String {
         Ok(context.reader.string(len as usize))
     }
 
-    fn ty() -> FieldType {
-        FieldType::STRING
+    fn ty(_fury: &Fury) -> i16 {
```
chaokunyang, line 41: What does `ty` means?

theweipeng: Yes. but type is a keyword, and type_id is a member of Any which is available for every type. ty is not a good  choise, Maybe we can use fury_type_id

chaokunyang (approved):
LGTM

PR #1797, opened 2024-08-11
feat(Rust): Support serialization and deserialization of trait objects

theweipeng:
### Support serialization and deserialization of trait objects
The relationship between a trait object and a concrete type is generated by a macro and stored in a Fury instance. We can access the serializer and deserializer using the  `Fury type ID`  and  `Rust type ID`.

Use as follows
```Rust
    #[impl_polymorph]
    trait Animal {
        fn get_name(&self) -> String;
    }

    #[derive(Fury, Debug)]
    #[polymorphic_traits("Animal")]
    struct Dog {
        name: String
    }

    impl Animal for Dog {
        fn get_name(&self) -> String {
            self.name.clone()
        }
    }

    #[derive(Fury)]
    struct Person {
        pet: Box<dyn Animal>
    }

    let p = Person {
        pet: Box::new(Dog {
            name: String::from("puppy")
        })
    };
    let mut fury = Fury::default();

    fury.register::<Person>(501);
    fury.register::<Dog>(500);
    let bin = fury.serialize(&p);
    let obj: Person = fury.deserialize(&bin).unwrap();
    assert_eq!(obj.pet.get_name(), "puppy");
```
We should add `impl_polymorph` for trait  which is use for serializing `Box<dyn xxxx>`  and `polymorphic_traits` for struct which is use for generate code to cast upcast concete type to trait object. Note that rust not support upcast trait yet(https://github.com/rust-lang/rust/issues/65991).

### DIFF
--- a/rust/fury-core/src/raw/maybe_trait_object.rs
+++ b/rust/fury-core/src/raw/maybe_trait_object.rs
@@ -1,0 +1,31 @@
+use mem::transmute;
+use std::any::TypeId;
+use std::mem;
+use crate::error::Error;
+
+pub struct MaybeTraitObject {
+    ptr: *const u8,
+    type_id: TypeId
+}
+
+impl MaybeTraitObject {
+    pub fn new<T: 'static>(value: T) -> MaybeTraitObject {
+        let ptr = unsafe { transmute::<Box<T>, *const u8>(Box::new(value)) };
+        let type_id = TypeId::of::<T>();
+        MaybeTraitObject {
+            ptr,
+            type_id
+        }
+    }
+
+    pub fn to_trait_object<T: 'static>(self) -> Result<T, Error> {
+        if self.type_id == TypeId::of::<T>() {
+            Ok(unsafe {
+                *(transmute::<*const u8, Box<T>>(self.ptr))
+            })
+        } else {
+            Err(Error::ConvertToTraitObjectError {})
+        }
+    }
+}
+
--- a/rust/fury-core/src/resolver/class_resolver.rs
+++ b/rust/fury-core/src/resolver/class_resolver.rs
@@ -18,99 +18,86 @@
 use super::context::{ReadContext, WriteContext};
 use crate::error::Error;
 use crate::fury::Fury;
-use crate::serializer::StructSerializer;
+use crate::serializer::Serializer;
 use std::any::TypeId;
 use std::{any::Any, collections::HashMap};
+use crate::raw::maybe_trait_object::MaybeTraitObject;
 
-pub struct Harness {
-    serializer: fn(&dyn Any, &mut WriteContext),
-    deserializer: fn(&mut ReadContext) -> Result<Box<dyn Any>, Error>,
-}
 
-impl Harness {
-    pub fn new(
-        serializer: fn(&dyn Any, &mut WriteContext),
-        deserializer: fn(&mut ReadContext) -> Result<Box<dyn Any>, Error>,
-    ) -> Harness {
-        Harness {
-            serializer,
-            deserializer,
-        }
-    }
-
-    pub fn get_serializer(&self) -> fn(&dyn Any, &mut WriteContext) {
-        self.serializer
-    }
-
-    pub fn get_deserializer(&self) -> fn(&mut ReadContext) -> Result<Box<dyn Any>, Error> {
-        self.deserializer
-    }
-}
+pub type TraitObjectDeserializer = fn(&mut ReadContext) -> Result<MaybeTraitObject, Error>;
 
 pub struct ClassInfo {
     type_def: Vec<u8>,
-    type_id: u32,
+    fury_type_id: u32,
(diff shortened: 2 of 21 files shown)


rust/tests/tests/test_custom_trait.rs
```diff
@@ -0,0 +1,81 @@
...
+use fury_derive::{impl_polymorph, Fury};
+use std::any::{Any, TypeId};
+
+#[impl_polymorph]
+trait Animal {
+    fn get_name(&self) -> String;
+}
+
+#[derive(Fury, Debug)]
+#[polymorphic_traits("Animal")]
```
chaokunyang, line 29: Could we avoid annotate `polymorphic_traits`？

rust/fury-core/src/raw/maybe_trait_object.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+impl MaybeTraitObject {
+    pub fn new<T: 'static>(value: T) -> MaybeTraitObject {
+        let ptr = unsafe { transmute::<Box<T>, *const u8>(Box::new(value)) };
+        let type_id = TypeId::of::<T>();
+        MaybeTraitObject { ptr, type_id }
+    }
+
+    pub fn to_trait_object<T: 'static>(self) -> Result<T, Error> {
```
jiacai2050, line 35: If user don't call this function, then `T` is leaked?

If it's, we can impl Drop for MaybeTraitObject to release memory..

rust/fury-core/src/resolver/class_resolver.rs
```diff
@@ -18,99 +18,78 @@
...
-        self.type_id_map.insert(TypeId::of::<T>(), id);
-        self.serialize_map
-            .insert(id, Harness::new(serializer::<T>, deserializer::<T>));
-        self.class_info_map.insert(TypeId::of::<T>(), class_info);
-    }
-
-    pub fn get_harness_by_type(&self, type_id: TypeId) -> Option<&Harness> {
-        self.get_harness(*self.type_id_map.get(&type_id).unwrap())
+    pub fn get_class_info_by_fury_type(&self, type_id: u32) -> &ClassInfo {
+        let type_id = self.fury_type_id_map.get(&type_id).unwrap();
```
jiacai2050, line 86: Too many unwrap here, how about return a Result?

theweipeng: Yes. The codebase contain too many unwrap, I will fix it one by one.

rust/tests/tests/test_custom_trait.rs
theweipeng: I'll try something else, see if I can get rid of this

rust/fury-core/src/raw/maybe_trait_object.rs
theweipeng: Great catch, I will fixed it.

rust/tests/tests/test_custom_trait.rs
jiacai2050: I think of one solution to this issue.

Before this PR, `dyn Any` is already supported, so solution below should work? Let me know if there is anything I'm missing.

> Suppose the trait already contains a `fn as_any(&self) -> &dyn Any` method, then we can use `trait_object.as_any().type_id()` to get underlying struct's type id, then we don't need this `polymorphic_traits`.

`as_any` solution is a frequently used tricks in rust, so I think adding this requirement is an acceptable choice.

theweipeng: > I think of one solution to this issue.
> 
> Before this PR, `dyn Any` is already supported, so solution below should work? Let me know if there is anything I'm missing.
> 
> > Suppose the trait already contains a `fn as_any(&self) -> &dyn Any` method, then we can use `trait_object.as_any().type_id()` to get underlying struct's type id, then we don't need this `polymorphic_traits`.
> 
> `as_any` solution is a frequently used tricks in rust, so I think adding this requirement is an acceptable choice.

We already supported `dyn Any` , but even if we have the `type id ` of  the underlying struct,   I don't quite understand how to cast to the `dyn object` through type_id.

My understanding of the solution is as follows

```Rust

impl Serializer for Box<dyn Foo> {
    deserialize() {
          let type_id_in_binary = fury.reader.i16();  // We read the type_id from binary, which was received from another app.
          let concrete_type_deserializer = fury.get_deserializer(type_id_in_binary); // We obtain the deserializer that was registered by the user.
          let concrete_instance: Box<dyn Any> = concrete_type_deserializer();     // Call the function pointer, and we get the instance of the underlying struct. The type of the result should be Any, as the function pointer should be compatible with all the deserializers.
         
        concrete_instance as Box<dyn Foo> // We can't do this, as Rust can only cast from primitive types or concrete types. The compiler can't deduce the type of `concrete_instance` at compile time, so it can't create the v-table. 
    }
}

```
 Let me know if there are some misunderstandings

PR #1809, opened 2024-08-17
fix(scala): fix nested type serialization in scala object type

### LINKED ISSUE #1801: Scala: wrong class name in generated java source for class inside nested objects
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

0.7.0

### Component(s)

Java, Other

### Minimal reproduce step

https://github.com/yoohaemin/fury-bugreport/blob/65788b6902939e71aa136959183c13b132861ce6/src/main/scala/Main.scala

You can clone the repo and try `sbt run` at the root of the repo.

serialization round-trip works

### What did you see instead?

generate java source failed to compile. See the linked file above for the message.

### Anything Else?

See: https://stackoverflow.com/questions/30809070/accessing-scala-nested-classes-from-java

This shows the correct format of referring to classes inside nested scala `object`s. In this case, it should be `test.A$B$C`. It generated `test.A.B$.C` (See line 40 of the generated code inside my example).

chaokunyang:
## Related issues

Closes #1801 

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/codegen/CodeGenerator.java
+++ b/java/fury-core/src/main/java/org/apache/fury/codegen/CodeGenerator.java
@@ -496,9 +496,15 @@ public class CodeGenerator {
     if (clz.isPrimitive()) {
       return true;
     }
-    if (clz.getCanonicalName() == null) {
+    try {
+      if (clz.getCanonicalName() == null) {
+        return false;
+      }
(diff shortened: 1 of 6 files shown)


chaokunyang:
image
strange error

theweipeng (approved):

PR #1817, opened 2024-08-26
feat(rust): support enum serialization

### LINKED ISSUE #1393: [Rust] add enum for fury macro derive
as title

jiacai2050:
## What does this PR do?

Support enum serialization for Rust.

## Related issues

Close #1393

### DIFF
--- a/rust/fury-derive/src/object/derive_enum.rs
+++ b/rust/fury-derive/src/object/derive_enum.rs
@@ -1,3 +1,20 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
 use proc_macro2::TokenStream;
 use quote::quote;
 use syn::DataEnum;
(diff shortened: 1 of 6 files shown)


rust/fury-derive/src/object/derive_enum.rs
```diff
@@ -0,0 +1,68 @@
...
+                #(
+                    Self::#variant_idents => {
+                        context.writer.var_int32(#variant_values);
+                    }
+                )*
+            }
+        }
+
+        fn reserved_space() -> usize {
+            4
```
jiacai2050, line 46: I'm not sure if this is right.

rust/fury-derive/src/object/derive_enum.rs
```diff
@@ -0,0 +1,68 @@
...
+// under the License.
+
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::DataEnum;
+
+pub fn gen_type_def(_data_enum: &DataEnum) -> TokenStream {
+    quote! {
+        fn type_def(fury: &fury_core::fury::Fury) -> Vec<u8> {
+            Vec::new()
```
jiacai2050, line 25: Since there is no fields in enum, so we can leave it empty?

rust/fury-derive/src/object/derive_enum.rs
```diff
@@ -0,0 +1,68 @@
...
+    quote! {
+       fn read(
+           context: &mut fury_core::resolver::context::ReadContext,
+       ) -> Result<Self, fury_core::error::Error> {
+           let v = context.reader.var_int32();
+           match v {
+               #(
+                   #variant_values => Ok(Self::#variant_idents),
+               )*
+               _ => panic!("unknown value"),
```
chaokunyang, line 64: Not related to this PR, but maybe it's time to think about how we better handle exceptional data. If we don't handle such dirty data, it's easy to construct data to attack the service

jiacai2050: Agree, the Rust codebase contains too many `unwrap`, we need  to refactor it to return `Result` before more feature request.

rust/fury-derive/src/object/derive_enum.rs
theweipeng: Yes, it is right

rust/fury-derive/src/object/derive_enum.rs
theweipeng: We can levae it empty since enum types do not require compatibility support

theweipeng (approved):
LGTM

PR #1821, opened 2024-08-29
fix(rust): use iter to avoid reallocate

jiacai2050:
## What does this PR do?

Use iterator + collect to avoid re-allocate vec/set/map.


theweipeng (approved):
LGTM

PR #1823, opened 2024-09-03
refactor(rust): simplify error handle

jiacai2050:
## What does this PR do?

Make error easy to use.

In most case, users don't care about error details, so too many fields in enum is hard to write, and hard to use. 

So I refactor it to include an `Other` field to be used as a general Error, most error can be mapped to it directly.

`Ref` is a special case, so I leave it as it's now.


theweipeng (approved):
LGTM

Issue #1842, opened 2024-09-13
[Java] Maximum size of Java arrays is not correct


fink-arthur:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

8.1.0

### Component(s)

Java

### Minimal reproduce step

`MemoryBuffer memoryBuffer = MemoryBuffer.newHeapBuffer(10);`
 `memoryBuffer.grow(Integer.MAX_VALUE);`

### What did you expect to see?

A MemoryBuffer of the maximum size

### What did you see instead?

`java.lang.OutOfMemoryError: Requested array size exceeds VM limit`

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #1843, opened 2024-09-13
fix(java): Fix for maximum size of java arrays

fink-arthur:
## What does this PR do?

Fixes the maximum size of Java arrays using Integer.MAX_VALUE when it should be Integer.MAX_VALUE - 8.
See this https://github.com/openjdk/jdk14u/blob/84917a040a81af2863fddc6eace3dda3e31bf4b5/src/java.base/share/classes/jdk/internal/util/ArraysSupport.java#L577 or https://www.baeldung.com/java-arrays-max-size

## Related issues

- #1842

No

- [ ] Does this PR introduce any public API change? No
- [ ] Does this PR introduce any binary protocol compatibility change? No


chaokunyang (approved):
LGTM, thanks

PR #1862, opened 2024-09-27
ban these sbt commands

pjfanning:
Releases of jars to Maven Central should be finished using repository.apache.org UI.

These sbt commands are still allowed:
* sonatypePrepare
* sonatypeBundleUpload
* sonatypeClose


chaokunyang (approved):
LGTM

PR #1877, opened 2024-10-09
feat(kotlin): Introduce kotlin package with stdlib collections and tests

wywen:
## What does this PR do?

This PR adds a base project for Kotlin on the JVM.
- It adds a pom.xml based on the java project.
- Adds a Serializer class that registers base collection classes not covered by the JVM.
- Adds code to test all documented collections in the kotlin stdlib via TestNG.
- Adds register and adds serializers for EmptyList, EmptySet, EmptyMap classes.
- This PR does not handle kotlin `withDefault` wrapper classes, as that requires serializing arbitrary lambdas.

## Related issues

- #683

This PR creates a new fury-kotlin JAR, no documentation yet from the main project.

## Benchmark
N/A


kotlin/README.md
```diff
@@ -0,0 +1,7 @@
+# Apache Fury™ Kotlin
+
+Provides Fury support for Kotlin Serialization on JVM:
```
chaokunyang, line 3: Could we clear that Fury has already support most kotlin types well, it's easy for users to think that fury only support those types

kotlin/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+- stdlib `collection`: `ArrayDeque`, `ArrayList`, `HashMap`,`HashSet`, `LinkedHashSet`, `LinkedHashMap`.
```
chaokunyang, line 5: Seems kotlin `ArrayList` is java `ArrayList`, can we clear classes like this?

wywen: OK! I've split the tested with and implementation section that hopefully clears up the issue!

I plan to add tests for the kotlin primitives. I think out of the box most of them are the same as Java, except for the unsigned types. Then in no particular order I plan to test Array, String, Enum, data class(kotlin equivalent to record) and out of the box stdlib types such as `Pair`, `Triple`.

chaokunyang:
Could you add header for following types?
image

kotlin/README.md
```diff
@@ -0,0 +1,21 @@
+# Apache Fury™ Kotlin
+
+This provides additional Fury support for Kotlin Serialization on JVM:
+
+Most standard kotlin types are already supported out of the box with the default java implementation.
```
chaokunyang, line 5:
```suggestion
Most standard kotlin types are already supported out of the box with the default Fury java implementation.
```

kotlin/README.md
chaokunyang: unsigned types are a java object type instead of primitive type in java. We may need to add supports for those types in future, otherwise the serialization for unsigned types will be slow

wywen:
> Could you add header for following types? image

Done!

chaokunyang (approved):
LGTM, thanks for this awesome work.
We can add serializers for kotlint Range, unsigned int in later PRs

PR #1885, opened 2024-10-14
feat(ci): support building python on windows

### LINKED ISSUE #798: [C++] Windows build support
Fury support macos and linux build, but windows is not supported.

Fury use bazel for building c++ code, windows building need some extra work.

An-DJ:
## What does this PR do?

Support building python on Windows.

1. Use (git) bash to run python steps on Windows (MYSY, MINGW)
2. Add missing headers
3. Rename FuryLogLevel::ERROR to FuryLogLevel::ERR. I don't know why this enum make build failed, but it is successed if renaming it to FuryLogLevel::ERR
4. Rename pyx built dynamic lib name '*.so' to ['*.pyd'](https://docs.python.org/3.10/faq/windows.html#is-a-pyd-file-the-same-as-a-dll)
5. Lock bazel version to `6.3.2` on Windows
6. Unify Python type `Integer` memoryview formats (typecodes) on different operating systems ([ref](https://docs.python.org/3/library/array.html))

|                 | int16 | int32 | int64 |
|-----------------|-------|-------|-------|
| Linux(64-bit)   | 'h'   | 'i'   | 'l'   |
| MacOS(64-bit)   | 'h'   | 'i'   | 'l'   |
| Windows(64-bit) | 'h'   | 'l'   | 'q'   |

## Related issues

Close #798 


chaokunyang:
Hi @An-DJ , it this PR ready for review?

An-DJ:
> Hi @An-DJ , it this PR ready for review?

No, CI result is fake, the build script didn't have been executed successfully. We need more works on windows platform...

I will continue this PR in the next few days (too busy in this week...)

An-DJ:
@chaokunyang This UT is strange.

```
@pytest.mark.parametrize("language", [Language.XLANG, Language.PYTHON])
def test_array_serializer(language):
    fury = Fury(language=language, ref_tracking=True, require_class_registration=False)
    for typecode in PyArraySerializer.typecode_dict.keys():
        arr = array.array(typecode, list(range(10)))

        # additional code to print array attribute
        print(arr.itemsize, arr.typecode)
        # end

        assert ser_de(fury, arr) == arr
    for dtype in Numpy1DArraySerializer.dtypes_dict.keys():
        arr = np.array(list(range(10)), dtype=dtype)
        new_arr = ser_de(fury, arr)
        assert np.array_equal(new_arr, arr)
        np.testing.assert_array_equal(new_arr, arr)
```

I test it on Ubuntu 20.04 and Windows 10, but there are different result:

```
# Ubuntu

pyfury/tests/test_serializer.py::test_array_serializer[Language.XLANG] 2 h
4 i
8 l
4 f
8 d
PASSED
pyfury/tests/test_serializer.py::test_array_serializer[Language.PYTHON] 2 h
4 i
8 l
4 f
8 d
PASSED
```

```
# Windows

=================== 1 failed, 1 passed, 3 warnings in 0.88s ===================
FAILED         [ 50%]
2 h
4 i
4 l
```

The point is that `l` type could be 4 or 8 bit on different platform.

In addition, in `_util.pyx`, there are also failed converted between `const int64_t` and `long`:

```
cdef inline uint8_t* get_address(v):
    view = memoryview(v)
    cdef str dtype = view.format
    cdef:
       # ...
        const int64_t[:] signed_long_data
       # ...
        uint8_t* ptr
    if dtype == "b":
        signed_char_data = v
        ptr = <uint8_t*>(&signed_char_data[0])
    elif dtype == "l":  # here
        signed_long_data = v
        ptr = <uint8_t*>(&signed_long_data[0])
    else:
        raise Exception(f"Unsupported buffer of type {type(v)} and format {dtype}")
    return ptr
```

Could you plz give me some advice?

chaokunyang:
Is Windows a bid endian machine?

An-DJ:
> Is Windows a big endian machine?

You means the github windows-2022 ci runner is hosted on a big endian machine?

No. According to the [related document](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#cloud-hosts-for-github-hosted-runners), for Linux and Windows runners, GitHub uses Dadsv5-series virtual machines. For more information, see [Dasv5 and Dadsv5-series](https://learn.microsoft.com/en-us/azure/virtual-machines/dasv5-dadsv5-series#dadsv5-series) in the Microsoft Azure documentation.

image


`AMD EPYC 7763v (Genoa) [x86-64]` support small endian only.

An-DJ:
@chaokunyang Actually we check the python array initialization method  `array.array` with different dtype, but each type code has its `Mininum size in bytes`, not a fixed size. 

IMHO it is the reason why it behaves differently on UNIX and Windows.

Ref: https://docs.python.org/3/library/array.html

image

We should find a strict way to fix each dtype size... right?

chaokunyang:
Yeah, you are right. Fury take `l` as 4 bytes, but it may be 8 bytes on windows. In such cases, we may need to expand every item from 4 bytes value to 8 bytes in a loop manually.

An-DJ:
Hi @chaokunyang 

Recently I have searched and found that windows(MSVC) standards basic [data type range](https://learn.microsoft.com/en-us/cpp/cpp/data-type-ranges?view=msvc-170). Also in [intel docs](https://www.intel.com/content/www/us/en/developer/articles/technical/size-of-long-integer-type-on-different-architecture-and-os.html).

Unfortunately, `long`, which used as `l` in Python, is stricted in 4 bytes on Windows, but 8 bytes on *nix. That is a gap between *nix and Windows.

So Apache Fury support `q` (in another words, `signed long long`) ONLY maybe a better choice. That type is 8 bytes size in both *nix and Windows.

I am not fimilair with Python array.array() users' habit, but this approach could be a simple solution to universe the memory model in Apache Fury Python support.

What do you think of that?

chaokunyang:
Great idea, I think we can use `q (signed long long)` instead of `l`.

chaokunyang:
> Great idea, I think we can use `q (signed long long)` instead of `l`. And raise Error when `l` array format type are passed.

cpp/fury/util/logging.h
```diff
@@ -35,7 +35,7 @@ enum class FuryLogLevel {
   DEBUG = -1,
   INFO = 0,
   WARNING = 1,
-  ERROR = 2,
+  ERR = 2,
```
PragmaTwice, line 38: Could you attach the error you face of the original `ERROR` identifier?

PragmaTwice: Hmm it seems here: https://github.com/apache/fury/actions/runs/11643078405/job/32423399925#step:5:1894

PragmaTwice: Just took a glance, and I think that maybe the Cython generated code defines a macro named `ERROR` and it's included before this header so that weird thing happens. Not familiar with related code here, maybe @penguin-wwy has ideas about this.

An-DJ:
Hi @chaokunyang

There is a friendly github action that could simplify the bazel installization. ⬇️

https://github.com/marketplace/actions/setup-bazel-environment

But I found that the Apache organization manages a allowed action namelist.

Could you plz add that action to the namelist? Then we can easily install bazel 6.3.2 on Windows.

Ref:
https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#allowing-select-actions-and-reusable-workflows-to-run

.python-version
```diff
@@ -0,0 +1 @@
+3.11.0
```
chaokunyang, line 1: is this committed to repo by accident?

An-DJ: > is this committed to repo by accident?

Deleted.

cpp/fury/util/logging.h
chaokunyang: from chatgpt:

MSVC and Macros: The MSVC environment automatically defines some macros that can interfere with your code. One such macro is ERROR, which is commonly defined within Windows headers or other libraries. When you use ERROR as a name for an enumeration or variable in your code, it collides with this pre-existing macro definition, causing syntax errors.


We can use ERR here, for better compatibility, maybe we should define the enum value with FURY prefix:
```c++

enum class FuryLogLevel {
  FURY_DEBUG = -1,
  FURY_INFO = 0,
  FURY_WARNING = 1,
  FURY_ERROR = 2,
  FURY_FATAL = 3
};
```

python/pyfury/_util.pyx
```diff
@@ -29,10 +29,13 @@ from libcpp cimport bool as c_bool
 from pyfury.includes.libutil cimport(
     CBuffer, AllocateBuffer, GetBit, SetBit, ClearBit, SetBitTo, CStatus, StatusCode, utf16HasSurrogatePairs
 )
+import os
 
 cdef int32_t max_buffer_size = 2 ** 31 - 1
 cdef int UTF16_LE = -1
 
+_WINDOWS = os.name == 'nt'
```
chaokunyang, line 37:
```suggestion
cdef c_bool _WINDOWS = os.name == 'nt'
```

chaokunyang (approved):
Looks great!

cpp/fury/util/logging.h
An-DJ: Good idea. Restyle it in #2016 . : )

PR #1890, opened 2024-10-18
feat(java): Refactor String serialization and deserialization

### LINKED ISSUE #1754: [Java] implement fast utf16 to utf8 conversion
Currently Fury use `java.lang.StringCoding#encode(java.nio.charset.Charset, char[], int, int)` to convert  utf16 to utf8.
```java
  static byte[] encode(Charset cs, char[] ca, int off, int len) {
        CharsetEncoder ce = cs.newEncoder();
        int en = scale(len, ce.maxBytesPerChar());
        byte[] ba = new byte[en];
        if (len == 0)
            return ba;
        boolean isTrusted = false;
        if (System.getSecurityManager() != null) {
            if (!(isTrusted = (cs.getClass().getClassLoader0() == null))) {
                ca =  Arrays.copyOfRange(ca, off, off + len);
                off = 0;
            }
        }
        ce.onMalformedInput(CodingErrorAction.REPLACE)
          .onUnmappableCharacter(CodingErrorAction.REPLACE)
          .reset();
        if (ce instanceof ArrayEncoder) {
            int blen = ((ArrayEncoder)ce).encode(ca, off, len, ba);
            return safeTrim(ba, blen, cs, isTrusted);
        } else {
            ByteBuffer bb = ByteBuffer.wrap(ba);
            CharBuffer cb = CharBuffer.wrap(ca, off, len);
            try {
                CoderResult cr = ce.encode(cb, bb, true);
                if (!cr.isUnderflow())
                    cr.throwException();
                cr = ce.flush(bb);
                if (!cr.isUnderflow())
                    cr.throwException();
            } catch (CharacterCodingException x) {
                throw new Error(x);
            }
            return safeTrim(ba, bb.position(), cs, isTrusted);
        }
    }
```

This invoke `sun.nio.cs.UTF_8.Encoder#encode`:
```java
        public int encode(char[] sa, int sp, int len, byte[] da) {
            int sl = sp + len;
            int dp = 0;
            int dlASCII = dp + Math.min(len, da.length);

            // ASCII only optimized loop
            while (dp < dlASCII && sa[sp] < '\u0080')
                da[dp++] = (byte) sa[sp++];

            while (sp < sl) {
                char c = sa[sp++];
                if (c < 0x80) {
                    // Have at most seven bits
                    da[dp++] = (byte)c;
                } else if (c < 0x800) {
                    // 2 bytes, 11 bits
                    da[dp++] = (byte)(0xc0 | (c >> 6));
                    da[dp++] = (byte)(0x80 | (c & 0x3f));
                } else if (Character.isSurrogate(c)) {
                    if (sgp == null)
                        sgp = new Surrogate.Parser();
                    int uc = sgp.parse(c, sa, sp - 1, sl);
                    if (uc < 0) {
                        if (malformedInputAction() != CodingErrorAction.REPLACE)
                            return -1;
                        da[dp++] = repl;
                    } else {
                        da[dp++] = (byte)(0xf0 | ((uc >> 18)));
                        da[dp++] = (byte)(0x80 | ((uc >> 12) & 0x3f));
                        da[dp++] = (byte)(0x80 | ((uc >>  6) & 0x3f));
                        da[dp++] = (byte)(0x80 | (uc & 0x3f));
                        sp++;  // 2 chars
                    }
                } else {
                    // 3 bytes, 16 bits
                    da[dp++] = (byte)(0xe0 | ((c >> 12)));
                    da[dp++] = (byte)(0x80 | ((c >>  6) & 0x3f));
                    da[dp++] = (byte)(0x80 | (c & 0x3f));
                }
            }
            return dp;
        }

```

This implementation is not effficient enough, we need a faster one.

Most chars in utf16 string are latin1 chars, we can calculate an approximate chars stats using SIMD or super word, then dispatch to different codepath:

```
int latin1Chars, len;
if (latin1Chars * 2 >= len) {
  // Check and write 8 bytes every time 
  fastUtf8Encode(utf16Bytes);
} else {
  // check byte by byte every time
  utf8Encode(utf16Bytes);
}
```

Multiple encoding strategy:
- If latin1 chars is greater than 50%, use this optimized utf16 encoding
- If latin1 chars is less than 50%, use utf16 encoding

Add a new Fury Config option such as `StringEncoding` enum:
```java
enum StingEncoding {
  NATIVE_STRING,
  LATIN1_UTF16,
  LATIN1_UTF8_16,
}
```
Users can set string encoding in FuryBuilder

### LINKED ISSUE #1868: [Question] large string has bigger serialized size
### Question

自定义类型对象有字段string，该字段是一个大文本，发现该对象序列化出来的体积比JDK序列化要大很多，如何解决？

HuangXingBo:
## Related issues

Closes #1868
Closes #1754

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
@@ -21,6 +21,7 @@ package org.apache.fury.serializer;
 
 import static org.apache.fury.type.TypeUtils.PRIMITIVE_CHAR_ARRAY_TYPE;
 import static org.apache.fury.type.TypeUtils.STRING_TYPE;
+import static org.apache.fury.util.StringUtils.MULTI_CHARS_NON_LATIN_MASK;
 
 import java.lang.invoke.CallSite;
 import java.lang.invoke.LambdaMetafactory;
@@ -43,6 +44,7 @@ import org.apache.fury.reflect.ReflectionUtils;
 import org.apache.fury.type.Type;
 import org.apache.fury.util.MathUtils;
 import org.apache.fury.util.Preconditions;
+import org.apache.fury.util.StringEncodingUtils;
 import org.apache.fury.util.StringUtils;
 import org.apache.fury.util.unsafe._JDKAccess;
 
@@ -149,37 +151,24 @@ public final class StringSerializer extends ImmutableSerializer<String> {
   public Expression writeStringExpr(Expression strSerializer, Expression buffer, Expression str) {
     if (isJava) {
       if (STRING_VALUE_FIELD_IS_BYTES) {
-        return new StaticInvoke(StringSerializer.class, "writeBytesString", buffer, str);
+        if (compressString) {
+          return new Invoke(strSerializer, "writeCompressedBytesString", buffer, str);
+        } else {
+          return new StaticInvoke(StringSerializer.class, "writeBytesString", buffer, str);
+        }
       } else {
         if (!STRING_VALUE_FIELD_IS_CHARS) {
           throw new UnsupportedOperationException();
         }
         if (compressString) {
-          return new Invoke(strSerializer, "writeCharsStringCompressed", buffer, str);
+          return new Invoke(strSerializer, "writeCompressedCharsString", buffer, str);
         } else {
-          return new Invoke(strSerializer, "writeCharsStringUncompressed", buffer, str);
+          return new StaticInvoke(StringSerializer.class, "writeCharsString", buffer, str);
         }
       }
     } else {
       return new Invoke(strSerializer, "writeUTF8String", buffer, str);
     }
-  }
-
-  // Invoked by jit
-  public void writeCharsStringCompressed(MemoryBuffer buffer, String value) {
-    final char[] chars = (char[]) Platform.getObject(value, STRING_VALUE_FIELD_OFFSET);
-    if (StringUtils.isLatin(chars)) {
-      writeCharsLatin(buffer, chars, chars.length);
-    } else {
-      writeCharsUTF16(buffer, chars, chars.length);
-    }
-  }
-
-  // Invoked by jit
-  public void writeCharsStringUncompressed(MemoryBuffer buffer, String value) {
-    int numBytes = MathUtils.doubleExact(value.length());
-    final char[] chars = (char[]) Platform.getObject(value, STRING_VALUE_FIELD_OFFSET);
-    buffer.writePrimitiveArrayWithSize(chars, Platform.CHAR_ARRAY_OFFSET, numBytes);
   }
 
   public String readString(MemoryBuffer buffer) {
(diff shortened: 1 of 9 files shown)


chaokunyang:
The code in StringSerializer started to grow a lot, maybe we could move UTF8/16 encoding to a new file `org.apache.fury.serializer.StringUTF8`

chaokunyang:
And maybe we should provide a StringEncoding option too:
```
enum StingEncoding {
  NATIVE_STRING,
  LATIN1_UTF16,
  LATIN1_UTF8_16,
}
```
Sometimes big serialized binary is accepptable to users

chaokunyang:
Another thing is that `org.apache.fury.serializer.StringSerializer#writeStringExpr/readStringExpr` should also be updated to pick up right encoding

java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
```diff
@@ -609,6 +638,95 @@ public void writeUTF8String(MemoryBuffer buffer, String value) {
...
+      } else {
+        if (bestCoder(bytes) == UTF8) {
+          writeBytesUTF8(buffer, bytes);
+        } else {
+          writeBytesUTF16(buffer, bytes);
+        }
+      }
+    } else {
+      final char[] chars = (char[]) Platform.getObject(value, STRING_VALUE_FIELD_OFFSET);
+      if (StringUtils.isLatin(chars)) {
```
chaokunyang, line 720: Maybe we could check latin1 in bestCoder for chars

java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
```diff
@@ -625,4 +743,534 @@ public String readUTF8String(MemoryBuffer buffer) {
...
+  private static boolean fastDecodeUTF8(byte[] src, int offset, int len, byte[] dst) {
+    final int end = offset + len;
+    int dp = 0;
+
+    while (offset < end) {
+      if (offset + 8 <= end
+          && (Platform.getLong(src, Platform.BYTE_ARRAY_OFFSET + offset) & 0x8080808080808080L)
+              == 0) {
+        // ascii only
+        for (int i = 0, pos = Platform.IS_LITTLE_ENDIAN ? dp : dp + 1; i < 8; ++i, pos += 2) {
```
chaokunyang, line 999: Could we try to unroll this loop into 8 lines of code and test whether will it be faster

java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
```diff
@@ -625,4 +743,534 @@ public String readUTF8String(MemoryBuffer buffer) {
...
+
+  private static boolean fastDecodeUTF8(byte[] src, int offset, int len, char[] dst) {
+    int end = offset + len;
+    int dp = 0;
+    while (offset < end) {
+      if (offset + 8 <= end
+          && (Platform.getLong(src, Platform.BYTE_ARRAY_OFFSET + offset) & 0x8080808080808080L)
+              == 0) {
+        // ascii only
+        for (int i = 0; i < 8; ++i) {
```
chaokunyang, line 1102: ditto for loop unroll

java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
```diff
@@ -609,6 +638,95 @@ public void writeUTF8String(MemoryBuffer buffer, String value) {
...
+    int endOffset =
+        Math.min(
+            Platform.BYTE_ARRAY_OFFSET + vectorizedBytes, Platform.BYTE_ARRAY_OFFSET + sampleNum);
+    int count = 0;
+    for (int offset = Platform.BYTE_ARRAY_OFFSET; offset < endOffset; offset += 8) {
+      long multiChars = Platform.getLong(bytes, offset);
+      if ((multiChars & MULTI_CHARS_NON_LATIN_MASK) == 0) {
+        count += 4;
+      } else {
+        for (int i = Platform.IS_LITTLE_ENDIAN ? 1 : 0; i < 8; ++i) {
```
chaokunyang, line 689: I'm not sure whether unrolling will be faster, maybe need a benchmark. One drawback is unrolling will make this method have bigger bytecode size. But if it's faster, we could afford that.

java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
HuangXingBo: Yes. it is a good suggestion.

java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
HuangXingBo: Yes, I checked the assembly code and it is indeed not loop-unrolled.

HuangXingBo:
@chaokunyang Thanks a lot for the comments. I think it is sufficient to continue using the `isCompress` configuration, which is relatively easy to understand for users. Introducing a new encoding configurations are difficult for users to understand. I will make changes based on the comments and submit a formal PR.

java/fury-core/src/main/java/org/apache/fury/util/StringEncodingUtils.java
```diff
@@ -0,0 +1,362 @@
...
+        charOffset += 4;
+      } else {
+        char c = src[charOffset++];
+        if (c < 0x80) {
+          dst[dp++] = (byte) c;
+        } else if (c < 0x800) {
+          dst[dp] = (byte) (0xc0 | (c >> 6));
+          dst[dp + 1] = (byte) (0x80 | (c & 0x3f));
+          dp += 2;
+        } else if (c >= '\uD800' && c <= Character.MAX_LOW_SURROGATE) {
```
chaokunyang, line 51: How big `convertUTF16ToUTF8`?  Can this method be inlined to its caller?

chaokunyang: Since this method is called only when there are many ascii chars, this branch is less frequent. If this big method can't be inlined, we can move this branch into a new method.

chaokunyang: I'm not sure which will be faster, maybe need some benchmark. Could you use MediaContent for benchmark?

java/fury-core/src/main/java/org/apache/fury/util/StringEncodingUtils.java
```diff
@@ -0,0 +1,362 @@
...
+package org.apache.fury.util;
+
+import static org.apache.fury.util.StringUtils.MULTI_CHARS_NON_LATIN_MASK;
+
+import org.apache.fury.memory.Platform;
+
+/** String Encoding Utils. */
+public class StringEncodingUtils {
+
+  public static int convertUTF16ToUTF8(char[] src, byte[] dst, int dp) {
```
chaokunyang, line 29: Maybe we need to add a javadoc to document that this method will be faster only when there are many assci chars in the string. Acutally I'm curious whether this method will be faster than JDK native implementation for other string

java/fury-core/src/main/java/org/apache/fury/util/StringEncodingUtils.java
```diff
@@ -0,0 +1,362 @@
...
+
+package org.apache.fury.util;
+
+import static org.apache.fury.util.StringUtils.MULTI_CHARS_NON_LATIN_MASK;
+
+import org.apache.fury.memory.Platform;
+
+/** String Encoding Utils. */
+public class StringEncodingUtils {
+
```
chaokunyang, line 28:
```suggestion
A fast encoding implementation to convert a utf16 encoded string into utf8 encoded binary when most chars in the string are ascii chars. 
```

java/fury-core/src/main/java/org/apache/fury/util/StringEncodingUtils.java
```diff
@@ -0,0 +1,362 @@
...
+          dst[dp] = (byte) (0xe0 | ((c >> 12)));
+          dst[dp + 1] = (byte) (0x80 | ((c >> 6) & 0x3f));
+          dst[dp + 2] = (byte) (0x80 | (c & 0x3f));
+          dp += 3;
+        }
+      }
+    }
+    return dp;
+  }
+
```
chaokunyang, line 77:
```suggestion
 
   /**
  * A fast decoding implementation to convert a utf8 encoded string into utf16 encoded 
  *  string when most chars in the string are ascii chars. 
  */
```

java/fury-core/src/main/java/org/apache/fury/util/StringEncodingUtils.java
HuangXingBo: The overall algorithm idea implemented by this method is based on the scalar version of simdutf, and the non-ascillary part is based on the implementation of fastjson2（The speedup ratio is twice of JDK）. Even in the all-Chinese scenario, it will be faster than the JDK implementation.

java/fury-core/src/main/java/org/apache/fury/util/StringEncodingUtils.java
HuangXingBo: This branch can be extracted, but I think for this kind of for loop with complex logic, iniline or not is not so critical, and even forcing inline of overly complex code can be a side effect.

chaokunyang: I see, thanks

java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
```diff
@@ -603,26 +722,121 @@ private static MethodHandle getJavaStringZeroCopyCtrHandle() {
...
+
+    for (int i = vectorizedChars; i < sampleNum; i++) {
+      if (chars[i] < 0x80) {
+        count++;
+      }
+    }
+
+    // ascii number > 50%, choose UTF-8
+    if (count >= sampleNum * 0.5) {
+      if (count == sampleNum && StringUtils.isLatin(chars)) {
```
chaokunyang, line 781: Could we use `StringUtils.isLatin(chars, sampleNum)`  instead?

java/fury-core/src/main/java/org/apache/fury/serializer/StringSerializer.java
```diff
@@ -235,75 +214,106 @@ public String readBytesString(MemoryBuffer buffer) {
...
+        return readCharsString(buffer);
       }
     }
   }
 
+  public void writeCompressedBytesString(MemoryBuffer buffer, String value) {
+    final byte[] bytes = (byte[]) Platform.getObject(value, STRING_VALUE_FIELD_OFFSET);
+    final byte coder = Platform.getByte(value, Offset.STRING_CODER_FIELD_OFFSET);
+    if (coder == LATIN1 || bestCoder(bytes) == UTF16) {
+      writeBytesString(buffer, value);
```
chaokunyang, line 310: Maybe we should make `writeBytesString` receive bytes and coder, and invoke like `writeBytesString(buffer, coder, bytes)`

HuangXingBo:
my private GHA has passed. https://github.com/HuangXingBo/fury/actions/runs/11442511611

chaokunyang (approved):
Looks awesome! Thank you so much!

PR #1891, opened 2024-10-18
feat(kotlin): Add unsigned array support and tests for arrays and strings

wywen:
## What does this PR do?

This PR adds tests for serializing/deserializing:
- Strings (same as Java)
- Primitive arrays in Kotlin (same as Java)
- Array<T> in kotlin (same as Java)
- Unsigned arrays in kotlin - `UByteArray`, `UIntArray`, `UShortArray`, `ULongArray`

Unsigned arrays in kotlin are currently marked experimental, and are subject to API changes (hence the annotations needed to suppress those warnings).

These types are implemented as a view over the signed arrays e.g. UByteArray is a view over ByteArray with contents reinterpreted as UByte, so serializers. The current implementation delegate to existing serializers for corresponding signed types.

The xlang type id is set to LIST for unsigned types.

## Related issues

#683

Yes. Unsigned primitives no longer need to be registered for `fury-kotlin`.

## Benchmark
N/A

### DIFF
--- a/kotlin/src/main/kotlin/org/apache/fury/serializer/kotlin/UnsignedArraySerializers.kt
+++ b/kotlin/src/main/kotlin/org/apache/fury/serializer/kotlin/UnsignedArraySerializers.kt
@@ -32,6 +32,11 @@ abstract class AbstractDelegatingArraySerializer<T, T_Delegate>(
     private val delegateClass: Class<T_Delegate>
 ) : Serializer<T> (fury, cls) {
 
+    // Lazily initialize the delegatingSerializer here to avoid lookup cost.
+    private val delegatingSerializer by lazy {
+        fury.classResolver.getSerializer(delegateClass)
+    }
+
     abstract fun toDelegateClass(value: T): T_Delegate
 
     abstract fun fromDelegateClass(value: T_Delegate): T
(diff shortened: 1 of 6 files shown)


kotlin/src/main/kotlin/org/apache/fury/serializer/kotlin/UnsignedArraySerializers.kt
```diff
@@ -0,0 +1,109 @@
...
+    override fun xwrite(buffer: MemoryBuffer, value: T) {
+        write(buffer, value)
+    }
+
+    override fun xread(buffer: MemoryBuffer): T {
+        return read(buffer)
+    }
+
+    override fun write(buffer: MemoryBuffer, value: T) {
+        val delegatingSerializer = fury.classResolver.getSerializer(delegateClass)
```
chaokunyang, line 52: Could we make this as a field of the DelegatingArraySerializer to reduce lookup cost?

chaokunyang: All those unsiged array type are final I guess

chaokunyang (approved):

kotlin/src/main/kotlin/org/apache/fury/serializer/kotlin/UnsignedArraySerializers.kt
chaokunyang: The lazy can be removed too

PR #1892, opened 2024-10-20
feat(java): add option to treat enum as string

orisgarno:
### Current implementation
For now fury read/write enum with ordinal approach by default. So when serializer serialize as 1, deserializer will return the 2nd enum (index 1)

#### For example :
non updated library on serializer:
```
enum SearchMode {
  CHEAPEST, EARLIEST
}
```

updated library on deserializer:

```
enum SearchMode {
  RECOMMENDED, CHEAPEST, EARLIEST
}
```
- if serializer serialize CHEAPEST, deserializer will return RECOMMENDED.

___
### New Option
This PR will allow fury to treat enum as string so it can have more flexibility like JSON serializer. 
#### For example:
non updated library on serializer:
```
enum SearchMode {
  CHEAPEST, EARLIEST
}
```
updated library on deserializer:
```
enum SearchMode {
  RECOMMENDED, CHEAPEST, EARLIEST
}
```
updated library with different class on deserializer:
```
enum SearchModeV2 {
  RECOMMENDED, CHEAPEST, EARLIEST
}
``` 
- if serializer serialize CHEAPEST, deserializer will return CHEAPEST.
- if serializer serialize CHEAPEST, deserializer for different class will still return CHEAPEST.


java/fury-core/src/main/java/org/apache/fury/config/Config.java
```diff
@@ -60,6 +60,7 @@ public class Config implements Serializable {
   private final boolean scalaOptimizationEnabled;
   private transient int configHash;
   private final boolean deserializeNonexistentEnumValueAsNull;
+  private final boolean treatEnumAsString;
```
chaokunyang, line 63: could we rename it as `serializeEnumByName` instead to keep more consistent naming style?

java/fury-core/src/main/java/org/apache/fury/serializer/EnumSerializer.java
```diff
@@ -20,40 +20,69 @@
...
+    }
   }
 
   @Override
   public Enum read(MemoryBuffer buffer) {
-    int value = buffer.readVarUint32Small7();
-    if (value >= enumConstants.length) {
-      return handleNonexistentEnumValue(value);
+    if (fury.getConfig().treatEnumAsString()) {
+      String metaStringBytes = META_STRING_RESOLVER.readMetaString(buffer);
```
chaokunyang, line 73: We could read and return MetaStringBytes directly and use it as the map key to avoid create a String object every time

java/fury-core/src/main/java/org/apache/fury/serializer/EnumSerializer.java
```diff
@@ -20,40 +20,69 @@
...
 import org.apache.fury.Fury;
 import org.apache.fury.memory.MemoryBuffer;
+import org.apache.fury.meta.MetaString;
+import org.apache.fury.resolver.MetaStringResolver;
 import org.apache.fury.util.Preconditions;
 
 @SuppressWarnings("rawtypes")
-public final class EnumSerializer extends ImmutableSerializer<Enum> {
+public class EnumSerializer extends ImmutableSerializer<Enum> {
+  private static final MetaStringResolver META_STRING_RESOLVER = new MetaStringResolver();
```
chaokunyang, line 33: could we use `fury.getMetaStringResolver` instead?

java/fury-core/src/main/java/org/apache/fury/serializer/EnumSerializer.java
```diff
@@ -20,40 +20,69 @@
...
   private final Enum[] enumConstants;
+  private final Map<String, Enum> enumStringRepresentation;
```
chaokunyang, line 35: If key is MetaStringBytes, could you use IdentityMap instead?

orisgarno: Hi @chaokunyang , thank you for your review.
I have tried to use IdentityMap as you mentioned above, but somehow it fail to match the key. do i missed something here?
image

java/fury-core/src/main/java/org/apache/fury/serializer/EnumSerializer.java
```diff
@@ -40,20 +50,54 @@ public EnumSerializer(Fury fury, Class<Enum> cls) {
...
+      int initialCapacity = (int) Math.ceil(enumConstants.length / 0.99f);
+
+      metaStringtoEnumRepresentation = new HashMap<>(initialCapacity, 0.99f);
+
+      for (Enum enumConstant : enumConstants) {
+        if (enumConstant != null) {
+          MetaString ms =
+              Encoders.GENERIC_ENCODER.encode(enumConstant.name(), MetaString.Encoding.UTF_8);
+          MetaStringBytes msb = metaStringResolver.getOrCreateMetaStringBytes(ms);
+          metaStringtoEnumRepresentation.put(msb, enumConstant);
```
chaokunyang, line 65: We could also create an array `MetaStringBytes[]` whose index are enum oridinal

java/fury-core/src/main/java/org/apache/fury/serializer/EnumSerializer.java
```diff
@@ -40,20 +50,54 @@ public EnumSerializer(Fury fury, Class<Enum> cls) {
...
+    } else {
+      metaStringtoEnumRepresentation = null;
+    }
   }
 
   @Override
   public void write(MemoryBuffer buffer, Enum value) {
-    buffer.writeVarUint32Small7(value.ordinal());
+    if (fury.getConfig().serializeEnumByName()) {
+      MetaString enumMetaString =
```
chaokunyang, line 76: If we create an array MetaStringBytes[] whose index are enum oridinal, we could get MetaStringBytes just from the oridinal. maybe this will make IdentifyMap work

orisgarno: do you mean like this?
it's still not able to find the key for identity map.
im no expert, but seems like because it doesnt refer to the same object while checking the key.
while it works if we use HashMap.

```
      metaStringBytesArr = new MetaStringBytes[enumConstants.length];
      for (Enum enumConstant : enumConstants) {
        if (enumConstant != null) {
          MetaString ms =
              Encoders.GENERIC_ENCODER.encode(enumConstant.name(), MetaString.Encoding.UTF_8);
          MetaStringBytes msb = metaStringResolver.getOrCreateMetaStringBytes(ms);
          metaStringtoEnumRepresentation.put(msb, enumConstant);
          metaStringBytesArr[enumConstant.ordinal()] = msb;
        }
      }
```
```
  @Override
  public void write(MemoryBuffer buffer, Enum value) {
    if (fury.getConfig().serializeEnumByName()) {
      MetaStringBytes metaStringBytes = metaStringBytesArr[value.ordinal()];
      metaStringResolver.writeMetaStringBytes(buffer, metaStringBytes);
    } else {
      buffer.writeVarUint32Small7(value.ordinal());
    }
  }
```
```
  @Override
  public Enum read(MemoryBuffer buffer) {
    if (fury.getConfig().serializeEnumByName()) {
      MetaStringBytes metaStringBytes = metaStringResolver.readMetaStringBytes(buffer);
      Enum e = metaStringtoEnumRepresentation.get(metaStringBytes);
      if (e != null) {
        return e;
      }
      return handleNonexistentEnumValue(metaStringBytes.decode(Encoders.GENERIC_DECODER));
    } else {
      int value = buffer.readVarUint32Small7();
      if (value >= enumConstants.length) {
        return handleNonexistentEnumValue(value);
      }
      return enumConstants[value];
    }
  }
```

chaokunyang: metaStringResolver may return a different object, we could use fury ObjectMap instead.  fury ObjectMap is a linear probe based map, which is faster than hashmap for smaller map size

java/fury-core/src/main/java/org/apache/fury/config/Config.java
```diff
@@ -133,6 +135,11 @@ public boolean deserializeNonexistentEnumValueAsNull() {
     return deserializeNonexistentEnumValueAsNull;
   }
 
+  /** deserialize and serialize enum by name. */
+  public boolean serializeEnumByName() {
```
chaokunyang, line 139: could we add this option to java guide options part in https://fury.apache.org/docs/guide/java_object_graph_guide/#furybuilder--options ,which correspond to  docs/guide/java_serialization_guide.md

java/fury-core/src/main/java/org/apache/fury/serializer/EnumSerializer.java
```diff
@@ -40,20 +50,54 @@ public EnumSerializer(Fury fury, Class<Enum> cls) {
...
+    if (fury.getConfig().serializeEnumByName()) {
+      // as we know the size of enum is fixed, initialize the size of map with that value
+      int initialCapacity = (int) Math.ceil(enumConstants.length / 0.99f);
+
+      metaStringtoEnumRepresentation = new HashMap<>(initialCapacity, 0.99f);
+
+      for (Enum enumConstant : enumConstants) {
+        if (enumConstant != null) {
+          MetaString ms =
+              Encoders.GENERIC_ENCODER.encode(enumConstant.name(), MetaString.Encoding.UTF_8);
```
chaokunyang, line 63: We'd better use `org.apache.fury.meta.MetaStringEncoder#encode(java.lang.String)` instead, it can get smaller encoded size for enum name than utf8 encoding

chaokunyang (approved):
LGTM, thanks for your contribution

chaokunyang:
@orisgarno Just found that org.apache.fury.serializer.NonexistentClassSerializers.NonexistentEnumClassSerializer should be updated too to support skip Nonexistent Enum types. Would you like to submit a PR?

orisgarno:
sure @chaokunyang will do.
thank you.

PR #1901, opened 2024-10-23
feat(kotlin): Add tests and support for built-in types from the kotlin stdlib

wywen:
## What does this PR do?

This PR adds tests and covers built-in classes from the kotlin stdlib.

Two specific serializers are added for builtins:
1) Duration
2) Uuid

## Related issues

#683

Adds builtin types for kotlin.

## Benchmark

N/A


kotlin/src/main/java/org/apache/fury/serializer/kotlin/KotlinSerializers.java
```diff
@@ -94,5 +100,48 @@ public static void registerSerializers(Fury fury) {
...
+        resolver.register(CharCategory.class);
+        resolver.register(CharDirectionality.class);
+        resolver.register(HexFormat.class);
+        resolver.register(MatchGroup.class);
+
+        // kotlin.time
+        resolver.register(DurationUnit.class);
+        resolver.register(Duration.class);
+        resolver.registerSerializer(Duration.class, new DurationSerializer(fury));
+        resolver.register(TestTimeSource.class);
```
chaokunyang, line 140: Do we need to register this test class by default?

wywen: Hmm I was debating this, let's remove for now.

chaokunyang (approved):
LGTM

PR #1904, opened 2024-10-24
fix(java): NonExistentEnum on mode serializeEnumByName

orisgarno:
## What does this PR do?
Handle NonExistentEnum on mode serializeEnumByName by returning UNKNOWN. since there are no relevancy anymore by using enum ordinal.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
@@ -34,6 +34,7 @@ import org.apache.fury.resolver.ClassInfo;
 import org.apache.fury.resolver.ClassInfoHolder;
 import org.apache.fury.resolver.ClassResolver;
 import org.apache.fury.resolver.MetaContext;
+import org.apache.fury.resolver.MetaStringResolver;
 import org.apache.fury.resolver.RefResolver;
 import org.apache.fury.serializer.NonexistentClass.NonexistentEnum;
 import org.apache.fury.type.Descriptor;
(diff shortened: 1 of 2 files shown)


java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
```diff
@@ -226,14 +227,21 @@ public Object read(MemoryBuffer buffer) {
...
     public NonexistentEnumClassSerializer(Fury fury) {
       super(fury, NonexistentEnum.class);
+      metaStringResolver = fury.getMetaStringResolver();
       enumConstants = NonexistentEnum.class.getEnumConstants();
     }
 
     @Override
     public Object read(MemoryBuffer buffer) {
+      if (fury.getConfig().serializeEnumByName()) {
+        metaStringResolver.readMetaStringBytes(buffer);
```
orisgarno, line 241: wondering, do i need to read the buffer first to make sure the buffer ready for next consumption?

chaokunyang: I think current implementation is fine.

orisgarno: ok, should i cancel and close this PR then?

chaokunyang: I mean your implementation here is right, no need to change

chaokunyang (approved):
LGTM

Issue #1887, opened 2024-10-16
[Question] Could use Python bindings instead of Cython


penguin-wwy:
### Question

From the code, the Buffer in Cython is just a wrapper around the Buffer in cpp. It might be beneficial to use Python bindings directly, as this could reduce performance overhead and the cost of code maintenance.

chaokunyang:
You mean implement Buffer using python only instead of CBuffer?

penguin-wwy:
My suggestion is to use [pybind11](https://github.com/pybind/pybind11) or directly write a [C-API](https://docs.python.org/3/c-api/index.html) to call the cpp `fury::Buffer`. The current Cython-generated C++ code still calls the `fury::Buffer`, which I think is unnecessary.

```cpp
static CYTHON_INLINE int64_t __pyx_f_6pyfury_5_util_6Buffer_get_int64(struct __pyx_obj_6pyfury_5_util_Buffer *__pyx_v_self, uint32_t __pyx_v_offset, CYTHON_UNUSED int __pyx_skip_dispatch) {
  ...

  /* "pyfury/_util.pyx":140
 *     cpdef inline int64_t get_int64(self, uint32_t offset):
 *         self.check_bound(offset, <int32_t>8)
 *         return self.c_buffer.get().GetInt64(offset)             # <<<<<<<<<<<<<<
 * 
 *     cpdef inline float get_float(self, uint32_t offset):
 */
  __pyx_r = __pyx_v_self->c_buffer.get()->GetInt64(__pyx_v_offset);
  goto __pyx_L0;

  /* function exit code */
  ...
}
```

chaokunyang:
That would be great, if we can get a big performance gain, we can compromise some code maintenance costs

chaokunyang:
Thanks for sharing so insightful experiments. I used to think pybind is fast since it avoid generating code like cython. It turns out it's the slowest, this superised me a little. Cython seems provide some annotations to help generate c++ code, are you going to employ such kind of optimization?

Superskyyy:
I suggest using Nanobind, nanobind is the next generation of pybind11 from the same guy that invented it. It is suggested officially to use nanobind over Pybind11. Python 3.8+ is not a problem since anything below is already subset by CPython.

PragmaTwice:
Yeah from my side generally cython is hard to debug and maintain, also lacks good IDE/LSP support.

[...]

PR #1905, opened 2024-10-24
feat(python): guide cython to optimize code generation

penguin-wwy:
<!--
**Thanks for contributing to Fury.**

**If this is your first time opening a PR on fury, you can refer to [CONTRIBUTING.md](https://github.com/apache/fury/blob/main/CONTRIBUTING.md).**

Contribution Checklist

    - The **Apache Fury (incubating)** community has restrictions on the naming of pr titles. You can also find instructions in [CONTRIBUTING.md](https://github.com/apache/fury/blob/main/CONTRIBUTING.md).

    - Fury has a strong focus on performance. If the PR you submit will have an impact on performance, please benchmark it first and provide the benchmark result here.
-->

## What does this PR do?

Optimize the C++ code generated by Cython by modifying the function implementations of the Buffer class.

Using `get_int8` as an example, the following C++ code is generated before modification.

```c++
static PyObject *__pyx_pf_6pyfury_5_util_6Buffer_26get_int8(struct __pyx_obj_6pyfury_5_util_Buffer *__pyx_v_self, uint32_t __pyx_v_offset) {
  ...
  __pyx_t_1 = __pyx_f_6pyfury_5_util_6Buffer_get_int8(__pyx_v_self, __pyx_v_offset, 1);
  if (unlikely(PyErr_Occurred())) __PYX_ERR(0, 119, __pyx_L1_error)
  __pyx_t_2 = __Pyx_PyInt_From_int8_t(__pyx_t_1);
  if (unlikely(!__pyx_t_2)) __PYX_ERR(0, 119, __pyx_L1_error)
  ...
```

Since get_int32 returns a C++ type, the cython needs to generate calling `PyLong_FromLong`, and guard code such as calling `PyErr_Occurred`.
However, it is known that int32 can always be used to generate a `PyLongObject` with `PyLong_FromLong`. Therefore, when we manually call this, the cython only needs to check if the return value is null, as shown in the following code.

```c++
  __pyx_t_1 = __pyx_f_6pyfury_5_util_6Buffer_get_int8(__pyx_v_self, __pyx_v_offset, 1);
  if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 130, __pyx_L1_error)
```


## Related issues

- #1887 

<!--
If any user-facing interface changes, please [open an issue](https://github.com/apache/fury/issues/new/choose) describing the need to do so and update the document if necessary.
-->

## Benchmark

Microbenchmark case:
```
def benchmark_get_bool(buffer):
    buffer.get_bool(0)
    buffer.get_bool(1)
    buffer.get_bool(100)
    buffer.get_bool(1000)

def benchmark_get_int32(buffer):
    buffer.get_int32(0)
    buffer.get_int32(1)
    buffer.get_int32(100)
    buffer.get_int32(1000)


def benchmark_get_float(buffer):
    buffer.get_float(0)
    buffer.get_float(1)
    buffer.get_float(100)
    buffer.get_float(1000)


def benchmark_read(buffer):
    buffer.reader_index = 0
    buffer.read_int8()
    buffer.read_int16()
    buffer.read_int24()
    buffer.read_int32()
    buffer.read_int64()
    buffer.read_float()
    buffer.read_double()
```

Result:
```
# before
python benchmark.py --affinity 0
.....................
benchmark_get_bool: Mean +- std dev: 220 ns +- 4 ns
.....................
benchmark_get_int32: Mean +- std dev: 276 ns +- 10 ns
.....................
benchmark_get_float: Mean +- std dev: 254 ns +- 2 ns
.....................
benchmark_read: Mean +- std dev: 409 ns +- 11 ns

# after
python benchmark.py --affinity 0                                                                                                                                                                                                                                                                                    
.....................
benchmark_get_bool: Mean +- std dev: 215 ns +- 4 ns
.....................
benchmark_get_int32: Mean +- std dev: 264 ns +- 9 ns
.....................
benchmark_get_float: Mean +- std dev: 243 ns +- 6 ns
.....................
benchmark_read: Mean +- std dev: 380 ns +- 4 ns
```

### DIFF
--- a/python/pyfury/_serialization.pyx
+++ b/python/pyfury/_serialization.pyx
@@ -1630,6 +1630,18 @@ cdef class BytesSerializer(CrossLanguageCompatibleSerializer):
         return fury_buf.to_pybytes()
 
 
+"""
+Collection serialization format:
+https://fury.apache.org/docs/specification/fury_xlang_serialization_spec/#list
+Has the following changes:
+* None has an independent NonType type, so COLLECTION_NOT_SAME_TYPE can also cover the concept of being nullable.
+* No flag is needed to indicate that the element type is not the declared type.
+"""
+cdef int8_t COLLECTION_DEFAULT_FLAG = 0b0
+cdef int8_t COLLECTION_TRACKING_REF = 0b1
+cdef int8_t COLLECTION_NOT_SAME_TYPE = 0b1000
+
+
 cdef class CollectionSerializer(Serializer):
     cdef ClassResolver class_resolver
     cdef MapRefResolver ref_resolver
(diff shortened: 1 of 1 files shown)


chaokunyang:
Hi @penguin-wwy , thanks for your effort. I see we replaced all specific types to generic pyobject. This may introduce overhead if we invoke from cython side. Are there some annotation to reduce checks for us automatically so we can avoiding change return types?

penguin-wwy:
> Hi @penguin-wwy , thanks for your effort. I see we replaced all specific types to generic pyobject. This may introduce overhead if we invoke from cython side. Are there some annotation to reduce checks for us automatically so we can avoiding change return types?

Yes, I found that this modification indeed disrupts cython's annotation optimization. However, I think we should write an appropriate Python microbenchmark, similar to what is done in integration_tests. This way, it will be easier to identify differences during optimization. I will draft this PR and proceed with the modifications once the microbenchmark is completed.

PR #1911, opened 2024-10-26
fix(java): child container deep copy

zhaommmmomo:
## What does this PR do?
fix child container bug

## Related issues
https://github.com/apache/fury/issues/1855

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/reflect/ReflectionUtils.java
+++ b/java/fury-core/src/main/java/org/apache/fury/reflect/ReflectionUtils.java
@@ -380,6 +380,19 @@ public class ReflectionUtils {
     return fields;
   }
 
+  public static List<Field> getFieldsWithoutSuperClasses(Class<?> cls, Set<Class<?>> superClasses) {
+    Preconditions.checkNotNull(cls);
+    List<Field> fields = new ArrayList<>();
+    Class<?> clazz = cls;
+    do {
+      if (!superClasses.contains(clazz)) {
+        Collections.addAll(fields, clazz.getDeclaredFields());
+      }
+      clazz = clazz.getSuperclass();
+    } while (clazz != null);
+    return fields;
+  }
+
   /** Get fields values from provided object. */
   public static List<Object> getFieldValues(Collection<Field> fields, Object o) {
     List<Object> results = new ArrayList<>(fields.size());
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
@@ -309,7 +309,7 @@ public abstract class AbstractObjectSerializer<T> extends Serializer<T> {
   public static InternalFieldInfo[] buildFieldsInfo(Fury fury, List<Field> fields) {
     List<Descriptor> descriptors = new ArrayList<>();
     for (Field field : fields) {
-      if (!Modifier.isStatic(field.getModifiers())) {
+      if (!Modifier.isTransient(field.getModifiers()) && !Modifier.isStatic(field.getModifiers())) {
         descriptors.add(new Descriptor(field, TypeRef.of(field.getGenericType()), null, null));
       }
     }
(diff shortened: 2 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/reflect/ReflectionUtils.java
```diff
@@ -380,6 +380,19 @@ public static List<Field> getFields(Class<?> cls, boolean searchParent) {
     return fields;
   }
 
+  public static List<Field> getFieldsWithoutSuperClasses(Class<?> cls, Set<Class<?>> superClasses) {
```
chaokunyang, line 383: You can use `org.apache.fury.reflect.ReflectionUtils#getFields(cls, false)` instead

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -256,6 +306,29 @@ private InternalFieldInfo[] buildFieldsInfo() {
     return fieldInfos;
   }
 
+  public static InternalFieldInfo[] buildFieldsInfo(Fury fury, List<Field> fields) {
```
chaokunyang, line 309: Could we just use `org.apache.fury.serializer.AbstractObjectSerializer#buildFieldInfos` instead. We should only copy those fields if we create object new instance following serialization process

java/fury-core/src/main/java/org/apache/fury/serializer/collection/ChildContainerSerializers.java
```diff
@@ -210,11 +216,14 @@ public Map newMap(MemoryBuffer buffer) {
     }
 
     @Override
-    public T copy(T originMap) {
-      if (objectSerializer == null) {
-        objectSerializer = new ObjectSerializer<>(fury, type, false);
+    public Map newMap(Map originMap) {
```
chaokunyang, line 219: Why can't we just delegate copy to objectSerializer? I think we should follow same pattern as serialization did

chaokunyang:
Hi @zhaommmmomo , I think we can just let copy for all serializers following same pattern of serialization. We should copy same fields as serialization. And we should use same methods for object creation, otherwise object may be in a inconsistent state

java/fury-core/src/main/java/org/apache/fury/serializer/collection/ChildContainerSerializers.java
chaokunyang: I see, objectSerializer can set fields for current type

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -256,6 +306,29 @@ private InternalFieldInfo[] buildFieldsInfo() {
...
+    List<Descriptor> descriptors = new ArrayList<>();
+    for (Field field : fields) {
+      if (!Modifier.isStatic(field.getModifiers())) {
```
chaokunyang, line 312: transient fields should be filtered too

chaokunyang: We can just `fury.getClassResolver().getAllDescriptorsMap(beanClass, false).values();` instead

java/fury-core/src/main/java/org/apache/fury/serializer/collection/ChildContainerSerializers.java
```diff
@@ -151,11 +154,14 @@ public Collection newCollection(MemoryBuffer buffer) {
     }
 
     @Override
-    public T copy(T originCollection) {
-      if (objectSerializer == null) {
-        objectSerializer = new ObjectSerializer<>(fury, type, false);
+    public Collection newCollection(Collection originCollection) {
+      Collection newCollection = super.newCollection(originCollection);
+      if (fieldInfos == null) {
+        List<Field> fields = ReflectionUtils.getFieldsWithoutSuperClasses(type, superClasses);
```
chaokunyang, line 160: We can just `fury.getClassResolver().getAllDescriptorsMap(beanClass, false).values(); ` or `org.apache.fury.reflect.ReflectionUtils#getFields` instead

chaokunyang (approved):
LGTM

Issue #1896, opened 2024-10-22
[Kotlin] add code lint CI for kotlin


chaokunyang:
### Feature Request

add code lint CI for kotlin

PR #1912, opened 2024-10-26
build(kotlin): Add spotless ktfmt and enable explicit api mode for library code.

wywen:
## What does this PR do?

This PR modifies the build to:
1) add ktfmt - Since the fury java uses GOOGLE style, the fury kotlin will also use GOOGLE style. The version of ktfmt used is the highest compatible with the spotless maven plugin.
2) explicit api mode
3) parameterized kotlin version in pom.xml so it is easy to change in future
4) Update README.md with build, format and testing instructions.

## Related issues

#1896 

None.

## Benchmark
N/A

### DIFF
--- a/kotlin/README.md
+++ b/kotlin/README.md
@@ -42,6 +42,7 @@ Additional Notes:
 mvn clean
 mvn -T10 compile
 ```
+
 ## Code Format
 
 ```bash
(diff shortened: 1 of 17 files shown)


chaokunyang (approved):
Looks awesome!

PR #1914, opened 2024-10-27
fix(java): Fix incorrect results of utf16 to utf8 conversion for latin1 but not ascii characters

HuangXingBo:
## What does this PR do?

Fix incorrect results of utf16 to utf8 conversion for latin1 but not ascii characters


chaokunyang (approved):
LGTM

PR #1925, opened 2024-11-04
feat(java): ReplaceResolveSerializer deep copy

zhaommmmomo:
## What does this PR do?
Adjusting the deep copy of ReplaceResolveSerializer.

obj -> writeReplace -> copy -> readResolve -> newObj

## Related issues
https://github.com/apache/fury/issues/1849

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/ReplaceResolveSerializer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/ReplaceResolveSerializer.java
@@ -139,6 +139,19 @@ public class ReplaceResolveSerializer extends AbstractObjectSerializer {
         }
       }
     }
+
+    Object readResolve(Object o) {
+      if (readResolveFunc != null) {
+        return readResolveFunc.apply(o);
+      } else {
+        try {
+          return readResolveMethod.invoke(o);
+        } catch (Exception e) {
+          Platform.throwException(e);
+          throw new IllegalStateException(e);
+        }
+      }
+    }
   }
 
   private static final ClassValue<ReplaceResolveInfo> REPLACE_RESOLVE_INFO_CACHE =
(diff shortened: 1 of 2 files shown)


java/fury-core/src/main/java/org/apache/fury/serializer/ReplaceResolveSerializer.java
```diff
@@ -44,7 +44,7 @@
  * will skip classname writing if object returned by `writeReplace` is different from current class.
  */
 @SuppressWarnings({"unchecked", "rawtypes"})
-public class ReplaceResolveSerializer extends Serializer {
+public class ReplaceResolveSerializer extends AbstractObjectSerializer {
```
chaokunyang, line 47: Since ReplaceResolveSerializer copy by copy returned object of `writePlace` method, it's unnecessary to extends  AbstractObjectSerializer

zhaommmmomo: > Since ReplaceResolveSerializer copy by copy returned object of `writePlace` method, it's unnecessary to extends AbstractObjectSerializer

okay

chaokunyang (approved):
Looks great!

chaokunyang:
maybe we could add more complexed copy tests such as copying`CustomReplaceClass1/CustomReplaceClass2/CustomReplaceClass3/CustomReplaceClass4/CustomReplaceClass5/CustomReplaceClass6/SimpleCollectionTest/Subclass1/Subclass2/Subclass3`. All test objects in ReplaceResolveSerializerTest could be used for complexed copying tests. In this way, we cann ensure future change won't break this feature.

zhaommmmomo:
> maybe we could add more complexed copy tests such as copying`CustomReplaceClass1/CustomReplaceClass2/CustomReplaceClass3/CustomReplaceClass4/CustomReplaceClass5/CustomReplaceClass6/SimpleCollectionTest/Subclass1/Subclass2/Subclass3`. All test objects in ReplaceResolveSerializerTest could be used for complexed copying tests. In this way, we cann ensure future change won't break this feature.

There are already some copy test cases for these classes in ReplaceResolveSerializerTest
image

chaokunyang:
Some cases are more complicated, maybe we need to add copy tests for all those objects

zhaommmmomo:
> Some cases are more complicated, maybe we need to add copy tests for all those objects
Okay, I'll add it later

zhaommmmomo:
> Some cases are more complicated, maybe we need to add copy tests for all those objects

Okay, I'll add it later

PR #1932, opened 2024-11-07
【Java】add deserial object with feature deserializeNonexistentClassNotWriteFullClassInfo feature

shuyufang:
…ullClassInfo

## What does this PR do?

we has some scene , after the class with the upstream system serialized, the downstream system can not deserialize without the class object, but current .withDeserializeNonexistentClass write the full class info. we want fury support deserialize object when none local class can deserializeNonexistentClassNotWriteFullClassInfo feature

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
@@ -28,6 +28,7 @@ import org.apache.fury.collection.LongMap;
 import org.apache.fury.collection.MapEntry;
 import org.apache.fury.collection.Tuple2;
 import org.apache.fury.collection.Tuple3;
+import org.apache.fury.config.Config;
 import org.apache.fury.memory.MemoryBuffer;
 import org.apache.fury.meta.ClassDef;
 import org.apache.fury.resolver.ClassInfo;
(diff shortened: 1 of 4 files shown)


java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
```diff
@@ -193,6 +194,7 @@ public Object read(MemoryBuffer buffer) {
       ClassFieldsInfo fieldsInfo = getClassFieldsInfo(classDef);
       ObjectSerializer.FinalTypeField[] finalFields = fieldsInfo.finalFields;
       boolean[] isFinal = fieldsInfo.isFinal;
+      Config config = fury.getConfig();
```
chaokunyang, line 197: We'd better get the option here, and pass that flag to `getFileName` method

java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
```diff
@@ -208,23 +210,35 @@ public Object read(MemoryBuffer buffer) {
...
     }
   }
 
+  public static String getFileName(String qualifiedFieldName, Config config) {
+    if (config.deserializeNonexistentClassNotWriteFullClassInfo()) {
+      int index = qualifiedFieldName.lastIndexOf(".");
+      if (index < 0) {
+        return qualifiedFieldName;
+      }
+      return qualifiedFieldName.substring(index + 1);
```
chaokunyang, line 236: This will create a new string everytime for deserialization, which is very slow. A better method is rename `fieldInfo#qualifiedFieldName` into `fieldInfo#fieldName`, and build it as qualified based on the option.

chaokunyang: And we should skip duplicated field name in super class if this option is enabled

chaokunyang:
@shuyufang Seems we can refactor `ClassDef#FieldInfo#qualifiedFieldName` into `fieldName`, and only add classname when parent class has duplicate field with subclass. In this way, we can skip add a new option

PR #1942, opened 2024-11-13
feat(python): Implement collection serialization protocol

penguin-wwy:
## What does this PR do?

Implement a new format for collection serialization in pyfury.

## Benchmark

```
fury_tuple: Mean +- std dev: [base] 259 us +- 6 us -> [collection] 256 us +- 5 us: 1.01x faster
fury_large_tuple: Mean +- std dev: [base] 92.7 ms +- 5.5 ms -> [collection] 63.7 ms +- 4.8 ms: 1.46x faster
fury_list: Mean +- std dev: [base] 277 us +- 6 us -> [collection] 267 us +- 3 us: 1.04x faster
fury_large_list: Mean +- std dev: [base] 92.8 ms +- 5.3 ms -> [collection] 66.5 ms +- 3.0 ms: 1.40x faster

Geometric mean: 1.21x faster
```

### DIFF
--- a/python/pyfury/_serialization.pyx
+++ b/python/pyfury/_serialization.pyx
@@ -1630,6 +1630,18 @@ cdef class BytesSerializer(CrossLanguageCompatibleSerializer):
         return fury_buf.to_pybytes()
 
 
+"""
+Collection serialization format:
+https://fury.apache.org/docs/specification/fury_xlang_serialization_spec/#list
+Has the following changes:
+* None has an independent NonType type, so COLLECTION_NOT_SAME_TYPE can also cover the concept of being nullable.
+* No flag is needed to indicate that the element type is not the declared type.
+"""
+cdef int8_t COLLECTION_DEFAULT_FLAG = 0b0
+cdef int8_t COLLECTION_TRACKING_REF = 0b1
+cdef int8_t COLLECTION_NOT_SAME_TYPE = 0b1000
+
+
 cdef class CollectionSerializer(Serializer):
     cdef ClassResolver class_resolver
     cdef MapRefResolver ref_resolver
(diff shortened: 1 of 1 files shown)


python/pyfury/_serialization.pyx
```diff
@@ -1644,29 +1656,143 @@ cdef class CollectionSerializer(Serializer):
     cpdef int16_t get_xtype_id(self):
         return -FuryType.LIST.value
 
+    cpdef int8_t write_header(self, Buffer buffer, value):
```
chaokunyang, line 1659: `cdef` is enough here, it will be invoked in ColllectionSerializer only

python/pyfury/_serialization.pyx
```diff
@@ -1644,29 +1656,143 @@ cdef class CollectionSerializer(Serializer):
...
+        cdef int8_t collect_flag = COLLECTION_DEFAULT_FLAG
+        elem_type = type(next(iter(value)))
```
chaokunyang, line 1661: This create iterator object, maybe we could just init it as None here

python/pyfury/_serialization.pyx
```diff
@@ -1644,29 +1656,143 @@ cdef class CollectionSerializer(Serializer):
...
+                self._write_int(buffer, value)
+            elif elem_type is bool:
+                self._write_bool(buffer, value)
+            elif elem_type is float:
+                self._write_float(buffer, value)
+            else:
+                if (collect_flag & COLLECTION_TRACKING_REF) == 0:
+                    self._write_same_type_no_ref(buffer, value)
+                else:
+                    self._write_same_type_ref(buffer, value)
```
chaokunyang, line 1692: We should write type only once, instead write type for every element in the loop

python/pyfury/_serialization.pyx
chaokunyang, line 1659: Currently golang didn't implement this protocol yet, could we enable it only for pure python serialization?

chaokunyang:
Hi @penguin-wwy , thanks for your hard work. This new protocol should be faster, you did't implement the new protocol fully. I left some comments, could you take a look at it?

python/pyfury/_serialization.pyx
penguin-wwy: done

python/pyfury/_serialization.pyx
```diff
@@ -1644,29 +1656,143 @@ cdef class CollectionSerializer(Serializer):
...
     cpdef write(self, Buffer buffer, value):
-        buffer.write_varint32(len(value))
+        if len(value) == 0:
+            buffer.write_varint64(0)
+            return
+        cdef int8_t collect_flag = self.write_header(buffer, value)
         cdef MapRefResolver ref_resolver = self.ref_resolver
         cdef ClassResolver class_resolver = self.class_resolver
+        if (collect_flag & COLLECTION_NOT_SAME_TYPE) == 0:
+            elem_type = type(next(iter(value)))
```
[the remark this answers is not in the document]
penguin-wwy: done

python/pyfury/_serialization.pyx
penguin-wwy: done

python/pyfury/_serialization.pyx
penguin-wwy: `write_header` will only be called in `Python`, and the `Xlang` format has not been modified at the moment

python/pyfury/_serialization.pyx
penguin-wwy: This is to extract the type of the first element and avoid checking `elem_type is None` in the loop. Since the collection might be a set, an iterator retrieves it.

penguin-wwy:
Benchmark data has been updated.

python/pyfury/_serialization.pyx
```diff
@@ -1644,29 +1664,143 @@ cdef class CollectionSerializer(Serializer):
...
+            buffer.write_varint64(0)
+            return
+        cdef pair[int8_t, int64_t] header_pair = self.write_header(buffer, value)
+        cdef int8_t collect_flag = header_pair.first
+        cdef int64_t elem_type_ptr = header_pair.second
+        cdef elem_type = <type>int2obj(elem_type_ptr)
         cdef MapRefResolver ref_resolver = self.ref_resolver
         cdef ClassResolver class_resolver = self.class_resolver
+        if (collect_flag & COLLECTION_NOT_SAME_TYPE) == 0:
+            if elem_type is str:
```
chaokunyang, line 1690: you still write null flag and type here, this should be skipped for same type and not-null elements. could you take `org.apache.fury.serializer.collection.AbstractCollectionSerializer#writeSameTypeElements` as a reference?

penguin-wwy: In Python, the type of `None` is `NoneType`, so if all elements are of the same type, either none of them are `None` or all of them are `None`.
```
>>> type(None)
<class 'NoneType'>
>>> type(object()) is type(None)
False
```

penguin-wwy: Done, write the type flag only once for primitive types

chaokunyang (approved):
LGTM, thanks very much

PR #1944, opened 2024-11-18
fix(java): Move schema caching to unsafe trait to avoid issues when using non-inferred schema.

wywen:
## What does this PR do?

This PR removes the java specific `ExtField` class from the schema and moves the extData mechanism to the internal UnsafeTrait class. This is necessary because `ExtField` is only created internally from `inferSchema` method used by java, and we would potentially import or derive schemas from other sources (e.g XLANG, handwritten schema, arrow-native source) -- those schemas will be incompatible when we attempt to retrieve the cached schema. Removing schema-caching will fix the issue, but create allocations, so after some discussion, we decided to move the mechanism to the internal UnsafeTrait class.

This implementation makes changes internal API:
- Derived classes from `UnsafeTrait` need to initialize the `extData` cache and define the number of extData slots needed.
- The internal `getStruct` method needs to define which slot we use to retrieve `extData`.

Other:

- REVERTED: pom.xml for fury-format will automatically run tests with appropriate --add-opens flag for arrow

N/A

### DIFF
--- a/java/fury-format/pom.xml
+++ b/java/fury-format/pom.xml
@@ -116,12 +116,6 @@
           </archive>
         </configuration>
       </plugin>
-      <plugin>
-          <artifactId>maven-surefire-plugin</artifactId>
-          <configuration>
-              <argLine>--add-opens=java.base/java.nio=ALL-UNNAMED</argLine>
-          </configuration>
-      </plugin>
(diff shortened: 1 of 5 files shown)


chaokunyang (approved):
LGTM

Issue #1950, opened 2024-11-20
[Question] [Java]: Configurable buffer size limit


theigl:
### Question

Since https://github.com/apache/fury/commit/01f1b66383a127a801f22ebe7fac5ca54d675878  Fury's internal `MemoryBuffer` is reset to 128k after every use.

The change makes sense because it reclaims memory that could otherwise sit unused, *but* it will cause massive allocations if most of the object graphs you serialize are >= 128k. In this case, a new buffer is allocated for every serialization.

In my application, most object graphs are 64k-512k leading to very frequent reallocation of the built-in buffer.

Currently, my only option is to manage a pool of `MemoryBuffers` myself and pass them to Fury instead of relying on the built-in buffer, but this makes using Fury *significantly more complex*.

Would it make sense to make the buffer size limit configurable? In my case, a 1MB limit would be ideal.

chaokunyang:
Make sense, would you like to add a new field at `FuryBuilder/Config` and use that value for Buffer reset

theigl:
@chaokunyang: I pushed a PR for this, but I'm not sure if it is the right approach. Do you have any feedback?

PR #1963, opened 2024-12-01
feat(java): configurable buffer size limit

theigl:
## What does this PR do?

This PR introduces a new configuration option `bufferSizeLimitBytes` that replaces the hard-coded default of 128kb.

## Related issues

#1950

The PR introduces a new configuration option `bufferSizeLimitBytes`.

## Discussion

This PR solves my problem, but I'm not sure if it is the right way to move forward. This is quite a low-level configuration option, but a potentially very important one. Every user whose average payload size is >=128kb, will need to increase this value for maximum performance. Maybe the default limit should be increased to something less conservative like 1MB, so fewer users will need to adjust this setting?


chaokunyang (approved):
LGTM

PR #1968, opened 2024-12-05
fix(java): Fix flakiness in ExpressionVisitorTest#testTraverseExpression

AmitPr:
## What does this PR do?

`ExpressionVisitorTest#testTraverseExpression` relies on ordered traversal. However, the traversal depends on the order of `getDeclaredFields()` to be deterministic (see below). The Java specification explicitly marks `getDeclaredFields` as a function that can return field names in any order. In the wild, this is most likely to manifest on different architectures with different JVM versions.

https://github.com/apache/fury/blob/54b62fb6ab5d7e557131efe07c7402c885f6e7c4/java/fury-core/src/main/java/org/apache/fury/reflect/ReflectionUtils.java#L368-L381

Using [NonDex](https://github.com/TestingResearchIllinois/NonDex), we can replicate the flaky behavior with the following command:

```
mvn edu.illinois:nondex-maven-plugin:2.1.7:nondex -pl fury-core/ -Dcheckstyle.skip -Dmaven.javadoc.skip -Dtest=org.apache.fury.codegen.ExpressionVisitorTest
```

The fix, in this case, is to simply use HashSet equality instead of an ordered List equality. The above NonDex command should succeed after applying this patch.

I do, however, note that there are other flaky tests (found by running NonDex on the entire `fury-core` project) that fail due to reliance on e.g. `PriorityQueue#toArray`, which is also explicity marked to return nondeterministically ordered arrays.

No

### DIFF
--- a/java/fury-core/src/test/java/org/apache/fury/codegen/ExpressionVisitorTest.java
+++ b/java/fury-core/src/test/java/org/apache/fury/codegen/ExpressionVisitorTest.java
@@ -29,7 +29,6 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-
 import org.apache.fury.codegen.Expression.Literal;
 import org.apache.fury.reflect.ReflectionUtils;
 import org.apache.fury.reflect.TypeRef;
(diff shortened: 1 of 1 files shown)


pandalee99:
Welcome! You can refer to this document: [CONTRIBUTING](https://github.com/apache/fury/blob/main/CONTRIBUTING.md)

chaokunyang (approved):
LGTM, thanks

pandalee99 (approved):

PR #1985, opened 2024-12-20
fix(java): only print warn message if scopedMetaShareEnabled is true …

zhfeng:
…and not in CompatibleMode

## Related issues

- https://github.com/quarkiverse/quarkus-fury/issues/51

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
+++ b/java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
@@ -400,7 +400,7 @@ public final class FuryBuilder {
       if (deserializeNonexistentClass == null) {
         deserializeNonexistentClass = false;
       }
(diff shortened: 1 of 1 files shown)


java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
```diff
@@ -400,7 +400,7 @@ private void finish() {
       if (deserializeNonexistentClass == null) {
         deserializeNonexistentClass = false;
       }
-      if (scopedMetaShareEnabled != null) {
+      if (scopedMetaShareEnabled) {
```
chaokunyang, line 403: This will throw null pointer exception, the default value is null

zhfeng: Fixed.

chaokunyang (approved):
LGTM

PR #1990, opened 2024-12-25
feat(c++): Support the UTF-8 to UTF-16 with SIMD

### LINKED ISSUE #1964: [C++] To support the utf8 utf16 and using simd to accelerate the optimization
### Feature Request

Earlier, I implemented utf16 to utf8 and used simd to speed up the process.

``` c++
std::string utf16ToUtf8(const std::u16string &utf16, bool is_little_endian)
```

Now, you also need to implement utf8 to utf16.
Because now in java, using utf8, 16, latin1 is becoming very common, now we need to implement it in py as well.
But for speed reasons, we might consider using a direct call to a c++ module.

For simd acceleration, it is necessary to adapt to x86 architecture, arm and risc-v.
The solution is that you can use the avx2 instruction set under x86 and so on...

Module lack utf8 utf16.

Implement utf8 to utf16, and use simd acceleration.

_No response_

### Additional context

_No response_

pandalee99:
## What does this PR do?
To support the utf8 utf16 and using simd to accelerate the optimization
``` c++
std::string utf16ToUtf8(const std::u16string &utf16, bool is_little_endian)
```

The logic of converting UTF-8 to UTF-16 isn't that complicated. but there are still lots of optimizations that I haven't come up with yet.

So, I'll first design a version that's a bit faster than the original one, and then think about how to make further optimizations.

Judging from the tests, the logic is correct:
``` text 
[----------] 9 tests from UTF8ToUTF16Test
[ RUN      ] UTF8ToUTF16Test.BasicConversion
[       OK ] UTF8ToUTF16Test.BasicConversion (0 ms)
[ RUN      ] UTF8ToUTF16Test.EmptyString
[       OK ] UTF8ToUTF16Test.EmptyString (0 ms)
[ RUN      ] UTF8ToUTF16Test.SurrogatePairs
[       OK ] UTF8ToUTF16Test.SurrogatePairs (0 ms)
[ RUN      ] UTF8ToUTF16Test.BoundaryValues
[       OK ] UTF8ToUTF16Test.BoundaryValues (0 ms)
[ RUN      ] UTF8ToUTF16Test.SpecialCharacters
[       OK ] UTF8ToUTF16Test.SpecialCharacters (0 ms)
[ RUN      ] UTF8ToUTF16Test.LittleEndian
[       OK ] UTF8ToUTF16Test.LittleEndian (0 ms)
[ RUN      ] UTF8ToUTF16Test.BigEndian
[       OK ] UTF8ToUTF16Test.BigEndian (0 ms)
[ RUN      ] UTF8ToUTF16Test.RoundTripConversion
[       OK ] UTF8ToUTF16Test.RoundTripConversion (0 ms)
```
image

And from the performance perspective, it's improved compared to serial processing:
image

The speed of execution has been significantly improved

Actually, this code doesn't use libraries like AVX2 or really apply SIMD to process. The main reason is that the structure of UTF-8 encoding is complex and not fixed. It involves multi-byte encoding, and we need to analyze it byte by byte when dealing with different bytes. So, without clear rules and a uniform length, it becomes really hard to directly parallelize the processing of each byte. During the process of converting UTF-8 to UTF-16, we have to handle characters of different lengths, ranging from 1 to 4 bytes, which makes it difficult to break it down into structures that can be directly applied to SIMD operations. 
There are also some code style changes, uniform writing

## Related issues

Close #1964 

### DIFF
--- a/cpp/fury/util/string_util.cc
+++ b/cpp/fury/util/string_util.cc
@@ -28,7 +28,6 @@
 #endif
 
 #include <chrono>
-#include <stdexcept>
 #include <string>
 
 namespace fury {
--- a/cpp/fury/util/string_util.cc
+++ b/cpp/fury/util/string_util.cc
@@ -71,7 +71,6 @@ std::u16string utf8ToUtf16SIMD(const std::string &utf8, bool is_little_endian) {
 
   while (i + 32 <= n) {
 
-    // Now process the characters in 'in' SIMD register
     for (int j = 0; j < 32; ++j) {
       uint8_t byte = utf8[i + j];
 
--- a/cpp/fury/util/string_util.cc
+++ b/cpp/fury/util/string_util.cc
@@ -33,272 +33,272 @@
 namespace fury {
 
 // Swap bytes to convert from big endian to little endian
-    inline uint16_t swapBytes(uint16_t value) {
-        return (value >> 8) | (value << 8);
-    }
+inline uint16_t swapBytes(uint16_t value) {
+  return (value >> 8) | (value << 8);
+}
 
-    inline void utf16ToUtf8(uint16_t code_unit, char *&output) {
-        if (code_unit < 0x80) {
-            *output++ = static_cast<char>(code_unit);
-        } else if (code_unit < 0x800) {
-            *output++ = static_cast<char>(0xC0 | (code_unit >> 6));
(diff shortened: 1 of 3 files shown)


chaokunyang (approved):
LGTM

Issue #1984, opened 2024-12-19
 java.lang.IndexOutOfBoundsException when use withStringCompressed(true)


wzhongke:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

0.9.0

### Component(s)

Java

### Minimal reproduce step

```java
public class Main {
    private static final ThreadSafeFury FURY = Fury.builder().withLanguage(Language.JAVA)
        .requireClassRegistration(false)
        .withNumberCompressed(true)
        .withStringCompressed(true)
        .withRefTracking(true)
        .withMetaCompressor(new DeflaterMetaCompressor())
        .buildThreadLocalFury();

    @Data
    public static class Simple {
        private String str;

        public Simple(String str) {
            this.str = str;
        }
    }

    public static void main(String[] args) {
        byte[] serialize = FURY.serialize(new Simple("STG@ON DEMAND Solutions@GeoComputing Switch/ Hub@Digi Edgeport/216 – 16 port Serial Hub"));
        FURY.deserialize(serialize);
    }
}
```

### What did you expect to see?

No error

### What did you see instead?

Exception in thread "main" org.apache.fury.exception.DeserializationException: Deserialize failed, read objects are: [Main.Simple(str=null)]
	at org.apache.fury.util.ExceptionUtils.handleReadFailed(ExceptionUtils.java:63)
	at org.apache.fury.Fury.deserialize(Fury.java:797)
	at org.apache.fury.Fury.deserialize(Fury.java:714)
	at org.apache.fury.ThreadLocalFury.deserialize(ThreadLocalFury.java:137)
	at com.lenovo.awakens.Main.main(Main.java:33)
Caused by: java.lang.IndexOutOfBoundsException: readerIndex(30) + length(81) exceeds size(123): org.apache.fury.memory.MemoryBuffer$BoundChecker@2584b82d
	at org.apache.fury.memory.MemoryBuffer$BoundChecker.fillBuffer(MemoryBuffer.java:186)
	at org.apache.fury.memory.MemoryBuffer.checkReadableBytes(MemoryBuffer.java:2412)
	at org.apache.fury.serializer.StringSerializer.readBytesUnCompressedUTF16(StringSerializer.java:434)
	at org.apache.fury.serializer.StringSerializer.readBytesString(StringSerializer.java:206)
[...]

PR #1991, opened 2024-12-30
fix(java): Fix the issue caused by not using readCompressedBytesString during deserialization when string compression is enabled.

Aliothmoon:
Fix the issue caused by not using readCompressedBytesString during deserialization when string compression is enabled.

## What does this PR do?

fix issue [#1984 ](https://github.com/apache/fury/issues/1984)


chaokunyang (approved):
LGTM, thanks

Issue #2000, opened 2025-01-08
[Nodejs] npm install failed on CI


chaokunyang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

main

### Component(s)

JavaScript

### Minimal reproduce step

https://github.com/apache/fury/actions/runs/12671304833/job/35312782636?pr=1999

### What did you expect to see?

no error

### What did you see instead?

![image](https://github.com/user-attachments/assets/c1d9a84c-7698-4b4f-b653-35ae46a6bc3c)

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #2001, opened 2025-01-09
feat(javascript): removing HPS

theweipeng:
## What does this PR do?
Removed HPS because we can now use `--expose_externalize_string` to detect the encoding of a string. 
Additionally, we have submitted a PR (https://github.com/nodejs/node/pull/56147) to Node.js for detecting string encoding. If this PR is merged, we will be able to achieve good performance without this flag.

## Related issues
#2000


chaokunyang:
@theweipeng Should we still keep `hps` since https://github.com/nodejs/node/pull/56147 will be available for newer nodejs version?

theweipeng:
It will be a long time before that PR committed in node repo be released. And hps module has far more performance than the way provided by that PR. So I will close this PR

PR #2002, opened 2025-01-09
chore(c++): Function inlining and code tuning

pandalee99:
## What does this PR do?
1、Inline `string_util.cc` function to `string_util.h`
2、Clearer judgment of conditions:`if defined(USE_IMMINTRIN_SIMD)`
3、Remove redundant code

### DIFF
--- a/cpp/fury/util/string_util.cc
+++ b/cpp/fury/util/string_util.cc
@@ -21,42 +21,19 @@
 
 #if defined(__x86_64__) || defined(_M_X64)
 #include <immintrin.h>
+#define USE_IMMINTRIN_SIMD
 #elif defined(__ARM_NEON) || defined(__ARM_NEON__)
 #include <arm_neon.h>
+#define USE_NEON_SIMD
 #elif defined(__riscv) && __riscv_vector
 #include <riscv_vector.h>
+#define USE_RISCV_VECTOR_SIMD
 #endif
 
 #include <chrono>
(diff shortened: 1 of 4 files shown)


cpp/fury/util/string_util.cc
```diff
@@ -21,43 +21,20 @@
 
 #if defined(__x86_64__) || defined(_M_X64)
 #include <immintrin.h>
+#define USE_IMMINTRIN_SIMD
```
chaokunyang, line 24: How about adding a `platform.h` and define macro in that file. 

And I perfer naming the macro as `FURY_HAS_SSE2/FURY_HAS_NEON/...`

PragmaTwice:
Maybe we can try some portable SIMD libraries, like
- https://github.com/xtensor-stack/xsimd
- https://github.com/jfalcou/eve/tree/main
- https://github.com/google/highway

.. instead of handwritten intrinsic calls.

chaokunyang:
I'm OK with each way. Maybe we could run some tests on small batch of data and larger data to see the performance difference.  [eve](https://github.com/jfalcou/eve) needs c++ 20, which is not an option. But for [xsimd](https://github.com/xtensor-stack/xsimd) and [highway](https://github.com/google/highway), we could do some tests, especially for small array whose length is less then `1000`.

chaokunyang (approved):

PR #2007, opened 2025-01-14
perf(python): optimize pystr deserialize perf

chaokunyang:

### DIFF
--- a/cpp/fury/util/array_util.h
+++ b/cpp/fury/util/array_util.h
@@ -1,0 +1,83 @@
+#include "fury/util/platform.h"
+
+namespace fury {
+#if defined(FURY_HAS_NEON)
+inline uint16_t getMaxValue(uint16_t* arr, size_t length) {
+  if (length == 0) {
+    return 0;  // Return 0 for empty arrays
+  }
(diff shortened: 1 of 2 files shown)


chaokunyang:
cc @penguin-wwy @pandalee99 @theweipeng

python/pyfury/_util.pyx
```diff
@@ -572,12 +573,15 @@ cdef class Buffer:
...
+        if size == 0:
+            return ""
         cdef const char * buf = <const char *>(self.c_buffer.get().data() + self.reader_index)
         self.reader_index += size
         cdef uint32_t encoding = header & <uint32_t>0b11
         if encoding == 0:
             # PyUnicode_FromASCII
-            return PyUnicode_DecodeLatin1(buf, size, "strict")
+            return <unicode>Fury_PyUnicode_FromUCS1(buf, size)
+            # return PyUnicode_DecodeLatin1(buf, size, "strict")
```
chaokunyang, line 584: If i use PyUnicode_DecodeLatin1 directly here, It's faster in macos, which is unexpected Since my implementation used the simd, and if i invoke PyUnicode_DecodeLatin1 directly in PyUnicode_FromUCS1, it's slower too. @penguin-wwy  do you have any ideas?

PR #2008, opened 2025-01-16
perf(python): optimize bytes buffer creation

chaokunyang:
## What does this PR do?

This pr optimize bytes buffer creation by a fast path to get address of `bytes` by `PyBytes_AsString` instead of using slow `memoryview`


pandalee99 (approved):
LGTM

PR #2010, opened 2025-01-18
feat(java): make 4 bytes utf16 size header optional for utf8 encoding

chaokunyang:
## What does this PR do?

Currently fury serialize utf8 string in java will write num bytes of utf16 first, so that the deserializaiton can save one copy. 
But C++ and golang does not need this information. This PR makes the 4 bytes utf16 size header optional for utf8 encoding, so theat the xlang serialiation can use the standard fury string serialization spec, and align to other languages.

For performance consideration, this PR introduce `writeNumUtf16BytesForUtf8Encoding` which can perserve current behaviour.

## Related issues

#1890

## Benchmark

This PR will introduce an extra copy for deserialization since we can't know the size of utf16 in advance before decoding utf8 string.


HuangXingBo (approved):

pandalee99 (approved):
The code is very nice! LGTM!

Issue #1998, opened 2025-01-08
[Java] Deserializing one pojo into another type


chaokunyang:
### Feature Request

Continuing the discussion in https://github.com/apache/fury/issues/1972#issuecomment-2569043551, we need to support deserializing one pojo into another type without the need to register class

### Is your feature request related to a problem? Please describe

#1972

## Step1 Extend ClassDef to replace type
```
  /**
   * Return a new {@link ClassDef} by replacing {@code definedClass} of all fields and current {@link ClassSpec} in hierarchy 
   * from {@code cls} into {@code targetCls}.
   */
  public ClassDef replace(Class<?> cls, Class<?> targetCls) {

  }
```

## Step2 

- Extend ClassResolver with a new readClassInfoWithMetaShare method with passed deserialized type
- The `readClassInfoWithMetaShare` should return the original `ClassDef` first, then invoke `ClassDef#replace` to return a new ClassDef
- Use that new ClassDef to create a new Deserializer to deserialize the data into passed deserialized type

### Additional context

#1996

chaokunyang:
cc @orisgarno

orisgarno:
will start this in few days

orisgarno:
@chaokunyang im currenty working on it.
but do you have any idea why we need the original class(cls) ```public ClassDef replace(Class<?> cls, Class<?> targetCls)```?
we can get from the classid from metashare? or do i miss something here?

chaokunyang:
We could add a method like `replaceRootClassTo(Class <?> cls)`

PR #2012, opened 2025-01-19
feat(java): deserialize one pojo into another type

orisgarno:
## What does this PR do?
replace class def if target class is different type with the actual serialized one, so it can be deserialized to another type
#1998


java/fury-core/src/main/java/org/apache/fury/meta/ClassSpec.java
```diff
@@ -22,7 +22,7 @@
 import org.apache.fury.type.TypeUtils;
 
 public class ClassSpec {
-  public final String entireClassName;
+  public String entireClassName;
```
chaokunyang, line 25: I prefer to keep Classes immutable abd return a new Classes instead

orisgarno: Sure ofc, im still stuck facing exception here and there.

orisgarno:
Hi @chaokunyang, need your help here.
Any idea whats wrong with the approach? seems like its able to change the parent class and serializer but somehow serializer failed to read the buffer properly? do i missed something?
image

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -1674,6 +1674,33 @@ public ClassInfo readClassInfo(MemoryBuffer buffer) {
     return classInfo;
   }
 
+  public ClassInfo readClassInfo(MemoryBuffer buffer, Class<?> targetClass) {
+    if (metaContextShareEnabled) {
+      ClassInfo classInfo = readClassInfoWithMetaShare(buffer, fury.getSerializationContext().getMetaContext());
+
+      // replace target class if needed
+      if (!targetClass.getName().equals(classInfo.getCls().getName())) {
```
chaokunyang, line 1682:
```suggestion
      if (targetClass != classInfo.getCls()) {
```

this will be much faster

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -1674,6 +1674,33 @@ public ClassInfo readClassInfo(MemoryBuffer buffer) {
...
+        ClassInfo targetClassInfo = getClassInfo(targetClass);
+        buildClassDef(targetClassInfo);
+        classInfo.classDef = ClassDef.replaceRootClassTo(this, targetClassInfo);
+        classInfo.serializer = createSerializer(targetClass);
```
chaokunyang, line 1686: This line won't work as you expect. It just create a serializer for raw `targetClass` without using type info from peer class

orisgarno: Ok I'll look into that first

chaokunyang:
I make it work in https://github.com/apache/fury/pull/2012/commits/52dfea2246e61879006e6daf0a112c16eac8a8ec , you can continue the work now

orisgarno:
> @orisgarno I make it work in [52dfea2](https://github.com/apache/fury/commit/52dfea2246e61879006e6daf0a112c16eac8a8ec) , you can continue the work now. And it seems the tests you write have some issues, please take a look at it.

i ran quick check that it doesnt use the generic class from the target. thats why error class cast, will take a look later

chaokunyang:
Hi @orisgarno , could you also update the user guide document about how to deserialize one type into another?

orisgarno:
> Hi @orisgarno , could you also update the user guide document about how to deserialize one type into another?

@chaokunyang this pr making it auto for the root class right? Anything specific you want to add?

chaokunyang:
I think https://github.com/apache/fury/blob/main/docs/guide/java_serialization_guide.md#copingmapping-object-from-one-type-to-another-type needs being updated

orisgarno:
> I think https://github.com/apache/fury/blob/main/docs/guide/java_serialization_guide.md#copingmapping-object-from-one-type-to-another-type needs being updated. Class registration for deserialization is unnecessary

my bad, i have removed the unnecessary part.

chaokunyang (approved):
LGTM

PR #2016, opened 2025-01-21
style(cpp): rename logging level with "FURY_" prefix

An-DJ:
## What does this PR do?

Thus the duplicated macro problem could be solved.

## Benchmark


chaokunyang (approved):
LGTM

PR #2024, opened 2025-01-25
feat(java): support streaming encode/decode to/from buffer for row format

### LINKED ISSUE #2019: [Java][Format] Should Encoder work with ByteBuffer (via MemoryBuffer)?
### Feature Request

We are evaluating integrating Fury row format with a Kafka application. The Kafka deserializer provides a signature
```
T Deserializer.deserialize(String topic, Headers headers, ByteBuffer data)
```
In order to use this with either
```
T Encoder.decode(byte[])
T RowEncoder.fromRow(BinaryRow row)
```
some slightly awkward adaptation is necessary.

In the first case, the adaptation is simple and automatic by a default method, but costs copying your buffer into a `byte[]` unnecessarily.
In the second case, I was able to adapt using the implementation as an example, but it feels like it would be good for the framework to provide this out of the box:

```
    public kafka.Deserializer<Rec> deserializer() {
        final var schema = rowEncoder.schema();
        final var schemaHash = DataTypes.computeSchemaHash(schema);
        return new kafka.Deserializer<Rec>() {
            @Override
            public Rec deserialize(final String topic, final byte[] data) {
                // This case is easy
                return rowEncoder.decode(data);
            }

            @Override
            public Rec deserialize(final String topic, final Headers headers, final ByteBuffer data) {
                // This case... not so much
                final MemoryBuffer buffer = MemoryUtils.wrap(data);
                final long peerSchemaHash = buffer.readInt64();
                if (peerSchemaHash != schemaHash) {
                    throw new ClassNotCompatibleException(
                            String.format(
                                    "Schema is not consistent, encoder schema is %s. "
                                            + "self/peer schema hash are %s/%s. "
                                            + "Please check writer schema.",
                                            schema, schemaHash, peerSchemaHash));
                }
                final BinaryRow row = new BinaryRow(schema);
                row.pointTo(buffer, buffer.readerIndex(), buffer.size());
                return rowEncoder.fromRow(row);
            }
        };
    }
```

For decode, a seemingly unnecessary byte[] copy could be avoided
For fromRow, some repeated low level code could be hidden inside the Fury framework

Add new methods:
```
Encoder.decode(MemoryBuffer buf)
RowEncoder.fromRow(MemoryBuffer buf)
```

It could be interesting to examine whether 
```
Encoder.encodeTo(T obj, MemoryBuffer buf)
```
is beneficial to add to avoid intermediate `byte[]` there too.

_No response_

### Additional context

Thank you for your consideration! We hope to adopt Fury into our application if the prototyping works out.

chaokunyang:
## What does this PR do?

support streaming encode/decode to/from buffer for row format

## Related issues

Closes #2019

### DIFF
--- a/java/fury-format/src/main/java/org/apache/fury/format/encoder/ArrayEncoderBuilder.java
+++ b/java/fury-format/src/main/java/org/apache/fury/format/encoder/ArrayEncoderBuilder.java
@@ -127,7 +127,8 @@ public class ArrayEncoderBuilder extends BaseBinaryEncoderBuilder {
     expressions.add(array);
 
     Expression.Reference fieldExpr = new Expression.Reference(FIELD_NAME, ARROW_FIELD_TYPE, false);
-    Expression listExpression = serializeForArray(array, arrayWriter, arrayToken, fieldExpr);
+    Expression listExpression =
+        serializeForArrayByWriter(array, arrayWriter, arrayToken, fieldExpr);
 
     expressions.add(listExpression);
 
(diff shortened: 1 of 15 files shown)


pandalee99 (approved):
LGTM,the perf very nice!

Issue #2031, opened 2025-01-28
[Java] Build a provider contains several fury config as a basic dataProvider for unit test and create a basic unit test use this dataProvider


Hen1ng:
### Feature Request

Create a comprehensive testcase that includes various branching conditions for map testing, covering all types of maps, and consolidate them using the same provider. This way, when testing, you only need to modify the configuration of one provider to run tests on all types of maps.

PR #2032, opened 2025-01-28
feat(java): Add basicMultiConfigFury dataprovider and add basic map unit test use the dataProvider

Hen1ng:
#2031 


java/fury-core/src/test/java/org/apache/fury/FuryTestBase.java
```diff
@@ -245,6 +245,29 @@ public static Object[][] javaFuryKVCompatible() {
     };
   }
 
+  @DataProvider
+  public static Object[][] basicMultiConfigFury() {
```
chaokunyang, line 249: Could We use guava cartesianProduct instead?

chaokunyang (approved):

Issue #1956, opened 2024-11-23
[Java] Type annotation hints for serialization


chaokunyang:
### Feature Request

We could provide some extra type information to mmake fury serialization faster.

The annotation can be like following:

```java
@TypeInfo(fieldsNullable = false, trackingRef = false, polymorphic = false)
class Foo {
  @FieldInfo(trackingRef = false)
  int[] intArray;
  @FieldInfo(polymorphic = true)
  Object object;
  @FieldInfo(tagId = 1, nullable = true)
  List<Object> objectList;
}
```

Or

```java
@Type(fieldsNullable = false, trackingRef = false, polymorphic = false)
class Foo {
  @Field(trackingRef = false)
  int[] intArray;
  @Field(polymorphic = true)
  Object object;
  @Field(tagId = 1, nullable = true)
  List<Object> objectList;
}


```

bank-ro:
Hi all : quick ping on `@ForyField(tagId = 1)`  support. Has _tagId_ annotation handling been merged yet?

We’re eager to start using it to trim schema metadata overhead (especially on objects with many fields).

From the serialization format documentation:
> If tag id is used, i.e. field name is written by an unsigned varint tag id.

chaokunyang:
Hi @bank-ro , this has been supported in #3021 and #3093

PR #2036, opened 2025-02-05
feat(java): FuryField annotation hints for struct serialization

Hen1ng:
## Related issues
#1956 


java/fury-core/src/main/java/org/apache/fury/annotation/FieldInfo.java
```diff
@@ -0,0 +1,21 @@
...
+
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface FieldInfo {
```
chaokunyang, line 11: Could we name it as `FuryField`, just like `JsonField`？

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -398,27 +402,55 @@ private static FinalTypeField buildFinalTypeField(Fury fury, Descriptor d) {
...
   }
 
   private static GenericTypeField buildContainerField(Fury fury, Descriptor d) {
     return new GenericTypeField(
         d.getTypeRef(),
         d.getDeclaringClass() + "." + d.getName(),
         d.getField() != null ? FieldAccessor.createAccessor(d.getField()) : null,
-        fury);
+        fury,
+        d.getField() != null ? d.getField().getAnnotation(FuryField.class) : null);
```
chaokunyang, line 415: How about parse Annotation in Descriptor? And pass descriptor into `GenericTypeField`

chaokunyang: We could pass Descriptor into GenericTypeField instead of pass `FieldAccessor`. Then creating `FieldAccessor` in `GenericTypeField`

java/fury-core/src/main/java/org/apache/fury/serializer/CompatibleSerializer.java
```diff
@@ -599,14 +616,15 @@ private Object readFieldValue(FieldResolver.FieldInfo fieldInfo, MemoryBuffer bu
       case ClassResolver.PRIMITIVE_DOUBLE_CLASS_ID:
         return buffer.readFloat64();
       case ClassResolver.STRING_CLASS_ID:
-        return fury.readJavaStringRef(buffer);
+        return fury.readJavaStringRef(buffer, fieldInfo.getFuryFieldAnnotationInfo().nullable);
```
chaokunyang, line 619: Could we add a nullbale field into `fieldInfo` directly to avoid two indirections?

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
```diff
@@ -495,104 +548,165 @@ static boolean writePrimitiveFieldValueFailed(
...
     if (!fury.isBasicTypesRefIgnored()) {
       return true; // let common path handle this.
     }
+    boolean nullable = fieldAnnotationInfo.nullable;
     // add time types serialization here.
     switch (classId) {
-      case ClassResolver.STRING_CLASS_ID: // fastpath for string.
-        fury.writeJavaStringRef(buffer, (String) (fieldValue));
+      case ClassResolver.STRING_CLASS_ID: // fast path for string.
+        fury.writeNullableJavaStringRef(buffer, (String) (fieldValue), nullable);
```
chaokunyang, line 563: This makes writeNullableJavaStringRef too complicated. How about:

```java 
{
String v =  (String) (fieldValue);
if (nullable) {
  fury.writeJavaStringRef(buffer, v);
} else {
  fury.writeString(buffer, v);
}
return false;
}
```

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
```diff
@@ -495,104 +548,165 @@ static boolean writePrimitiveFieldValueFailed(
    * @return true if field value isn't written by this function.
    */
   static boolean writeBasicObjectFieldValueFailed(
-      Fury fury, MemoryBuffer buffer, Object fieldValue, short classId) {
+      Fury fury,
```
chaokunyang, line 551: How about adding a new `writeBasicNullableObjectFieldValueFailed` method? In this method, we handle nullability, and in `writeBasicObjectFieldValueFailed`, all fields are not null

chaokunyang:
We should also support annotation in `org.apache.fury.builder.BaseObjectCodecBuilder`, it can be left in a follow-up PR

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
chaokunyang: `org.apache.fury.builder.BaseObjectCodecBuilder` also needs to read this annotation data

java/fury-core/src/main/java/org/apache/fury/annotation/FuryField.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface FuryField {
+
+  /** Whether field is nullable. */
+  boolean nullable() default true;
```
chaokunyang, line 32: How about setting default value to false

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
Hen1ng: Call fury.writeString(buffer, v) separately; This method will ignore the case that stringSerializer.needToWriteRef() is true, resulting in this refResolver.writeRefOrNull(buffer, str) method not being called. Is this as expected?

chaokunyang: How about
```java
if (nullable || trackStringRef) {
  fury.writeJavaStringRef(buffer, v);
} else {
  fury.writeString(buffer, v);
}
```

chaokunyang:
@Hen1ng This PR has conflict with main branch, could you merge main branch into this branch and resolve conflict?

Hen1ng:
> We should also support annotation in `org.apache.fury.builder.BaseObjectCodecBuilder`, it can be left in a follow-up PR

I will

java/fury-core/src/main/java/org/apache/fury/annotation/FuryField.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface FuryField {
+
+  /** Whether field is nonNull, default false. */
+  boolean nonNull() default false;
```
chaokunyang, line 32: I prefer naming it to `nullable` and default value is false

java/fury-core/src/main/java/org/apache/fury/annotation/FuryField.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface FuryField {
+
+  /** Whether field is nonNull, default false. */
```
chaokunyang, line 31: Update javadoc too if renamed to nullable

java/fury-core/src/main/java/org/apache/fury/resolver/FieldResolver.java
```diff
@@ -748,6 +749,7 @@ public static class FieldInfo {
     protected final ClassResolver classResolver;
     private final FieldAccessor fieldAccessor;
     private final ClassInfoHolder classInfoHolder;
+    private boolean nonNull;
```
chaokunyang, line 752:
```suggestion
    private boolean nullable;
```

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -107,30 +108,71 @@ static Object readFinalObjectFieldValue(
...
+      Serializer<Object> serializer,
+      boolean nonNull) {
+    if (!nonNull) {
+      return binding.readNullable(buffer, serializer);
+    } else {
+      return binding.read(buffer, serializer);
+    }
+  }
+
+  static void writeNullable(
```
chaokunyang, line 136: How about adding those methods in `SerializationBinding` interface?

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -146,14 +188,16 @@ static Object readContainerFieldValue(
...
     } else {
-      byte headFlag = buffer.readByte();
-      if (headFlag == Fury.NULL_FLAG) {
-        fieldValue = null;
-      } else {
-        generics.pushGenericType(fieldInfo.genericType);
-        fieldValue = binding.readContainerFieldValue(buffer, fieldInfo);
-        generics.popGenericType();
+      boolean nonNull = fieldInfo.nonNull;
+      if (!nonNull) {
```
chaokunyang, line 192:
```suggestion
      if (fieldInfo.nullable) {
```

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -954,12 +1110,20 @@ public String toString() {
...
           + ", classId="
           + classId
-          + ", qualifiedFieldName="
+          + ", qualifiedFieldName='"
           + qualifiedFieldName
-          + ", field="
-          + (fieldAccessor != null ? fieldAccessor.getField() : null)
+          + ", fieldAccessor="
+          + fieldAccessor
+          + ", nonNull="
```
chaokunyang, line 1125:
```suggestion
          + ", nullable="
```

java/fury-core/src/test/java/org/apache/fury/FuryTest.java
```diff
@@ -355,6 +357,148 @@ public void testSerializePackageLevelBeanJIT() {
     serDeCheckSerializer(fury, o, "PackageLevelBean");
   }
 
+  @Data
+  public static class BeanM {
```
chaokunyang, line 361: Could we add move it into FuryAnnotationTest.java?

java/fury-core/src/main/java/org/apache/fury/serializer/AbstractObjectSerializer.java
```diff
@@ -826,69 +962,50 @@ protected T newBean() {
...
-        TypeRef<?> typeRef, short classId, String qualifiedFieldName, FieldAccessor fieldAccessor) {
-      this.typeRef = typeRef;
+    private InternalFieldInfo(Descriptor d, short classId) {
+      this.typeRef = d.getTypeRef();
       this.classId = classId;
-      this.qualifiedFieldName = qualifiedFieldName;
-      this.fieldAccessor = fieldAccessor;
+      this.qualifiedFieldName = d.getDeclaringClass() + "." + d.getName();
+      this.fieldAccessor = d.getField() != null ? FieldAccessor.createAccessor(d.getField()) : null;
+      this.nullable = d.getFuryField() == null || d.getFuryField().nullable();
```
chaokunyang, line 993: Need to check whether type of field is primitive

java/fury-core/src/main/java/org/apache/fury/serializer/NonexistentClassSerializers.java
```diff
@@ -138,10 +138,12 @@ public void write(MemoryBuffer buffer, Object v) {
       }
       for (ObjectSerializer.GenericTypeField fieldInfo : fieldsInfo.otherFields) {
         Object fieldValue = value.get(fieldInfo.qualifiedFieldName);
+        boolean nullable = fieldInfo.nullable;
+        // todo question why not use binding
```
chaokunyang, line 142: We do need to use binding, I forget to update it

chaokunyang: Maybe we could left that in a new PR

chaokunyang (approved):

Issue #1935, opened 2024-11-07
[Python][Protocol] Chunk by chunk predictive map serialization protocol 


chaokunyang:
### Feature Request

 Chunk by chunk predictive map serialization protocol  can be 2x faster than current one in pyfury. we should implement this new protocol.

See #925 for more details

### Describe the solution you'd like

https://fury.apache.org/docs/specification/fury_xlang_serialization_spec/#map has a formulized spec:

> All Map serializers must extend `AbstractMapSerializer`.

Format:

```
| length(unsigned varint) | key value chunk data | ... | key value chunk data |
```

#### map key-value chunk data

Map iteration is too expensive, Fury won't compute the header like for list since it introduce
[considerable overhead](https://github.com/apache/fury/issues/925).
Users can use `MapFieldInfo` annotation to provide the header in advance. Otherwise Fury will use first key-value pair
to predict header optimistically, and update the chunk header if the prediction failed at some pair.

Fury will serialize the map chunk by chunk, every chunk has 255 pairs at most.

```
|    1 byte      |     1 byte     | variable bytes  |
+----------------+----------------+-----------------+
| chunk size: N  |    KV header   |   N*2 objects   |
```

KV header:

- If track key ref, use the first bit `0b1` of the header to flag it.
- If the key has null, use the second bit `0b10` of the header to flag it. If ref tracking is enabled for this
  key type, this flag is invalid.
- If the key types of map are different, use the 3rd bit `0b100` of the header to flag it.
- If the actual key type of the map is not the declared key type, use the 4rd bit `0b1000` of the header to flag it.
- If track value ref, use the 5th bit `0b10000` of the header to flag it.
- If the value has null, use the 6th bit `0b100000` of the header to flag it. If ref tracking is enabled for this
  value type, this flag is invalid.
- If the value types of the map are different, use the 7rd bit `0b1000000` header to flag it.
- If the value type of map is not the declared value type, use the 8rd bit `0b10000000` of the header to flag it.

If streaming write is enabled, which means Fury can't update written `chunk size`. In such cases, map key-value data
format will be:

```
|    1 byte      | variable bytes  |
+----------------+-----------------+
|    KV header   |   N*2 objects   |
```

`KV header` will be a header marked by `MapFieldInfo` in java. For languages such as golang, this can be computed in
advance for non-interface types most times. The implementation can generate different deserialization code based read
header, and look up the generated code from a linear map/list.

#### Why serialize chunk by chunk?

[...]

PR #2037, opened 2025-02-05
feat(python):  chunk based map serialization in pure python

pandalee99:
## What does this PR do?

I implemented chunk based map serialization in pure python, and there are some differences from cython. and, cython has strict type detection, here I added put_uint8 and other methods, and removed some fast path.

## Related issues
#1935


python/pyfury/_serializer.py
```diff
@@ -285,6 +287,7 @@ class MapSerializer(Serializer):
         "ref_resolver",
         "key_serializer",
         "value_serializer",
+        "use_chunk_serialize",
```
chaokunyang, line 290: this is unnecessary, we can just use chunk serialization directly

pandalee99: Good idea

python/pyfury/_serializer.py
```diff
@@ -294,39 +298,219 @@ def __init__(self, fury, type_, key_serializer=None, value_serializer=None):
...
+            buffer.put_int8(chunk_size_offset - 1, chunk_header)
+            key_serializer_type = type(key_serializer)
+            value_serializer_type = type(value_serializer)
+            chunk_size = 0
+            while True:
+                if (key is None or value is None or
+                        type(key) is not key_cls or type(value) is not value_cls):
+                    break
+                if not key_write_ref or not ref_resolver.write_ref_or_null(buffer, key):
+                    if key_cls is str:
```
chaokunyang, line 382: The pure python implementation is used for debug only, we can remove all fastpath, just keep the most simple implementation

pandalee99: But users might not be smart enough, and if there's a bit of inconsistency in the code, they won't be able to clearly recognize the problem. What do you think?

chaokunyang: Users will use cython mode only, python mode is used only when there are some deserialization error which  needs debug line by line

python/pyfury/_fury.py
```diff
@@ -82,11 +82,15 @@
 # Key is null, value type is declared type, and ref tracking for value is disabled.
 NULL_KEY_VALUE_DECL_TYPE = KEY_HAS_NULL | VALUE_DECL_TYPE
 # Key is null, value type is declared type, and ref tracking for value is enabled.
-NULL_KEY_VALUE_DECL_TYPE_TRACKING_REF = KEY_HAS_NULL | VALUE_DECL_TYPE | TRACKING_VALUE_REF
+NULL_KEY_VALUE_DECL_TYPE_TRACKING_REF = (
```
chaokunyang, line 85: could we reserve those parameters in _serializer.py, it's used by map serializer only

chaokunyang:
image
chunk map serialization still fail when cython disabled @pandalee99

pandalee99:
hi, shawn! @chaokunyang 
In the pure python, MapSerializer is  be done.

But there are a error that are not about MapSerializer. Because of the add of `ci/run_ci.sh`, there are error left over from history `test_serialize_arrow():`, can you help me solve them?

image

chaokunyang:
Hi @pandalee99 , please replace read_buffer_object with:
```python
    def read_buffer_object(self, buffer) -> Buffer:
        in_band = buffer.read_bool()
        if in_band:
            size = buffer.read_varuint32()
            buf = buffer.slice(buffer.reader_index, size)
            buffer.reader_index += size
            return buf
        else:
            assert self._buffers is not None
            return next(self._buffers)
```

chaokunyang (approved):
LGTM

PR #2042, opened 2025-02-06
feat(java): zstd meta compressor

orisgarno:
## What does this PR do?
create zstd metacompressor as an option. let zstd access the src arr instead of copy to new array.


chaokunyang:
Could we add a new fury java module named as `fury-extension`, and move those code to that module? In this way, we could reduce the dependencies in fury-core

java/fury-core/src/main/java/org/apache/fury/meta/ZstdMetaCompressor.java
```diff
@@ -0,0 +1,68 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fury.meta;
+
+import com.github.luben.zstd.Zstd;
+import com.github.luben.zstd.ZstdException;
+
+public class ZstdMetaCompressor implements MetaCompressor {
```
chaokunyang, line 25: Could we add a test with `org.apache.fury.config.FuryBuilder#withMetaCompressor` set to this compressor?

orisgarno: do you mean this?
java/fury-extensions/src/test/java/org/apache/fury/meta/ClassDefEncoderTest.java
or any specific in mind?

java/fury-core/pom.xml
```diff
@@ -58,6 +58,10 @@
       <artifactId>slf4j-api</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>com.github.luben</groupId>
+      <artifactId>zstd-jni</artifactId>
+    </dependency>
```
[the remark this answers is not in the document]
chaokunyang: Yes, d1a2930 is OK. Could you remove `zstd-jni` from fury-core module, it's used in fury-extensions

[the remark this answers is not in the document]
orisgarno: Oops forgot to remove that. Done

java/fury-extensions/src/main/java/org/apache/fury/meta/ZstdMetaCompressor.java
```diff
@@ -0,0 +1,68 @@
...
+    Zstd.compressByteArray(
+        compressedData,
+        0,
+        (int) maxCompressedSize,
+        data,
+        offset,
+        size,
+        Zstd.defaultCompressionLevel());
+
+    return compressedData;
```
chaokunyang, line 45:
```suggestion
    long size = Zstd.compressByteArray(
        compressedData,
        0,
        (int) maxCompressedSize,
        data,
        offset,
        size,
        Zstd.defaultCompressionLevel());

    return Arrays.copyOf(compressedData, (int) size);
```

orisgarno: Can you share why its better to have its copies? @chaokunyang

chaokunyang: I am not sure whether `maxCompressedSize` is exactly the compressed size fo zstd, or it's a n estimated  max size

orisgarno: After read zstd code, I got the idea why. Nice catch. 
Lgtm

orisgarno: Let me check for the decompress as well

orisgarno: Updated

java/pom.xml
```diff
@@ -105,6 +107,11 @@
         <artifactId>janino</artifactId>
         <version>${janino.version}</version>
       </dependency>
+      <dependency>
```
chaokunyang, line 110: could we move this dependency version info into fury-extensions module? we try to keep this pom simplistic

java/pom.xml
```diff
@@ -81,6 +82,7 @@
     <maven.compiler.target>1.8</maven.compiler.target>
     <guava.version>32.1.2-jre</guava.version>
     <janino.version>3.1.10</janino.version>
+    <zstd.version>1.5.6-9</zstd.version>
```
chaokunyang, line 85: ditto

chaokunyang (approved):
LGTM

Issue #162, opened 2023-05-09
[Java] Serialization support for java.util.BitSet


chaokunyang:
**Is your feature request related to a problem? Please describe.**
Add serialization support for java.util.BitSet

PR #2050, opened 2025-02-08
feat(java):Add fastpath serialize and deserialize

Hen1ng:
## What does this PR do?

Add fastpath serialize and deserialize for hashmap and arraylist type


java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -604,6 +604,16 @@ private void writeData(MemoryBuffer buffer, ClassInfo classInfo, Object obj) {
       case ClassResolver.STRING_CLASS_ID:
         stringSerializer.writeJavaString(buffer, (String) obj);
         break;
+      case ClassResolver.ARRAYLIST_CLASS_ID:
```
chaokunyang, line 607: The `STRING_CLASS_ID` and `ARRAYLIST_CLASS_ID` is not continuous, it may make the switch clause degenerate  from switch jump, which make the execution slower.

Maybe we should remote this fastpath from copyObject too

Hen1ng: i made writeData public, and use javap -c Fury.class. Obtain the following result
public void writeData(org.apache.fury.memory.MemoryBuffer, org.apache.fury.resolver.ClassInfo, java.lang.Object);
    Code:
       0: aload_2
       1: invokevirtual #162                // Method org/apache/fury/resolver/ClassInfo.getClassId:()S
       4: tableswitch   { // 14 to 34
                    14: 104
                    15: 118
                    16: 132
                    17: 146
                    18: 160
                    19: 196
                    20: 210
                    21: 228
                    22: 242
                    23: 327
                    24: 327
                    25: 327
                    26: 327
                    27: 327
                    28: 327
                    29: 327
                    30: 327
                    31: 327
                    32: 327
                    33: 257
                    34: 292
               default: 327
          }

string class id is 22, arraylist classid is 33, and hashmap class id is 34, the result shows that JVM automatically fills the gaps before 22-33 and still uses the tableswitch bytecode instruction, which has a complexity of O (1)

chaokunyang: Awesome, I didn't know javac can do such optimization

chaokunyang (approved):

PR #2056, opened 2025-02-11
perf(python): upgrade pyarrow version to 15.0.0

pandalee99:
## What does this PR do?

 pip install -v -e .
Using pip 22.0.4 from D:\code\PycharmProjects\project\venv\lib\site-packages\pip (python 3.9)
Obtaining file:///D:/code/fury/fury/python
  Running command python setup.py egg_info

  A module that was compiled using NumPy 1.x cannot be run in
  NumPy 2.0.2 as it may crash. To support both 1.x and 2.x
  versions of NumPy, modules must be compiled with NumPy 2.0.
  Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

  If you are a user of the module, the easiest solution will be to
  downgrade to 'numpy<2' or try to upgrade the affected module.
  We expect that some modules will need time to support NumPy 2.

  Traceback (most recent call last):  File "<string>", line 2, in <module>
    File "<pip-setuptools-caller>", line 34, in <module>
    File "D:\code\fury\fury\python\setup.py", line 32, in <module>
      import pyarrow as pa
    File "D:\code\PycharmProjects\project\venv\lib\site-packages\pyarrow\__init__.py", line 65, in <module>
      import pyarrow.lib as _lib
  AttributeError: _ARRAY_API not found
  Traceback (most recent call last):
    File "<string>", line 2, in <module>
    File "<pip-setuptools-caller>", line 34, in <module>
    File "D:\code\fury\fury\python\setup.py", line 32, in <module>
      import pyarrow as pa
    File "D:\code\PycharmProjects\project\venv\lib\site-packages\pyarrow\__init__.py", line 65, in <module>
      import pyarrow.lib as _lib
    File "pyarrow\lib.pyx", line 36, in init pyarrow.lib
  ImportError: numpy.core.multiarray failed to import
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> See above for output.

  note: This error originates from a subprocess, and is likely not a problem with pip.


penguin-wwy:
FYI, Python3.13 has updated `pyarrow==18.0.0`, which requires `numpy > 2.0`. This should be added to the README.

python/README.md
```diff
@@ -6,7 +6,10 @@ Fury is a blazingly-fast multi-language serialization framework powered by just-
 
 ```bash
 cd python
-pip install pyarrow==14.0.0 Cython wheel numpy pytest
+# For Python versions less than 3.13, numpy 2 is not currently supported.
+pip install pyarrow==14.0.0 Cython wheel 'numpy<2' pytest
```
chaokunyang, line 10: Could we uninstall numpy first, then when we install pyarrow, it will install correct numpy automatically.

We could left a comment about why we need to uninstall numpy first

pandalee99: 'then when we install pyarrow, it will install correct numpy automatically.'
I don't think so.
when i try to :
``` cmd
pip uninstall -y numpy
pip install pyarrow==14.0.0 numpy 
```

image

but, if numpy2 is existing, 
make cmd:
``` cmd
pip install pyarrow==14.0.0 'numpy<2' 
```

image

it will :
Installing collected packages: numpy, pyarrow
  Attempting uninstall: numpy
    Found existing installation: numpy 2.0.2
    Uninstalling numpy-2.0.2:
      Successfully uninstalled numpy-2.0.2
Successfully installed numpy-1.26.4 pyarrow-14.0.0


After practice, I think this is OK

chaokunyang: pyarrow 15.0.0 has specified that numpy version should be less than 2. https://github.com/apache/arrow/blob/a61f4af724cd06c3a9b4abd20491345997e532c0/python/setup.py#L452

maybe we could upgrade arrow version in java and python

pandalee99: as you wish

chaokunyang:
github ci config y'all needs being updated too @pandalee99

chaokunyang (approved):

PR #2058, opened 2025-02-13
feat(javascript): simplify the definition of object type description

theweipeng:
## What does this PR do?
It is hard to describe the type of a object, we should write some code like this:
```JavaScript
Type.object("example.foo", {
    a: Type.string()
})
```
 The `Type.xxx` APIs are very flexible but not easy to use.

This PR add improve these APIS to help simplifing the type definition. Now we can definition object type like this:

```JavaScript
    @Type.object("example.foo")
    class Foo {
      @Type.int32()
      a: number;
    }

    const { serialize, deserialize } = fury.registerSerializer(Foo);
    const foo = new Foo();
    foo.a = 123;
    const input = serialize(foo);
    const result = deserialize(
      input
    );

    expect(result instanceof Foo);
    expect(result).toEqual({ a: 123 })

```

The result of  `Type.xxx()` serves both as a decorator function and as a description object. It is totally compatible with `Type.object("example.foo", {
    a: Type.string()
})`

Note that decoration APIs can not detect the type of a anonymous object. providing anonymous object like `{a: 123}` still will cause error.


chaokunyang (approved):

Forchapeatl:
@theweipeng , please open more javascript issues

theweipeng:
> @theweipeng , please open more javascript issues

OK， I will open more

PR #2075, opened 2025-02-19
feat(javascript): Impl xlang

theweipeng:
## What does this PR do?
1.  Rename `TypeDescription` to  `ClassInfo`  to keep consistent with Java/Python.
2.  Replace `Tag` by `Namespace` and `TypeName` which is used for marking structs in xlang protocol.
3. Add MetaStringResolver to manage  the reading and writing of`Namespace` and `TypeName`.
4. Add  `TypeMetaResolver` which is defined in (https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#type-meta). Note that this part of protocol may change in the furture.
5. Add `Evoluation Mode` which is defined in (https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#schema-evolution), Now we can deserialize in `Evoluation Mode`
6. Add a `constructClass` config to control the construction of Class. Constructor often introduces a remote code execution vulnerability, if `constructClass` is disabled, we will construct the object to a anonymous object without call the constructor of the class.
7.   Refactor the MetaStringEncoder MetaStringDecoder and MetaString to make it perfectly consistent with Java/Python
8. Bump version to 0.11.0-SNAPSHOT

constructClass Disabled:
```javascript
    @Type.struct("example.foo")
    class Foo {
      @Type.int32()
      a: number;
    }
    fury.registerSerializer(Foo);

    const foo = new Foo({constructClass: false});
    foo.a = 123;
    const input = fury.serialize(foo)
    const result = fury.deserialize(input);
    expect(!(result instanceof Foo)); // The result is constructed from Object, not Foo.
    expect(result).toEqual({ a: 123 }) // We can access the props normality
```

Evoluation Usage:

```javascript
    const fury = new Fury({
        mode: Mode.Compatible
    });    

    @Type.struct("example.foo")
    class Foo {
        @Type.string()
        bar: string;

        @Type.int32()
        bar2: number;

        setBar(bar: string) {
            this.bar = bar;
            return this;
        }

        setBar2(bar2: number) {
            this.bar2 = bar2;
            return this;
        }
    }

    const { serialize } = fury.registerSerializer(Foo);
    const bin = serialize(new Foo().setBar("hello").setBar2(123));


    @Type.struct("example.foo")
    class Foo2 {
        @Type.string()
        bar: string;
    }

    const fury2 = new Fury({
        mode: Mode.Compatible,
        hooks: {
            afterCodeGenerated: (code: string) => {
                return beautify.js(code, { indent_size: 2, space_in_empty_paren: true, indent_empty_lines: true });
              }        
            }
    });    
    const { deserialize  } = fury2.registerSerializer(Foo2);
    const r = deserialize(bin);
    expect(r).toEqual({
        bar: "hello",
        bar2: 123,
    })
```

 It's too much code for a single PR, Todos in the codebase will be resolve by other PRS.


javascript/packages/fury/lib/classInfo.ts
```diff
@@ -0,0 +1,672 @@
...
+    super();
+    return Object.setPrototypeOf(f, new.target.prototype);
+  }
+}
+
+/**
+ * T is for type matching
+ */
+// eslint-disable-next-line
+export class ClassInfo<T = unknown> extends ExtensibleFunction {
```
chaokunyang, line 66: How about TypeInfo? Seems it's more general?

chaokunyang (approved):
LGTM

PR #2083, opened 2025-02-26
feat(javascript): Speed up string serializer

theweipeng:
## What does this PR do?
The PR (https://github.com/nodejs/node/commit/6cb0690fccd6ed902f785edd1ae5b55855035998) for string detection committed in the node repository is merged. Now we can speed up string processing without hps module. 
There are now three ways of string serialization, the first is with hps module which has super high performance; the second is for higher versions of nodejs which can detect string encoding at zero cost; the third is the normal way which is used for lower version of nodejs and browser env.


chaokunyang (approved):
LGTM

PR #2101, opened 2025-03-15
perf(java): Improve performance by using System.arraycopy to copy between byte arrays

Cybermaxke:
## What does this PR do?

Try to use `System.arraycopy` when copying between byte arrays. This significantly increases performance, in our use case serialization is twice as fast (serializing/deserializing 1M complex strucutures using (de)serializeJavaObject in parallel). When using Platform.copyMemory, the JDK is doing a lot of type checking to see if it's a byte array, causing a visible slowdown. `Unsafe.memoryCopy` implementation was changed in Java 9 from a native method to a method calling the internal unsafe method containing the reflection checks.

There are a few places where System.arraycopy is already used, referring to it being faster on certain JDK implementations. Tested our use case on a macbook with arm64 cpu against zulu 21 and open-jdk 23. 

image

## Benchmark
I tried to reproduce the performance impact through one on existing benchmarks `UserTypeSerializeSuite.fury_serialize`, however I didn't see a difference here.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
@@ -313,17 +313,23 @@ public final class MemoryBuffer {
   }
 
   public void get(int index, byte[] dst, int offset, int length) {
-    final long pos = address + index;
-    if ((index
-            | offset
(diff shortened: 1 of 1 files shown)


chaokunyang (approved):
LGTM

PR #2102, opened 2025-03-16
feat(java): use sha256 to check disallowed.txt tamper

### LINKED ISSUE #2100: [Java] add check for multiple disallowed.txt
### Feature Request

A malicious lib provider could sneakily include a fury/disallowed.txt file in their lib. Users might have this malicious lib on their classpath. Could override the fury built-in one.

_No response_

_No response_

_No response_

### Additional context

_No response_

chaokunyang:
## What does this PR do?

 use sha256 to check disallowed.txt tamper
## Related issues

Closes #2100 

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/DisallowedList.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/DisallowedList.java
@@ -34,6 +34,7 @@ import org.apache.fury.exception.InsecureException;
 /** A class to record which classes are not allowed for serialization. */
 class DisallowedList {
   private static final String DISALLOWED_LIST_TXT_PATH = "fury/disallowed.txt";
+  // when disallowed.txt changed, update this hash by print the result of `calculateSHA256`
   private static final String SHA256_HASH =
       "d418999c49b0aa83b8bde8a79aa758a16adb5599e384f842db65dbcd633c541b";
   private static final Set<String> DEFAULT_DISALLOWED_LIST_SET;
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/DisallowedList.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/DisallowedList.java
@@ -39,10 +39,6 @@ class DisallowedList {
       "d418999c49b0aa83b8bde8a79aa758a16adb5599e384f842db65dbcd633c541b";
   private static final Set<String> DEFAULT_DISALLOWED_LIST_SET;
 
-  public static void main(String[] args) {
-    System.out.println(calculateSHA256(DEFAULT_DISALLOWED_LIST_SET));
-  }
-
   static {
     try (InputStream is =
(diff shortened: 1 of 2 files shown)


java/fury-core/src/main/java/org/apache/fury/resolver/DisallowedList.java
```diff
@@ -24,15 +24,25 @@
...
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 import org.apache.fury.exception.InsecureException;
 
 /** A class to record which classes are not allowed for serialization. */
 class DisallowedList {
   private static final String DISALLOWED_LIST_TXT_PATH = "fury/disallowed.txt";
+  // when disallowed.txt changed, update this hash by print the result of `calculateSHA256`
+  private static final String SHA256_HASH =
+      "d418999c49b0aa83b8bde8a79aa758a16adb5599e384f842db65dbcd633c541b";
```
pjfanning, line 39: could this be calculated by the build process and this class generated?

pjfanning: * it's also probably true that if the build fails after fury/disallowed.txt is modified (without the SHA upgrade) then that should be enough
* could we remove the main method just have the a unit test that compares the generated SHA to the hardcoded one?
* main methods in classes is not usually regarded as good
* the test could be run instead to get the new SHA, so a coder could grab that and updated the hardcoded SHA

chaokunyang: I removed the main method. The chekc happs when class init. If check failed, the class init will fail, existing tests in DisallowedListTest will all fail too. Seems adding new tests is unnecesary @pjfanning 

caculate the hash at build process will make the build complex. Since this is not updated frequently, how about updating it manually every time?

java/fury-core/src/main/java/org/apache/fury/resolver/DisallowedList.java
```diff
@@ -52,6 +61,33 @@ class DisallowedList {
     }
   }
 
+  private static String calculateSHA256(Set<String> disallowedSet) {
```
pjfanning, line 64: I think this would simpler if calculated the digest of the file bytes. That preserves the line endings so no need to add back separators.

java/fury-core/src/main/java/org/apache/fury/resolver/DisallowedList.java
pjfanning: Sure

java/fury-core/src/main/java/org/apache/fury/resolver/DisallowedList.java
chaokunyang: changed to use file. new contributors can use `sha256sum disallowed.txt` to get sum

pandalee99 (approved):

pjfanning (approved):
lgtm

Issue #2107, opened 2025-03-18
[Python] Inconsistent struct hash calculation between Java and Python implementations


LouisLou2:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

Commit: bc6a0b586aa2ab99e3a0059ee645c7b9901e92a3

### Component(s)

Python

### Minimal reproduce step

1. Check out the repository at commit bc6a0b586aa2ab99e3a0059ee645c7b9901e92a3
2. Run the CrossLanguageTest.java tests, specifically the testSerializeSimpleStruct case
3. The test fails with an error indicating hash inconsistency between Java and Python:
```
ERROR:root:Execute ['test_serialize_simple_struct', '/home/leo/projects/fury_newest_dev/fury/java/fury-core/test_serialize_simple_struct'] failed with Hash 16360 is not consistent with 550 for class <class '__main__.ComplexObject2'>
```

This issue occurs across all sub-test cases in CrossLanguageTest.java, with each test failing due to struct hash inconsistency during cross-language communication.

### What did you expect to see?

The struct hash calculation should be consistent between Java and Python implementations, allowing cross-language serialization tests to pass successfully.

### What did you see instead?

Tests fail with hash inconsistency errors like:
```
pyfury.error.ClassNotCompatibleError: Hash 16360 is not consistent with 550 for class <class '__main__.ComplexObject2'>
```

### Anything Else?

I manually modified the Python source code to bypass the hash verification, and the subsequent steps completed without issues. This indicates that while the hash calculation is inconsistent, the actual data structure and serialization format are compatible between the two implementations.

This suggests it could be a bug in the hash calculation algorithm implementation in one or both languages, or possibly that the Python implementation is lagging behind the Java implementation in terms of feature parity or updates to the hashing algorithm.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #2108, opened 2025-03-19
fix(python): inconsistent struct hash calculation between Java and Python

LouisLou2:
### Problem
The struct hash calculation implementations in Java and Python are inconsistent. When `classinfo` is `None` in the Python implementation, the method returns early without updating the hash value, whereas the Java implementation continues with a default hash value of 0.

### Solution
Modified the `visit_customized` method in the Python implementation to initialize `hash_value` to 0 before checking if `classinfo` is `None`, ensuring the hash computation continues regardless of the `classinfo` status. This approach aligns with the Java implementation behavior.

Changes:
- Initialize `hash_value` to 0 at the beginning of the method
- Replace the early return with a conditional assignment to `hash_value`
- Always call `_compute_field_hash` at the end of the method

### Testing
Verified the fix by running the `CrossLanguageTest.java` tests that were previously failing due to this issue. The struct hash values now match between Java and Python implementations.

## Related issues
- #2107 


chaokunyang:
Code lint check failed, please use `black` to format _struct.py

chaokunyang (approved):
Looks great

PR #2136, opened 2025-04-01
fix(java): fix nested chunk map serialization error when generics exists

### LINKED ISSUE #2135: chunk based map serialization doesn't work when codegen disabled
### Search before asking

### Version

main

### Component(s)

Java

### Minimal reproduce step
I've tested with an instance of a rather complex class. Same results were observed with the following super simple class:
```
public class CustomClass implements Serializable {
	Map<String, Object> values = new LinkedHashMap<>();
	
	public Map<String, Object> getValues() {
		return this.values;
	}
}
```
Here are my test results.
#### With codegen=false
It seems to just work now :)

## With codegen=true
object is an instance of a custom class CustomClass.
CustomClass.values is a LinkedHashMap<String, Object> so it's easier to put a null value at some desired index/position
CustomClass.values is empty at the start of each code example
Since I use codegen, the JVM is restarted before each test just in case. Of course if the generated classes are cached somehow, this is a moot point...

##### 1.1 Non null value
Works ok
##### 1.2 Null value
```
// Throws exception: 
// readerIndex(192) + length(1) exceeds size(192): org.apache.fury.memory.MemoryBuffer$BoundChecker@15d50d64
object.getValues().put("86rbsj6si", null);
fury.deserialize(fury.serialize(object));
```
##### 2.1 No null values
Works ok.
##### 2.2 Null values
Null value added last
```
// Throws exception Cannot invoke "Object.getClass()" because "<local7>" is null
object.getValues().put("0q20z6qfk", "Some string value");
object.getValues().put("1mws7xdj5", 1741046400000L);
object.getValues().put("4a0yvgjjw", Map.of("value", 1.0, "currencyCode", "EUR"));
object.getValues().put("86rbsj6si", null);
fury.deserialize(fury.serialize(object));
```
Null value added not first, not last
```
// Same, throws exception: Cannot invoke "Object.getClass()" because "<local7>" is null
object.getValues().put("0q20z6qfk", "Some string value");
object.getValues().put("1mws7xdj5", 1741046400000L);
object.getValues().put("86rbsj6si", null);
object.getValues().put("4a0yvgjjw", Map.of("value", 1.0, "currencyCode", "EUR"));
fury.deserialize(fury.serialize(object));
```
Weirdest test case, null value added first:
```
// Works ok!
object.getValues().put("86rbsj6si", null);
object.getValues().put("0q20z6qfk", "Some string value");
object.getValues().put("1mws7xdj5", 1741046400000L);
object.getValues().put("4a0yvgjjw", Map.of("value", 1.0, "currencyCode", "EUR"));
fury.deserialize(fury.serialize(object));
```
Just for the sake of completion, I've also tested with one additional nested level
```
// Expected this to work (first amount map entry has a null value), and indeed, it works :)
object.getValues().put("0q20z6qfk", "Some string value");
object.getValues().put("1mws7xdj5", 1741046400000L);
Map<String, Object> amount = new LinkedHashMap<>();
amount.put("value", null);
amount.put("counterValue", 1.0);
amount.put("currencyCode", "EUR");
object.getValues().put("4a0yvgjjw", amount);
```
```
// Expected this to fail (last amount map entry has a null value), but it works?!...
object.getValues().put("0q20z6qfk", "Some string value");
object.getValues().put("1mws7xdj5", 1741046400000L);
Map<String, Object> amount = new LinkedHashMap<>();
amount.put("counterValue", 1.0);
amount.put("currencyCode", "EUR");
amount.put("value", null);
object.getValues().put("4a0yvgjjw", amount);
```

So, with codegen, sometimes it works, sometimes it fails with BoundChecker exception and sometimes it fails with a NPE within some generated code...
Not sure how you can make sense of this, but I ran those tests multiple times with consistent results on my dev machine (win11, OpenJDK 21.0.5_11 from Adoptium)

/

### What did you see instead?

/

### Anything Else?

_No response_

chaokunyang:
## What does this PR do?

 fix nested chunk map serialization error when generics exists
## Related issues

Closes #2135 

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/collection/AbstractMapSerializer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/collection/AbstractMapSerializer.java
@@ -769,6 +769,8 @@ public abstract class AbstractMapSerializer<T> extends Serializer<T> {
       }
     } else {
       for (int i = 0; i < chunkSize; i++) {
+        // increase depth to avoid read wrong outer generic type
+        fury.incDepth(1);
         Object key =
             trackKeyRef
                 ? binding.readRef(buffer, keySerializer)
@@ -777,6 +779,7 @@ public abstract class AbstractMapSerializer<T> extends Serializer<T> {
             trackValueRef
                 ? binding.readRef(buffer, valueSerializer)
                 : binding.read(buffer, valueSerializer);
+        fury.incDepth(-1);
         map.put(key, value);
         size--;
       }
--- a/java/fury-core/src/test/java/org/apache/fury/serializer/collection/MapSerializersTest.java
+++ b/java/fury-core/src/test/java/org/apache/fury/serializer/collection/MapSerializersTest.java
@@ -990,4 +990,18 @@ public class MapSerializersTest extends FuryTestBase {
     o.map2 = ofArrayList(ofHashMap("k2", "2"));
     serDeCheck(fury, o);
   }
+
+  @Data
+  public static class NestedMapCollectionGenericTestClass {
+    public Map<String, Object> map = new HashMap<>();
+  }
+
+  @Test(dataProvider = "enableCodegen")
+  public void testNestedMapCollectionGeneric(boolean enableCodegen) {
+    NestedMapCollectionGenericTestClass obj = new NestedMapCollectionGenericTestClass();
+    obj.map = new LinkedHashMap<>();
+    obj.map.put("obj", ofHashMap("obj", 1, "b", ofArrayList(10)));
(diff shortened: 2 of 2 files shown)


theweipeng (approved):

theigl:
@chaokunyang: I'm still seeing this error in 0.10.1. I guess this wasn't backported. WIll this fix be included in 0.10.2 or 0.11.0?

chaokunyang:
> @chaokunyang: I'm still seeing this error in 0.10.1. I guess this wasn't backported. WIll this fix be included in 0.10.2 or 0.11.0?

Hi @theigl , This fix will be included in 0.10..2 and 0.11.0

PR #2139, opened 2025-04-06
refactor(java): refactor object serializer for unifying xlang/java serialization in java

chaokunyang:

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
@@ -38,6 +38,7 @@ import org.apache.fury.reflect.FieldAccessor;
 import org.apache.fury.resolver.ClassInfo;
 import org.apache.fury.resolver.ClassResolver;
 import org.apache.fury.resolver.RefResolver;
+import org.apache.fury.resolver.TypeResolver;
 import org.apache.fury.type.Descriptor;
 import org.apache.fury.type.DescriptorGrouper;
 import org.apache.fury.type.Generics;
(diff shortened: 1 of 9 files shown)


theweipeng (approved):

java/fury-core/src/main/java/org/apache/fury/serializer/ObjectSerializer.java
```diff
@@ -73,19 +74,23 @@ public final class ObjectSerializer<T> extends AbstractObjectSerializer<T> {
...
+  private final SerializationBinding binding;
+  private final TypeResolver typeResolver;
 
   public ObjectSerializer(Fury fury, Class<T> cls) {
     this(fury, cls, true);
   }
 
   public ObjectSerializer(Fury fury, Class<T> cls, boolean resolveParent) {
     super(fury, cls);
+    binding = SerializationBinding.createBinding(fury);
```
LiangliangSui, line 86: **NIT:**

We can create `SerializationBinding` directly in `Fury.java` instead of repeatedly creating it in `ObjectSerializer`, `NonexistentClassSerializer`, `MetaSharedSerializer` serializers, so that `SerializationBinding` can be reused, because the parameters required by `SerializationBinding.createBinding` are only `Fury`.

chaokunyang: `SerializationBinding` is a internal interface, I want to keep it visible only in serializer package. Createing SerializationBinding directly in Fury.java will make it visible to users.

LiangliangSui: Okay, I got it.

PR #2128, opened 2025-03-28
fix(java): fix disallowed.txt check in windows

chaokunyang:
## What does this PR do?
- add windows ci for java21
- Fix sha256 check error on windows:
Image
## Related issues

Closes #2100

## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

PR #2142, opened 2025-04-11
fix(java): fix DisallowedList calculate hash in Windows

moooonk:
## What does this PR do?
It appears that [PR #2128](https://github.com/apache/fury/pull/2128) did not truly resolve this issue, as it only modified the test case code without addressing the actual runtime logic. This PR re-fixes the problem.


LiangliangSui:
Maybe @chaokunyang  is more interested in this.

BTW, is there a similar error after the fix in https://github.com/apache/fury/pull/2128?

chaokunyang (approved):

PR #2146, opened 2025-04-13
feat(java): unify java and xlang object serialization

chaokunyang:
## What does this PR do?

This PR unify the java and xlang object serialization in java:
- Remove StructSerializer
- Unify struct hash compute between java and python
- Align fields sort between java and python
- unify the java and xlang object serialization in java


java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
```diff
@@ -2032,6 +2040,74 @@ public void setCodeGenerator(ClassLoader[] loaders, CodeGenerator codeGenerator)
...
+    if (isMap(cls)) {
+      return Types.MAP;
+    }
+    if (fury.getXtypeResolver().isRegistered(cls)) {
+      return fury.getXtypeResolver().getClassInfo(cls).getXtypeId();
+    } else {
+      if (ReflectionUtils.isMonomorphic(cls)) {
+        throw new UnsupportedOperationException(cls + " is not supported for xlang serialization");
+      }
+      return 128;
```
LiangliangSui, line 2107: What does 128 represent? Could you explain it?

LiangliangSui:
@chaokunyang Here is a suggestion.

Currently, there are many `writeXXX`, `xwriteXXX`, `readXXX` and `xreadXXX` methods in `Fury.java`. Is it possible for us to define an abstract class to summarize these interfaces by following the example of https://github.com/apache/fury/pull/2139?

In this way, we can delete the `Fury#xwriteXXX` and `Fury#xreadXXX` methods(At the same time, it can also reduce the logic within Fury.java). The cross-language details can be handled in the respective implementation subclasses of this abstract class.

java/fury-core/src/main/java/org/apache/fury/resolver/ClassResolver.java
chaokunyang: Used a constant to represent it, it indicate a unknown type

chaokunyang:
> @chaokunyang Here is a suggestion.
> 
> Currently, there are many `writeXXX`, `xwriteXXX`, `readXXX` and `xreadXXX` methods in `Fury.java`. Is it possible for us to define an abstract class to summarize these interfaces by following the example of #2139?
> 
> In this way, we can delete the `Fury#xwriteXXX` and `Fury#xreadXXX` methods(At the same time, it can also reduce the logic within Fury.java). The cross-language details can be handled in the respective implementation subclasses of this abstract class.

This is what `SerializationBinding` does. And I don't think we can remove things like  `Fury#xwriteXXX` and `Fury#xreadXXX`. It's not only used in `SerializationBinding`, but also used outside of it. `SerializationBinding` is desgined for internal usage only, I don't want to expost it users. And `Fury` itself is the obly entrance class, I don't think we should add other entrance class. Another issue for abstract class is the runtime cost, the polymorphic virtual method call cost is very expensive,, which is we try to avoid

pandalee99 (approved):

LiangliangSui:
> This is what SerializationBinding does. And I don't think we can remove things like Fury#xwriteXXX and Fury#xreadXXX. It's not only used in SerializationBinding, but also used outside of it. 

Will users directly use functions like `Fury#xwriteXXX` and `Fury#xreadXXX` externally, perhaps when creating custom serializers?

chaokunyang:
> > This is what SerializationBinding does. And I don't think we can remove things like Fury#xwriteXXX and Fury#xreadXXX. It's not only used in SerializationBinding, but also used outside of it.
> 
> Will users directly use functions like `Fury#xwriteXXX` and `Fury#xreadXXX` externally, perhaps when creating custom serializers?

Yes, they will use such API for customize xlang serializer

PR #2154, opened 2025-04-18
fix(java): Fix logger error

### LINKED ISSUE #2152: [Java] ArrayIndexOutOfBoundsException when logging error
### Search before asking

### Version

0.10.1

### Component(s)

Java

### Minimal reproduce step

I can't reproduce this at the moment because it happened on one of our test environments, but the stacktrace is quite clear.

Logging an error should log the error instead of throwing an exception.

### What did you see instead?

```
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
        at org.apache.fury.logging.FuryLogger.log(FuryLogger.java:153)
        at org.apache.fury.logging.FuryLogger.error(FuryLogger.java:133)
        at org.apache.fury.pool.ThreadPoolFury.execute(ThreadPoolFury.java:84)
        at org.apache.fury.pool.ThreadPoolFury.deserialize(ThreadPoolFury.java:138)
```

`ThreadPoolFury` has the following code:

```java
    } catch (Exception e) {
      LOG.error(e.getMessage(), e);
    }
```

In `FuryLogger` this method is called:

```java
  public void error(String msg, Throwable t) {
    if (LoggerFactory.getLogLevel() >= ERROR_LEVEL) {
      log(ERROR_LEVEL, msg, new Object[] {t}, true);
    }
  }
```

It passes the throwable as the sole `args` to the log function. Now if the exception message contains more than one placeholder like `{}`, the `ArrayIndexOutOfBoundsException` is thrown here:

```java
char c = msg.charAt(i);
  if (c == '{' && msg.charAt(i + 1) == '}') {
	builder.append(args[count++]);
	i++;
} else {
	builder.append(c);
}
```

In my opinion, the logger should *not* attempt to interpolate random placeholders in messages of thrown exceptions. 

### Anything Else?

_No response_

chaokunyang:
## Related issues

Closes #2152 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,7 +240,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.8, 3.12, 3.13]
-        os: [ubuntu-20.04, macos-13, macos-14, windows-2022]
+        os: [ubuntu-latest, macos-13, macos-14, windows-2022]
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


pandalee99 (approved):
LGTM

PR #2161, opened 2025-04-19
fix(java): fix xlang container field deserialization type error

### LINKED ISSUE #2105: [Java] Set deserialized as ArrayList in XLANG mode disrupts collection semantics
### Search before asking

### Version

0.10.0

### Component(s)

Java

### Minimal reproduce step

```java
package org.apache.fury;

import org.apache.fury.config.Language;

import java.util.HashSet;
import java.util.Set;

public class Main {
  public static class SomeClass {
    Set<String> set = new HashSet<>();
  }

    Fury fury = Fury.builder()
            .withLanguage(Language.XLANG)
            .withRefTracking(true)
            .ignoreBasicTypesRef(true)
            .ignoreTimeRef(true)
            .ignoreStringRef(false)
            .build();
    fury.register(SomeClass.class, "MySomeClass");

    SomeClass someClass = new SomeClass();
    byte[] bytes = fury.serialize(someClass);
    SomeClass obj = (SomeClass)fury.deserialize(bytes);
    System.out.println(obj.set.getClass());
  }
}
```

I expected the deserialized object to maintain the same collection type as the original object - a HashSet in this case. Since the field is declared as a Set<String>, it should be deserialized as a Set implementation like HashSet, preserving the collection's semantics (uniqueness of elements, hash-based operations, etc.).

### What did you see instead?

The deserialized object contains an ArrayList instead of a HashSet:
```bash
class java.util.ArrayList
```
The Set is converted to an ArrayList during deserialization, which breaks the expected behavior of the Set interface (no duplicate elements, different performance characteristics for operations like contains()).

### Anything Else?

I've traced the issue to the StructSerializer::getGenericType method which contains the following logic:

```java
if (resolver.isMap(cls)) {
  t.setSerializer(
      ReflectionUtils.isAbstract(cls)
          ? new MapSerializer(fury, HashMap.class)
          : resolver.getSerializer(cls));
} else if (resolver.isCollection(cls)) {
  t.setSerializer(
      ReflectionUtils.isAbstract(cls)
          ? new CollectionSerializer(fury, ArrayList.class)
          : resolver.getSerializer(cls));
} else if (cls.isArray()) {
  t.setSerializer(new ArraySerializers.ObjectArraySerializer(fury, cls));
}
```
When the serializer detects a field is a Collection and its class is abstract (which includes interfaces like Set), it defaults to using ArrayList as the implementation rather than preserving the original collection type.

While this approach makes the code work (the elements are preserved), it changes the semantics of the collection. This can lead to unexpected behavior for code that relies on specific collection implementations or interfaces.

chaokunyang:
## What does this PR do?

fix xlang container field deserialization type error

## Related issues

Closes #2105 

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/resolver/XtypeResolver.java
+++ b/java/fury-core/src/main/java/org/apache/fury/resolver/XtypeResolver.java
@@ -26,6 +26,7 @@ import static org.apache.fury.meta.Encoders.PACKAGE_DECODER;
 import static org.apache.fury.meta.Encoders.PACKAGE_ENCODER;
 import static org.apache.fury.meta.Encoders.TYPE_NAME_DECODER;
 import static org.apache.fury.resolver.ClassResolver.NO_CLASS_ID;
+import static org.apache.fury.serializer.collection.MapSerializers.*;
 import static org.apache.fury.type.TypeUtils.qualifiedName;
 
 import java.lang.reflect.Type;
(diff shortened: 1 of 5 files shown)


LiangliangSui (approved):
LGTM

PR #2162, opened 2025-04-19
fix(java): fix nested chunk map serialization error when generics exists

chaokunyang:
## Related issues

#2136


pandalee99 (approved):
LGTM, this fix so good!

Issue #2110, opened 2025-03-19
[Java] Inconsistent readVarUint36Small() behavior depending on remaining buffer size


LouisLou2:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

0.10.0

### Component(s)

Java

### Minimal reproduce step

```java
public static void main(String[] args) {
  // With buffer size 9 - outputs 68719476735
  MemoryBuffer buffer1 = MemoryBuffer.newHeapBuffer(9);
  int off1 = buffer1._unsafePutVarUint36Small(0, 68719476735L);
  long res1 = buffer1.readVarUint36Small();
  System.out.println(res1);  // Outputs: 68719476735
  
  // With buffer size 8 - outputs 34359738367 (incorrect)
  MemoryBuffer buffer2 = MemoryBuffer.newHeapBuffer(8);
  int off2 = buffer2._unsafePutVarUint36Small(0, 68719476735L);
  long res2 = buffer2.readVarUint36Small();
  System.out.println(res2);  // Outputs: 34359738367
}
```

### What did you expect to see?

The same value (68719476735) should be read regardless of the initial buffer size. The reading behavior should be consistent and should not depend on the remaining buffer size.

### What did you see instead?

Different results depending on buffer size:

With buffer size 9: correct output 68719476735
With buffer size 8: incorrect output 34359738367

### Anything Else?

I found that `readVarUint36Small()` has two different code paths depending on the remaining buffer size:

1. When `size - readIdx >= 9`, it uses a "fast" path that can read up to 36 bits of data
2. When `size - readIdx < 9`, it falls back to `readVarUint36Slow()` which only reads up to 35 bits (5 groups of 7 bits)

In `readVarUint36Small()` fast path, the last bit extraction does:
```java
// 0xff0000000: 0b11111111 << 28
result |= (bulkValue >>> 4) & 0xff0000000L;
```
This extracts 8 bits for the high order bits.

Meanwhile, the slow path in `readVarUint36Slow()` adds 7 bits up to 5 times, reaching a maximum of 35 bits.

This inconsistency causes different values to be read depending only on the buffer size, not on the actual data. There needs to be alignment between the "fast" and "slow" paths - either both should read 35 bits or both should read 36 bits.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #2179, opened 2025-04-24
fix(java): ensure readVarUint36Small reads full bits regardless of remaining buffer size

LouisLou2:
## What does this PR do?

This PR fixes the inconsistent behavior of `readVarUint36Small()` when reading from buffers of different remaining sizes.  
- Aligns the fast path and slow path so both read the full 36 bits.  
- Updates the bit‐extraction in the slow path (`readVarUint36Slow()`) to match the fast path’s handling of the high‐order bits.  

## Related issues

- #2110

## Does this PR introduce any user‐facing change?

No user‐facing changes are introduced.  

## Benchmark

No measurable performance impact is expected, as the fix only adjusts bit extraction logic without altering code paths.  
Benchmark

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
+++ b/java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
@@ -1800,7 +1800,7 @@ public final class MemoryBuffer {
           result |= (b & 0x7F) << 21;
           if ((b & 0x80) != 0) {
             b = readByte();
-            result |= (b & 0x7F) << 28;
+            result |= (b & 0xFF) << 28;
           }
         }
       }


java/fury-core/src/main/java/org/apache/fury/memory/MemoryBuffer.java
```diff
@@ -1800,7 +1800,7 @@ private long readVarUint36Slow() {
           result |= (b & 0x7F) << 21;
           if ((b & 0x80) != 0) {
             b = readByte();
-            result |= (b & 0x7F) << 28;
+            result |= (b & 0xFF) << 28;
```
chaokunyang, line 1803: Could we add an unit test?

LouisLou2: I'll add one later.

chaokunyang (approved):

PR #2173, opened 2025-04-23
fix(kotlin): fix kotlin compile error

chaokunyang:
## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

PR #2188, opened 2025-04-24
fix(kotlin): fix kotlin compile error (#2173)

LiangliangSui:
## What does this PR do?

cherrypick https://github.com/apache/fury/pull/2173 into releases-0.10


pandalee99 (approved):
我也遇到了这个问题

chaokunyang:
We should revert this PR, 0.10 branch won't compile with this PR @LiangliangSui @pandalee99 .

And it would be better if we add a github ci for kotlin to check whether we really fixed an issue

pandalee99:
> We should revert this PR, 0.10 branch won't compile with this PR @LiangliangSui @pandalee99 .
> 
> And it would be better if we add a github ci for kotlin to check whether we really fixed an issue

sure, it makes sense

PR #2191, opened 2025-04-26
Support furyField nullable in codeGen pattern

Hen1ng:

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/builder/BaseObjectCodecBuilder.java
+++ b/java/fury-core/src/main/java/org/apache/fury/builder/BaseObjectCodecBuilder.java
@@ -333,8 +333,9 @@ public abstract class BaseObjectCodecBuilder extends CodecBuilder {
     ctx.addImports(AbstractCollectionSerializer.class, AbstractMapSerializer.class);
   }
 
-  protected Expression serializeFor(Expression inputObject, Expression buffer, TypeRef<?> typeRef) {
-    return serializeFor(inputObject, buffer, typeRef, false);
+  protected Expression serializeForNullable(
+      Expression inputObject, Expression buffer, TypeRef<?> typeRef, boolean nullable) {
+    return serializeForNullable(inputObject, buffer, typeRef, null, false, nullable);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/builder/BaseObjectCodecBuilder.java
```diff
@@ -375,6 +371,45 @@ protected Expression serializeFor(
...
+      if (typeRef.isPrimitive()) {
+        return serializeForNotNull(inputObject, buffer, typeRef, serializer, generateNewMethod);
+      }
+      if (nullable) {
+        Expression action =
+            new ListExpression(
+                new Invoke(
+                    buffer,
+                    "writeByte",
+                    new Literal(Fury.NOT_NULL_VALUE_FLAG, PRIMITIVE_BYTE_TYPE)),
```
chaokunyang, line 401: Literal.ofByte

java/fury-core/src/main/java/org/apache/fury/builder/BaseObjectCodecBuilder.java
```diff
@@ -375,6 +371,45 @@ protected Expression serializeFor(
...
+                serializeForNotNull(inputObject, buffer, typeRef, serializer, generateNewMethod));
+        return new If(
+            eqNull(inputObject),
+            new Invoke(buffer, "writeByte", new Literal(Fury.NULL_FLAG, PRIMITIVE_BYTE_TYPE)),
```
chaokunyang, line 405: Literal.ofByte(Fury.NULL_FLAG)

java/fury-core/src/main/java/org/apache/fury/builder/BaseObjectCodecBuilder.java
```diff
@@ -1441,6 +1489,26 @@ private Expression readNullable(
...
+      Expression buffer,
+      TypeRef<?> typeRef,
+      Function<Expression, Expression> callback,
+      Supplier<Expression> deserializeForNotNull,
+      boolean nullable) {
+    if (nullable) {
+      Expression notNull =
+          neq(
+              inlineInvoke(buffer, "readByte", PRIMITIVE_BYTE_TYPE),
+              new Literal(Fury.NULL_FLAG, PRIMITIVE_BYTE_TYPE));
```
chaokunyang, line 1502: Literal.ofByte(Fury.NULL_FLAG)

java/fury-core/src/main/java/org/apache/fury/builder/CompatibleCodecBuilder.java
```diff
@@ -341,9 +341,13 @@ private Expression writeEmbedTypeFieldValue(
       Expression bean, Expression buffer, FieldInfo fieldInfo) {
     Descriptor descriptor = createDescriptor(fieldInfo);
     walkPath.add(descriptor.getDeclaringClass() + descriptor.getName());
+    boolean nullable = false;
+    if (!descriptor.getTypeRef().isPrimitive()) {
```
chaokunyang, line 345: Could we add a nullable method in Descriptor?

chaokunyang: I think we can have a nullable/trackingRef field and getter in Descriptor.

We can also have a Descriptor which can build Descriptor with different field values easily.

```java
Descriptor desc = ...;
new DescriptorBuilder(desc).nullable(true).trackingRef(false).build();
desc.builder().nullable(true).trackingRef(false).build();
```

java/fury-core/src/main/java/org/apache/fury/type/Descriptor.java
```diff
@@ -123,6 +127,7 @@ private Descriptor(Field field, Method readMethod) {
     this.writeMethod = null;
     this.typeRef = null;
     this.furyField = this.field.getAnnotation(FuryField.class);
+    this.nullable = furyField == null || furyField.nullable();
```
chaokunyang, line 130: If furyField is null, nullable should be `!typeRef.isPrimitive()`

Hen1ng: typeRef is null in this constructor

chaokunyang: Then it should be `field.getType().isPrimitive()`

chaokunyang (approved):

chaokunyang:
@Hen1ng could we add a test with codegen enabled and annotation added for field?

Issue #2204, opened 2025-05-06
[Java][Bug] Failed to deserialize Thrift object


X-czh:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

- Fury: 0.10.2 & 0.11.0 & master
- JDK: 8u442

### Component(s)

Java

### Minimal reproduce step

```java
Fury fury = Fury.builder().withLanguage(Language.JAVA).withCodegen(true).build();
fury.register(ThriftObject.class);
ThriftObject o = new ThriftObject();
// The default initial buffer size of 4096 cannot reproduce
fury.deserialize(new FuryInputStream(new ByteArrayInputStream(fury.serialize(o)), 2));
```

### What did you expect to see?

No exception.

### What did you see instead?

```
org.apache.fury.exception.DeserializationException: Deserialize failed, read objects are: [VidsAndFlightHitRules(vid:null, rules:null)]

	at org.apache.fury.util.ExceptionUtils.handleReadFailed(ExceptionUtils.java:63)
	at org.apache.fury.Fury.deserialize(Fury.java:810)
	at org.apache.fury.Fury.deserialize(Fury.java:829)
	at org.apache.fury.Fury.deserialize(Fury.java:822)
	at com.bytedance.ad.abtest.etl.realtime.util.FuryTest.testFury(FuryTest.java:38)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
[...]

PR #2205, opened 2025-05-06
fix(java): ensure FuryObjectInputStream.read never returns 0 when length>0 #2204

X-czh:
## What does this PR do?

Ensure FuryObjectInputStream.read never returns 0 when length>0, so that when outer logic attempts to read in a while loop, it won't get stuck or throw exception.

## Related Issues
#2204 

No.

## Benchmark

No measurable performance impact is expected.


chaokunyang (approved):
LGTM

Issue #2118, opened 2025-03-24
[Python] TimestampSerializer fails on Windows for early time near epoch


LouisLou2:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

Fury: Latest development branch
OS: Windows 11
Python: 3.12
The issue is not present on Linux

### Component(s)

Python

### Minimal reproduce step

The issue occurs in CrossLanguageTest.java when running on Windows. The error happens when Fury tries to serialize datetime objects with timestamps close to the epoch (1970-01-01).

The root cause is in python/pyfury/_serialization.pyx at line 1192:
```python
timestamp = int(value.timestamp() * 1000000)
```
The timestamp() method on naive datetime objects fails on Windows for early dates. This can be reproduced with:
```python
import datetime

# This fails on Windows
d = datetime.datetime.fromtimestamp(timestamp=100)  # Early date: Jan 1, 1970 00:01:40
t = d.timestamp()  # OSError: [Errno 22] Invalid argument

# These work fine on Windows
d1 = datetime.datetime.fromtimestamp(timestamp=1000000)  # Later date
t1 = d1.timestamp()  # Works fine

d2 = datetime.datetime.fromtimestamp(timestamp=100, tz=datetime.timezone.utc)  # With timezone
t2 = d2.timestamp()  # Works fine
```

### What did you expect to see?

Cross-language serialization tests should pass on both Linux and Windows platforms.

### What did you see instead?

On Windows, serialization fails with:
```bash
ERROR:root:Execute ['test_cross_language_serializer', 'C:\\Users\\86511\\AppData\\Local\\Temp\\test_cross_language_serializer2918297548183494188data'] failed with [Errno 22] Invalid argument
Traceback (most recent call last):
  File "d:\projects\fury_dev\fury\python\pyfury\tests\test_cross_language.py", line 622, in <module>
    func(*args[1:])
  File "d:\projects\fury_dev\fury\python\pyfury\tests\test_cross_language.py", line 337, in test_cross_language_serializer
    fury.serialize(obj, buffer=new_buf)
  File "python/pyfury/_serialization.pyx", line 639, in pyfury._serialization.Fury.serialize
  File "python/pyfury/_serialization.pyx", line 688, in pyfury._serialization.Fury._serialize
  File "python/pyfury/_serialization.pyx", line 746, in pyfury._serialization.Fury.xserialize_ref
[...]

PR #2209, opened 2025-05-07
fix(python): TimestampSerializer fails on Windows for naive datetimes near epoch

LouisLou2:
## What does this PR do?

This PR resolves an `OSError: [Errno 22] Invalid argument` in `TimestampSerializer` on Windows when serializing naive `datetime` objects (no `tzinfo`) close to the Unix epoch. This issue stems from `datetime.datetime.timestamp()` failing for such dates on Windows.

The fix applies a workaround for Windows when handling naive `datetime` objects:
1.  The naive `datetime` is made timezone-aware by setting its `tzinfo` to `datetime.timezone.utc`.
2.  The POSIX timestamp is then retrieved.
3.  This timestamp is adjusted by the local timezone offset (considering DST) to produce the correct microsecond timestamp for serialization.

This approach ensures correct timestamp generation on Windows for these edge cases without affecting other platforms or timezone-aware `datetime` objects. The fix is implemented in both `_serialization.pyx` and `_serializer.py`.

## Related issues
- #2118


chaokunyang (approved):
Looks great

PR #2218, opened 2025-05-11
feat:  Align object array to collection serialization protocol v2

david1437:
## Related issues

- #1229

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
@@ -51,6 +51,7 @@ public class ArraySerializers {
     private final Class<T> innerType;
     private final Serializer componentTypeSerializer;
     private final FuryArrayAsListSerializer collectionSerializer;
+    private ArrayAsList list;
     private final GenericType collectionGenericType;
     private final ClassInfoHolder classInfoHolder;
     private final int[] stubDims;
@@ -80,13 +81,8 @@ public class ArraySerializers {
           this.collectionGenericType = null;
         } else {
           this.componentTypeSerializer = fury.getClassResolver().getSerializer(componentType);
-          if (dimension > 1) {
-            this.collectionSerializer = new FuryArrayAsListSerializer(fury);
-            this.collectionGenericType = buildCollectionGenericType(dimension);
-          } else {
-            this.collectionSerializer = null;
-            this.collectionGenericType = null;
-          }
+          this.collectionSerializer = new FuryArrayAsListSerializer(fury);
+          this.collectionGenericType = buildCollectionGenericType(dimension);
         }
       } else {
         // TODO add ClassInfo cache for non-final component type.
@@ -95,6 +91,7 @@ public class ArraySerializers {
         this.collectionGenericType = null;
       }
       this.stubDims = new int[dimension];
+      this.list = null;
       classInfoHolder = fury.getClassResolver().nilClassInfoHolder();
     }
 
@@ -105,10 +102,18 @@ public class ArraySerializers {
       Serializer componentSerializer = this.componentTypeSerializer;
       if (this.collectionSerializer != null) {
         fury.getGenerics().pushGenericType(this.collectionGenericType);
-        ArrayAsList list = new ArrayAsList(0);
+        ArrayAsList list = this.list;
+        if (list == null) {
+          list = new ArrayAsList(0);
+        } else {
+          this.list = null;
+        }
         list.setArray(arr);
+        fury.incDepth(1);
         this.collectionSerializer.write(buffer, list);
+        fury.incDepth(1);
         fury.getGenerics().popGenericType();
+        this.list = list;
       } else {
         int header = componentSerializer != null ? 0b1 : 0b0;
         buffer.writeVarUint32Small7(len << 1 | header);
--- a/java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
+++ b/java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
@@ -101,7 +101,10 @@ public class ArraySerializers {
       RefResolver refResolver = fury.getRefResolver();
       Serializer componentSerializer = this.componentTypeSerializer;
       if (this.collectionSerializer != null) {
-        fury.getGenerics().pushGenericType(this.collectionGenericType);
+        buffer.writeVarUint32Small7(len);
+        if (len == 0) {
+          return;
+        }
         ArrayAsList list = this.list;
         if (list == null) {
           list = new ArrayAsList(0);
@@ -110,9 +113,10 @@ public class ArraySerializers {
         }
         list.setArray(arr);
         fury.incDepth(1);
+        fury.getGenerics().pushGenericType(this.collectionGenericType);
         this.collectionSerializer.write(buffer, list);
-        fury.incDepth(-1);
         fury.getGenerics().popGenericType();
+        fury.incDepth(-1);
         this.list = list;
       } else {
         int header = componentSerializer != null ? 0b1 : 0b0;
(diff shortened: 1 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
```diff
@@ -208,6 +251,30 @@ private Object[] newArray(int numElements) {
       }
       return value;
     }
+
+    private GenericType buildCollectionGenericType(int dimensions) {
```
david1437, line 255: I am hoping there is a better way to resolve this dyamically rather then using switch? Any suggestions?

chaokunyang: I think we can write using a loop, and using `TypeRef.of` to build generic type

chaokunyang: It's like:
```java

 Tuple2<Class<?>, Integer> info = TypeUtils.getArrayComponentInfo(declaredClass);
      List<TypeRef<?>> typeRefs = new ArrayList<>(info.f1 + 1);
      typeRefs.add(collectionTypeRef);
      for (int i = 0; i < info.f1; i++) {
        typeRefs.add(TypeUtils.getElementType(typeRefs.get(i)));
      }
      Collections.reverse(typeRefs);
      for (int i = 1; i < typeRefs.size(); i++) {
        TypeRef<?> arrayType = typeRefs.get(i - 1);
        TypeRef<?> typeRef =
            TypeRef.of(
                Array.newInstance(arrayType.getRawType(), 1).getClass(),
                typeRefs.get(i).getExtInfo());
        typeRefs.set(i, typeRef);
      }
```

See more details in `ClassDef#buildFieldType`

david1437: Apologies if it is a dumb question, but where is collectionTypeRef coming from in this case? I was thinking it might be TypeUtils.COLLECTION_TYPE but even in this case its adding one element for looping over multiple dimensions so would still fail

chaokunyang: My bad, in your case, I think you can write like:
```java
TypeRef arrayTypeRef = TypeRef.of(componentClass);
for (int i = 0; i < dims -1; i++ ) {
  arrayTypeRef = collectionOf(arrayTypeRef);
}
```

java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
```diff
@@ -85,52 +103,69 @@ public void write(MemoryBuffer buffer, T[] arr) {
...
-        for (T t : arr) {
-          if (!refResolver.writeRefOrNull(buffer, t)) {
-            componentSerializer.write(buffer, t);
-          }
-        }
+      if (this.collectionSerializer != null) {
+        fury.getGenerics().pushGenericType(this.collectionGenericType);
+        ArrayAsList list = new ArrayAsList(0);
+        list.setArray(arr);
+        this.collectionSerializer.write(buffer, list);
```
chaokunyang, line 110: We need to invoke `fury.incDepth` before  and after invoking `collectionSerializer.write`

java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
```diff
@@ -85,52 +103,69 @@ public void write(MemoryBuffer buffer, T[] arr) {
...
-      buffer.writeVarUint32Small7(len << 1 | header);
-      if (componentSerializer != null) {
-        for (T t : arr) {
-          if (!refResolver.writeRefOrNull(buffer, t)) {
-            componentSerializer.write(buffer, t);
-          }
-        }
+      if (this.collectionSerializer != null) {
+        fury.getGenerics().pushGenericType(this.collectionGenericType);
+        ArrayAsList list = new ArrayAsList(0);
```
chaokunyang, line 108: Could we cache this object as a field.  Nested serialization is a rare case, if we cache it as a field, we could save object creation cost. 

We could write like :
```java
ArrayAsList list = this.list;
if (list == null) {
  list = new ArrayAsList(0);
} else {
  this.list = null
}

// before serialiazation

...


// after serialization

this.list = list

```

java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
```diff
@@ -69,12 +76,23 @@ public ObjectArraySerializer(Fury fury, Class<T[]> cls) {
...
           this.componentTypeSerializer = null;
+          this.collectionSerializer = null;
+          this.collectionGenericType = null;
         } else {
           this.componentTypeSerializer = fury.getClassResolver().getSerializer(componentType);
+          if (dimension > 1) {
+            this.collectionSerializer = new FuryArrayAsListSerializer(fury);
+            this.collectionGenericType = buildCollectionGenericType(dimension);
+          } else {
+            this.collectionSerializer = null;
```
chaokunyang, line 87: Why it's null for 1d object array?

david1437: I thought for 1D / Non nested generics that we intended to use the normal default serialization path that was in place before?

chaokunyang: That is only for primitive array. Other array still use Collection serialization protocol. `List<String>` and `String[]` should have same layout

david1437: Ok I  am under impression T[] can never have primitives only wrapper primitive classes should those use predvious method as well ie. Double[] ?

java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
david1437: I tried with 

incDepth(1)
...
incDepth(1)

Should I be using -1 after serialization to reset the depth?

chaokunyang: incDepth(-1) should be invoked after serialization

chaokunyang:
> Any suggestions on running the CI suites locally so I can work on debugging the issues?

You can execute:

```bash
cd java
mvn -T10 clean install -DskipTests
cd fury-core
mvn test -Dtest=org.apache.fury.serializer.ArraySerializersTest 
```

Or you can use IDE such jetbrains intellj to debug test one by one, I prefer this method:
image

david1437:
So I have been running the ArraySerializerTests and they have all been passing so a bit confused on why the CICD tests are failing

chaokunyang:
@david1437  you could execute:
```bash
cd java
mvn -T10 clean install -DskipTests
cd fury-core
mvn -T16 test`
```

All failed tests will be printed in the end

java/fury-core/src/main/java/org/apache/fury/serializer/ArraySerializers.java
```diff
@@ -147,39 +191,49 @@ public void xwrite(MemoryBuffer buffer, T[] arr) {
...
-          } else {
-            elem = refResolver.getReadObject();
-          }
-          value[i] = elem;
+      if (this.collectionSerializer != null) {
+        fury.getGenerics().pushGenericType(this.collectionGenericType);
+        value = this.collectionSerializer.read(buffer).toArray();
+        fury.getGenerics().popGenericType();
+        if (this.innerType.isPrimitive() || TypeUtils.isBoxed(this.innerType)) {
+          return Arrays.copyOf(value, value.length, this.type);
```
david1437, line 201: Considering this does a copy it does not feel good is there a better approach for handling this? without this i get a class cast exception?

david1437:
After resolving the issues around primitives and depth increment it seems only a couple tests are failing, Would appreciate some help on those specific failing cases: JdkProxySerializerTest and NonexistentClassSerializersTest

PR #2221, opened 2025-05-12
fix(ci): fix releases-0.10.3 ci

pandalee99:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -20,7 +20,7 @@ name: Publish Fury
 on:
   push:
     tags:
-      - "v*"
+      - "disabled*"
 
 jobs:
   release-python:


chaokunyang (approved):

.github/workflows/release.yaml
```diff
@@ -20,7 +20,7 @@ name: Publish Fury
 on:
   push:
     tags:
-      - "v*"
+      - "disabled*"
```
LiangliangSui, line 23: What is our intention in doing this? disabled CI in releases-0.10.3 ？

chaokunyang: We don't release pyfury binary in 0.10

LiangliangSui: I see. We had a deceptive shutdown. In fact, there is no "disabled" branch. I thought "disabled" was a switch before.

Issue #2222, opened 2025-05-12
[Java] add DescriptorBuilder for easy build and copying Descriptor


chaokunyang:
### Feature Request

Implement builder interface like:
```java
Descriptor desc = ...; 
new DescriptorBuilder(desc).nullable(true).trackingRef(false).build(); 
desc.builder().nullable(true).trackingRef(false).build();
```

PR #2229, opened 2025-05-13
feat(java): add DescriptorBuilder for easy build and copying Descriptor

Hen1ng:
## Related issues
#2222 

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/type/Descriptor.java
+++ b/java/fury-core/src/main/java/org/apache/fury/type/Descriptor.java
@@ -85,7 +85,9 @@ public class Descriptor {
   private final Field field;
   private final Method readMethod;
   private final Method writeMethod;
-  private final FuryField furyField;
+  private FuryField furyField;
+  private boolean nullable;
+  private boolean trackingRef;
 
   public Descriptor(Field field, TypeRef<?> typeRef, Method readMethod, Method writeMethod) {
     this.field = field;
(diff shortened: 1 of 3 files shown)


java/fury-core/src/main/java/org/apache/fury/type/DescriptorBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+    this.typeName = typeName;
+    return this;
+  }
+
+  public DescriptorBuilder name(String name) {
+    this.name = name;
+    return this;
+  }
+
+  public DescriptorBuilder snakeCaseName(String snakeCaseName) {
```
chaokunyang, line 79: This should be consistent with name, when building descriptor, we set `snakeCaseName` automatically in the constructor.

java/fury-core/src/main/java/org/apache/fury/type/Descriptor.java
```diff
@@ -143,14 +145,28 @@ private Descriptor(
...
+        builder.field,
+        builder.readMethod,
+        builder.writeMethod);
+    this.nullable = builder.nullable;
+    this.trackingRef = builder.trackingRef;
+    this.type = builder.type;
+    this.furyField = builder.furyField;
+  }
+
   public Descriptor copy(Method readMethod, Method writeMethod) {
```
chaokunyang, line 164: Could we also add a `copyBuilder()` method in Descriptor to simplify the usage. And could we add a test?

chaokunyang (approved):

PR #2242, opened 2025-05-16
test(benchmark): add msgpack serialization/deserialization to benchmark

Kn0688:
ObjectType must be MEDIA_CONTENT or SAMPLE.

## What does this PR do?

 add msgpack serialization/deserialization to benchmark

## Related issues

no

no


java/benchmark/src/main/java/org/apache/fury/benchmark/UserTypeSerializeSuite.java
```diff
@@ -23,16 +23,7 @@
...
-import org.apache.fury.benchmark.state.FstState;
-import org.apache.fury.benchmark.state.FuryState;
-import org.apache.fury.benchmark.state.HessionState;
-import org.apache.fury.benchmark.state.JDKState;
-import org.apache.fury.benchmark.state.JsonbState;
-import org.apache.fury.benchmark.state.KryoState;
-import org.apache.fury.benchmark.state.ObjectType;
-import org.apache.fury.benchmark.state.ProtoBuffersState;
-import org.apache.fury.benchmark.state.ProtostuffState;
+import org.apache.fury.benchmark.state.*;
```
chaokunyang, line 26: could you use `mvn spotless:apply` to format code? if it doesn't optimize imports, change it manualy

we do not use `*` for import

java/benchmark/src/main/java/org/apache/fury/benchmark/util/MsgpackUtil.java
```diff
@@ -0,0 +1,599 @@
...
+        default:
+          messageUnpacker.skipValue();
+          break;
+      }
+    }
+
+    return mediaContent;
+  }
+
+  private static void packMedia(MessagePacker messagePacker, Media media) throws IOException {
```
chaokunyang, line 116: Is there an automatic way to serialize it ? This is basically a manually-written serializer. We are benchmarking with the serializer you written instead of msgpack itself. We only use the msgpack protocol

chaokunyang: If there is no automatic way, maybe we could add some comments in this class and the benchmark readme

java/benchmark/src/main/java/org/apache/fury/benchmark/UserTypeSerializeSuite.java
Kn0688: Yes, I have already used ```mvn spotless:apply``` to format code.But It is an IDE issue. I'll fix it.
image

java/benchmark/src/main/java/org/apache/fury/benchmark/util/MsgpackUtil.java
Kn0688: The msgpack's official provides [msgpack-jackson](https://github.com/msgpack/msgpack-java/tree/main/msgpack-jackson) lib, but the performance is relatively poor. So, generate a basic [handwritten code](https://github.com/apache/fury/tree/main/java/benchmark/src/main/java/org/apache/fury/benchmark/util/MsgpackUtil.java) using qwen3(LLM). Then modify it.

Below is a performance comparison.
g2

chaokunyang (approved):

PR #2243, opened 2025-05-16
feat(java): row encoder supports custom types and collections

stevenschlansker:
## What does this PR do?

Extend Java Row Format to allow registering custom datatypes (e.g. UUID as Int128) and collection factories (e.g. `SortedSet<UUID>` as `new TreeSet<UUID>(customComparator)` )
Additionally supports arrays of custom types e.g. `UUID[]`

Since the type inference is in `fury-core` but I wanted to keep new features scoped to `fury-format`, I had to add a small plugin interface to core so that format can add types dynamically without affecting existing core behavior.

## Related issues

https://github.com/apache/fury/issues/2208

The `Encoders` class has new `registerCustomCodec` and `registerCustomCollectionFactory` methods.
All custom types are written with the existing protocol as embedded memory buffers just like any other field, but with a custom byte representation, so there should be no wire compatibility concerns.

## Benchmark

There should be no change to performance in existing use cases. The code is carefully written to have no runtime impact if not used. Custom types are invoked via static methods or instance method on static final fields, which should be easily inlined by jit for minimum overhead.

Here is example generated code to help show this:
https://gist.github.com/stevenschlansker/ed7dae863e78d3c87e30bdea39fa8dea

### DIFF
--- a/java/fury-format/README.md
+++ b/java/fury-format/README.md
@@ -11,3 +11,45 @@ Fury row format is heavily inspired by spark tungsten row format, but with chang
 - Support adding fields without breaking compatibility.
 
 The initial fury java row data structure implementation is modified from spark unsafe row/writer.
+
+It is possible to register custom type handling and collection factories for the row format -
+see Encoders.registerCustomCodec and Encoders.registerCustomCollectionFactory.
+
+A short example:
+
+```
+@Data
+public static class UuidType {
+  public UUID f1;
+  public UUID[] f2;
+  public SortedSet<UUID> f3;
+}
+
+static class UuidEncoder implements CustomCodec.MemoryBufferCodec<UUID> {
+  @Override
+  public MemoryBuffer encode(final UUID value) {
+    final MemoryBuffer result = MemoryBuffer.newHeapBuffer(16);
+    result.putInt64(0, value.getMostSignificantBits());
+    result.putInt64(8, value.getLeastSignificantBits());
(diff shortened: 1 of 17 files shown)


java/fury-format/README.md
```diff
@@ -11,3 +11,6 @@ Fury row format is heavily inspired by spark tungsten row format, but with chang
 - Support adding fields without breaking compatibility.
 
 The initial fury java row data structure implementation is modified from spark unsafe row/writer.
+
+It is possible to register custom type handling and collection factories for the row format -
+see Encoders.registerCustomCodec and Encoders.registerCustomCollectionFactory.
```
chaokunyang, line 16: Could we add an example here, maybe in next PR?

chaokunyang (approved):
Looks awesome! Excelent work

chaokunyang:
image

Code lint failed, Please use ` mvn spotless:apply` to format code

chaokunyang:
image

Some generated code failed to compile

stevenschlansker:
Fixed the last bug and applied formatting.

stevenschlansker:
One question: is there any important difference between `Field.notNullable(fieldName, ArrowType.Binary.INSTANCE)` vs `DataTypes.primitiveArrayField(fieldName, DataTypes.int8())`?

chaokunyang:
> primitiveArrayField

I think both are OK

Issue #2245, opened 2025-05-19
[C++] Improve and enhance the fury-cpp benchmark module


pandalee99:
### Feature Request

Improve and enhance the fury-cpp benchmark module 

There are still some problems with this module. For example, the performance test is not sufficient. I hope to continue to improve and enhance fury-cpp.

### Is your feature request related to a problem? Please describe

#2033

lzaeh:
This looks interesting—may I work on it?

PR #2246, opened 2025-05-19
fix(C++): Fix reversed parameters and incorrect calls in string_util benchmark

lzaeh:
## What does this PR do?
This PR fixes several minor issues and improves the implementation of string utility benchmarks:
- Corrected reversed function parameters.
- Fixed an incorrect function call.
- Replaced some usages of auto with explicit types for better readability and maintainability.
- Refactored random number generation by encapsulating std::mt19937 in a thread-local helper function.

One part of the original code used default random generation methods, which could result in uneven value distributions and non-deterministic behavior across platforms. To improve randomness quality and thread-safety, I replaced it with a `thread_local` encapsulated `std::mt19937` generator. This ensures better reproducibility and consistency during benchmarking.

After these changes, the code compiles and runs correctly. Benchmark results are stable and as expected. (See image below.)

Screenshot 2025-05-19 at 17 24 58

## Related issues
- #2245

### DIFF
--- a/cpp/fury/benchmark/benchmark_string_util.cc
+++ b/cpp/fury/benchmark/benchmark_string_util.cc
@@ -148,17 +148,16 @@ std::u16string generateUtf16(size_t length) {
   return result;
 }
 
-
 /*
- * Number of test strings to generate.
- * Each benchmark will run on `num_tests` samples.
+ *  Number of test strings to generate.
+ *  Each benchmark will run on `num_tests` samples.
  */
 const size_t num_tests = 1000;
 
 /*
- * Target length when generating individual strings.
- * For single-byte encodings (ASCII, Latin-1), this is exact.
- * For multibyte encodings (UTF-8, UTF-16), actual string size may vary slightly.
+ *  Target length when generating individual strings.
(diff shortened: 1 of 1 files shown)


cpp/fury/benchmark/benchmark_string_util.cc
```diff
@@ -34,15 +34,21 @@
  */
 
 // Generate random bytes (0x00 to 0xFF)
+
```
pandalee99, line 37: I think this space should be deleted

cpp/fury/benchmark/benchmark_string_util.cc
```diff
@@ -156,8 +162,8 @@ const size_t string_length = 1000;
 std::vector<std::string> generateAsciiString(size_t num_tests,
                                              size_t string_length) {
   std::vector<std::string> test_strings;
-  for (size_t i = 0; i < string_length; ++i) {
-    test_strings.push_back(generateUtf8(num_tests));
+  for (size_t i = 0; i < num_tests; ++i) {
```
pandalee99, line 165: If you think num_tests and string_length are different, could you add a comment?

pandalee99 (approved):
LGTM

PR #2116, opened 2025-03-23
feat(java): support nested bean in array/collection/map for row format

chaokunyang:
## What does this PR do?

support nested bean in array/collection/map for row format

## Related issues

Closes #2106

## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

PR #2250, opened 2025-05-21
feat(java): fury row encoder now supports implementing interfaces with simple value type

### LINKED ISSUE #2223: [Question] [fury-format] Are there any libs related to binary operations of rows?
### Question

Operate binary without deserialization. For example, String equals,  array contains, map exists, etc

stevenschlansker:
## What does this PR do?

You can now call `Encoders.registerLazyBeanInterface` and Fury will generate a bean implementation type for you.
This should allow easy integration with Immutables library for example - I will be testing that over the coming days.

## Related issues

Fixes #2223

New api to register interface types. Since this was not implemented before, there should be no breaking change.

### DIFF
--- a/java/fury-core/src/main/java/org/apache/fury/type/TypeUtils.java
+++ b/java/fury-core/src/main/java/org/apache/fury/type/TypeUtils.java
@@ -336,7 +336,7 @@ public class TypeUtils {
     if (type.getClass() == Class.class) {
       return (Class<?>) type;
     } else {
-      return getRawType(typeRef.getType());
+      return getRawType(type);
     }
   }
 
@@ -549,6 +549,10 @@ public class TypeUtils {
         .where(new TypeParameter<V>() {}, valueType);
   }
 
+  public static boolean isContainer(Class<?> cls) {
+    return isCollection(cls) || isMap(cls) || cls.isArray();
+  }
+
   public static boolean isCollection(Class<?> cls) {
     return cls == ArrayList.class || Collection.class.isAssignableFrom(cls);
   }
@@ -569,7 +573,11 @@ public class TypeUtils {
   }
 
   public static boolean isBean(Type type, CustomTypeRegistry customTypes) {
-    return isBean(TypeRef.of(type), new LinkedHashSet<>(), customTypes);
+    return isBean(TypeRef.of(type), customTypes);
+  }
+
+  public static boolean isBean(TypeRef<?> type, CustomTypeRegistry customTypes) {
+    return isBean(type, new TypeResolutionContext(customTypes));
   }
 
   /**
@@ -578,18 +586,18 @@ public class TypeUtils {
    * nested class is ok.
    */
   public static boolean isBean(TypeRef<?> typeRef) {
-    return isBean(typeRef, new LinkedHashSet<>());
+    return isBean(typeRef, new TypeResolutionContext(CustomTypeRegistry.EMPTY));
   }
 
-  private static boolean isBean(TypeRef<?> typeRef, LinkedHashSet<TypeRef<?>> walkedTypePath) {
-    return isBean(typeRef, walkedTypePath, CustomTypeRegistry.EMPTY);
+  public static boolean isBean(Class<?> clz, TypeResolutionContext ctx) {
+    return isBean(TypeRef.of(clz), ctx);
   }
 
-  private static boolean isBean(
-      TypeRef<?> typeRef,
-      LinkedHashSet<TypeRef<?>> walkedTypePath,
-      CustomTypeRegistry customTypes) {
+  public static boolean isBean(TypeRef<?> typeRef, TypeResolutionContext ctx) {
     Class<?> cls = getRawType(typeRef);
+    if (ctx.isSynthesizedBeanType(cls)) {
+      return true;
+    }
     if (Modifier.isAbstract(cls.getModifiers()) || Modifier.isInterface(cls.getModifiers())) {
       return false;
     }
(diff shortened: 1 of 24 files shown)


java/fury-format/src/test/java/org/apache/fury/format/encoder/ImplementInterfaceTest.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  @Test
+  public void testInterfaceTypes() {
+    final InterfaceType bean1 = new ImplementInterface(42, "42");
+    final RowEncoder<InterfaceType> encoder = Encoders.bean(InterfaceType.class);
+    final BinaryRow row = encoder.toRow(bean1);
+    final MemoryBuffer buffer = MemoryUtils.wrap(row.toBytes());
+    row.pointTo(buffer, 0, buffer.size());
+    final InterfaceType deserializedBean = encoder.fromRow(row);
```
chaokunyang, line 84: Is it possible to pass `InterfaceType.class` to `fromRow`? In this way, users don't have to invoke `Encoders.registerLazyBeanInterface`. This will simplify the usage.

We can cache generated `InterfaceTypeImpl` using Weak/SoftMap

java/fury-format/README.md
```diff
@@ -13,7 +13,8 @@ Fury row format is heavily inspired by spark tungsten row format, but with chang
 The initial fury java row data structure implementation is modified from spark unsafe row/writer.
 
 It is possible to register custom type handling and collection factories for the row format -
-see Encoders.registerCustomCodec and Encoders.registerCustomCollectionFactory.
+see Encoders.registerCustomCodec and Encoders.registerCustomCollectionFactory. For an interface,
+Fury can synthesize a simple value implementation, see Encoders.registerLazyBeanInterface.
```
chaokunyang, line 17: Could we add code example in this file?

java/fury-format/src/test/java/org/apache/fury/format/encoder/ImplementInterfaceTest.java
stevenschlansker: You are right, registration is not strictly required.
I was concerned about the implementation being too eager to implement all interfaces it sees, if you accidentally have something like `Supplier` or other non-data interface in your type hierarchy.

But it will probably just cause compile errors anyway, so for v1 we can just assume all interfaces referred to from another interface are implementable, and trust the user to only give us types that are sensible. Thus there is now no change to public API.

stevenschlansker: Weak / soft map seems to not be necessary, we can still create the codec up-front, since we know all interfaces referred to from the given input interface.

stevenschlansker:
I removed the need to register interface types, and updated the docs to include an example.

java/fury-core/src/main/java/org/apache/fury/type/TypeUtils.java
```diff
@@ -737,7 +744,7 @@ private static LinkedHashSet<Class<?>> listBeansRecursiveInclusive(
         TypeRef<?> elementType = getElementType(typeToken);
         LinkedHashSet<TypeRef<?>> newPath = new LinkedHashSet<>(walkedTypePath);
         newPath.add(elementType);
-        beans.addAll(listBeansRecursiveInclusive(elementType.getClass(), newPath, customTypes));
```
stevenschlansker, line 740: I think this is a pre-existing bug - `elementType.getClass()` is always `TypeRef.class`, should this be `getRawType` instead?

stevenschlansker: maybe introduced in #2116

java/fury-format/src/main/java/org/apache/fury/format/encoder/RowEncoderBuilder.java
```diff
@@ -86,12 +90,19 @@ public RowEncoderBuilder(TypeRef<?> beanType) {
       // non-public class is not accessible in other class.
       clsExpr =
           new Expression.StaticInvoke(
-              Class.class, "forName", CLASS_TYPE, false, Literal.ofClass(beanClass));
```
stevenschlansker, line 89: `Class.forName` accepts a String literal, not a Class literal.

java/fury-core/src/main/java/org/apache/fury/type/TypeUtils.java
```diff
@@ -735,9 +742,12 @@ private static LinkedHashSet<Class<?>> listBeansRecursiveInclusive(
         }
       } else if (isCollection(type)) {
         TypeRef<?> elementType = getElementType(typeToken);
+        while (isCollection(elementType.getRawType()) || elementType.getRawType().isArray()) {
```
stevenschlansker, line 745: this update is necessary otherwise `List<List<Double>>` tries to inspect `List<Double>` as the element, but really it should be `Double`

java/fury-format/src/test/java/org/apache/fury/format/encoder/ImplementInterfaceTest.java
stevenschlansker: Well, I actually did not get this to work today. Either the type checks are too loose and include random things like `interface List<E>` or they are too strong and do not include the interface type I am interested in. Having a list of known safe types to implement made this easy. I can take another look tomorrow.

chaokunyang (approved):
Looks great, thanks for this excelent work!

PR #2254, opened 2025-05-27
feat(java): row format supports Optional<T>

stevenschlansker:
## What does this PR do?

Support `Optional<T>` in java row format
Stored as an inline nullable field - so this implementation chooses not to differentiate between `(Optional) null` and `Optional.empty()` and always deserialize as `empty`.

New type supported - docs updated


chaokunyang (approved):
This looks great and elegant !

PR #2256, opened 2025-05-27
feat(java): row format supports Record types

stevenschlansker:
## What does this PR do?

Support Record types in row format

Treated as Bean types, except we have to use canonical constructor instead of unsafe field setting

Test only runs on Java 17+

New type support in java row format

### DIFF
--- a/ci/run_ci.sh
+++ b/ci/run_ci.sh
@@ -146,10 +146,11 @@ integration_tests() {
 
 jdk17_plus_tests() {
   java -version
+  export JDK_JAVA_OPTIONS="--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED"
   echo "Executing fury java tests"
   cd "$ROOT/java"
   set +e
-  mvn -T10 --batch-mode --no-transfer-progress test install -pl '!fury-format,!fury-testsuite'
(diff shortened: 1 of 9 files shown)


ci/run_ci.sh
```diff
@@ -146,10 +146,11 @@ integration_tests() {
 
 jdk17_plus_tests() {
   java -version
+  export JDK_JAVA_OPTIONS="--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED"
   echo "Executing fury java tests"
   cd "$ROOT/java"
   set +e
-  mvn -T10 --batch-mode --no-transfer-progress test install -pl '!fury-format,!fury-testsuite'
+  mvn -T10 --batch-mode --no-transfer-progress install
```
stevenschlansker, line 153: Is there a reason fury-format is excluded here? I think it means integration tests can run with the wrong fury-format jar - one from a snapshot repo, instead of the one from the PR build

chaokunyang: > Is there a reason fury-format is excluded here? I think it means integration tests can run with the wrong fury-format jar - one from a snapshot repo, instead of the one from the PR build

It should not be skipped, we skipped before just because running arrow on JDK17+ in earlier version is tricky

PR #2266, opened 2025-05-29
chore(java): skip xlang tests when pyfory is not available

stevenschlansker:
## What does this PR do?

right now a Java dev has to contend with failing tests that require Python
by skipping the xlang tests when pyfory is not available, a Java developer can build without needing to set up pyfory

### DIFF
--- a/java/fory-test-core/src/main/java/org/apache/fory/test/TestUtils.java
+++ b/java/fory-test-core/src/main/java/org/apache/fory/test/TestUtils.java
@@ -21,10 +21,13 @@ package org.apache.fory.test;
 
 import java.io.BufferedReader;
 import java.io.InputStreamReader;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
+import org.testng.SkipException;
 
 public class TestUtils {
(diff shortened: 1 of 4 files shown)


chaokunyang:
I prefer to keep it in fury-format module, but skip it if pyfory is not installed, or we can use a env variable to skip it, and enable the tests in ci via this env variable

chaokunyang:
CrossLanguageTest in fory-core has same issue

java/fory-test-core/src/main/java/org/apache/fory/test/TestUtils.java
```diff
@@ -74,4 +77,17 @@ public static boolean executeCommand(
...
     }
   }
+
+  public static void verifyPyforyInstalled() {
+
+    if (executeCommand(
+        Arrays.asList(
+            "python",
+            "-c",
+            "import importlib, sys; sys.exit(0 if importlib.util.find_spec(\"pyfury\") is None else 1)"),
```
chaokunyang, line 87:
```suggestion
            "import importlib, sys; sys.exit(0 if importlib.util.find_spec(\"pyfory\") is None else 1)"),
```

stevenschlansker:
Ok, I updated the code to skip the xlang tests if outside CI and pyfory is not available.

PR #2269, opened 2025-05-30
fix(go): fix metastringbytes inconsistency issue

### LINKED ISSUE #2268: [go] The byte results generated by using MetaStringResolver are inconsistent
### Search before asking

### Version

0.10.3

### Component(s)

Go

### Minimal reproduce step

The processing methods in Python modules and Java modules are the same, but not in go

The byte streams are inconsistent

### What did you see instead?

Inconsistency

### Anything Else?

_No response_

lzaeh:
## What does this PR do?
**Summary:**
 Fixes inconsistencies between the Go implementation of `MetaStringResolver` and the reference Cython implementation. Specifically, it corrects encoding logic, hash computation, and ensures buffer output matches Python’s behavior. Also adds comprehensive test coverage for empty strings, long strings, Chinese, and Japanese strings in both Go and Python.

**Details:**

1. Adjusted `SmallStringThreshold` in `resolver.go` to match the Cython implementation (from 8 to 16), ensuring proper classification of short vs long strings.
2. Corrected encoding logic for both small and large strings. Updated hash computation to use 64-bit MurmurHash3 with seed 47, aligning with the Cython version.
3. Added an explicit check for empty strings in `meta_string_encoder.go` to default their encoding to `UTF_8`, avoiding incorrect assignment of `LOWER_SPECIAL` as seen previously.
4. Added new test cases in `test_metastring_resolver.py`, and created a Go test file `metastring_resolver_test.go` covering edge cases such as empty strings, multilingual strings, and long strings.

**How did I discover the issue**

I began by studying the Cython implementation of `MetaStringResolver` to understand the intended encoding and serialization logic. After becoming familiar with the Python logic, I reviewed the Go implementation and began debugging.

- **First**, I noticed the `SmallStringThreshold` was inconsistent: Python uses 16, but Go had it set to 8. This caused strings like `"hello, world"` (12 bytes) to be treated as *long* strings in Go (leading to additional hashcode bytes written), while Python treated them as *short* strings — leading to buffer mismatches.
image

  *Fix: Updated `SmallStringThreshold` to 16 in Go.*

- **Second**, I found mismatches in the encoding byte. After investigating the source of the encoding logic in Go, I discovered the computation did not match the Python logic, and it assigned incorrect encoding values.
image

  *Fix: Rewrote the encoding logic to match Python's criteria.*

- **Third**, during testing of long strings, I found the hashcode between Go and Python differed. Upon inspection:
  - Go used a 128-bit Murmur3 hash with no seed.
  - Python used a 64-bit Murmur3 hash with seed 47.
image

  *Fix: Replaced Go's hash function with a 64-bit Murmur3 with seed 47, and replicated the same bit-shifting and encoding masking.*

- **Fourth**, while testing empty strings, I found that Go assigned them the encoding `0x01` (`LOWER_SPECIAL`), while Python correctly assigned `0x00` (`UTF_8`). Tracing through the encoding logic, I saw that Go's encoding decision relied on flags computed from a `for` loop — which was skipped for empty strings, causing all flags to remain `true` and defaulting to `LOWER_SPECIAL`.
image
The last image actually shows the result after a successful debug. Sorry for the confusion it may have caused.

  *Fix: Added an explicit empty string check in `ComputeEncoding`, returning `UTF_8` early.*

- All these changes were validated by inspecting serialized buffers, and outputs were compared against Python’s output step-by-step.
## Related issues
close #2268


go/fory/metastring_resolver_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/apache/fory/go/fory/meta"
+)
+
+func TestMetaStringResolver(t *testing.T) {
```
pandalee99, line 28: I think `go/fory/metastring_resolver_test.go` should be: `go/fory/test/metastring_resolver_test.go`, because it's a bit strange to put a separate test file in the main directory.

python/pyfory/tests/test_metastring_resolver.py
```diff
@@ -23,14 +23,44 @@
 def test_metastring_resolver():
     resolver = MetaStringResolver()
     encoder = MetaStringEncoder("$", "_")
+
+    # Test 1: Regular English string
     metastr1 = encoder.encode("hello, world")
     metabytes1 = resolver.get_metastr_bytes(metastr1)
-    buffer = Buffer.allocate(32)
+    buffer = Buffer.allocate(512)
```
pandalee99, line 30: Well, although it's fine to change the default memory allocation, I think it's unnecessary

go/fory/metastring_resolver_test.go
lzaeh: Yes sir, I will do it right now.

pandalee99 (approved):
LGTM, your thinking and patience are admirable.

PR #2274, opened 2025-06-02
chore(java): Use `crossLanguage` instead of `language` in Fory

CherishCai:
## What does this PR do?
The Fory class fields `language` and `crossLanguage` are redundant and can be quickly determined using `crossLanguage` without the need for `language`.


CherishCai:
Additionally, the field 'peerLanguage' in the Fory class currently not needed. Will it have any future use? Perhaps it will be used for log printing?

chaokunyang:
> Additionally, the field 'peerLanguage' in the Fory class currently not needed. Will it have any future use? Perhaps it will be used for log printing?

This will be used in future, fory can choose better serialization method for faster deserialization in peer based on this field

Issue #2160, opened 2025-04-19
[Python] Support meta share for type forward/backward compatibility


chaokunyang:
### Feature Request

Meta share is used for support type forward/backward compatibility. Currently this is only supported in native java/javascript serializaiton.

We should support  meta share for xlang serialization format, so that when a class in java/python/... add/delete a field, the deserializers in python can still succeed

urlyy:
Hi, I plan to try to handle this `Python` issue as a preparation for the `Rust` implementation in `OSPP`.

urlyy:
Hi, I just looked at the implementation of metashare and compatible in Java version. But looking at the Python code, it seems that the implementation of "compatible based on metashare" hasn't even started yet. Starting from scratch is a bit challenging. Do you have any suggestions? I plan to start by implementing a simple meta_context, related resolvers, and serializers based on the protocol documentation.

urlyy:
@chaokunyang 
Hi, I have some questions: 
1. `meta_share` is used when a Fury object processes some objects of the same class multiple times. So if I use a Fury object sometimes deal with the parent class and sometimes the subclass, does that mean I shouldn't use meta_share? Are there any other situations where I need to be compatible but should not use meta_share?
2. Is `schema evolution` the implementation approach of "forward/backward compatibility"?
3. May I start now? I want to finish an issue related to compatibility before the student application deadline for OSPP.

chaokunyang:
> [@chaokunyang](https://github.com/chaokunyang) Hi, I just read through the specification document and have some questions to ask:
> 
> 1. `meta_share` is used when a Fury object processes some objects of the same class multiple times. So if I use a Fury object sometimes deal with the parent class and sometimes the subclass, does that mean I shouldn't use meta_share? Are there any other situations where I need to be compatible but should not use meta_share?
> 2. Is `schema evolution` the implementation approach of "type forward/backward compatibility"?
> 3. Does this issue involve `codegen`?
> 
> Thank you!

Hi @urlyy , for your quesiont:
1. If you use Fury to deal with the parent class, fury need to write meta for parent class. If you use Fury to deal with the subclass, fury will write meta for subclass, but note that meta for subclass includes meta for parent class
2. You are right, `schema evolution spec` is the  the implementation approach of "type forward/backward compatibility"
[...]

PR #2278, opened 2025-06-02
feat(python): support xlang metashare (WIP)

urlyy:
## What does this PR do?
Prepare to implement xlang metashare encode&decode in pyfory. 
Now finish `field_info` and `type_info(TypeDef)` in python not cython. 
I will continue to finish type_resolver for registry.

other small change:
- Add `Buffer::write_uint8` for writing 1 byte header.
- Add metastring encoding options for field_name and namespace metastring encode.

## Some Question
- Still confused about naming. For me, I want to change all `class_` prefix to `type_`, like `class_info -> type_info`, `class_resolver -> type_resolver`. So it will be unified. 
- There is not `writeVarUint32Small7` in pyfury cython. Now I use `write_varuint32`. Should I add this function?
- Now I just finish the xlang TypeDefMeta specification. As the java version is too complex to imitate, I use my way to implement this: not use outer Encoder/Decoder/Resolver class, just implement the  `xread/xwrite` logic in `type_info/field_info` class. If this is not good, I will rewrite this part.
- Is Java now support testing with python using Xlang&Compatible mode? 

## Related issues
#2160

WIP

## Benchmark

TODO

### DIFF
--- a/python/pyfory/meta/meta_share.py
+++ b/python/pyfory/meta/meta_share.py
@@ -1,0 +1,39 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from enum import IntEnum
+
+class CompatibleMode(IntEnum):
+    SCHEMA_CONSISTENT = 1
+    COMPATIBLE = 2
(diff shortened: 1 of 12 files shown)


python/pyfory/meta/meta_share.py
```diff
@@ -0,0 +1,39 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+from enum import IntEnum
+
+class CompatibleMode(IntEnum):
+    SCHEMA_CONSISTENT = 1
+    COMPATIBLE = 2
+
+class MetaContext:
```
chaokunyang, line 24: We'd better to put this class in _serialization.pyx for better performance

chaokunyang:
It would be great if we could change all class_ prefix to type_, like class_info -> type_info, class_resolver -> type_resolver. This will make the code more clean. How about opening another PR for this? After that PR is merged, we could merge from main branch again to resovle the conflicts

urlyy:
> It would be great if we could change all class_ prefix to type_, like class_info -> type_info, class_resolver -> type_resolver. This will make the code more clean. How about opening another PR for this? After that PR is merged, we could merge from main branch again to resovle the conflicts

ok

pandalee99:
@urlyy Hi, are you considering continuing to implement this PR?

urlyy:
> @urlyy Hi, are you considering continuing to implement this PR?

Sorry😢, I've been a bit busy these days, and the PR is a bit troublesome for me. I don't think I'll continue working on this PR.

Issue #2259, opened 2025-05-28
Deserialize big object using Fury frequently, triggering a Full GC


ZhiQiang-Tiger:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

Fury version:0.6.0 & 0.10.1
JDK version: jdk11
OS: Linux

### Component(s)

Java

### Minimal reproduce step

When deserialization of a large object fails, handleReadFailed will log a large object. If the condition happens frequently, it will result in frequent Full GC.

![Image](https://github.com/user-attachments/assets/52d5011b-c99c-417f-b8d0-8aaa3cc14308)

### What did you expect to see?

I'd like to add a new configuration to "FuryBuilder" when building fury instance to control the amount of error messages printed, a user-selectable logging level of detail.
For example, define some enum instances like "Whole object, partial object properties, nothing printed".

### What did you see instead?

![Image](https://github.com/user-attachments/assets/556e3ac6-e1c9-4de2-b325-eff2ba737e72)

![Image](https://github.com/user-attachments/assets/c8a11820-c538-42b1-b41b-f4e9a02e7831)

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

chaokunyang:
Please reopen this issue if it still persists on fory 0.14.1

PR #2263, opened 2025-05-29
chore: rename apache fury to apache fory

chaokunyang:
## What does this PR do?

Based on the discussion and vote in Fury PPMC and ASF trademark group, Apache Fury will renamed to Apache Fory.

This PR renames all fury in the codebase to fory.

## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

PR #2284, opened 2025-06-03
feat(java): add log downgrade configuration

ZhiQiang-Tiger:
## What does this PR do?

Users can choose the level of error messages to be printed.
## Related issues

Deserialize big object using Fury frequently, triggering a Full GC when logging failed object #2259

The user can select the log mode by adding the configuration ExceptionLogMode via FuryBuilder. When log sampling, set the sampling step by logSampleStep.
- [ ] add enum class ExceptionLogMode.
- [ ] add sample step param.


java/fury-core/pom.xml
```diff
@@ -51,7 +51,16 @@
     <maven.compiler.target>8</maven.compiler.target>
     <fury.java.rootdir>${basedir}/..</fury.java.rootdir>
   </properties>
-
+    <distributionManagement>
```
chaokunyang: Please remove this

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -19,47 +19,18 @@
...
-import java.util.function.Function;
-import javax.annotation.concurrent.NotThreadSafe;
 import org.apache.fury.builder.JITContext;
 import org.apache.fury.collection.IdentityMap;
-import org.apache.fury.config.CompatibleMode;
-import org.apache.fury.config.Config;
-import org.apache.fury.config.FuryBuilder;
-import org.apache.fury.config.Language;
-import org.apache.fury.config.LongEncoding;
+import org.apache.fury.config.*;
```
chaokunyang: Fory don't use `*`  for imports, please execute `mvn spotless:apply & mvn checkstyle:check`

java/fury-core/src/test/java/org/apache/fury/LogDowngradeTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.testng.annotations.Test;
+
+import java.util.Base64;
+
+
+public class LogDowngradeTest extends FuryTestBase{
+    @Test
+    public void testLog(){
+        Fury fury = Fury.builder()
+                .exceptionLogMode(ExceptionLogMode.SAMPLE_PRINT, 2)
```
chaokunyang:
```suggestion
                .withExceptionLogMode(ExceptionLogMode.SAMPLE_PRINT, 2)
```

java/fury-core/src/test/java/org/apache/fury/LogDowngradeTest.java
```diff
@@ -0,0 +1,48 @@
...
+//        byte[] serialize = fury.serialize(testObj);
+//        String string = Base64.getEncoder().encodeToString(serialize);
+//        System.out.println(string);
+        String str = "AgDCAgDEAv8UdGVzdEL9/f3+Af4B/gH+Af4B/gH+Af3+AQ==";
+        byte[] decode = Base64.getDecoder().decode(str);
+        Object deserialize = fury.deserialize(decode);
+        System.out.println(deserialize);
+    }
+
+    public TestObj builderTestObj(){
```
chaokunyang: I think we don't need this. We could just serialize `BeanA.createBeanA(2)`, and overwrite last bytes, the deserialization will fail.

java/fury-core/src/main/java/org/apache/fury/util/ExceptionUtils.java
```diff
@@ -57,15 +59,36 @@ public static StackOverflowError trySetStackOverflowErrorMessage(
...
   }
 
+  public static List<Object> getExceptionObjects(Fury fury) {
+      ObjectArray readObjects = ((MapRefResolver) fury.getRefResolver()).getReadObjects();
+      // carry with read objects for better trouble shooting.
+      List<Object> objects = Arrays.asList(readObjects.objects).subList(0, readObjects.size);
+      switch (fury.getExceptionLogMode()) {
+        case NONE_PRINT:
+          return new ArrayList<>();
+        case SAMPLE_PRINT:
```
chaokunyang, line 77: how about limiting max samples instead of configuring sample step?  Using step will print little things for small object, and still print lots of things for big object

java/fury-core/src/main/java/org/apache/fury/config/FuryBuilder.java
```diff
@@ -142,6 +145,20 @@ public FuryBuilder serializeEnumByName(boolean serializeEnumByName) {
     return this;
   }
 
+  /** exception log level
+   * choose the log level, print the error message.
+   * */
+  public FuryBuilder exceptionLogMode(ExceptionLogMode exceptionLogMode) {
```
chaokunyang:
```suggestion
  // -1 indicates all sample,s 0 indcates no samples, a positive int indicates max samples
  public FuryBuilder withDeserializationExceptionMaxRecordSamples(int maxSamples) {
```

java/fury-core/src/main/java/org/apache/fury/util/ExceptionUtils.java
ZhiQiang-Tiger: > how about limiting max samples instead of configuring sample step? Using step will print little things for small object, and still print lots of things for big object

I think that the user will select a suitable log mode by judging the situation of the object. In other words，our design is scalable and we can employ multiple logging strategies to solve different situation.

java/fury-core/src/main/java/org/apache/fury/config/ExceptionLogMode.java
```diff
@@ -0,0 +1,7 @@
+package org.apache.fury.config;
```
LiangliangSui, line 19: The newly added files need to have a license header inserted at the top.

ZhiQiang-Tiger: > The newly added files need to have a license header inserted at the top.

okay, I will fix it later.

java/fury-core/src/main/java/org/apache/fury/util/ExceptionUtils.java
chaokunyang: Users can know how big of theri object graph. Iif we use step, the issue won't be resolved in some cases. Or we could increase a new strategy which limit max samples, and let that mode as the default mode

chaokunyang:
@ZhiQiang-Tiger Please check branch from latest main branch, we renamed fury to fory in #2263

java/fury-core/src/main/java/org/apache/fury/Fury.java
```diff
@@ -1679,6 +1681,13 @@ public boolean compressLong() {
     return config.compressLong();
   }
 
+  public ExceptionLogMode getExceptionLogMode() {
+    return config.getExceptionLogMode();
+  }
+  public int getLogSampleStep() {
```
chaokunyang, line 1687: Please format code by `mvn spotless:apply`

chaokunyang: And I think we don't need to duplicate fields here, we could just pass value directly from Config

java/fury-core/src/main/java/org/apache/fury/util/ExceptionUtils.java
```diff
@@ -57,15 +59,35 @@ public static StackOverflowError trySetStackOverflowErrorMessage(
...
       Platform.throwException(t);
       throw new IllegalStateException("unreachable");
     }
   }
 
+  public static List<Object> getExceptionObjects(Fury fury) {
+      ObjectArray readObjects = ((MapRefResolver) fury.getRefResolver()).getReadObjects();
+      // carry with read objects for better trouble shooting.
+      List<Object> objects = Arrays.asList(readObjects.objects).subList(0, readObjects.size);
+      switch (fury.getExceptionLogMode()) {
```
chaokunyang, line 74:
```suggestion
      switch (fury.getConfig().getExceptionLogMode()) {
```

java/fury-core/src/main/java/org/apache/fury/Fury.java
ZhiQiang-Tiger: > And I think we don't need to duplicate fields here, we could just pass value directly from Config

ok. At the time it was to regulate

Issue #2270, opened 2025-05-31
[Java]  raise an exception with error message when when registering serializer for Map/List


chaokunyang:
### Feature Request

When  registering serializer for Map/List, one should extend `AbstractMapSerializer/AbstractListSerializer`. If the customized serializer class extend `Serializer` directly. When serializing, Fury will raise CastException.

### Is your feature request related to a problem? Please describe

When  registering serializer for Map/List, we should check serializer type. If it's not extended from `AbstractMapSerializer/AbstractListSerializer`, we should raise exception in ClassResolver, and the error message should tell users what they should do

### Additional context

#2251

chaokunyang:
cc @OmCheeLin

OmCheeLin:
@chaokunyang ok, got it.

Issue #2251, opened 2025-05-25
[Java] Customized serializer for a `Collection` causes `ClassCastException`


gudzpoz:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fury/issues) and found no similar issues.

### Version

Fury: `'org.apache.fury:fury-core:0.10.2'`
JDK: GraalVM 23.0.2+7.1 (build 23.0.2+7-jvmci-b01)

### Component(s)

Java

### Minimal reproduce step

A collection class with a custom serializer:
```java
    public static final class EmptyList extends AbstractList<Object> {
        @Override
        public Object get(int index) {
            throw new IndexOutOfBoundsException();
        }
        @Override
        public int size() {
            return 0;
        }

        public static final class EmptySerializer extends Serializer<EmptyList> {
            public EmptySerializer(Fury fury) {
                super(fury, EmptyList.class);
            }
            @Override
            public void write(MemoryBuffer buffer, EmptyList value) {
                // no-op
            }
            @Override
            public EmptyList read(MemoryBuffer buffer) {
                return new EmptyList();
            }
        }
    }
```
When it is put in another class:
```java
    public record SomeRecord(EmptyList specialList, List<Object> normalList) {
    }
```
It will produce `ClassCastException` when serialized:
```java
    @Test
    public void testCollection() {
        Fury fury = Fury.builder()
                .withLanguage(Language.JAVA)
                .build();
        fury.register(EmptyList.class);
        fury.registerSerializer(EmptyList.class, EmptyList.EmptySerializer.class);
        fury.register(SomeRecord.class);
        SomeRecord someRecord = new SomeRecord(new EmptyList(), List.of());
        byte[] bytes = fury.serialize(someRecord); // <-- ClassCastException here
        Object output = fury.deserialize(bytes);
        SomeRecord someRecord2 = assertInstanceOf(SomeRecord.class, output);
        assertEquals(someRecord, someRecord2);
    }
```

### What did you expect to see?

1. I expect the `testCollection` test above to pass without throwing exceptions.
2. I expect being able to write a custom serializer for *any* class. Forcing the usage of `AbstractCollectionSerializer` for all collections is not ideal and there are plenty of special cases when one wants a special serializer.

### What did you see instead?

```
2025-05-26 03:42:55 INFO  Fury:159 [Test worker] - Created new fury org.apache.fury.Fury@2072acb2
[...]

PR #2291, opened 2025-06-03
fix(java): raise exception when registering invalid serializer for Map/List

OmCheeLin:
## What does this PR do?
raise an exception with error message when when registering serializer for Map/List.

## Related issues

- #2251
- #2270


java/fory-core/src/main/java/org/apache/fory/util/ValidateSerializer.java
```diff
@@ -0,0 +1,53 @@
...
+package org.apache.fory.util;
+
+import java.util.Collection;
+import java.util.Map;
+import org.apache.fory.serializer.Serializer;
+import org.apache.fory.serializer.collection.AbstractCollectionSerializer;
+import org.apache.fory.serializer.collection.AbstractMapSerializer;
+
+public class ValidateSerializer {
+  public static void validateSerializer(
```
chaokunyang: How about merging this class into SerializationUtils.java, and make SerializationUtils.java public. Seems we don't need a separate class for this validation

chaokunyang (approved):
LGTM

PR #2297, opened 2025-06-04
fix(java): row encoders custom codec accepts class supertype

stevenschlansker:
## What does this PR do?
`Encoders.registerCustomCodec` now allows class token to be supertype of T
this matters for e.g. `registerCustomCodec(Id.class, new CustomCodec<Id<T>, ...>);` since `Id.class` is not `Id<T>` but `Id` is super of `Id<T>`


stevenschlansker:
Actually, I'm not sure this is the right fix. Sorry for the noise

PR #2304, opened 2025-06-06
feat(java): row encoder supports synthesizing interfaces nested inside of records

stevenschlansker:
## What does this PR do?

Row encoder synthesized interfaces now work inside enclosing record classes.
The previous logic was over-complicated to work around the overly aggressive expanding of types. With the fix in #2265 this is no longer necessary and we can simplify the logic around detecting interfaces to synthesize, in a way that better supports nesting too.


chaokunyang (approved):
LGTM

PR #2305, opened 2025-06-06
feat(java): row encoder supports custom rewriting values without changing their type

stevenschlansker:
## What does this PR do?

add support for row encoder CustomCodec to read or write replace values

we use this to e.g. replace Fory generated interface types with our own implementation, generated via Immutables ( https://immutables.github.io/ )


java/fory-core/src/main/java/org/apache/fory/type/TypeUtils.java
```diff
@@ -631,9 +631,12 @@ public static boolean isBean(TypeRef<?> typeRef, TypeResolutionContext ctx) {
                   // do field modifiers and getter/setter validation here, not in getDescriptors.
                   // If Modifier.isFinal(d.getModifiers()), use reflection
                   // private field that doesn't have getter/setter will be handled by reflection.
-                  return ctx.getCustomTypeRegistry().hasCodec(cls, t.getRawType())
-                      || isSupported(t, newTypePath)
-                      || isBean(t, newTypePath);
```
stevenschlansker, line 636: This `isBean` is unreachable because `isSupported` already checks `isBean`

chaokunyang (approved):
Looks awesome! 

Could we add some detailed document in https://github.com/apache/fory/blob/main/docs/guide/row_format_guide.md, so that other users can use this excelent feature.

I even think our rencent improvements to fury row format deserve a blog

Issue #2307, opened 2025-06-07
[Xlang] ClassCastException during deserialization due to MetaStringBytes cache collision when class name encodings differ or produce identical keys


LouisLou2:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fory/issues) and found no similar issues.

### Version

latest commit (8d028b47d900a31d86c7b67c37f952ac36b30d6a)

### Component(s)

Java

### Minimal reproduce step

*   The following Java code demonstrates the issue:

```java
package org.apache.fory;

import org.apache.fory.config.Language;

public class Main {
  private static class SomeClass1 {
    int number;
    // Adding a field to make it slightly different in structure, though name is the primary issue
    String name = "SomeClass1";
  }

  private static class SomeClass2 {
    int number;
    String name = "SomeClass2";
  }

  public static void main(String[] args) {
    SomeClass1 object1 = new SomeClass1();
    object1.number = 10;
    SomeClass2 object2 = new SomeClass2();
    object2.number = 20;

    Fory fory = Fory.builder().withLanguage(Language.XLANG)
      .build();

    // Register two classes with names that only differ by case.
    // XLANG might handle these in a way that leads to a hash collision
    // for the MetaStringBytes cache key if encoding is not considered.
    fory.register(SomeClass1.class, "aclass");
    fory.register(SomeClass2.class, "Aclass"); // Note the capital 'A'

    System.out.println("Serializing object1 (SomeClass1 as \"aclass\")");
    byte[] bytes1 = fory.serialize(object1);
    System.out.println("Deserializing object1...");
    SomeClass1 deserializedObject1 = (SomeClass1) fory.deserialize(bytes1);
    System.out.println("Deserialized object1.number: " + deserializedObject1.number);

    System.out.println("Serializing object2 (SomeClass2 as \"Aclass\")");
    byte[] bytes2 = fory.serialize(object2);
    System.out.println("Deserializing object2...");
    // This is where the ClassCastException occurs
    SomeClass2 deserializedObject2 = (SomeClass2) fory.deserialize(bytes2);
    System.out.println("Deserialized object2.number: " + deserializedObject2.number);
  }
}
```

### What did you expect to see?

I expected the program to run without exceptions.
    `object1` should be successfully serialized and deserialized back into a `SomeClass1` instance.
    `object2` should be successfully serialized and deserialized back into a `SomeClass2` instance.
    The output should show the correct numbers for both deserialized objects.

### What did you see instead?

[...]

PR #2308, opened 2025-06-07
fix(java): Use (long, long, byte) key for MetaStringBytes cache to prevent collisions

LouisLou2:
## What does this PR do?

This PR fixes a `ClassCastException` during deserialization caused by an incorrect `MetaStringBytes` being retrieved from the cache. The previous cache key `(long, long)` in `MetaStringResolver` (for small meta strings) didn't include the string's `encoding` byte, leading to collisions when different strings (e.g., "aclass" vs "Aclass") hashed to the same two longs.

**Changes:**

1.  **Introduced `LongLongByteMap`:** A new map using a `(long k1, long k2, byte k3)` key.
2.  **Updated `MetaStringResolver`:** Modified relevant methods (e.g., `createSmallMetaStringBytes`, `readSmallMetaStringBytes`) to use `LongLongByteMap`, incorporating the `encoding` byte into the cache key for small meta strings.
3.  **Renamed Test:** `LongLongMapTest.java` renamed to `LongLongByteMapTest.java` and updated to test the new map.

## Related issues

- #2307 


chaokunyang (approved):
Looks great!

Issue #2286, opened 2025-06-03
[Java] Support codegen for xlang serialization in java for strict mode


chaokunyang:
### Feature Request

Currently all xlang serialization don't support codegen:
```java

  @Data
  public static class Foo {
    public Integer f1;
    public String f2;
    public List<String> f3;
    public Map<String, Integer> f4;
    public Bar f5;
  }

  @Data
  public static class Bar {
    public Integer f1;
    public String f2;
  }


  @Test
  public void testMetaSharedBasic() {
    Fory fory =
        Fory.builder()
            .withLanguage(Language.XLANG)
            .withCodegen(false)
            .withCompatibleMode(CompatibleMode.SCHEMA_CONSISTENT)
            .build();
    fory.register(Foo.class, "example.foo");
    fory.register(Bar.class, "example.bar");
    serDeCheck(fory, Bar.create());
    serDeCheck(fory, Foo.create());
  }
```

we should support it for better performance:
```java
  @Test
  public void testMetaSharedBasic() {
    Fory fory =
        Fory.builder()
            .withLanguage(Language.XLANG)
            .withCodegen(true)
            .withCompatibleMode(CompatibleMode.SCHEMA_CONSISTENT)
            .build();
    fory.register(Foo.class, "example.foo");
    fory.register(Bar.class, "example.bar");
    serDeCheck(fory, Bar.create());
    serDeCheck(fory, Foo.create());
  }
```

### Describe the solution you'd like

We need to change:
- BaseObjectCodecBuilder.java
- ObjectCodecBuilder.java

Hen1ng:
@chaokunyang I am very interested in this issue, could you please assign it to me？

chaokunyang:
@Hen1ng Assigned to you! Looking forward to your contribution

PR #2312, opened 2025-06-08
feat(java): Support codegen for xlang serialization in java for strict mode

Hen1ng:
## Related issues
#2286 


java/fory-core/src/main/java/org/apache/fory/builder/BaseObjectCodecBuilder.java
```diff
@@ -240,6 +240,9 @@ public String genCode() {
     ctx.addField(ctx.type(Fory.class), FORY_NAME);
     Expression encodeExpr = buildEncodeExpression();
     Expression decodeExpr = buildDecodeExpression();
+
+    Expression xlangEncodeExpr = buildXlangEncodeExpression();
```
chaokunyang: we should do this only fury is configured to use xlang mode, and when using xlang mode, do not generate code for java native serialization, i.e. don't generate code for `read/write`

chaokunyang:
Could we add some unit tests for xlang jit mode?

java/fory-core/src/main/java/org/apache/fory/builder/CodecBuilder.java
```diff
@@ -123,6 +123,8 @@ public CodecBuilder(CodegenContext ctx, TypeRef<?> beanType) {
   /** Returns an expression that serialize java bean of type {@link CodecBuilder#beanClass}. */
   public abstract Expression buildEncodeExpression();
 
+  public abstract Expression buildXlangEncodeExpression();
```
chaokunyang: Could we keep buildXlangEncodeExpression and buildXlangDecodeExpression in BaseObjectCodecBuilder.java?

java/fory-core/src/test/java/org/apache/fory/CrossLanguageTest.java
```diff
@@ -83,7 +84,7 @@ public class CrossLanguageTest extends ForyTestBase {
 
   @BeforeClass
   public void isPyforyInstalled() {
-    TestUtils.verifyPyforyInstalled();
+    //    TestUtils.verifyPyforyInstalled();
```
chaokunyang: Seems this is unrelated to this PR, could we revert this change?

java/fory-core/src/test/java/org/apache/fory/CrossLanguageTest.java
```diff
@@ -798,4 +799,55 @@ public void testEnumField() throws java.io.IOException {
     Assert.assertEquals(xserDe(fory, a), a);
     structRoundBack(fory, a, "test_enum_field");
   }
+
+  @Test
+  public void testCodeGen() {
```
chaokunyang: Could we split it into two tests, one for SCHEMA_CONSISTENT, and another for COMPATIBLE.

And perhaps we could create a new test file `CodegenXlangTest.java` under `org.apache.fory.xlang` package, and move this test into that file. `CrossLanguageTest` is a little big currently

java/fory-core/src/main/java/org/apache/fory/resolver/XtypeResolver.java
```diff
@@ -65,15 +68,7 @@
...
-import org.apache.fory.serializer.ArraySerializers;
-import org.apache.fory.serializer.EnumSerializer;
-import org.apache.fory.serializer.LazySerializer;
-import org.apache.fory.serializer.NonexistentClass;
-import org.apache.fory.serializer.NonexistentClassSerializers;
-import org.apache.fory.serializer.ObjectSerializer;
-import org.apache.fory.serializer.SerializationUtils;
-import org.apache.fory.serializer.Serializer;
-import org.apache.fory.serializer.Serializers;
+import org.apache.fory.serializer.*;
```
chaokunyang: Fory doesn't use * imports, please expand all imports

java/fory-core/src/main/java/org/apache/fory/resolver/XtypeResolver.java
```diff
@@ -232,16 +228,126 @@ private void register(
...
+   * already related with a class, this method should try to reuse that class info, otherwise jit
+   * callback to update serializer won't take effect in some cases since it can't change that
+   * classinfo.
+   */
+  public <T> void setSerializer(
+      Class<T> cls, Serializer<T> serializer, String namespace, String typeName, int xtypeId) {
+    addSerializer(cls, serializer, namespace, typeName, xtypeId);
+  }
+
+  /** Ass serializer for specified class. */
```
chaokunyang: what does `ass` means here?

java/fory-core/src/main/java/org/apache/fory/resolver/XtypeResolver.java
```diff
@@ -232,16 +228,126 @@ private void register(
...
+      }
+    } else {
+      if (codegen) {
+        LOG.info("Object of type {} can't be serialized by jit", cls);
+      }
+      switch (fory.getCompatibleMode()) {
+        case SCHEMA_CONSISTENT:
+          return ObjectSerializer.class;
+        case COMPATIBLE:
+          return shareMeta ? ObjectSerializer.class : CompatibleSerializer.class;
```
chaokunyang: CompatibleSerializer.class is not used for xlang serialization, we used meta shade mode for xlang forward/backward compatibility, CompatibleSerializer use kv format instead

java/fory-core/src/main/java/org/apache/fory/resolver/XtypeResolver.java
```diff
@@ -232,16 +228,126 @@ private void register(
...
+      } else {
+        try {
+          classCtx.add(cls);
+          Class<? extends Serializer> sc;
+          switch (fory.getCompatibleMode()) {
+            case SCHEMA_CONSISTENT:
+              sc =
+                  fory.getJITContext()
+                      .registerSerializerJITCallback(
+                          () -> ObjectSerializer.class,
```
chaokunyang, line 279: I prefer creating serializer lazily, since the hash compute for type meta will read names of all registered classes. And when class registration is invoked, if the serializer are created egaerly, some classes don't have a change to let the registered name/tag being used for hash computing

java/fory-core/src/main/java/org/apache/fory/resolver/XtypeResolver.java
Hen1ng: shareMeta ?  MetaSharedSerializer.class : ObjectSerializer.class
is this correctly？

Issue #2277, opened 2025-06-02
[Java] support multi-dimensional array field serialization in xlang meta shared mode


chaokunyang:
### Feature Request

Currently we don't support multi-dimensional array field serialization in xlang meta shared mode

If we serialize such fields, we will get exception:
```java
  @Test
  public void testMDArrayField() {
    Fory fory =
        Fory.builder()
            .withLanguage(Language.XLANG)
            .withCodegen(false)
            .withCompatibleMode(CompatibleMode.COMPATIBLE)
            .build();
    // TODO support multi-dimensional array serialization
    fory.register(MDArrayFieldStruct.class, "example.a");
    MDArrayFieldStruct s = new MDArrayFieldStruct();
    s.arr = new int[][] {{1, 2}, {3, 4}};
    serDeCheck(fory, s);
  }
```
```java
java.lang.NullPointerException
	at org.apache.fory.serializer.ObjectSerializer.writeOtherFields(ObjectSerializer.java:141)
	at org.apache.fory.serializer.ObjectSerializer.write(ObjectSerializer.java:134)
	at org.apache.fory.serializer.ObjectSerializer.xwrite(ObjectSerializer.java:152)
	at org.apache.fory.serializer.LazySerializer.xwrite(LazySerializer.java:65)
	at org.apache.fory.Fory.xwriteData(Fory.java:613)
	at org.apache.fory.Fory.xwriteRef(Fory.java:548)
	at org.apache.fory.Fory.xwrite(Fory.java:421)
	at org.apache.fory.Fory.serialize(Fory.java:329)
	at org.apache.fory.Fory.serialize(Fory.java:281)
	at org.apache.fory.ForyTestBase.serDe(ForyTestBase.java:320)
	at org.apache.fory.ForyTestBase.serDeCheck(ForyTestBase.java:313)
```

OmCheeLin:
Please assign it to me, I would like to work on this.

PR #2314, opened 2025-06-08
feat(java): support multi-dimensional array field serialization in xlang meta shared mode

OmCheeLin:
## What does this PR do?
support multi-dimensional array field serialization in xlang meta shared mode

## Related issues

#2277


java/fory-core/src/main/java/org/apache/fory/resolver/XtypeResolver.java
```diff
@@ -466,6 +466,10 @@ private ClassInfo buildClassInfo(Class<?> cls) {
     return info;
   }
 
+  private boolean handleArray(Class<?> cls) {
+    return !TypeUtils.getArrayComponent(cls).isPrimitive() || !cls.getComponentType().isPrimitive();
```
chaokunyang: Seems this can be simplified into:

```suggestion
    return !cls.getComponentType().isPrimitive();
```

chaokunyang: And I prefer inline it into the caller, instead of creating a new method

chaokunyang (approved):
LGTM

PR #2320, opened 2025-06-10
fix(java): fix row encoder synthesized interface lazy-decoding of Optional

stevenschlansker:
## What does this PR do?

fix incorrect handling of Optional types nested inside of synthesized bean interfaces

If we build another 0.11.0 rc, it would be nice to include this fix if possible.


chaokunyang:
We can plan a new release next week

PR #2321, opened 2025-06-10
fix(java): fix row format handling Optional of type with custom codec

stevenschlansker:
## What does this PR do?

Fix the logic for type with custom codec wrapped in Optional, currently causes compile error

Please review #2320 first; this PR depends on that fix

And please consider for a 0.11.0-rc2 if that happens, otherwise 0.11.1


chaokunyang (approved):
LGTM

PR #2333, opened 2025-06-12
feat(java): array row encoder supports set of interface type

stevenschlansker:
## What does this PR do?

improve array row encoder to support custom collections that are not List, for example TreeSet
also supports interface types as value


stevenschlansker:
I do not understand the Python build failure - it does not seem related?

chaokunyang:
I try to fix it in #2334

PR #2335, opened 2025-06-13
fix(python): correct compute_string_hash usage in StructHashVisitor.visit_customized

### LINKED ISSUE #2332: QuickStart python demo run failed
### Search before asking

### Version

main

### Component(s)

Python

### Minimal reproduce step

After following the steps in the BUILD file to compile pyfory, I ran the Python demo from the Quickstart guide and encountered the error shown in the attached image. I noticed that the TypeInfo defined in both /fory/python/pyfory/_serialization.pyx and /fory/python/pyfory/_registry.py does not contain the namespace or typename fields. I would like to know whether this is a bug that needs to be fixed or if this behavior is expected.

Image

Here are the steps I took:
```bash
cd python
pip uninstall -y numpy
pip install pyarrow==15.0.0 Cython wheel pytest
pip install -v -e .
python3 test.py
```

test.py：
```python
from typing import Dict
import pyfory

class SomeClass:
    f1: "SomeClass"
    f2: Dict[str, str]
    f3: Dict[str, str]

fory = pyfory.Fory(ref_tracking=True)
fory.register_type(SomeClass, typename="example.SomeClass")
obj = SomeClass()
obj.f2 = {"k1": "v1", "k2": "v2"}
obj.f1, obj.f3 = obj, obj.f2
data = fory.serialize(obj)
# bytes can be data serialized by other languages.
print(fory.deserialize(data))
```

check whether it is a bug or bugfix

### What did you see instead?

run test.py successly

### Anything Else?

_No response_

lzaeh:
## What does this PR do?

Fixed the incorrect arguments passed to compute_string_hash in StructHashVisitor.visit_customized when calculating the hash value.
- Aligned this change with the Java implementation after consulting the project founder.
- Exposed Cython’s TypeResolver in Python so its internal _resolver could provide the namespace and typename decoders.
- Marked the _resolver field as readonly to allow external read access without permitting modification.
- Added the issue author’s example to the test suite for validation.
- Noted in the test file comments that objects with self-references cannot be compared using ==; instead, comparison of serialized data is recommended.

## Related issues
close #2332

### DIFF
--- a/python/pyfory/_serialization.pyx
+++ b/python/pyfory/_serialization.pyx
@@ -85,9 +85,9 @@ cdef int8_t NOT_NULL_VALUE_FLAG = -1
 # this flag indicates that the object is a referencable and first read.
 cdef int8_t REF_VALUE_FLAG = 0
 # Global MetaString decoder for namespace bytes to str
-namespace_decoder = MetaStringDecoder(".", "_")
+namespace_decoder = MetaStringDecoder(".","_")
 # Global MetaString decoder for typename bytes to str
-typename_decoder  = MetaStringDecoder("$", "_")
+typename_decoder  = MetaStringDecoder("$","_")
 
 
 @cython.final
@@ -424,7 +424,7 @@ cdef class TypeInfo:
         if self.namespace_bytes is None:
             return ""
         return self.namespace_bytes.decode(namespace_decoder)
-    
+
     cpdef str decode_typename(self):
         if self.typename_bytes is None:
             return ""
--- a/python/pyfory/_serialization.pyx
+++ b/python/pyfory/_serialization.pyx
@@ -85,9 +85,9 @@ cdef int8_t NOT_NULL_VALUE_FLAG = -1
 # this flag indicates that the object is a referencable and first read.
 cdef int8_t REF_VALUE_FLAG = 0
 # Global MetaString decoder for namespace bytes to str
-namespace_decoder = MetaStringDecoder(".","_")
+namespace_decoder = MetaStringDecoder(".", "_")
 # Global MetaString decoder for typename bytes to str
-typename_decoder  = MetaStringDecoder("$","_")
+typename_decoder = MetaStringDecoder("$", "_")
 
 
 @cython.final
@@ -436,7 +436,7 @@ cdef class TypeResolver:
     cdef:
         readonly Fory fory
         readonly MetaStringResolver metastring_resolver
-        readonly object _resolver
+        object _resolver
         vector[PyObject *] _c_registered_id_to_type_info
         # cls -> TypeInfo
         flat_hash_map[uint64_t, PyObject *] _c_types_info
--- a/python/pyfory/_serialization.pyx
+++ b/python/pyfory/_serialization.pyx
@@ -551,6 +551,7 @@ cdef class TypeResolver:
     cpdef inline reset_write(self):
         pass
 
+
 @cython.final
 cdef class Fory:
     cdef readonly object language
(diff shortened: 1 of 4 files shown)


python/pyfory/_struct.py
```diff
@@ -259,7 +259,13 @@ def visit_customized(self, field_name, type_, types_path=None):
         if typeinfo is not None:
             hash_value = typeinfo.type_id
             if TypeId.is_namespaced_type(typeinfo.type_id):
-                hash_value = compute_string_hash(typeinfo.namespace + typeinfo.typename)
+                namespace_str = typeinfo.namespace_bytes.decode(
+                    self.fory.type_resolver._resolver.namespace_decoder
```
chaokunyang: how about making `namespace_decoder` and `typename_decoder` as a module-level variable, and 
create `decode_namespace/decode_typename` method in `TypeInfo`? We could import those decoders in the created functions

lzaeh: Sounds great—I will go ahead and do that.

lzaeh: > how about making `namespace_decoder` and `typename_decoder` as a module-level variable, and create `decode_namespace/decode_typename` method in `TypeInfo`? We could import those decoders in the created functions



> how about making `namespace_decoder` and `typename_decoder` as a module-level variable, and create `decode_namespace/decode_typename` method in `TypeInfo`? We could import those decoders in the created functions

Sir, I’ve implemented the two decoders as module-level variables according to my understanding—please feel free to point out anything that may be inappropriate. I’m just concerned I might have misunderstood, as while rewriting the code I kept wondering if I should have moved MetaStringDecoder from the Python implementation into the Cython layer.

chaokunyang: We don't need to move MetaStringDecoder into cython, it's not in critical path, keeping it in python is more easy to maintain

python/pyfory/_serialization.pyx
```diff
@@ -415,13 +420,23 @@ cdef class TypeInfo:
...
+        return self.typename_bytes.decode(typename_decoder)
+
 
 @cython.final
 cdef class TypeResolver:
     cdef:
         readonly Fory fory
         readonly MetaStringResolver metastring_resolver
-        object _resolver
+        readonly object _resolver
```
chaokunyang: Could we remove this?

chaokunyang:
Please format code by `black python`

python/pyfory/_serialization.pyx
lzaeh: > Could we remove this?
of course

python/pyfory/tests/test_cross_language.py
```diff
@@ -466,6 +466,35 @@ def test_serialize_simple_struct(data_file_path):
...
+    assert data1 == data2
+    # bytes can be data serialized by other languages.
+    # due to the self-referential nature of this object,
+    # direct `==` comparison will fail.
+    # 1. Serialize `obj1` to `data1`
+    # 2. Deserialize `data1` to `obj2`
+    # 3. Serialize `obj2` to `data2`
+    # If `data1 == data2`, the round-trip preserves value equivalence.
+    # print(data1)
+    # print(data2)
```
pandalee99, line 493: Maybe it's not very reasonable to use `print`

lzaeh: > Maybe it's not very reasonable to use `print`

ok sir, im gonna remove those.

chaokunyang (approved):
LGTM

PR #2336, opened 2025-06-13
perf(java): Improve zone offset deserialization performance by overriding JDK caching

Cybermaxke:
## What does this PR do?

The JDK `ZoneOffset.ofTotalSeconds(...)` method uses a concurrent hash map for caching zone offsets with values for each 15 minutes (UTC, +01:00, +01:15, etc.), this causes a noticeable overhead when deserializing a lot of zone offsets. This PR overrides that caching for the single byte representation using an array, which is in line with the 15 minute JDK caching.


chaokunyang (approved):
Looks great

PR #2338, opened 2025-06-13
feat(java): store OptionalInt, OptionalLong, OptionalDouble same as nullable values

stevenschlansker:
## What does this PR do?

currently, optional wrappers are stored as a bean with two fields - present and value, taking 32 bytes
with this change, stored as nullable primitive, taking 8 bytes plus 1 bit for null

this is a breaking change to the row format, but since support is relatively recent, hopefully that is ok?

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/builder/CodecBuilder.java
+++ b/java/fory-core/src/main/java/org/apache/fory/builder/CodecBuilder.java
@@ -127,6 +127,10 @@ public abstract class CodecBuilder {
     return ctx.sourcePublicAccessible(cls);
   }
 
+  protected boolean fieldNullable(Descriptor descriptor) {
+    return false;
+  }
+
   protected Expression tryInlineCast(Expression expression, TypeRef<?> targetType) {
     return tryCastIfPublic(expression, targetType, true);
   }
@@ -162,9 +166,6 @@ public abstract class CodecBuilder {
     return expression;
   }
 
-  // left null check in sub class encode method to reduce data dependence.
-  private final boolean fieldNullable = false;
-
   protected Reference getRecordCtrHandle() {
     String fieldName = "_record_ctr_";
     Reference fieldRef = fieldMap.get(fieldName);
@@ -192,8 +193,9 @@ public abstract class CodecBuilder {
     TypeRef<?> fieldType = descriptor.getTypeRef();
     Class<?> rawType = descriptor.getRawType();
     String fieldName = descriptor.getName();
+    boolean fieldNullable = fieldNullable(descriptor);
     if (isInterface) {
-      return new Invoke(inputBeanExpr, descriptor.getName(), fieldName, fieldType);
+      return new Invoke(inputBeanExpr, descriptor.getName(), fieldName, fieldType, fieldNullable);
     }
     if (isRecord) {
       return getRecordFieldValue(inputBeanExpr, descriptor);
@@ -243,6 +245,7 @@ public abstract class CodecBuilder {
       fieldType = OBJECT_TYPE;
     }
     String fieldName = descriptor.getName();
+    boolean fieldNullable = fieldNullable(descriptor);
     if (Modifier.isPublic(beanClass.getModifiers())) {
       Preconditions.checkNotNull(descriptor.getReadMethod());
       return new Invoke(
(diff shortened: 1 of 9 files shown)


java/fory-core/src/main/java/org/apache/fory/builder/CodecBuilder.java
```diff
@@ -162,9 +162,6 @@ protected Expression tryCastIfPublic(
     return expression;
   }
 
-  // left null check in sub class encode method to reduce data dependence.
-  private final boolean fieldNullable = false;
```
stevenschlansker, line 166: I don't understand how this field works. It seems to never be assigned, so this codec builder always thinks all fields are not-null, which seems wrong

stevenschlansker:
I am not sure how to resolve the ObjectCodec thinking all fields are not nullable.
With the code as it is, the ObjectCodecBuilder always thinks reading fields will be a not-null result, which is not the case. However, when I apply what I think is a good fix - take the field nullability from Descriptor rather than always assume false - it seems to break the object codec generation method size limit, like

```
java.lang.AssertionError: Method writeFields2$ for class Struct4ForyRefCodec_2 has size 344 > 325 expected [true] but found [false]
	at org.testng.Assert.fail(Assert.java:110)
	at org.apache.fory.builder.ObjectCodecBuilderTest.lambda$4(ObjectCodecBuilderTest.java:145)

```

I would appreciate advice on how to correctly solve the problem with `CodecBuilder.getFieldValue` returning an expression always marked as not-null, even in cases like:

```
class Struct { OptionalLong f1; } 
struct.f1 // may be null
```

java/fory-core/src/main/java/org/apache/fory/builder/CodecBuilder.java
chaokunyang: ObjectCodecBuilder check the nullabiity separately for efficiency. We can make this as a method:
```java
protected boolean fieldNullable() {
   return true;
}
```

and overwrite iit n BaseObjectBuilder :
```java
@Overwrite
protected boolean fieldNullable() {
   return false;
}
```

java/fory-format/src/main/java/org/apache/fory/format/encoder/BaseBinaryEncoderBuilder.java
```diff
@@ -164,15 +167,21 @@ protected Expression serializeFor(
...
-          new Invoke(inputObject, "orElse", TypeUtils.OBJECT_TYPE, new Expression.Null(elemType));
+          new Invoke(
+              inputObject,
+              "orElse",
+              "",
+              TypeUtils.OBJECT_TYPE,
+              true,
+              new Expression.Null(elemType));
       Expression unwrapped =
-          new If(ExpressionUtils.eqNull(inputObject), new Expression.Null(elemType), orNull);
```
stevenschlansker, line 175: It seems there is usually a local variable holding whether the input value is null, but `ExpressionUtils.eqNull` does not use it - `Expression.IsNull` does. But then I found it was somewhat broken, so there's a few fixes included to make this work better.

This reduces dead code computing if a field is null that is not used, then we re-compute whether it is null again later for the `if()` statement

stevenschlansker:
Ok I think I got this working now. Includes some fixups to `isNull` handling to reduce the size of generated code.

java/fory-core/src/main/java/org/apache/fory/codegen/Expression.java
```diff
@@ -2287,7 +2288,8 @@ public ExprCode doGenCode(CodegenContext ctx) {
       String i = ctx.newName("i");
       String elemValue = ctx.newName("elemValue");
       Expression elementExpr =
-          action.apply(new Reference(i), new Reference(elemValue, elementType, false));
+          action.apply(
+              new Reference(i), new Reference(elemValue, elementType, !elementType.isPrimitive()));
```
chaokunyang: Could we add an `elementNullable` field and pass it when creating `ForEach` esxpression? And condition that `elementNullable` here?

For `java/fory-core/src/main/java/org/apache/fory/builder/BaseObjectCodecBuilder#writeContainerElements`, we don't need `nullable` variable since we check and computed nullability ahead in `writeElementsHeader`

java/fory-format/src/main/java/org/apache/fory/format/encoder/BaseBinaryEncoderBuilder.java
chaokunyang: Introduce a local variable made teh code bloat a little, so if this variable is used only once, fory just use `ExpressionUtils.eqNull` directly. Such optimization is used for very performance-sensitive cases. We can keep use `Expression.IsNull` to simplify the code

stevenschlansker: At least in these cases, the isNull variable is computed either way (although maybe later removed as dead code by jit) so might as well use it since we spend the method space anyway

java/fory-core/src/main/java/org/apache/fory/codegen/Expression.java
stevenschlansker: Updated

chaokunyang (approved):
Looks great!

PR #2230, opened 2025-05-14
feat(go): fury-go implements adaptation and optimization for new xlang

pandalee99:
## What does this PR do?
1、Cross-language interoperability of furygo has been achieved (except for dynamic Pointers).
2、The new collection serialization has been implemented, and dynamic writing will be performed based on the object header
3、The chunk based map serialization has been implemented
4、The int type is written to a dynamic varint
5、The string becomes a dynamic string and will be written according to the encoding
It is hoped that the dynamic pointer problem can be solved in the next pr.

## Related issues
Close #2200
Close #2201
Close #2202
Close #2014

## Related PR

Relate #2198 

## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

Issue #2202, opened 2025-05-05
[Go] chunk based map serialization implementation for fury go


chaokunyang:
### Feature Request

chunk based map serialization spec  https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#map  are very efficient for static typed languages. 

Without this spec, fury go needs to write meta for every key and value, which is very efficient. But if we use chunk based map serialization, fury can write meta for key and value only once if those meta are same

Issue #2201, opened 2025-05-05
[Go] new collection serialization for fury go


chaokunyang:
### Feature Request

New collection serialization spec https://fury.apache.org/docs/specification/fury_xlang_serialization_spec#list is much more efficient for static styped languages. Before this spec, fury go will write many meta for every element in the collection. With new spec, those meta will be written only once

PR #2339, opened 2025-06-15
perf(Go/python): Type system compatibility upgrade

pandalee99:
## What does this PR do?

In this PR, as the mentor, I completed the type system upgrade for furygo together with @lzaeh. The key tasks include:

1、Setting up the basic Go framework and defining upgrade objectives @pandalee99 
2、Implementing basic named structs and identifying issues to resolve @pandalee99 
3、Carrying out cross-language transformation for the MAP serializer @lzaeh 
4、Conducting cross-language transformation for the Slice serializer @lzaeh
5、Detecting and fixing errors in structure hash calculation @lzaeh
6、Solving cross-language serialization issues for structures @lzaeh
7、Resolving type system registration problems @lzaeh

## Related issues
#2014 #2200 #2201 #2202 #2268

## Relate PR

#2198 #2230 

@lzaeh, could you please elaborate on the changes made in this PR and share your insights on the type system upgrade for furygo? It would be great if you can walk us through the key challenges you encountered, the solutions implemented, and the impact of these changes on the project's functionality and cross-language compatibility. Looking forward to your detailed explanation!


lzaeh:
## Major Modifications & Explanations

Although this PR passes all current tests, the implementation approach is still debatable and some issues remain. Specifically, while it passes the existing test cases, the current test coverage is insufficient. For instance, in the case of pure array cross-language serialization, deserialization on the Go side still fails. Further investigation and testing will be conducted to explore and resolve this issue.

This PR mainly addresses the following:

- Deprecated the `allocate ID` mechanism  
- Aligned the serialization and deserialization of named structs  
- Fixed map serialization/deserialization issues  
- Partially resolved issues related to array and slice serialization

------

### 1. Behavior of Named Struct Registration

Currently, when manually registering a named struct, the registration function **does not** immediately perform actual type registration. Instead, registration is deferred until the first time the struct is serialized. At first, I mistakenly believed this to be an intentional lazy registration mechanism.

However, consider this scenario: both Python and Go sides register the same named struct, and a serialized struct instance is sent from Python to Go. If the Go side only performs deserialization (without prior serialization), its type system hasn't completed the registration, leading to deserialization failure.

Therefore, just like in Python, performing a **complete type registration at registration time** is essential. This PR adjusts the Go implementation accordingly to ensure consistency with other languages.

------

### 2. Handling of Allocate ID

After communicating with Pandalee, we confirmed that the `allocate ID` mechanism is problematic in cross-language scenarios.

Many local test cases—especially those involving arrays or struct pointers—rely on the `allocate ID` mechanism to pass. But once tested across languages, serious compatibility issues arise, which have now been resolved during the type registration process.

During local named struct tests, I discovered that for the same struct, the pointer type's typeID is dynamically allocated via `allocate ID`. This allows nested structs (either value or pointer) to be recognized correctly locally.

At registration time, we can distinguish between value and pointer types based on the sign of the typeID.

If we don’t differentiate further, we risk overwriting type information: e.g., both `A` and `*A` (value and pointer types) may share the same `namespaceBytes` and `typenameBytes`, and whichever is registered last will overwrite the other. This results in deserialization errors (e.g., returning a pointer when a value is expected).

To fix this, the `readTypeInfo` phase reconstructs the correct `typeInfo` based on the sign of the typeID, ensuring accurate deserialization logic.

A related issue is discussed in Section 6: "Array vs Slice".

------

### 3. Unified Reference Tracking Logic

To align with the `need_to_write_ref` behavior, I added an explicit `need_to_write_ref` method to every serializer.

While this approach may not be elegant, it ensures correct reference handling for struct types. Future refinements may optimize this triggering mechanism.

Currently, reference tracking for struct types is supported, but its implementation still needs to be strengthened, corrected, and is open to further discussion.

------

### 4. Problems in Map Serialization Logic

Following Pandalee’s suggestion, I temporarily unified the implementation using `mapSerializer` to improve overall functionality.

The original `write` and `read` logic of `mapSerializer` was inconsistent with other languages. I have aligned it based on Python’s implementation.

------

### 5. Maps Inside Named Structs vs. Maps Everywhere Else

Based on Python’s serialization logic, we observe the following:

#### For Map fields not inside a struct:

Python writes the full type info for each key/value, ensuring the deserialization side (other languages) can reconstruct them correctly.

For example, when writing a chunk, Python writes typeID info for the leading key-value pair. Since the map type is resolved via typeID, the initial `mapSerializer` starts with `none` for both key and value serializers. After each chunk, the kv serializers are reset to `none`, which simplifies deserialization.

#### For Map fields within a named struct:

Python creates a concrete type like `map[int]string`. As key/value types are known, **typeID info is not written**.

In this case, the deserialization side reconstructs the correct key/value types using the `mapSerializer` that was registered with the named struct. As long as the target language also registers the same struct, deserialization works.

#### From Java’s perspective:

If a map is **not** inside a struct, and no type-specific `mapSerializer` exists, how can the deserializer determine kv types?

Even if the typeID gives hints, the container `map` still needs to be instantiated. A common fallback is to use `Map<Object, Object>` (i.e., `map[interface{}]interface{}` in Go), letting the user cast types afterward.

------

### Strategy in Go:

1. For **maps inside named structs**, directly instantiate concrete typed maps.  
2. If the user specifies a concrete map type during deserialization, use that type.
3. For all **other maps**, fallback to `map[interface{}]interface{}` during deserialization.  
4. Introduced a temporary flag `mapInStruct` in `mapSerializer` to indicate whether the current serializer originates from a struct field.  
5. While `mapInStruct` is conceptually sound, the implementation still has room for refinement due to my limited coding experience.

------

### 6. Array vs Slice

1. In Fory-Go’s original type system, the treatment of arrays and slices was rather odd: slices seemed to “occupy” the type-space that should have belonged to arrays.

2. In Go itself arrays are very strict—an array is an array, a slice is a slice, and two arrays whose lengths differ are considered entirely different types.

3. For slices I decided to keep the previous behavior, but with one important adjustment: across languages a slice, even when it has a concrete element type, must never reuse the type-id that was historically assigned to its corresponding array type. The only correct choice is to give every slice the generic type-id LIST so that hash-calculation for structs and the overall cross-language semantics stay perfectly aligned.

4. For arrays I use the element-specific slice type-id during serialization, but I convert the value on the fly so I can reuse the existing slice serializer. During deserialization I patched several slice serializers so that, when the incoming parameters clearly indicate whether we are rebuilding an array or a slice (for example when the user explicitly requests an array container or when a struct field is an array), they now handle both cases correctly.

5. I added a batch of array-only tests written purely in Go, with no conversion helpers, and they all pass—proving that the updated slice serializers now cover both arrays and slices in one coherent implementation.

------

### 7. Support for Multi-Dimensional Slices

- During serialization: handled recursively as `[]interface{}`  
- During deserialization: restored as `[]interface{}` and requires user-side conversion to target type

------

### 8. Other Notes

- Where needed, I aligned Go’s serialization and deserialization logic with Python’s to ensure cross-language compatibility.
- I’ve corrected the field ordering in named structs, but haven’t yet addressed the time‐enum cases due to limited familiarity with the type system. Moreover, our current alignment approach—designed to avoid disrupting the existing code structure—may be suboptimal (it might be doing too much sorting). We’ll look into optimizing it further going forward.
- Modifications to Some Tests:
   - In cross-language complex struct tests, when aligning struct field types, F11 should be an array.
   - In local testing, I temporarily commented out the []int8 type because the existing serializer resources appear to be incorrect
   - In cross-language struct tests, I chose to test structs by value. If I were to test with pointer types, then when Python serializes to Go—Python only transmits values—Go would deserialize the bytes into a value type, unaware that the test’s starting point was a pointer, and the resulting type mismatch would cause the test to fail. Unifying tests around value types therefore makes the most sense. Bidirectional testing of pointer types must be done with a language system that natively supports pointers.
   - Temporarily added a type conversion function, as discussed above regarding map and array handling.
- Cases that actually require conversion functions:
   - []interface{} 
     - When a []interface{} contains an array, both cross-language and Go-specific semantics can, in principle, rely on the type-id to distinguish whether the value is an array or a slice.
     - In this version, however, I’ve chosen to favor slices. Specifically, I want slices to be handled by concrete, element-specific serializers, rather than using the generic LIST type-id. This improves efficiency and type fidelity for slices.
     - Because of that choice, once we encounter a type-id corresponding to an “array collection,” we can no longer tell whether the original value was a Go array or a slice. The type-id alone is not enough to disambiguate them under Go semantics.
     - Similarly, if a []interface{} contains a map, we cannot infer the map’s concrete type from the serialized data. This also requires conversion logic during deserialization.
     - As a result, arrays appearing inside a []interface{}—or in cases where the user has not specified a concrete deserialization target—will be deserialized into their corresponding slice types. These cases currently require post-processing or conversion to restore the intended array structure.
   - Multi-dimensional slices
 - Previous discussions about array types were mostly focused on Go-specific semantics, while cross-language behavior remained incomplete—mainly because the test coverage for certain types was lacking. In the context of arrays vs. slices, I mentioned that slices should use typeid = 21 during serialization. However, at this point, I’ve only applied that rule within struct serialization. From a semantic standpoint, this is clearly incomplete and potentially problematic.

It's a bit frustrating—although I’ve made some progress, various issues have surfaced along the way. That said, I’ll keep track of them and address each one step by step going forward.

pandalee99:
cc @chaokunyang

chaokunyang:
Seems there are many lint related changes not related to this PR. Could you create a new PR for those changes? We can merge that PR first to make this PR more easy for review

pandalee99:
> Seems there are many lint related changes not related to this PR. Could you create a new PR for those changes? We can merge that PR first to make this PR more easy for review

@chaokunyang 
I've removed irrelevant code style changes.
Now I think it's time for a review. These are highly relevant feature changes.
This pr was really complicated and I set up a lot of tests to verify the functionality was correct.

chaokunyang (approved):
LGTM

PR #2353, opened 2025-06-18
chore(java): ProtobufDispatcher should trigger log warning at most once

ClaudioConsolmagno:
## What does this PR do?

This PR addresses these 2 issues:
- It's possible the `ProtobufSerializer not loaded...` log warning to fire even if you aren't trying to serialise a Protobuf class.
- The warning is fired every time `ProtobufSerializer.getSerializerClass()` is called which can result in the log line to be constantly fired and displayed in logs.

With these changes, the log warning can only be fired once and only the first time you try to serialise a Protobuf class.

## Related issues

This has been discussed in https://github.com/apache/fory/discussions/2349

## Benchmark

I believe it should have an insubstantial benchmark effect.


chaokunyang (approved):
Looks great

PR #2358, opened 2025-06-21
feat(java): row encoder array deserialization into List supports lazy access

stevenschlansker:
this way if you use only some indexes, you don't pay deserialization costs for any unread indexes

When List subtype is not specified, now we use our own lazy list. Since the user did not specify type, it should be ok, and you can always specify `ArrayList` to materialize the list eagerly


stevenschlansker:
Still a couple test cases to fix

stevenschlansker:
Ok, should be good to go now. The overhead should be low in the common case (just a check for null in the read path)

chaokunyang (approved):
Looks great!

PR #2375, opened 2025-07-04
chore(java): Remove redundant code

LiangliangSui:

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/serializer/SerializationBinding.java
+++ b/java/fory-core/src/main/java/org/apache/fory/serializer/SerializationBinding.java
@@ -227,42 +227,22 @@ interface SerializationBinding {
 
     @Override
     public void writeNullable(MemoryBuffer buffer, Object obj) {
-      if (obj == null) {
-        buffer.writeByte(Fory.NULL_FLAG);
-      } else {
-        buffer.writeByte(NOT_NULL_VALUE_FLAG);
-        writeNonRef(buffer, obj);
-      }
+      fory.writeNullable(buffer, obj);
     }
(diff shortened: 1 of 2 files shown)


java/fory-core/src/main/java/org/apache/fory/serializer/SerializationBinding.java
```diff
@@ -227,42 +227,22 @@ public void writeNonRef(MemoryBuffer buffer, Object obj, ClassInfoHolder classIn
 
     @Override
     public void writeNullable(MemoryBuffer buffer, Object obj) {
-      if (obj == null) {
-        buffer.writeByte(Fory.NULL_FLAG);
-      } else {
-        buffer.writeByte(NOT_NULL_VALUE_FLAG);
-        writeNonRef(buffer, obj);
-      }
+      fory.writeNullable(buffer, obj);
```
chaokunyang, line 230: Manually inline have better performance. JVM JIT nto always inline those calls

LiangliangSui:
Does readNullable also need to be manually inline?

https://github.com/apache/fory/blob/a9e5d999f58b2a0ec8f22ead9a20040d45e39a73/java/fory-core/src/main/java/org/apache/fory/serializer/SerializationBinding.java#L381-L384

chaokunyang:
> Does readNullable also need to be manually inline?
> 
> https://github.com/apache/fory/blob/a9e5d999f58b2a0ec8f22ead9a20040d45e39a73/java/fory-core/src/main/java/org/apache/fory/serializer/SerializationBinding.java#L381-L384

Yes, we'd better do that

PR #2378, opened 2025-07-04
chore(Scala): add Scala PolymorphicTest

pjfanning:
## What does this PR do?

It wasn't clear from the docs how Fory handled polymorphism.
The final test fails, so I set it to be ignored. deserializeJavaObject does not work if the class you pass to it is a superclass.

### DIFF
--- a/scala/src/test/scala/org/apache/fory/serializer/scala/PolymorphicTest.scala
+++ b/scala/src/test/scala/org/apache/fory/serializer/scala/PolymorphicTest.scala
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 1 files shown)


scala/src/test/scala/org/apache/fory/serializer/scala/PolymorphicTest.scala
```diff
@@ -0,0 +1,68 @@
...
+      val c = Circle(5.0)
+      fory.deserialize(fory.serialize(c), classOf[Shape]) shouldEqual c
+    }
+    "serialize/deserialize circle class (Java object format)" in {
+      val c = Circle(5.0)
+      fory.deserializeJavaObject(fory.serializeJavaObject(c), classOf[Circle]) shouldEqual c
+    }
+    "serialize/deserialize circle class (Java object format - shape hint)" ignore {
+      val c = Circle(5.0)
+      fory.deserializeJavaObject(fory.serializeJavaObject(c), classOf[Shape]) shouldEqual c
```
pjfanning, line 65: test fails - so set to ignore

chaokunyang: For polymorphicc class, fory needs final class be passed. Or use serialize/deserialize app i directly

pjfanning: I'm using the plain serialize/deserialize in practice but I thought I'd try serializeJavaObject/deserializeJavaObject to see how they worked.

chaokunyang: `deserializeJavaObject` needs teh exact concreate type be passed. Actually I think we can provide better error message here if wrong type is passed

Issue #2294, opened 2025-06-04
[Python] support __reduce__ for python native serialization


chaokunyang:
### Feature Request

https://docs.python.org/3/library/pickle.html#object.__reduce__ is a advanced magic method for customizing python serialization, fory should support this method too.

### Additional context

#2293

Issue #2293, opened 2025-06-04
[Python] support __getstate__/__setstate__ for python native serialization


chaokunyang:
### Feature Request

https://docs.python.org/3/library/pickle.html#object.__getstate__  is magic method in python for custimizing serialization state, fory should support that.

PR #2384, opened 2025-07-05
feat(python): Make PickleSerializer a CrossLanguageCompatibleSerializer

esafak:
## What does this PR do?

*   Enable correct serialization/deserialization of objects using `__getstate__` and `__setstate__` when Fory's Python serialization falls back to the standard pickle mechanism.
*   Modify `_registry.py` to allow `PickleSerializer` fallback for `Language.XLANG` when `require_type_registration=False`.
*   Update `serializer.py` to make `PickleSerializer` inherit from `CrossLanguageCompatibleSerializer`, ensuring `xwrite` and `xread` delegate to `write` and `read`.
*   Add tests in `test_serializer.py` to verify `__getstate__` and `__setstate__` behavior during Fory's pickle fallback.

## Related issues

Completes #2293, #2294

### DIFF
--- a/python/pyfory/_registry.py
+++ b/python/pyfory/_registry.py
@@ -479,8 +479,9 @@ class TypeResolver:
             return type_info
         elif not create:
             return None
-        if self.require_registration and \
-           (self.language != Language.PYTHON or not issubclass(cls, Enum)):
+        if self.require_registration and (
+           self.language != Language.PYTHON or not issubclass(cls, Enum)
+        ):
             raise TypeUnregisteredError(f"{cls} not registered")
         logger.info("Type %s not registered", cls)
         serializer = self._create_serializer(cls)
@@ -490,7 +491,7 @@ class TypeResolver:
                 type_id = TypeId.NAMED_ENUM
             elif type(serializer) is PickleSerializer:
                 type_id = PickleSerializer.PICKLE_TYPE_ID
-            if not self.require_registration: # This applies only if language is PYTHON
+            if not self.require_registration:
                 if isinstance(serializer, DataClassSerializer):
                     type_id = TypeId.NAMED_STRUCT
         elif self.language == Language.XLANG:
--- a/python/pyfory/_registry.py
+++ b/python/pyfory/_registry.py
@@ -480,7 +480,7 @@ class TypeResolver:
         elif not create:
             return None
         if self.require_registration and (
-           self.language != Language.PYTHON or not issubclass(cls, Enum)
+            self.language != Language.PYTHON or not issubclass(cls, Enum)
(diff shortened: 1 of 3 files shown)


chaokunyang:
Languages like java/go can't deserialize binary serialized by pickle, maybe we don't need it be CrossLanguageCompatibleSerializer? And we could invoke `__getstate__/__setstate__` indrectly in fory to make fory suppoert such objects instead of fallback to pickle. In the long run, I think fory can be a drop-in replacement for pickle but with better performance

esafak:
How about this plan?

1. Define two new serializers:
   - A "Stateful Serializer" for objects that support __getstate__ and __setstate__.
   - A "Reduce Serializer" for objects that implement __reduce__ or __reduce_ex__.

2. For objects with __getstate__/__setstate__:
   - In the serialization function, check if both methods are available.
   - If yes, call __getstate__ to obtain the object state and then serialize that state using Fory's native methods.
   - For deserialization, create a new instance without calling __init__, then use the deserialized state with __setstate__.

3. For objects with __reduce__/__reduce_ex__:
   - Detect if these methods exist and give them higher precedence over __getstate__/__setstate__.
   - Support basic cases where __reduce__ returns either just a global name or a tuple (callable, args).
   - Extend to handle objects that include state or more complex information (like iterators or custom functions).

4. For handling constructors (especially for immutable types):
   - Check for __getnewargs_ex__ or __getnewargs__ during instance creation.
   - Serialize the constructor arguments so that during deserialization, the new instance is created with the correct arguments.

5. Testing & Backwards Compatibility:
   - Implement incremental tests to cover each serialization case (basic state, __reduce__ cases, and __getnewargs*__ cases).
   - Ensure that existing functionality still falls back to the Python pickle behavior when necessary, but aim to use Fory’s native XLANG format for better cross-language support and performance.

In short, the idea is to leverage Fory’s native serialization for objects that support customization, gradually replacing pickle's behavior with a more transparent and cross-language-friendly approach while maintaining compatibility with edge cases.

chaokunyang:
> How about this plan?
> 
> In short, the idea is to leverage Fory’s native serialization for objects that support customization, gradually replacing pickle's behavior with a more transparent and cross-language-friendly approach while maintaining compatibility with edge cases.

Looks great! 👍  

With this plan, fory will be able to serialize any objects in python if it can be serialized by pickle, but with fory serialization format, the serialization will be faster.

Another thing is to support local classes and lambda like cloudpickle. But we can do that after finished this plan

esafak:
> Another thing is to support local classes and lambda like cloudpickle.

Could you create an issue and describe how you think it should be done?

PR #2385, opened 2025-07-05
chore(Scala): remove var that can be mutated

pjfanning:
Not good in scala to expose a `var` like that can be used to mutate state from outside the class


chaokunyang (approved):
LGTM

PR #2388, opened 2025-07-05
refactor(python): replace Black and Flake8 with Ruff

esafak:
Contribution Checklist

## Why?

I was running the existing linting script on my last PR and it was failing because it was incompatible with the new version of `importlib-metadata` I had installed. First I considered unpinning or updating the versions of the linting libraries, then I realized it would be better to simply get rid of them in favor of [ruff](https://astral.sh/ruff), a modern Rust-based linter and formatter which aims to replace flake8 and black.

Note that this PR includes a few ruff reformatted python files.

## What does this PR do?

- Removed Black and Flake8 configurations and dependencies.
- Integrated Ruff for both formatting and linting, updating scripts and configuration files accordingly.
- Adjusted documentation and examples in `README.md` and `CONTRIBUTING.md` to reflect the switch to Ruff.
- Improved simplicity and consistency in formatting validation steps.
- Migrates linter configuration into `pyproject.toml`

No

## Benchmark

Ruff is faster :)

## Notes

The following [pycodestyle ignores](https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes) are dropped support for:
- E121: continuation line under-indented for hanging indent
- E123: closing bracket does not match indentation of opening bracket’s line
- E126: continuation line with same indent as next logical line
- E704: multiple statements on one line (def)
- W503: line break before binary operator
- W504: line break after binary operator

And [this bugbear warning](https://github.com/PyCQA/flake8-bugbear?tab=readme-ov-file#list-of-warnings):
- B001: Do not use bare except:, it also catches unexpected events like memory errors, interrupts, system exit, and so on. Prefer except


chaokunyang (approved):
Ruff looks great

PR #2390, opened 2025-07-06
chore(Scala): refactor scala test so it asserts something

pjfanning:
I presume the test used to fail with an exception but now it passes, it is better not to have printlns and to actually compare the result of the deserialization to the original.


pandalee99 (approved):
LGTM

PR #2394, opened 2025-07-07
chore(Scala): Create DurationTest.scala

pjfanning:
* the way that Scala Durations have been represented internally has changed between recent Scala versions
* it is useful to test that serialization/deserialization works
* it would be a breaking change but it might be useful for fory-scala to have a dedicated serializer for Scala Durations that converted the duration to ISO-8601 duration string (eg `PT1H` for 1 hour).
* this would mean that we wouldn't be relying on the internal state of class instances but have a format that is independent of the Duration class internal state but that still is easy for us to deserialize and match to meaningful new Scala Duration instance when we deserialize


chaokunyang (approved):
LGTM, we could add a new DurationSerializer for scala Duration in a new PR

Issue #2392, opened 2025-07-06
[Java] add support so that deserializeJavaObject will throw an exception if the encoded data is of wrong type


pjfanning:
### Feature Request

I have an ignored test case in #2378 where Fory returns null instead of failing with an exception.
The test case passes with the wrong type in the input params.
If we like nulls (and I hate them), can we add a ForyBuilder option to allow users to opt out of ever getting nulls from deserializeJavaObject?

chaokunyang:
It should throw an exception instead of returning null, we need to fix it.

chaokunyang:
deserializeJavaObject is removed

PR #2399, opened 2025-07-08
chore(Java): handle deserialization scenario that currently throws an IndexOutOfBoundsException

pjfanning:
This test fails. I had expected it to fail with null when wrong type was used (because of #2392 but this fails differently from that - the use case where I saw a null returned was a Scala use case)

Fails with IndexOutOfBoundsException while I would expect a nicer exception. Maybe should be a Fory DeserializationException.
```
java.lang.IndexOutOfBoundsException: readerIndex(1) + length(2) exceeds size(7): org.apache.fory.memory.MemoryBuffer$BoundChecker@4e268090

	at org.apache.fory.memory.MemoryBuffer$BoundChecker.fillBuffer(MemoryBuffer.java:186)
	at org.apache.fory.memory.MemoryBuffer.checkReadableBytes(MemoryBuffer.java:2440)
	at org.apache.fory.ForyTest_Struct2ForyCodec_0.read(ForyTest_Struct2ForyCodec_0.java:59)
	at org.apache.fory.Fory.readDataInternal(Fory.java:1055)
	at org.apache.fory.Fory.deserializeJavaObject(Fory.java:1223)
	at org.apache.fory.Fory.deserializeJavaObject(Fory.java:1200)
	at org.apache.fory.ForyTest.testDeserializeJavaObjectWrongType(ForyTest.java:689)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/util/ExceptionUtils.java
+++ b/java/fory-core/src/main/java/org/apache/fory/util/ExceptionUtils.java
@@ -63,8 +63,7 @@ public class ExceptionUtils {
       throw new DeserializationException(objects, t);
     } else if (t instanceof Exception) {
       throw new DeserializationException(
-          String.format("Failed to deserialize input"),
-          t);
+          "Failed to deserialize input", t);
     } else {
       Platform.throwException(t);
       throw new IllegalStateException("unreachable");
--- a/java/fory-core/src/main/java/org/apache/fory/util/ExceptionUtils.java
+++ b/java/fory-core/src/main/java/org/apache/fory/util/ExceptionUtils.java
@@ -25,6 +25,7 @@ import java.util.List;
 import org.apache.fory.Fory;
 import org.apache.fory.collection.ObjectArray;
 import org.apache.fory.exception.DeserializationException;
+import org.apache.fory.exception.ForyException;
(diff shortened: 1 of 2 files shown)


chaokunyang:
How about catching all exceptions and rethrow exception with DeserializationException and use root exception as cause?

java/fory-core/src/main/java/org/apache/fory/util/ExceptionUtils.java
```diff
@@ -61,6 +61,9 @@ public static RuntimeException handleReadFailed(Fory fory, Throwable t) {
       // carry with read objects for better trouble shooting.
       List<Object> objects = Arrays.asList(readObjects.objects).subList(0, readObjects.size);
       throw new DeserializationException(objects, t);
+    } else if (t instanceof Exception) {
```
pjfanning: @chaokunyang I could modify this method to add an extra param that is the class instance that deserializeJavaObject method has access to - this would mean that I could improve the exception message by including the class name. I could overload the existing method by having a new method that takes the extra param and leave this version intact.

chaokunyang (approved):
LGTM

Issue #2402, opened 2025-07-10
[Python] support serialize local function


chaokunyang:
### Feature Request

Pickle doesn't support serialize local class object, for example:

```python
def f():
    from dataclasses import dataclass

    @dataclass
    class LocalClass:
        a: int
        b: str

    local_class_instance = LocalClass(a=1, b="2")
    return local_class_instance


import pickle

pickle.dumps(f())
```

will get error:
```python
AttributeError                            Traceback (most recent call last)
Cell In[6], line 15
     10     return local_class_instance
     13 import pickle
---> 15 pickle.dumps(f())

AttributeError: Can't pickle local object 'f.<locals>.LocalClass'
```

Fory can serialize such object:
```python
In [23]: fory.register_type(type(o))

In [24]: o
Out[24]: f.<locals>.LocalClass(a=1, b='2')

In [25]: fory.register_type(type(o))

In [26]: fory.serialize(o)
Out[26]: b'\xd4b\x06\x02\xff\x8f\x8a\x01\x97@\x00\x00\xff\x02\xff\x042'

In [27]: fory.deserialize(o)
```

But when it comes to function or lambda, fory doesn't support:
```python
In [35]: lam = lambda x: x*2

In [36]: fory.register_type(type(lam))
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[36], line 1
----> 1 fory.register_type(type(lam))

File python/pyfory/_serialization.pyx:648, in pyfory._serialization.Fory.register_type()

File python/pyfory/_serialization.pyx:467, in pyfory._serialization.TypeResolver.register_type()

File ~/Desktop/chaokun/fury_open_source/python/pyfory/_registry.py:284, in TypeResolver.register_type(self, cls, type_id, namespace, typename, serializer)
    275 def register_type(
    276     self,
    277     cls: Union[type, TypeVar],
   (...)
    282     serializer=None,
    283 ):
--> 284     return self._register_type(
    285         cls,
    286         type_id=type_id,
    287         namespace=namespace,
    288         typename=typename,
    289         serializer=serializer,
    290     )

File ~/Desktop/chaokun/fury_open_source/python/pyfory/_registry.py:324, in TypeResolver._register_type(self, cls, type_id, namespace, typename, serializer, internal)
    322     raise TypeError(f"{cls} registered already")
    323 register_type = self._register_xtype if self.fory.language == Language.XLANG else self._register_pytype
--> 324 return register_type(
    325     cls,
[...]

PR #2404, opened 2025-07-10
feat(python): Add FunctionSerializer for function serialization

### LINKED ISSUE #2301: [Python] support serialize lambda/closure function
### Feature Request

support serialize lambda/closure function

cloudpickle support serialize lambda/closure function, but if we use cloudpickle to serialize such objects, the object graph will be handled in two different places. And the object reference relationship will be breaked, and we may serialize some nested obejcts twice, one in fury and the other in cloudpickle. 

In application which depends on strict objects reference relationship, such deserialization will give wrong object and make the application not work

We need to support serializing lambda/closure function natively in fory

_No response_

### Additional context

_No response_

esafak:
## What does this PR do?

* Implement `FunctionSerializer` to handle serialization and deserialization of function objects.
* Utilize `cloudpickle` for robust function serialization, including closures and lambdas.
* Introduce new test cases in `test_function.py` to cover various function types:
    * Lambda functions (simple and with closures).
    * Regular functions.
    * Nested functions.
* Register `FunctionSerializer` in `_registry.py` for automatic detection and handling of function types during serialization.
* Ensure compatibility with both cross-language and Python-only modes.

## Related issues

Closes #2301 and #2402 

## Benchmark

Not done

### DIFF
--- a/python/pyfory/serializer.py
+++ b/python/pyfory/serializer.py
@@ -849,6 +849,39 @@ class ReduceSerializer(CrossLanguageCompatibleSerializer):
             raise ValueError(f"Invalid reduce data format: {reduce_data[0]}")
 
 
+class FunctionSerializer(CrossLanguageCompatibleSerializer):
+    """Serializer for function objects using cloudpickle."""
+
+    def __init__(self, fory, cls):
+        super().__init__(fory, cls)
+        try:
+            import cloudpickle
+
+            self.cloudpickle = cloudpickle
+        except ImportError:
+            raise ImportError("cloudpickle is required for function serialization")
+
+    def xwrite(self, buffer, value):
+        # Use cloudpickle to serialize the function
+        pickled_data = self.cloudpickle.dumps(value)
+        buffer.write_bytes_and_size(pickled_data)
+
+    def xread(self, buffer):
+        # Use cloudpickle to deserialize the function
+        pickled_data = buffer.read_bytes_and_size()
+        return self.cloudpickle.loads(pickled_data)
+
+    def write(self, buffer, value):
+        # For Python-only mode, also use cloudpickle
+        pickled_data = self.cloudpickle.dumps(value)
+        buffer.write_bytes_and_size(pickled_data)
+
+    def read(self, buffer):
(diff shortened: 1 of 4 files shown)


python/pyfory/serializer.py
```diff
@@ -849,6 +849,39 @@ def read(self, buffer):
...
+        try:
+            import cloudpickle
+
+            self.cloudpickle = cloudpickle
+        except ImportError:
+            raise ImportError("cloudpickle is required for function serialization")
+
+    def xwrite(self, buffer, value):
+        # Use cloudpickle to serialize the function
+        pickled_data = self.cloudpickle.dumps(value)
```
chaokunyang: Could we support function serialization natively in fory. The function isntance may capture some closures objects, if we use cloudpickle/pickle to serializethe function, such objects will escape from fory object graph serialization. If the object is referenced outside of the function, it will be serialized by fory too, which will create two objects at deserialization. But this is wrong since we have only one reference at seriaization.

And another thing is that is also escape from fory class registration security checks

In the long run, we need to remove pickle/cloudpickle dependencies, to make fory as a drop-in replacement serialization framework.

esafak: I replaced them with `fory.(de)serialize_ref` and marshal.

python/pyfory/serializer.py
```diff
@@ -940,6 +943,222 @@ def read(self, buffer):
...
+        # Create and serialize a dictionary with only the necessary globals
+        globals_to_serialize = {name: globals_dict[name] for name in global_names if name in globals_dict}
+        self.fory.serialize_ref(buffer, globals_to_serialize)
+
+        # Handle additional attributes
+        attrs = {}
+        for attr in dir(func):
+            if attr.startswith("__") and attr.endswith("__"):
+                continue
+            if attr in ("__code__", "__name__", "__defaults__", "__closure__", "__globals__", "__module__", "__qualname__"):
```
chaokunyang: Could we create a private module variable to cache this set?

chaokunyang (approved):
Looks great!

Issue #1299, opened 2024-01-03
Consider to rewrite shell scripts in Python


PragmaTwice:
TBH these scripts are hard to maintain (especially considering the difference between bash and zsh), we need a more structure language like python to make them readable and maintanable.

LiangliangSui:
I totally agree with this, it can mask the differences in script syntax (e.g. windows script syntax is very unfriendly), and it is easier to write and code readable.

chaokunyang:
Sounds good to me, but rewriting those scripts into python is not an easy work, and most scripts are used by CI only. I suggest we start from rewriting the scripts used by contributors everyday.

chaokunyang:
This remind me of `PyTorch`. When you filing an issue, you can execute a python script from `PyTorch`, which can collect system information as much as it can, and print it. Then you can paste such information into the issue

LiangliangSui:
So can we start by rewriting the C++ CI Build script in Windows? such as `run_ci.py`

chaokunyang:
Yep, we can start from rewriting the C++ CI Build script in Windows

PR #2406, opened 2025-07-10
refactor(ci): Migrate CI runner to Python scripts

esafak:
At the suggestion of @LiangliangSui and @PragmaTwice, we propose a staged migration from shell scripts to Python per language. Please review the PR carefully; it was computer assisted.

## What does this PR do?

* Refactor CI runner from shell scripts to a Python-based system.
* Introduces `ci/run_ci.py` as the main entry point for CI tasks.
* Organizes CI logic into modular Python files within `ci/tasks/`.
* Replaces shell script calls with Python function calls for each language.
* Updates GitHub Actions workflows (`.github/workflows/ci.yml`) to use the new Python runner.
* Enhances argument parsing for better control over CI tasks (e.g., specifying Java versions).
* Adds support for various Java versions, including specific scenarios like Windows Java 21 and GraalVM.
* Includes new CI tasks for Go and code formatting.
* Consolidates and cleans up existing CI logic from shell scripts.
* Introduces environment variables (e.g., `USE_PYTHON_CPP`) to allow gradual migration and fallback to shell scripts.

## Related issues

Completes #1299

## Benchmark

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,13 +59,11 @@ jobs:
         with:
           python-version: 3.8
       - name: Install bazel
-        run: ./ci/run_ci.sh install_bazel
-      - name: Install python
-        run: ./ci/run_ci.sh install_python
-      - name: Install pyfory
-        run: ./ci/run_ci.sh install_pyfory
+        run: python ./ci/run_ci.py cpp --install-deps-only
+      - name: Install python dependencies
+        run: pip install pyarrow==15.0.0 Cython wheel pytest setuptools -U
       - name: Run CI with Maven
-        run: ./ci/run_ci.sh java${{ matrix.java-version }}
+        run: python ./ci/run_ci.py java --version ${{ matrix.java-version }}
       - name: Upload Test Report
         uses: actions/upload-artifact@v4
         if: ${{ !cancelled() }}
@@ -94,13 +92,11 @@ jobs:
         with:
           python-version: 3.8
       - name: Install bazel
-        run: ./ci/run_ci.sh install_bazel
-      - name: Install python
-        run: ./ci/run_ci.sh install_python
-      - name: Install pyfory
-        run: ./ci/run_ci.sh install_pyfory
+        run: python ./ci/run_ci.py cpp --install-deps-only
+      - name: Install python dependencies
+        run: pip install pyarrow==15.0.0 Cython wheel pytest setuptools -U
       - name: Run CI with Maven
-        run: ./ci/run_ci.sh java${{ matrix.java-version }}
+        run: python ./ci/run_ci.py java --version ${{ matrix.java-version }}
 
   java21_windows:
     name: Windows Java 21 CI
@@ -117,9 +113,13 @@ jobs:
         with:
           java-version: ${{ matrix.java-version }}
           distribution: "temurin"
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.8
       - name: Run CI with Maven
         shell: bash
-        run: ./ci/run_ci.sh windows_java21
+        run: python ./ci/run_ci.py java --version windows_java21
 
   graalvm:
     name: GraalVM CI
@@ -141,7 +141,7 @@ jobs:
           python-version: 3.8
       - name: Build native image and run
         shell: bash
-        run: ./ci/run_ci.sh graalvm_test
+        run: python ./ci/run_ci.py java --version graalvm
 
   kotlin:
     name: Kotlin CI
@@ -158,10 +158,14 @@ jobs:
         with:
           java-version: ${{ matrix.java-version }}
           distribution: "temurin"
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.8
       - name: Install fory java
         run: cd java && mvn -T10 --no-transfer-progress clean install -DskipTests && cd -
       - name: Run Kotlin CI
-        run: ./ci/run_ci.sh kotlin
+        run: python ./ci/run_ci.py kotlin
 
   scala:
     name: Scala CI
@@ -191,8 +195,12 @@ jobs:
         with:
           java-version: 8
           distribution: "temurin"
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.8
       - name: Run CI
-        run: ./ci/run_ci.sh integration_tests
+        run: python ./ci/run_ci.py java --version integration_tests
 
   javascript:
     name: JavaScript CI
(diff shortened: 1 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -59,13 +59,11 @@ jobs:
         with:
           python-version: 3.8
       - name: Install bazel
-        run: ./ci/run_ci.sh install_bazel
-      - name: Install python
-        run: ./ci/run_ci.sh install_python
-      - name: Install pyfory
-        run: ./ci/run_ci.sh install_pyfory
+        run: python ./ci/run_ci.py cpp --install-deps-only
```
chaokunyang, line 62: will `install-deps-only` option install bazel?

.github/workflows/ci.yml
```diff
@@ -94,13 +92,11 @@ jobs:
...
+      - name: Install python dependencies
+        run: pip install pyarrow==15.0.0 Cython wheel pytest setuptools -U
```
chaokunyang, line 97: If we move this to run_ci.py, we could comute arrow version from python version. i.e. using pyarrow 15 when running py3.8, otehrwise using pyarrrow 18.0

.github/workflows/ci.yml
esafak: Yes, because `cpp/run()` calls `install_cpp_deps` which installs bazel, regardless of `install_deps_only`:

```python
def run(install_deps_only=False):
    """Run C++ CI tasks.

    Args:
        install_deps_only: If True, only install dependencies without running tests.
    """
    logging.info("Running C++ CI tasks")
    common.install_cpp_deps()

    if install_deps_only:
        logging.info("Skipping tests as --install-deps-only was specified")
        return
    ...
```
https://github.com/esafak/fory/blob/101f085198fabcc1abcbda1c2697034ce9d4f6cd/ci/tasks/cpp.py#L22-L33

```python
def install_cpp_deps():
    """Install dependencies for C++ development."""
    # Check the Python version and install the appropriate pyarrow version
    python_version = platform.python_version()
    if python_version.startswith("3.13"):
        exec_cmd("pip install pyarrow==18.0.0")
        exec_cmd("pip install numpy")
    else:
        exec_cmd(f"pip install pyarrow=={PYARROW_VERSION}")
        # Automatically install numpy
    exec_cmd("pip install psutil")
    install_bazel()
```
https://github.com/esafak/fory/blob/101f085198fabcc1abcbda1c2697034ce9d4f6cd/ci/tasks/common.py#L179-L179

.github/workflows/ci.yml
esafak: I agree. I think there might be some redundancy here since we also do
```python
def install_cpp_deps():
    """Install dependencies for C++ development."""
    # Check the Python version and install the appropriate pyarrow version
    python_version = platform.python_version()
    if python_version.startswith("3.13"):
        exec_cmd("pip install pyarrow==18.0.0")
        exec_cmd("pip install numpy")
    else:
        exec_cmd(f"pip install pyarrow=={PYARROW_VERSION}")
        # Automatically install numpy
    exec_cmd("pip install psutil")
    install_bazel()
```
https://github.com/esafak/fory/blob/101f085198fabcc1abcbda1c2697034ce9d4f6cd/ci/tasks/common.py#L168-L179

I should probably let the GHA run `pip install -U pytest setuptools wheel cython` and install arrow/numpy in python. What do you think?

chaokunyang: I prefer do it in the script, this script can also be used by users to run tests locally

.github/workflows/ci.yml
chaokunyang: I see, thanks

.github/workflows/ci.yml
```diff
@@ -191,8 +195,12 @@ jobs:
         with:
           java-version: 8
           distribution: "temurin"
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.8
       - name: Run CI
-        run: ./ci/run_ci.sh integration_tests
+        run: python ./ci/run_ci.py java --version integration_tests
```
chaokunyang, line 203: It looks a little odd to let integration_tests as a version

.github/workflows/ci.yml
```diff
@@ -295,13 +298,11 @@ jobs:
...
           python-version: 3.8
       - name: Install bazel
-        run: ./ci/run_ci.sh install_bazel
-      - name: Install python
-        run: ./ci/run_ci.sh install_python
-      - name: Install pyfory
-        run: ./ci/run_ci.sh install_pyfory
+        run: python ./ci/run_ci.py cpp --install-deps-only
+      - name: Install python dependencies
+        run: pip install pyarrow==15.0.0 Cython wheel pytest setuptools -U
```
chaokunyang, line 303: ditto, maybe put it into python script

.github/workflows/ci.yml
esafak: I agree. I try to minimize GHA usage.

.github/workflows/ci.yml
esafak: We could rename it `version_or_test` ?

.github/workflows/ci.yml
chaokunyang: Could we also rename `Install bazel` to install cpp deps?

.github/workflows/ci.yml
chaokunyang: how about add a --module  or --component to specify which module to run in java, if no such module specified, we run all modules for java

ci/run_ci.py
```diff
@@ -15,202 +15,270 @@
...
+#   USE_PYTHON_FORMAT     - Format implementation
+#
+# By default, JavaScript, Rust, and C++ use the Python implementation,
+# while Java, Kotlin, Python, Go, and Format use the shell script implementation.
+
+# Environment variables to control which languages use the Python implementation
+# Default to the status quo before migration:
+# - JavaScript, Rust, and C++ use Python implementation
+# - Java, Kotlin, Python, Go, and Format use shell script implementation
+USE_PYTHON_CPP = os.environ.get("USE_PYTHON_CPP", "1") == "1"
```
chaokunyang, line 66: Where we set `USE_PYTHON_CPP` to 1, I don't see such code in this PR

ci/run_ci.py
```diff
@@ -15,202 +15,270 @@
...
+            "8",
+            "11",
+            "17",
+            "21",
+            "24",
+            "windows_java21",
+            "integration_tests",
+            "graalvm",
+        ],
+        default="17",
```
chaokunyang, line 167: we need to use java8, jars built usxing higher jdk may not run on lower jdk version.

ci/run_ci.py
esafak: You can override it if with env vars if you want, otherwise it defaults to 1, set right here.

chaokunyang (approved):
I'll update the remaining part in another PR

PR #2412, opened 2025-07-13
feat(scala): support scala object with default values

### LINKED ISSUE #1683: [Scala] support default not-null value in COMPATIBLE mode.
Currently, for scala, add new field and deserialize from old binary data will get a `null` as new field value. But avoid null is a good practice in scala language.
```scala
  val personBytes = readBytesFromFile("person_v1") // there are old version `Person(1,true,some text)`
  //  append field as new person: case class Person(a: Int, b: Boolean, c: String, d: String = "default d")
  val deserPerson = fury.deserializeJavaObject(personBytes, classOf[Person])
  println(s"deserPerson: ${deserPerson}") // deserPerson: Person(1,true,some text,null)
```

I think it's better using default value our a empty value to set the new field. Such as give the result:
```scala
// deserPerson: Person(1,true,some text, "default d")
```
And if there not a default value in the field define, can give a empty value. For String is `""` will be better for null.

If the new field is a structure, can using a default value to instance this one.
`case class Foo(a: String, b: Int)` can setting the default value as `Foo("", 0)`

However, for some performance care scenario. Using `null` should be better and handler by develop.

I advice add a new configuration to decide using the default value or `null` for new field.

chaokunyang:
## What does this PR do?

Support scala case default values. The implementation consists of:

1. **`ScalaCaseClassUtils`**: Utility class for detecting Scala case classes and retrieving default values from companion object methods.

2. **`MetaSharedSerializer`**: Modified to cache default value field information and apply default values to missing fields just before returning the deserialized object.

3. **`ScalaDefaultValueField`**: Internal field info class that caches field accessors and default values for efficient application during deserialization.

4. **Cached Processing**: During serializer initialization, default values are discovered and cached. During deserialization, cached default values are applied to any null fields that have corresponding default parameters.

This feature enhances Fury's compatibility with Scala case classes and provides better support for evolving schemas with default parameters.

## Related issues

Closes #1683

### DIFF
--- a/docs/guide/scala_guide.md
+++ b/docs/guide/scala_guide.md
@@ -163,13 +163,13 @@ val opt1: Option[Long] = None
 println(fory.deserialize(fory.serialize(opt1)))
 ```
 
-## Scala Case Class Default Values Support
+## Scala Class Default Values Support
 
-Fory now supports Scala case class default values during deserialization when using meta-shared serialization mode. This feature allows for forward/backward compatibility when case classes have default parameters.
+Fory now supports Scala  class default values during deserialization when using meta-shared serialization mode. This feature allows for forward/backward compatibility when case classes have default parameters.
 
 ### Overview
 
-When a Scala case class has default parameters, the Scala compiler generates companion object methods like `apply$default$1`, `apply$default$2`, etc. that return the default values. Fory can now detect these methods and use them when deserializing objects where certain fields are missing from the serialized data.
+When a Scala class has default parameters, the Scala compiler generates companion object methods like `apply$default$1`, `apply$default$2`, etc. that return the default values. Fory can now detect these methods and use them when deserializing objects where certain fields are missing from the serialized data.
 
 ### Example
 
--- a/docs/guide/scala_guide.md
+++ b/docs/guide/scala_guide.md
@@ -162,3 +162,83 @@ println(fory.deserialize(fory.serialize(opt)))
 val opt1: Option[Long] = None
 println(fory.deserialize(fory.serialize(opt1)))
 ```
+
+## Scala Case Class Default Values Support
+
+Fury now supports Scala case class default values during deserialization when using meta-shared serialization mode. This feature allows for forward/backward compatibility when case classes have default parameters.
+
+### Overview
+
+When a Scala case class has default parameters, the Scala compiler generates companion object methods like `apply$default$1`, `apply$default$2`, etc. that return the default values. Fury can now detect these methods and use them when deserializing objects where certain fields are missing from the serialized data.
+
+### Example
+
+Consider a Scala case class with default values:
+
+```scala
+case class SomeClass(v: List[IdAnyVal], x: Int = 1)
+```
+
+When this class is serialized and then deserialized in a context where the `x` field is missing from the serialized data, Fury will automatically use the default value `1` for the `x` field.
+
+### How It Works
+
+1. **Detection**: Fury detects if a class is a Scala case class by looking for its companion object (class name + "$") and checking for the presence of an `apply` method.
+
+2. **Default Value Discovery**: Fury scans the companion object for methods named `apply$default$1`, `apply$default$2`, etc., which correspond to default values for parameters.
+
+3. **Field Mapping**: During deserialization, Fury identifies fields that exist in the target class but are missing from the serialized data.
+
+4. **Value Application**: After reading all available fields from the serialized data, Fury applies default values to any missing fields just before returning the deserialized object.
+
+### Usage
+
+This feature is automatically enabled when:
+
+- Meta-shared serialization is enabled (`withMetaShare(true)`)
+- The target class is detected as a Scala case class
+- A field is missing from the serialized data but exists in the target class
+
+No additional configuration is required.
+
+### Limitations
+
+- **Parameter Name Matching**: The field-to-parameter mapping relies on parameter names being preserved during compilation. This requires the `-parameters` compiler flag to be enabled for optimal results.
+
+- **Companion Object Access**: The feature requires access to the companion object class. If the companion object is not accessible due to visibility modifiers, default values cannot be retrieved.
+
+- **Method Handle Creation**: The feature uses Java method handles to invoke default value methods. This requires appropriate access permissions.
+
+### Example Test
+
+```scala
+case class TestClass(v: String, x: Int = 1, y: Double = 2.0)
+
+val fory = Fory.builder()
+  .withMetaShare(true)
+  .withScalaOptimizationEnabled(true)
+  .build()
+
+// Serialize with all fields
+val original = TestClass("test", 42, 3.14)
+val serialized = fory.serialize(original)
+
+// Deserialize - missing fields will use default values
+val deserialized = fory.deserialize(serialized)
+// deserialized.x will be 1 (default)
+// deserialized.y will be 2.0 (default)
+```
+
+### Implementation Details
+
+The implementation consists of:
+
+1. **`ScalaCaseClassUtils`**: Utility class for detecting Scala case classes and retrieving default values from companion object methods.
+
+2. **`MetaSharedSerializer`**: Modified to cache default value field information and apply default values to missing fields just before returning the deserialized object.
+
+3. **`ScalaDefaultValueField`**: Internal field info class that caches field accessors and default values for efficient application during deserialization.
+
+4. **Cached Processing**: During serializer initialization, default values are discovered and cached. During deserialization, cached default values are applied to any null fields that have corresponding default parameters.
+
+This feature enhances Fury's compatibility with Scala case classes and provides better support for evolving schemas with default parameters.
--- a/docs/guide/scala_guide.md
+++ b/docs/guide/scala_guide.md
@@ -185,6 +185,7 @@ Fory supports default values for:
 1. **Detection**: Fory detects if a class is a Scala class by checking for the presence of default value methods (`apply$default$N` or `$default$N`).
 
 2. **Default Value Discovery**:
+
    - For case classes: Fory scans the companion object for methods named `apply$default$1`, `apply$default$2`, etc.
    - For regular Scala classes: Fory scans the class itself for methods named `$default$1`, `$default$2`, etc.
 
@@ -291,7 +292,7 @@ val deserialized = fory.deserialize(serialized, classOf[ConfigurationWithDefault
 object NestedClasses {
   // Class WITHOUT default values (for serialization)
   case class SimplePerson(name: String)
-  
+
   // Class WITH default values (for deserialization)
   case class Address(street: String, city: String = "DefaultCity")
   case class PersonWithDefaults(name: String, address: Address = Address("DefaultStreet"))
--- a/docs/guide/scala_guide.md
+++ b/docs/guide/scala_guide.md
@@ -185,7 +185,6 @@ Fory supports default values for:
 1. **Detection**: Fory detects if a class is a Scala class by checking for the presence of default value methods (`apply$default$N` or `$default$N`).
 
 2. **Default Value Discovery**:
-
    - For case classes: Fory scans the companion object for methods named `apply$default$1`, `apply$default$2`, etc.
    - For regular Scala classes: Fory scans the class itself for methods named `$default$1`, `$default$2`, etc.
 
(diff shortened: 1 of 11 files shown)


pandalee99 (approved):

pjfanning:
I code mainly in Scala and the issue description in #1683 doesn't make much sense to me,
I also am the main maintainer of jackson-module-scala so I have some experience of serde and Scala.

`case class Foo(a: String, b: Int)` defaulting to `Foo("", 0)` - I think this is strange

I think deserialization should just fail if there are no field values specified in the serialized data.

Defaulting the field values to null is strange in Scala but so is inferring the values.

Scala supports this: `case class Foo(a: String = "", b: Int = 0)`
In this case, the class maintainer has specified the default values and Fory deser should respect them - but only if they are set.

I would be ok with Moandic types being defaulted to empty values even if the class does not specify an explicit default but we should respect that default if set.

Examples:
* Option should default to None
* List and Set and Map (collections) should default to the appropriate empty collection

docs/guide/scala_guide.md
```diff
@@ -162,3 +162,61 @@ println(fory.deserialize(fory.serialize(opt)))
...
+
+```scala
+case class SomeClass(v: List[IdAnyVal], x: Int = 1)
+```
+
+When this class is serialized and then deserialized in a context where the `x` field is missing from the serialized data, Fury will automatically use the default value `1` for the `x` field.
+
+### How It Works
+
+1. **Detection**: Fury detects if a class is a Scala case class by looking for its companion object (class name + "$") and checking for the presence of an `apply` method.
```
pjfanning: * I don't think is a good way to do this
* case classes extend scala.Product (this is automatic, you don't need to declare this) - so check if the class extends this

pjfanning: if you want to support Scala classes that are not case classes, then we might need something else to support them

docs/guide/scala_guide.md
```diff
@@ -162,3 +162,61 @@ println(fory.deserialize(fory.serialize(opt)))
 val opt1: Option[Long] = None
 println(fory.deserialize(fory.serialize(opt1)))
 ```
+
+## Scala Case Class Default Values Support
+
+Fury now supports Scala case class default values during deserialization when using meta-shared serialization mode. This feature allows for forward/backward compatibility when case classes have default parameters.
```
pjfanning: Fory not Fury

pjfanning: there are other examples of Fury in this doc too

docs/guide/scala_guide.md
```diff
@@ -162,3 +162,61 @@ println(fory.deserialize(fory.serialize(opt)))
...
+### Overview
+
+When a Scala case class has default parameters, the Scala compiler generates companion object methods like `apply$default$1`, `apply$default$2`, etc. that return the default values. Fury can now detect these methods and use them when deserializing objects where certain fields are missing from the serialized data.
+
+### Example
+
+Consider a Scala case class with default values:
+
+```scala
+case class SomeClass(v: List[IdAnyVal], x: Int = 1)
```
pjfanning: Can we default List, Map, Set and Option to their empty values instead of null? If an explicit default is provided, use that but without one, empty values are good for these types.
Assuming 0 or "" for numbers and strings, I don't think that would be a good idea.

docs/guide/scala_guide.md
```diff
@@ -162,3 +162,61 @@ println(fory.deserialize(fory.serialize(opt)))
...
+
+// Serialize with all fields
+val original = TestClass("test", 42, 3.14)
+val serialized = fory.serialize(original)
+
+// Deserialize - missing fields will use default values
+val deserialized = fory.deserialize(serialized)
+// deserialized.x will be 1 (default)
+// deserialized.y will be 2.0 (default)
+```
```
pjfanning, line 313: this example is strange - I would expect `TestClass("test", 42, 3.14)` to serialize with the explicit values and deserialize to the explicit values - ie  `TestClass("test", 42, 3.14)`

TBH with Fory, how would you even serialize something that has missing field values?

In JSON, you could have external users who create their own JSON that they expect you to deserialize and in that case, `{ "v": "value" }` is a valid input for TestClass and the deserializer should return `TestClass("value", 1, 2.0)` because the defaults are specified in the class definition.
But `{ "v": "value", "x": 10 }` should deserialize as `TestClass("value", 10, 2.0)` because the input data has an explicit x=10.

docs/guide/scala_guide.md
chaokunyang: thanks, just updated

java/fory-core/src/main/java/org/apache/fory/util/ScalaDefaultValueUtils.java
```diff
@@ -0,0 +1,576 @@
...
+   * Gets the companion object instance for a Scala class, supporting both top-level and nested
+   * classes.
+   *
+   * @param cls the Scala class
+   * @return the companion object instance
+   * @throws RuntimeException if the companion object cannot be accessed
+   */
+  private static Object getCompanionObject(Class<?> cls) {
+    // Try the standard approach for top-level case classes
+    String companionClassName = cls.getName() + "$";
```
pjfanning, line 278: the code to get the companion class is duplicated in multiple methods - can you create a shared method and reuse it?

scala/src/test/scala/org/apache/fory/serializer/scala/ScalaDefaultValueTest.scala
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fory.serializer.scala
```
pjfanning, line 20: It is messy to add a scala dependency in a Java project.
It would be better to add this test in fory-scala and to run the test with all the supported scala versions. Scala 3 has a very different code base to Scala 2.12 but I would still expect the Java reflection based code you are using to work ok.

scala/src/test/scala/org/apache/fory/serializer/scala/ScalaDefaultValueTest.scala
```diff
@@ -0,0 +1,178 @@
...
+        val deserialized = fory.deserialize(serialized).asInstanceOf[RegularScalaClass]
+        deserialized shouldEqual original
+        deserialized.name shouldEqual "John"
+        deserialized.age shouldEqual 30
+        deserialized.city shouldEqual "New York"
+      }
+
+      s"handle missing fields with default values in regular Scala class in $modeName" in {
+        val fory = createFory(codegen)
+        val original = new RegularScalaClass("Jane") // age=25, city="Unknown"
```
pjfanning: Surely this test case already passes without any of the code changes in this PR?
Scala will set the age and city when you create the RegularScalaClass instance. So those values will be serialized and the values are there to be deserialized.

The only case where I see the new code being useful is if someone writes a Scala class and then modifies that class to add a new field with a default.

Someone might be using an old version of the class definition in their runtime. They serialize a class instance with Fory. That serialized data is sent to another runtime and that has the newer class definition with the extra field. In that case, Fory might be able to deserialize the data because it recognises that the serialized data is missing a field and the class definition tells us what to use as a default for that field.

Is this a use case that Fory wants to support? What if the serial Uid changes on the class?

scala/src/test/scala/org/apache/fory/serializer/scala/ScalaDefaultValueTest.scala
chaokunyang: It's in scala dir of fory, not a dir of fory-core.

docs/guide/scala_guide.md
chaokunyang: My bad. I updated the tests but forgot the document, it should be something like:
```scala
case class CaseClassMultipleDefaultsNoDefaults(v: String) // No x and y fields at all
case class CaseClassMultipleDefaultsWithDefaults(v: String, x: Int = 1, y: Double = 2.0)
// Serialize object WITHOUT x and y fields (they don't exist in source class)
val original = CaseClassMultipleDefaultsNoDefaults("test")
val serialized = fory.serializeJavaObject(original)

// Deserialize into class WITH default values for x and y
val deserialized = fory.deserializeJavaObject(serialized, classOf[CaseClassMultipleDefaultsWithDefaults])
```

I will update the doc later

docs/guide/scala_guide.md
chaokunyang: We could do this, but if user's code has some special logic conditioned on the nullability, this will break their code.

docs/guide/scala_guide.md
chaokunyang: We can support it, it also have static methods for default values:
```scala
class RegularScalaClassWithDefaults(val name: String, val age: Int = 25, val city: String = "Unknown")

chaokunyang@B-M856MD6M-2341 scala % javap -c RegularScalaClassWithDefaults
Warning: File ./RegularScalaClassWithDefaults.class does not contain class RegularScalaClassWithDefaults
Compiled from "ScalaDefaultValueTest.scala"
public class org.apache.fory.serializer.scala.RegularScalaClassWithDefaults {
  public static java.lang.String $lessinit$greater$default$3();
    Code:
       0: getstatic     #21                 // Field org/apache/fory/serializer/scala/RegularScalaClassWithDefaults$.MODULE$:Lorg/apache/fory/serializer/scala/RegularScalaClassWithDefaults$;
       3: invokevirtual #23                 // Method org/apache/fory/serializer/scala/RegularScalaClassWithDefaults$.$lessinit$greater$default$3:()Ljava/lang/String;
       6: areturn

  public static int $lessinit$greater$default$2();
    Code:
       0: getstatic     #21                 // Field org/apache/fory/serializer/scala/RegularScalaClassWithDefaults$.MODULE$:Lorg/apache/fory/serializer/scala/RegularScalaClassWithDefaults$;
       3: invokevirtual #27                 // Method org/apache/fory/serializer/scala/RegularScalaClassWithDefaults$.$lessinit$greater$default$2:()I
       6: ireturn

```

scala/src/test/scala/org/apache/fory/serializer/scala/ScalaDefaultValueTest.scala
chaokunyang: Yes, it's exactly the case that fory want to support. I updated the code and documents, everything should looks OK now.

chaokunyang: serial Uid changes won't have an effect, fory don't use serial Uid

java/fory-core/src/main/java/org/apache/fory/util/ScalaDefaultValueUtils.java
chaokunyang: Updated, no duplicates now

chaokunyang:
Hi @pjfanning , I've addressed all coments, everything should looks OK now, could you help review it again?

scala/src/test/scala/org/apache/fory/util/ScalaDefaultValueUtilsTest.scala
```diff
@@ -0,0 +1,382 @@
...
+       
+       if (fields.nonEmpty) {
+         val field = fields(0)
+         field.getFieldName should not be null
+         field.getDefaultValue should not be null
+         field.getFieldAccessor should not be null
+         field.getClassId should be >= 0.toShort
+       } else {
+         // If no fields are returned, that's also valid
+         fields.length shouldEqual 0
```
pjfanning: `fields shouldBe empty` - more idiomatic scalatest syntax - the failure on this would list the fields elements if it failed

`fields.length shouldEqual 0` will fail with 1 not equal to 0 - not very helpful.

pjfanning: although this seems hardly even useful here because line 370 already assures that this assertion will always pass

chaokunyang: Updated, should be OK now

PR #2414, opened 2025-07-15
feat(java): introduce Compact Row Codec

### LINKED ISSUE #2337: Row format better data packing
### Feature Request

I have been working to convert an application from Fory serialization format to use Row Format for better compatibility and hopefully less space usage (since each row does not have to save information about the data structure)

However, the savings are not as much as I would have liked. I think part of the problem is that the row format seems to be quite wasteful in the fixed fields area - all fields are aligned to 8 bytes, even boolean or short, so there is a lot of zero-padding, based on the `BinaryRowWriter` implementation.

Is this necessary? Or would it be possible that we can improve the row format to have less padding, when fields are known to have smaller sizes?

_No response_

_No response_

_No response_

### Additional context

_No response_

stevenschlansker:
## What does this PR do?

Introduce alternate "compact" row encoding that better uses knowledge of fixed-size types and sacrifices alignment to save space.

Introduce new Builder pattern to avoid explosion of `Encoders` static methods as more features are added to row format.

Optimizations include:
* struct stores fixed-size fields (e.g. Int128. FixedSizeBinary) inline in fixed-data area without offset + size
* struct of all fixed-sized fields is itself considered fixed-size to store in other struct or array
* struct skips null bitmap if all fields are non-nullable
* struct sorts fields by fixed-size for best-effort (but not guaranteed) alignment
* struct can use less than 8 bytes for small data (int, short, etc)
* struct null bitmap stored at end of struct to borrow alignment padding if possible
* array stores fixed-size fields inline in fixed-data area without offset+size
* array header uses 4 bytes for size (since Collection and array are only int-sized) and leaves remaining 4 bytes for start of null bitmap

Fixups include:
* toString better handles varbinary / fixed-binary (hex dump of first 256 bytes)
* start making Javadoc for row format

Compromises:
* less alignment could increase access time, but this is opt-in. and I think on modern processors it is not such a big deal.
* increased complexity of offset lookup, try to pre-compute in an array when possible and use `StableValue` when it is GA

Not compatible with existing row format.

## Related issues

Fixes #2337

New API for new Compact codec. Existing codec unchanged.

### DIFF
--- a/java/fory-format/README.md
+++ b/java/fory-format/README.md
@@ -8,11 +8,60 @@ Fory row format is heavily inspired by spark tungsten row format, but with chang
 - Decimal use arrow decimal format.
 - Variable-size field can be inline in fixed-size region if small enough.
 - Allow skip padding by generate Row using aot to put offsets in generated code.
-- Support adding fields without breaking compatibility.
 
-The initial fory java row data structure implementation is modified from spark unsafe row/writer.
+The initial Fory java row data structure implementation is modified from spark unsafe row/writer.
 
 See `Encoders.bean` Javadoc for a list built-in supported types.
+
+## Row Format Java
+
+To begin using the row format from Java, start with the `Encoders` class:
+
+```
+// Many built-in types and collections are supported
+public record MyRecord(int key, String value) {}
+
+// The encoder supplier is relatively expensive to create
+// It is thread-safe and should be re-used
+Supplier<RowEncoder<MyRecord>> encoderFactory =
+    Encoders.buildBeanCodec(MyRecord.class)
+            .build();
+
+// Each individual encoder is relatively cheap to create
+// It is not thread-safe, but may be reused by the same thread
+var encoder = encoderFactory.get();
+byte[] encoded = encoder.encode(new MyRecord(42, "Test"));
+
+MyRecord deserialized = encoder.decode(encoded);
+```
+
+## Compact Format
+
+The default row format is cross-language compatible and alignment-padded for maximum performance.
+When data size is a greater concern, the compact format provides an alternate encoding that uses
+significantly less space.
+
+Enable the compact codec on the encoder builder:
+
+```
+Supplier<RowEncoder<MyRecord>> encoderFactory =
+    Encoders.buildBeanCodec(MyRecord.class)
+            .compactEncoding()
+            .build();
+```
+
+Optimizations include:
+
+- struct stores fixed-size fields (e.g. Int128. FixedSizeBinary) inline in fixed-data area without offset + size
+- struct of all fixed-sized fields is itself considered fixed-size to store in other struct or array
+- struct skips null bitmap if all fields are non-nullable
+- struct sorts fields by fixed-size for best-effort (but not guaranteed) alignment
+- struct can use less than 8 bytes for small data (int, short, etc)
+- struct null bitmap stored at end of struct to borrow alignment padding if possible
+- array stores fixed-size fields inline in fixed-data area without offset+size
+- array header uses 4 bytes for size (since Collection and array are only int-sized) and leaves remaining 4 bytes for start of null bitmap
+
+## Custom Type Registration
 
 It is possible to register custom type handling and collection factories for the row format -
 see Encoders.registerCustomCodec and Encoders.registerCustomCollectionFactory. For an interface,
@@ -45,7 +94,7 @@ static class UuidEncoder implements CustomCodec.MemoryBufferCodec<UUID> {
 static class SortedSetOfUuidDecoder implements CustomCollectionFactory<UUID, SortedSet<UUID>> {
   @Override
   public SortedSet<UUID> newCollection(final int size) {
-    return new TreeSet<>(UnsignedUuidComparator.INSTANCE);
+    return new TreeSet<>();
   }
 }
 
--- a/java/fory-format/src/main/java/org/apache/fory/format/encoder/BufferResettingMapEncoder.java
+++ b/java/fory-format/src/main/java/org/apache/fory/format/encoder/BufferResettingMapEncoder.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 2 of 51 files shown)


stevenschlansker:
Hi @chaokunyang , I believe this PR is ready for review now. No rush, I know it is a lot to go through. Happy to explain or change anything that does not make sense.

We run this now in our testing environment successfully with about a 30% - 50% space savings on our datasets. Will be taking compact codec to production over the coming weeks.

chaokunyang:
@stevenschlansker Great! 30% - 50% space savings is a huge gain! I will review this PR ASAP, and release it in `0.13.0`

java/fory-format/src/main/java/org/apache/fory/format/row/binary/CompactBinaryRow.java
```diff
@@ -0,0 +1,157 @@
...
+    return new CompactBinaryArray(field);
+  }
+
+  @Override
+  protected BinaryMap newMap(final Field field) {
+    return new CompactBinaryMap(field);
+  }
+
+  @Override
+  protected int isNullBitmapOffset() {
```
chaokunyang: this looks like a method return `boolean`, how abouting naming it as `nullabilityBitmapOffset/nullBitmapOffset`

java/fory-format/src/main/java/org/apache/fory/format/encoder/CodecFormat.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.arrow.vector.types.pojo.Schema;
+import org.apache.fory.format.row.binary.BinaryArray;
+import org.apache.fory.format.row.binary.BinaryMap;
+import org.apache.fory.format.row.binary.BinaryRow;
+import org.apache.fory.format.row.binary.writer.BaseBinaryRowWriter;
+import org.apache.fory.format.row.binary.writer.BinaryArrayWriter;
+import org.apache.fory.memory.MemoryBuffer;
+import org.apache.fory.reflect.TypeRef;
+
+interface CodecFormat {
```
chaokunyang: Will the name `Encoding` better align with `Encoder`?

chaokunyang (approved):
LGTM overall, I left some minor comments. And could we also add some documents to `row_format_guide.md` in later PR. 

Another thing for compact is to inline string. For string less that 7 bytes, we can put it into fixed-size region. Many string are less than 7 bytes, I think this can also brings some space savings

java/fory-format/src/main/java/org/apache/fory/format/encoder/Encoder.java
```diff
@@ -30,11 +30,36 @@
...
+  /**
+   * Encode to a buffer with an embedded size. Variants without embedded size are not compatible.
+   */
   void encode(MemoryBuffer buffer, T obj);
+
+  /**
+   * Encode to a buffer without an embedded size. Variants with embedded size are not compatible.
+   * Returns number of bytes written to the buffer.
+   */
+  int bareEncode(MemoryBuffer buffer, T obj);
```
chaokunyang: Is it possible to merge `bareEncode` and `encode` into one method?

chaokunyang: this is only called on map/struct/array, merge into one method and use a branch don't introduce lots of runtime cost, but we can make the interface more simple

java/fory-format/src/main/java/org/apache/fory/format/encoder/BufferResettingMapEncoder.java
```diff
@@ -0,0 +1,107 @@
...
+package org.apache.fory.format.encoder;
+
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.fory.format.row.binary.BinaryArray;
+import org.apache.fory.format.row.binary.BinaryMap;
+import org.apache.fory.format.row.binary.writer.BinaryArrayWriter;
+import org.apache.fory.memory.MemoryBuffer;
+import org.apache.fory.memory.MemoryUtils;
+
+class BufferResettingMapEncoder<T> implements MapEncoder<T> {
```
chaokunyang, line 30: Could we add some documents to clarify the situations when we need to use this and the difference between this and BaseMapEncoder

java/fory-format/src/main/java/org/apache/fory/format/encoder/BaseMapEncoder.java
```diff
@@ -0,0 +1,129 @@
...
+package org.apache.fory.format.encoder;
+
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.fory.format.row.binary.BinaryArray;
+import org.apache.fory.format.row.binary.BinaryMap;
+import org.apache.fory.format.row.binary.writer.BinaryArrayWriter;
+import org.apache.fory.memory.MemoryBuffer;
+import org.apache.fory.memory.MemoryUtils;
+
+class BaseMapEncoder<M> implements MapEncoder<M> {
```
chaokunyang: `BaseXXX` looks like an abstract parent class, could we use a different name?

java/fory-format/src/main/java/org/apache/fory/format/encoder/BufferResettingMapEncoder.java
chaokunyang: How about `BinaryMapEncoder` to relfect the relationship with BinaryRow, BinaryArray naming in codebase

java/fory-format/src/main/java/org/apache/fory/format/encoder/BaseMapEncoder.java
chaokunyang: How about `BinaryMapEncoder` to relfect the relationship with BinaryRow, BinaryArray naming in codebase

similiar naming for other encoders

chaokunyang: And since we have different kinds of encoder for same type, we'd better to add some javadoc to document the distinctions

stevenschlansker:
Thank you for your review, I will make changes in the coming days.

java/fory-format/README.md
```diff
@@ -8,12 +8,52 @@ Fory row format is heavily inspired by spark tungsten row format, but with chang
...
+
+- struct stores fixed-size fields (e.g. Int128. FixedSizeBinary) inline in fixed-data area without offset + size
+- struct of all fixed-sized fields is itself considered fixed-size to store in other struct or array
+- struct skips null bitmap if all fields are non-nullable
+- struct sorts fields by fixed-size for best-effort (but not guaranteed) alignment
+- struct can use less than 8 bytes for small data (int, short, etc)
+- struct null bitmap stored at end of struct to borrow alignment padding if possible
+- array stores fixed-size fields inline in fixed-data area without offset+size
+- array header uses 4 bytes for size (since Collection and array are only int-sized) and leaves remaining 4 bytes for start of null bitmap
+
```
chaokunyang, line 63: could we add a minmal example about how to use compact mode?

stevenschlansker:
ok @chaokunyang I think I've addressed all the feedback. I like the idea of inline strings but I think it makes sense to defer that to a future PR at this point since all these changes are ready now

chaokunyang:
Great work!

Issue #2409, opened 2025-07-12
[Python] Remove pickle dependecy from python serialization


chaokunyang:

esafak:
`cloudpickle` is used as `self.pickler = Pickler(self.buffer)` in [handle_unsupported_write](https://github.com/apache/fory/blob/main/python/pyfory/_fory.py#L442-L447):
```python
def handle_unsupported_write(self, buffer, obj):
        if self._unsupported_callback is None or self._unsupported_callback(obj):
            buffer.write_bool(True)
            self.pickler.dump(obj)
        else:
            buffer.write_bool(False)
```
What would you like to do with it?

chaokunyang:
> `cloudpickle` is used as `self.pickler = Pickler(self.buffer)` in [handle_unsupported_write](https://github.com/apache/fory/blob/main/python/pyfory/_fory.py#L442-L447):
> 
> def handle_unsupported_write(self, buffer, obj):
>         if self._unsupported_callback is None or self._unsupported_callback(obj):
>             buffer.write_bool(True)
>             self.pickler.dump(obj)
>         else:
>             buffer.write_bool(False)
> What would you like to do with it?

@esafak We need to support the serialization of any types in fory  that can be serialized by pickle/cloudpickle

PR #2417, opened 2025-07-15
feat: Replace cloudpickle with native Fory serialization

esafak:
## What does this PR do?

* Removed cloudpickle dependency and related logic.
* Replaced cloudpickle serialization with Fory's native serialization for unsupported types.
* Updated serializers for Python arrays, NumPy arrays, and Objects to use native serialization.
* Modified tests to reflect the removal of cloudpickle and use of native serialization.
* Added global helper functions for testing to avoid issues with local function serialization.
* Updated README.md with a note about bazel caching issues.

## Related issues

Contributes to #2409 

## Notes

Next `PickleSerializer` will be removed bit by bit.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -16,6 +16,12 @@ pip install pyarrow==15.0.0 Cython wheel pytest
 pip install -v -e .
 ```
 
+If the last steps fails with an error like `libarrow_python.dylib: No such file or directory`,
+you are probably suffering from bazel's aggressive caching; the sought library is longer at the
+temporary directory it was the last time bazel ran. To remedy this run
+
+> bazel clean --expunge
+
 ### Environment Requirements
 
 - python 3.8+
--- a/python/pyfory/_fory.py
+++ b/python/pyfory/_fory.py
@@ -37,10 +37,6 @@ try:
 except ImportError:
     np = None
 
-from cloudpickle import Pickler
-
-from pickle import Unpickler
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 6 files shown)


esafak:
@chaokunyang I did not realize the xlang tests were being skipped in Python. I set `ENABLE_CROSS_LANGUAGE_TESTS` and now pytest complained that `fixture 'data_file_path' not found` is missing. I could not find a definition for it now or in the past. How is it supposed to work?

chaokunyang:
It's called from java CrossLanguageTest.java. Java serialize object into bytes and write to file, then execute python tests and pass that filepath as inputs

esafak:
I think a comment the python xlang tests explain this would have helped, along with documentation of how to run them locally.

chaokunyang:
> I think a comment the python xlang tests explain this would have helped, along with documentation of how to run them locally.

Yes, a comment would be helpful, thanks for this suggestion, I added a comment in https://github.com/apache/fory/pull/2420

chaokunyang:
Hi @esafak , would you like to continue working on this? I did a benchmark several days before. It shows that pyfory is faster than cpickle and has smaller size compared to pickle for [data in cpython benchmark](https://github.com/python/pyperformance/blob/main/pyperformance/data-files/benchmarks/bm_pickle/run_benchmark.py) . If we can drop-in replace pickle, it would be very promising.

esafak:
I'm busy, so feel free to take over.

chaokunyang:
> I'm busy, so feel free to take over.

I implemented in https://github.com/apache/fory/pull/2629

Issue #2441, opened 2025-08-06
[Java]deserialize exception when map put null value


spaceric:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fory/issues) and found no similar issues.

### Version

0.11.2

### Component(s)

Java

### Minimal reproduce step

    public static Student initStudentList2() {

        Student student1 = new Student();
        student1.setName("jack");
        student1.setAge(20);

        Map<String, Long> score1 = new HashMap<>();
        score1.put("math", null);
        score1.put("english", null);
        student1.setScore(score1);

        return student1;
    }

    @GetMapping("/stu/rw2")
    public String foryRW2(){
        try {

            Student details = initStudentList2();

            long start = System.currentTimeMillis();
            byte[] bytes = ForyUtils.FORY.serialize(details);

            Student deserialize = ForyUtils.FORY.deserialize(bytes, Student.class);

            System.out.println("time：" + (System.currentTimeMillis() - start));
            return JSON.toJSONString(deserialize);

        } catch (Exception e) {
            System.out.println(e);
            throw e;
        }
    }

### What did you expect to see?

deserialize correct

### What did you see instead?

java.lang.IndexOutOfBoundsException: readerIndex(54) + length(1) exceeds size(54): org.apache.fory.memory.MemoryBuffer$BoundChecker@61b615b2
	at org.apache.fory.memory.MemoryBuffer$BoundChecker.fillBuffer(MemoryBuffer.java:186) ~[fory-core-0.11.2.jar:0.11.2]
	at org.apache.fory.memory.MemoryBuffer.readUnsignedByte(MemoryBuffer.java:1342) ~[fory-core-0.11.2.jar:0.11.2]
	at org.apache.fory.serializer.collection.AbstractMapSerializer.readNullChunkKVFinalNoRef(AbstractMapSerializer.java:794) ~[fory-core-0.11.2.jar:0.11.2]
	at com.lqh.lqhtestweb.fory.StudentForyCodec_0.readFields2$(StudentForyCodec_0.java:382) ~[na:na]
	at com.lqh.lqhtestweb.fory.StudentForyCodec_0.read(StudentForyCodec_0.java:451) ~[na:na]
	at org.apache.fory.Fory.readDataInternal(Fory.java:1055) ~[fory-core-0.11.2.jar:0.11.2]
	at org.apache.fory.Fory.readRef(Fory.java:929) ~[fory-core-0.11.2.jar:0.11.2]
[...]

PR #2444, opened 2025-08-06
fix(java): crash when map only contains null value

LouisLou2:
## Related issues
- #2441


chaokunyang:
Could we add an unit test for this bug?

LouisLou2:
Can you take a look at this Rust CI error and see what the issue is? @chaokunyang

chaokunyang:
> Can you take a look at this Rust CI error and see what the issue is? @chaokunyang

It's code lint error, we need to fix those line error according to the error message

Issue #2435, opened 2025-07-31
[Question] There appear to be a few hurdles for Android support


LouisLou2:
I'm trying to use Fury on Android(Using Kotlin) and ran into some initialization issues. I have a quick question about platform support.

It seems there are two main blockers for Android:

Java Version Check: The static initializer in _JDKAccess fails with a NumberFormatException because System.getProperty("java.specification.version") returns "0.9" on Android.

```java
static {
  String property = System.getProperty("java.specification.version"); // for Android, always "0.9"
  if (property.startsWith("1.")) {
    property = property.substring(2);
  }
  String jmvName = System.getProperty("java.vm.name", "");
  IS_OPEN_J9 = jmvName.contains("OpenJ9");
  JAVA_VERSION = Integer.parseInt(property); // exception: 0.9 can't be parsed to an Integer
  
  //...
}
```

sun.misc.Unsafe API: Fury relies heavily on sun.misc.Unsafe, which isn't available on standard Android runtimes.

Given these, are there any plans or known workarounds to get Fury running on Android?

### Additional Context

https://youtrack.jetbrains.com/issue/KT-44089/Java-version-checking-doesnt-work-on-Android

neboskreb:
@LouisLou2 Check this related issue: #1101

PR #2445, opened 2025-08-06
refactor(java): [Don't Merge for Now]Modify the current memory manipulation to be compatible with Android

LouisLou2:
## What does this PR do?
This commit refactors memory management logic as an initial step towards making the project compatible with the Android platform.

**Key points**:
- This is an exploratory change to propose a potential solution. The design may require further refinement for clarity.
- These changes are strictly limited to memory operations. Full Android support will require other modifications.
- Detailed documentation explaining the rationale will be provided later.

## Related issues
- #2435

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/annotation/NotForAndroid.java
+++ b/java/fory-core/src/main/java/org/apache/fory/annotation/NotForAndroid.java
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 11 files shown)


chaokunyang:
Andriod String seems be implemented as a native class, we can't get `value` and `coder` field from String object. 

Here are two tricks to keep maxmimal performance:
- Use `public void getBytes (int srcBegin, int srcEnd, byte[] dst,  int dstBegin)` to get bytes for `JDK9+`
- Use `String.toCharArray` for JDK8

Don't use `String.charAt`, it's much more slow than get chars array first and check by a loop. JVM can employ some optimzation for simple loops if it doesn't contains nested methods call

Issue #2350, opened 2025-06-18
[Java] MemeryBuffer support custom direct allocator


MrChang0:
### Feature Request

`MemeryBuffer` grow need `ByteBuffer` Supplier make it allocate memory from a provoided memory pool when grow.

chaokunyang:
Seems we can provide global static method to set memory allocator. Could you also check whether there are other places needs to use this allocator?

MrChang0:
global static memory allocator is good idea. After checked other places like `ByteBufferSerializer` and `BufferObject` that allocate new buffers. I think those places don't need allocator, Keeping it simple is better, they should just handle serialization directly.

PR #2457, opened 2025-08-11
feat: growBuffer support DirectByteBuffer

Kn0688:
## What does this PR do?

The growBuffer method of MemoryBuffer support DirectByteBuffer

no

## Benchmark

no

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/memory/MemoryBuffer.java
+++ b/java/fory-core/src/main/java/org/apache/fory/memory/MemoryBuffer.java
@@ -1229,7 +1229,6 @@ public final class MemoryBuffer {
     writerIndex = newIdx;
   }
 
-  /** For off-heap buffer, this will make a heap buffer internally. */
   public void grow(int neededSize) {
     int length = writerIndex + neededSize;
     if (length > size) {
@@ -1237,7 +1236,6 @@ public final class MemoryBuffer {
     }
   }
 
-  /** For off-heap buffer, this will make a heap buffer internally. */
   public void ensure(int length) {
     if (length > size) {
       growBuffer(length);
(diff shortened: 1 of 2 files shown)


java/fory-core/src/main/java/org/apache/fory/memory/MemoryBuffer.java
```diff
@@ -1249,6 +1247,13 @@ private void growBuffer(int length) {
         length < BUFFER_GROW_STEP_THRESHOLD
             ? length << 2
             : (int) Math.min(length * 1.5d, Integer.MAX_VALUE - 8);
+    if (heapMemory == null) {
+      ByteBuffer bb = ByteBuffer.allocateDirect(newSize);
```
chaokunyang, line 1251: This is very slow. We can not use this api to allocate direct buffer. How about set a global direct buffer allocator instead and use that to allocate buffer. It's also what https://github.com/apache/fory/issues/2350 want to have

Kn0688: Do you mean using  allocator like netty's [PooledByteBufAllocator](https://mp.weixin.qq.com/s/Vv_XV61osu9N-jCRN-Y5Kw)?

chaokunyang: The implememtation may use such allocator, but fory could just define an interface, and let users to implement such `ByteBufferAllocator`. 

Another thing we need to figure out is memory deallocation of such buffers. Since we allocate memory from a `BufferAllocator`, when this `MemoryBuffer` is not used, we should return the buffer back to the pool. This is the tricky thing to do. Even for `java.nio.ByteBuffer`, this is not tackled well. `ByteBuffer` use soft/weak reference to trigger callback when there are no reference to `ByteBuffer`. But it's not predictable, this is why PooledByteBufAllocator use explict release API `ByteBuf#release`. If we add allocator, we should also add `release` method to `MemoryBuffer`

Kn0688: Got it. I will submit a new PR for this feature.

Issue #2480, opened 2025-08-15
manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl not installable on linux


chaokunyang:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fory/issues) and found no similar issues.

### Version

main

### Component(s)

Python

### Minimal reproduce step

```
pip install pyfory -i http://test.pypi.org/simple/   --trusted-host test.pypi.org  --no-cache-dir
```

Name like `pyfory-0.12.0a1-cp310-cp310-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl` won't work:
https://test-files.pythonhosted.org/packages/ab/4b/302f02fe781b727446250841b15a93e805ab3b3881fcd68a179622125e1c/pyfory-0.12.0a1-cp310-cp310-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl

Name like `pyfury-0.10.2-cp310-cp310-manylinux1_x86_64.whl` will work:
https://files.pythonhosted.org/packages/9e/41/5a97ae7714152354743ccd2b005b503d095a8c63ef54345025dc4aa77b56/pyfury-0.10.2-cp310-cp310-manylinux1_x86_64.whl

### What did you expect to see?

install suceed

### What did you see instead?

```
$pip install pyfory -i http://test.pypi.org/simple/   --trusted-host test.pypi.org  --no-cache-dir
Looking in indexes: http://test.pypi.org/simple/
ERROR: Could not find a version that satisfies the requirement pyfory (from versions: none)
ERROR: No matching distribution found for pyfory
```

### Anything Else?

#2381

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

chaokunyang:
@esafak linux wheel don't work either with #2381

esafak:
Which Linux version? It seems to be working per the latest pr

chaokunyang:
> Which Linux version? It seems to be working per the latest pr

My linux version is `4.9.151-015.ali3000.alios7.x86_64`. It seems my pip version is a little old, after I upgreade pip, the wheel can be installed. I think maybe we need to release wheel for manylinux2014, it's supported in more pip versions

esafak:
[...]

PR #2487, opened 2025-08-18
ci: Migrate python to cibuildwheel

esafak:
This is an attempt to simplify CI

## What does this PR do?

* Add Python 3.13 to pyproject.toml supported versions.
* Configure cibuildwheel for Linux and Musl Linux builds, including architecture support for x86_64 and aarch64.
* Introduce new GitHub Actions workflows for building containerized and native wheels, enhancing CI/CD capabilities.
* Update existing workflows to use cibuildwheel for building wheels across various Python versions and architectures.

## Related issues

#2480


esafak:
cibuildwheel does not work well with bazel out of the box. We would need to create our own images derived from manylinux. Something for later.

PR #2494, opened 2025-08-20
fix(java): Encoders.mapEncoder(TypeRef, TypeRef, TypeRef, Fory) should load bean classes

stevenschlansker:
We dynamically select the key and value type, and tried using this overload of Encoders.mapEncoder
With the single-arg mapEncoder invocation, bean codec classes are loaded with `token4BeanLoad`
    
But this overload never loads the bean classes leading to unexpected exceptions during Map codec compile since bean codecs are not loaded

### DIFF
--- a/java/fory-format/src/test/java/org/apache/fory/format/encoder/MapEncoderTest.java
+++ b/java/fory-format/src/test/java/org/apache/fory/format/encoder/MapEncoderTest.java
@@ -134,9 +134,9 @@ public class MapEncoderTest {
   @Test
   public void testKVStructMap() {
     Map<SimpleFoo, SimpleFoo> map = ImmutableMap.of(SimpleFoo.create(), SimpleFoo.create());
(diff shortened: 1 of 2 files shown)


java/fory-format/src/test/java/org/apache/fory/format/encoder/MapEncoderTest.java
```diff
@@ -134,9 +134,9 @@ public void testSimpleNestStructWithMapEncoder() {
   @Test
   public void testKVStructMap() {
     Map<SimpleFoo, SimpleFoo> map = ImmutableMap.of(SimpleFoo.create(), SimpleFoo.create());
-    MapEncoder encoder = Encoders.mapEncoder(new TypeRef<Map<SimpleFoo, SimpleFoo>>() {});
```
stevenschlansker, line 137: IDE warnings due to raw type `MapEncoder`

chaokunyang (approved):

PR #2510, opened 2025-08-25
fix(ci): fix build wheel bash on windows

chaokunyang:


esafak:
I think this will not be enough; see #2511

PR #2506, opened 2025-08-24
fix(ci): Use $ROOT/dist for wheel distribution

esafak:
## Why

The python wheels were not being found.

## What does this PR do?

* Update deploy.sh to use the $ROOT/dist path for storing wheels.
* This change ensures consistent wheel distribution across different platforms.
* Fixes an issue where wheels were being placed in relative ../dist directories.

## Related issues

Fixes #2505

PR #2511, opened 2025-08-25
ci: Fix Windows wheel creation

esafak:
## Why?

* Wheels were not created on Windows because it does not use bash by default
* Once that works, downloading them will fail because `actions/download-artifact` requires the `run-id` to access artifacts from different workflows.

## What does this PR do?

* Add `actions: read` permission to release workflow.
* Configure `download-artifact` action to use `run-id` and `pattern` for downloading wheels from a previous workflow run.
* Enable `merge-multiple` for downloading artifacts.
* Add `shell: bash` to build step in native release workflow.

## Related issues

#2506

### DIFF
--- a/.github/workflows/build-native-release.yml
+++ b/.github/workflows/build-native-release.yml
@@ -42,6 +42,7 @@ jobs:
         run: ./ci/run_ci.sh install_bazel_windows
         shell: bash
       - name: Build wheel
+        shell: bash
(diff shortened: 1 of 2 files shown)


chaokunyang (approved):

chaokunyang:
The publish still fail:
image

https://github.com/apache/fory/actions/runs/17198889531/job/48785824375

PR #2523, opened 2025-08-25
feat(CI): Cache npm, add node 24, lock file

esafak:
## Why?

JS CI fails a lot due to npm.

## What does this PR do?

* Caches npm to hopefully avoid the problem.
* Replaces node 16 with 24

## Notes
Per [the documentation](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#working-with-lockfiles):
> The action follows [actions/cache](https://github.com/actions/cache/blob/main/examples.md#node---npm) guidelines, and caches global cache on the machine instead of node_modules, so cache can be reused between different Node.js versions.

This means all node versions share a cache per operating system. Since all node versions run simultaneously, one version writes while the others read, due to contention. In the latest run, here are the winners (check the _Post Use Node.js_ step):

* Windows: [node 18](https://github.com/apache/fory/actions/runs/17217046296/job/48843158759)
* MacOS: [node 24](https://github.com/apache/fory/actions/runs/17217046296/job/48843158761)
* Ubuntu: [node 24](https://github.com/apache/fory/actions/runs/17217046296/job/48843158783)

### DIFF
--- a/.github/workflows/build-containerized-pr.yml
+++ b/.github/workflows/build-containerized-pr.yml
@@ -33,3 +33,8 @@ jobs:
       - uses: actions/checkout@v5
       - name: Build and test wheels
         run: ./ci/build_linux_wheels.py --arch ${{ runner.arch }}
+      - name: Upload wheels as artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: pyfory-wheels-containerized-${{ matrix.os }}
+          path: dist/*.whl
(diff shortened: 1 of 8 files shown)


chaokunyang (approved):

chaokunyang:
package-lock.json is an auto-generated file, do we need to check in this file？ @theweipeng . If we run nimp install, this file will change most of times

chaokunyang:
https://github.com/apache/arrow-js/blob/main/yarn.lock and https://github.com/apache/opendal/blob/main/bindings/nodejs/pnpm-lock.yaml checked lock file too

esafak:
npm install will update the lock file if you change the dependencies. That is to be expected.

theweipeng (approved):
LGTM

Issue #2527, opened 2025-08-26
Python wheel release incomplete


chaokunyang:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fory/issues) and found no similar issues.

### Version

main

### Component(s)

Python

### Minimal reproduce step

git push apache v0.12.1-a4

### What did you expect to see?

Release 30 wheels  like https://test.pypi.org/manage/project/pyfory/release/0.12.0a2/ :
<img width="870" height="1862" alt="image" src="https://github.com/user-attachments/assets/8201cef9-3279-4812-adc2-640876fedba3" />
 

### What did you see instead?

The release is not right still:

https://test.pypi.org/manage/project/pyfory/release/0.12.1a3/
<img width="1896" height="1630" alt="image" src="https://github.com/user-attachments/assets/426e148a-4bb2-44b6-84f2-0dbf8d19aa8d" />

Lots of wheels are missed, it should have 30 wheel files, but we only uploaded 10 wheels.

The current released macos wheels are incomplete too:

<img width="1730" height="142" alt="image" src="https://github.com/user-attachments/assets/4a22aebf-4cac-4ec6-8afa-9815c023f5b1" />

it can only install pyfory==0.12.0a2:
<img width="2754" height="224" alt="image" src="https://github.com/user-attachments/assets/4c96a7db-73ad-441a-b3b1-fe62a1e681e6" />

We should upload wheels like https://test.pypi.org/manage/project/pyfory/release/0.12.0a2/ :
<img width="870" height="1862" alt="image" src="https://github.com/user-attachments/assets/8201cef9-3279-4812-adc2-640876fedba3" />

### Anything Else?
#2520
#2521
#2522
### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

chaokunyang:
@esafak maybe we need to compare current release pipeline with https://github.com/apache/fory/tree/v0.12.0 and cherry-pick missed features

esafak:
The problem is that I can't test the release workflows. Just testing GHA in general is very difficult. That's why I use dagger pipelines in my own code; I can run them locally. Anyway, I have a new PR that should improve things.

PR #2532, opened 2025-08-26
fix(ci): Use ref_name tag in Windows, name release runs, verify wheel versions, use all python versions for release

esafak:
## Why?

The Python release pipeline is broken:
* Windows wheels do not use the ref tag (pyfory-0.12.1-cp39-cp39-win_amd64.whl vs pyfory-0.12.1a3-cp310-cp310-macosx_15_0_universal2.whl)
* Linux does not release wheels for the full set of Python versions.

## What does this PR do?

* Introduce `ci/tasks/python_container_build_script.sh` for containerized wheel builds.
* Update `ci/build_linux_wheels.py` to use the new container script and define Python versions.
* Enhance `ci/deploy.sh` to force wheel version on Windows.
* Add version verification logic to `.github/workflows/build-native-pr.yml`.
* Update GitHub Actions workflows (`release-python.yaml`, `build-native-release.yml`, `build-containerized-release.yml`, `build-native-pr.yml`) for consistency and artifact naming.
* Refactor `ci/build_linux_wheels.py` to pass `GITHUB_REF_NAME` to Docker for version verification.

## Related issues

#2527

### DIFF
--- a/ci/tasks/python_container_build_script.sh
+++ b/ci/tasks/python_container_build_script.sh
@@ -33,7 +33,8 @@ verify_version() {
     # Check if GITHUB_REF_NAME is available and use it for verification
     if [ -n "$GITHUB_REF_NAME" ]; then
         # Strip leading 'v' if present
-        local expected_version="${GITHUB_REF_NAME#v}"
+        local expected_version
+        expected_version="$(ci/deploy.sh parse_py_version $GITHUB_REF_NAME)"
         echo "Expected version: $expected_version"
 
         if [ "$installed_version" != "$expected_version" ]; then


ci/deploy.sh
```diff
@@ -85,6 +85,21 @@ build_pyfory() {
     else
       $PYTHON_CMD setup.py bdist_wheel --dist-dir="$ROOT/dist"
     fi
+  elif [[ "$OSTYPE" == "msys" || "$OSTYPE" == "win32" ]]; then
+
+    if [ -n "$GITHUB_REF_NAME" ]; then
+      # Strip leading 'v' if present
+      version="${GITHUB_REF_NAME#v}"
```
chaokunyang: This may not generate right verson, the release.py have special handle for version, which change `0.12.1-rc1` to `0.12.1rc1`:
```python
def _update_python_version(lines, v: str):
    for index, line in enumerate(lines):
        if "__version__ = " in line:
            v = v.replace("-alpha", "a")
            v = v.replace("-beta", "b")
            v = v.replace("-rc", "rc")
            lines[index] = f'__version__ = "{v}"\n'
            break
```

ci/deploy.sh
```diff
@@ -85,6 +85,21 @@ build_pyfory() {
...
+      echo "Using version from GITHUB_REF_NAME: $version"
+
+      # Windows tends to drop alpha/beta markers - force it through setup.cfg
```
chaokunyang: We don;t have this for v0.12.0, but hte released wheels for windows are right:
image

https://test.pypi.org/project/pyfory/0.12.0a2/#files

The version are read from :
```
[tool.setuptools.dynamic]
version = {attr = "pyfory.__version__"}
```

And it's set in `ci/release.py`:

```python
def _update_python_version(lines, v: str):
    for index, line in enumerate(lines):
        if "__version__ = " in line:
            v = v.replace("-alpha", "a")
            v = v.replace("-beta", "b")
            v = v.replace("-rc", "rc")
            lines[index] = f'__version__ = "{v}"\n'
            break
```

Could you check whether you set this version right?

esafak: If you look at [the windows artifacts being uploaded recently](https://github.com/apache/fory/actions/runs/17227356645/job/48874547955#annotation:4:142), you'll see the version is incorrect.

ci/deploy.sh
esafak: I can't call python in bash, so I'll implement this function in it.

chaokunyang: `version="${version//-alpha/a}" && version="${version//-beta/b}" && version="${version//-rc/rc}"` this should work

ci/deploy.sh
chaokunyang: I compared the release yaml with v0.12.0.  v0.12.0 didn't pass `tag` to `ci/deploy.sh bump_py_version`.

`bump_py_version` strip `v` from tag. 

```bash
bump_py_version() {
  local version="$1"
  if [ -z "$version" ]; then
    # Get the latest tag from the current Git repository
    version=$(git describe --tags --abbrev=0)
    # Check if the tag starts with 'v' and strip it
    if [[ $version == v* ]]; then
      version="${version:1}"
    fi
  fi
  python "$ROOT/ci/release.py" bump_version -l python -version "$version"
}
```

If you pass tag explicitly, yo uneed top update the script:
```bash
bump_py_version() {
  local version="$1"
  if [ -z "$version" ]; then
    # Get the latest tag from the current Git repository
    version=$(git describe --tags --abbrev=0)
  fi
  # Check if the tag starts with 'v' and strip it
  if [[ $version == v* ]]; then
    version="${version:1}"
  fi
  python "$ROOT/ci/release.py" bump_version -l python -version "$version"
}
```

ci/deploy.sh
```diff
@@ -88,8 +88,10 @@ build_pyfory() {
   elif [[ "$OSTYPE" == "msys" || "$OSTYPE" == "win32" ]]; then
 
     if [ -n "$GITHUB_REF_NAME" ]; then
-      # Strip leading 'v' if present
       version="${GITHUB_REF_NAME#v}"
+      version="${version//-alpha/a}"
```
chaokunyang: How about refactoring `bump_py_version` into two functions, one is `get_py_version`, another is `bump_py_version`, which just invoke `python "$ROOT/ci/release.py" bump_version -l python -version $(get_py_version)`, then we can use `get_py_version` here

ci/deploy.sh
esafak: OK, I'll do that. [The current run](https://github.com/apache/fory/actions/runs/17255959530) created the wheel `pyfory-0.13.0.dev0-cp38-cp38-win_amd64.whl`. Is that right?

ci/deploy.sh
esafak: I don't see how; it is already called `bump_py_version`. shall we do that in another pr?

.github/workflows/build-native-release.yml
```diff
@@ -45,13 +45,25 @@ jobs:
...
         run: |
           python -m pip install --upgrade pip
           pip install dist/*.whl
-          python -c "import pyfory; print(pyfory.__version__)"
-      - name: Upload artifacts
+          INSTALLED_VERSION=$(python -c "import pyfory; print(pyfory.__version__)")
+          echo "Installed version: $INSTALLED_VERSION"
+          # Verify version matches the tag (strip leading 'v' if present)
+          EXPECTED_VERSION="${{ github.ref_name }}"
+          EXPECTED_VERSION=${EXPECTED_VERSION#v}
```
chaokunyang, line 59: this version seems not right either, maybe need same logic to replace `-alpha/beta/rc`?

ci/deploy.sh
chaokunyang: It's right

ci/deploy.sh
chaokunyang: Maybe like:
```bash
parse_py_version() {
  local version="$1"
  if [ -z "$version" ]; then
    # Get the latest tag from the current Git repository
    version=$(git describe --tags --abbrev=0)
  fi
  # Check if the tag starts with 'v' and strip it
  if [[ $version == v* ]]; then
    version="${version:1}"
  fi
  version="${version//-alpha/a}"
  version="${version//-beta/b}"
  version="${version//-rc/rc}"
  echo "$version"
}

bump_py_version() {
  local version
  version=$(parse_py_version "$1")
  python "$ROOT/ci/release.py" bump_version -l python -version "$version"
}
```

chaokunyang: then you can use `parse_py_version` in anypalce where you verify version

ci/tasks/python_container_build_script.sh
```diff
@@ -0,0 +1,86 @@
...
+# Function to verify the installed version against expected version
+verify_version() {
+    local installed_version=$1
+
+    echo "Installed version: $installed_version"
+
+    # Check if GITHUB_REF_NAME is available and use it for verification
+    if [ -n "$GITHUB_REF_NAME" ]; then
+        # Strip leading 'v' if present
+        local expected_version="${GITHUB_REF_NAME#v}"
```
chaokunyang: Do we need to invoke `parse_py_version` here?

chaokunyang: And maybe we can create another PR with name `v0.12.1-rc1` just for trigger this logic to ensure everything just works.

esafak: do you expect it to match `pyfory.__version__`?

chaokunyang: Yes, it should be the same version

chaokunyang (approved):

chaokunyang:
@esafak 
https://github.com/apache/fory/actions/runs/17257151658/job/48971290720
image

wheel version on windows still wrong, seems version bump is not executed:

image

chaokunyang:
Let me add `shell: bash` and try it again

esafak:
I don't think that's it; it ran. It did not pick up the version from setup.cfg. Do you know why it did not pick up the bumped version in the first place?

chaokunyang:
I set `bash` nad the build succed for windows: https://github.com/apache/fory/actions/runs/17257282611/job/48971637490

image

image


But linux wheel build failed:
https://github.com/apache/fory/actions/runs/17257282639
image

PR #2534, opened 2025-08-27
fix(python): fix bump_wheel version

chaokunyang:

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -58,6 +58,7 @@ parse_py_version() {
   version="${version//-alpha/a}"
   version="${version//-beta/b}"
   version="${version//-rc/rc}"
+  version="${version//-//}"
   echo "$version"
 }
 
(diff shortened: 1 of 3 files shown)


chaokunyang:
@esafak

ci/deploy.sh
```diff
@@ -58,6 +58,7 @@ parse_py_version() {
   version="${version//-alpha/a}"
   version="${version//-beta/b}"
   version="${version//-rc/rc}"
+  version="${version//-//}"
```
esafak, line 61: I think -rc becomes redundant.

esafak (commented):
good idea with bash

ci/deploy.sh
chaokunyang: It's a convention to name release candicate as `v$major}.${minor}.${fix}-rc{number}`

esafak: I meant the next line makes this one redundant.

chaokunyang: Sometimes we have tag like `v0.12.1-a1`. Such patterns are not handled by above lines

PR #2540, opened 2025-08-27
fix(java): row format buffer recycling leaves offset and size for null values

stevenschlansker:
when reusing buffers during encoding, if a field changes from not-null to null, the offset and size from previous encode operations will remain in the output buffer

this causes hash mismatch for the same input object depending on previous serialized objects

### DIFF
--- a/java/fory-format/src/main/java/org/apache/fory/format/row/binary/writer/BinaryWriter.java
+++ b/java/fory-format/src/main/java/org/apache/fory/format/row/binary/writer/BinaryWriter.java
@@ -121,10 +121,10 @@ public abstract class BinaryWriter {
   }
 
   /**
-   * Since writer is used for one-pass writer, same field won't be writer twice. There is no need to
-   * put zero into the corresponding field when set null.
+   * Writer might recycle buffers, so implementations should clear data from a previous not-null
+   * write when setting null to avoid information leaks.
    */
-  public final void setNullAt(int ordinal) {
+  public void setNullAt(int ordinal) {
     BitUtils.set(buffer, startIndex + bytesBeforeBitMap, ordinal);
   }
 
--- a/java/fory-format/src/main/java/org/apache/fory/format/row/binary/writer/BinaryRowWriter.java
+++ b/java/fory-format/src/main/java/org/apache/fory/format/row/binary/writer/BinaryRowWriter.java
@@ -128,6 +128,12 @@ public class BinaryRowWriter extends BinaryWriter {
     writeDecimal(ordinal, value, (ArrowType.Decimal) schema.getFields().get(ordinal).getType());
   }
 
+  @Override
+  public void setNullAt(int ordinal) {
+    super.setNullAt(ordinal);
+    write(ordinal, 0L);
+  }
+
   public BinaryRow getRow() {
     BinaryRow row = new BinaryRow(schema);
     int size = size();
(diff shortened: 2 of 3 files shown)


java/fory-format/src/main/java/org/apache/fory/format/row/binary/writer/BinaryWriter.java
```diff
@@ -121,10 +121,10 @@ protected final void zeroOutPaddingBytes(int numBytes) {
   }
 
   /**
-   * Since writer is used for one-pass writer, same field won't be writer twice. There is no need to
-   * put zero into the corresponding field when set null.
+   * Writer might recycle buffers, so implementations should clear data from 
+   * a previous not-null write when setting null to avoid information leaks.
    */
-  public final void setNullAt(int ordinal) {
+  public void setNullAt(int ordinal) {
```
chaokunyang, line 127: This will make it as a polymorphic method, which introduce call method overhead.

The encode should reset the buffer, maybe something went wrong there

stevenschlansker: As far as I can tell, the `RowEncoder.toRow(T obj)` method will reset the buffer, but `byte[] encode(T obj)` method does not reset the buffer at all, but recycles the buffer without clearing.

Is there a better way to fix this problem?

stevenschlansker: @chaokunyang , when you get some time, could you please recommend a fix you like?

I think the most straightforward fix is to change all encode methods like `byte[] encode(T)` to also use a new buffer every time, like `toRow(T)` does. This is nice because then it is not possible to incorrectly leak information, but that could be a performance penalty to reallocate the buffer.

Alternately, we can change the `setNullAt` contract on `BinaryRowWriter` to also require `setOffsetAndSize(offset, 0, 0)` from the caller (and update code generator to do this) but this is a little suspect because then the contract for `BinaryWriter.setNullAt` is different depending on the specific subtype (row vs array vs map) writer you use.

Or, do you have another idea to fix this? Thanks!

stevenschlansker: Also, I did a little digging with compiler directives to print inlining decisions.
It looks like despite the method theoretically being polymorphic, in practice, the JIT is able to inline it anyway since there are not any actual polymorphic dispatches in the code. (At least, not in my test cases) - row encoder always calls row writer, array encoder calls array writer, etc

Example:
```
2025-09-03 01:51:44 INFO  JaninoUtils:121 [main] - Compile [RowEncoderTest_BarRowCodec149387948_1353170030] take 53 ms
                              @ 46   org.apache.fory.format.row.binary.writer.BinaryRowWriter::setNullAt (13 bytes)   inline (hot)
                                @ 2   org.apache.fory.format.row.binary.writer.BinaryWriter::setNullAt (18 bytes)   inline (hot)
```

So maybe removing `final` is not as harmful as you think? I do appreciate the desire for maximum performance of course, so let me know what you think is the best fix.

chaokunyang: I will take a look today, and share my findings to you

chaokunyang: Hi @stevenschlansker , I take a look today. And I found althoug the result bytes of `encoder.encode(bar)` are different, both can be deserialized correctly.

This is actually a performance optimization trick, since we called `setNullAt`, the `offsetAndSize` won't be accessed, so we skipped zero it out.

We can't invoke zero out on BianryArrayWriter since for it use less bytes for numeric array with element size less than 8.

Your fix is right.

chaokunyang (approved):

java/fory-format/src/main/java/org/apache/fory/format/row/binary/writer/BinaryWriter.java
stevenschlansker: Thank you!

PR #2545, opened 2025-08-28
feat(Rust): Support skipping unneeded field bytes when deserializing in compatible mode

urlyy:
## What does this PR do?
Support skipping unneeded field bytes when deserializing in compatible mode.

## Related issues
#2531

### DIFF
--- a/rust/fory-core/src/serializer/nonexistent.rs
+++ b/rust/fory-core/src/serializer/nonexistent.rs
@@ -1,0 +1,108 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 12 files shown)


urlyy:
Oh, maybe this filename isn’t quite appropriate? `nonexistent.rs` should probably be `skip.rs` instead. Since it skips not only when the type or field is **nonexistent** but also when it doesn’t **match**.

rust/fory-derive/src/object/read.rs
```diff
@@ -113,14 +113,7 @@ fn deserialize_compatible(fields: &[&Field]) -> TokenStream {
                     _ => {
                         // skip bytes
                         println!("no need to deserialize {:?}:{:?}", _field.field_name.as_str(), _field.field_type);
```
urlyy, line 115: Maybe need to be replaced as `println!("skip {:?}:{:?}", _field.field_name.as_str(), _field.field_type);`. Or just remove this. I just want to print a warning or log to user.

rust/fory-core/src/serializer/nonexistent.rs
```diff
@@ -0,0 +1,108 @@
...
+                    Ok(())
+                } else if ref_flag == (RefFlag::Null as i8) {
+                    Err(anyhow!("Try to deserialize non-option type to null"))?
+                } else if ref_flag == (RefFlag::Ref as i8) {
+                    Err(Error::Ref)
+                } else {
+                    Err(anyhow!("Unknown ref flag, value:{ref_flag}"))?
+                }
+            } else {
+                unreachable!()
```
urlyy, line 89: Option or Enum or some other will enter here. The implementation and unit tests will be added in other PRs.

urlyy:
@chaokunyang Please review.

PR #2551, opened 2025-08-28
fix(java): array encoder seems to waste 8 bytes in encode(T)

stevenschlansker:
the bytes at the end are always 0. are these bytes necessary?


chaokunyang:
It‘s unnecessary

chaokunyang (approved):

Issue #2192, opened 2025-04-27
[OSPP][Go] Implement type forward/backward compatibility for Fury Go


chaokunyang:
### Feature Request

1) Background:
Fury is a high-performance serialization library that supports cross-language binary protocols. Currently, Go-based applications are gradually increasing. Go language is widely used for its performance, while also facing serialization challenges that need to be compatible with different versions and type structures. For Go language serialization, ensuring forward and backward compatibility of types is crucial for long-term project evolution and multi-version support.
(2) Existing Work:
Currently, Fury has implemented efficient serialization and deserialization in languages such as Java/Python/NodeJS, but related work in Rust is still in its early stages, only supporting serialization and deserialization when object type schemas are strictly consistent. As projects expand and complexity grows, the serialization scheme needs to be more complete and efficient to support version evolution.
(3) Existing Shortcomings:
The current Fury Rust serialization mechanism does not fully support forward and backward compatibility of structural types. When types are updated, there is a lack of adaptation mechanisms to handle added or removed fields to ensure the stability and consistency of serialized data. At the same time, the efficient sharing mechanism of metadata has not yet formed best practices.
(4) Desired Improvements:
We hope to implement a type-compatible mechanism based on metadata sharing, enabling Fury Rust serialization to support dynamic type updates while ensuring compatibility between different versions. Through efficient metadata sharing methods, implement intelligent serialization handling for added, modified, and removed fields to reduce the development burden and operational risks brought by version changes.
(5) Final Project Implementation Goals:
The ultimate goal is to implement a serialization framework for Fury Rust that supports forward and backward compatibility of types. This framework will utilize the metadata sharing mechanism of the Fury binary serialization protocol. The implementation can reference Fury Java/NodeJS and other language implementations, ultimately automatically adapting to changes in type structures, ensuring that data between different versions can be correctly serialized and deserialized, and enhancing the breadth and depth of Fury's application in the Go ecosystem.

LouisLou2:
@chaokunyang, thanks for opening this! I noticed a small inconsistency in the description: the title and context say “Fury Go,” but several places in the body refer to “Fury Rust.” We may need to straighten this out.

chaokunyang:
> [@chaokunyang](https://github.com/chaokunyang), I noticed a small inconsistency in the description: the title and context say “Fury Go,” but several places in the body refer to “Fury Rust.” We may need to straighten this out.

Updated

PR #2554, opened 2025-08-31
feat(go): add type meta encoding for meta share

junjiexh:
## Why?

type forward/backward compatible serialization is critical for online service which different service update their data schema and deploy at different time.

encoding format can address this : [type def encoding spec](https://fory.apache.org/docs/specification/fory_xlang_serialization_spec#type-def)

## What does this PR do?
This PR implements type definition encoding and decoding that conforms to the specification. This functionality can be integrated into the main serialization flow to enable metadata sharing mode.

TODOs:
- Add encoding support for collection types (map, slice, set, etc.)
- Integrate into the main flow to enable metadata sharing
- Add support for compressed type definitions

## Related issues

#2192 

- [x] Does this PR introduce any public API change? **no**
- [x] Does this PR introduce any binary protocol compatibility change? **no**

### DIFF
--- a/go/fory/type.go
+++ b/go/fory/type.go
@@ -19,14 +19,13 @@ package fory
 
 import (
 	"fmt"
+	"github.com/apache/fory/go/fory/meta"
 	"hash/fnv"
 	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
 	"time"
-
-	"github.com/apache/fory/go/fory/meta"
 )
 
 type TypeId = int16


go/fory/type.go
```diff
@@ -19,13 +19,14 @@ package fory
...
 	"fmt"
-	"github.com/apache/fory/go/fory/meta"
 	"hash/fnv"
 	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/apache/fory/go/fory/meta"
```
pandalee99: I prefer not to make changes to this style

junjiexh: will fix that

Issue #2559, opened 2025-09-01
[Question] [Java] Why are serializers recreated?


theigl:
### Question

After running in production for several hours and no new Fory instance being created for several hours because usage is relatively uniform, I still see high load from new serializers being initialized. Here is an example:

<img width="942" height="485" alt="Image" src="https://github.com/user-attachments/assets/8ed68986-cf34-43e2-899c-a3c758200f9b" />

This class has been serialized tens of thousands of times already and almost certainly by all available instances of Fory. 

Any idea how this could happen? Is the `ClassResolver` reset at some point? Could registrations be evicted by GC? We are serializing thousands of different classes. Is there some limit on class registration?

My config looks like this:

```java
final ForyBuilder builder = Fory.builder();
builder.withName(name);
builder.withClassLoader(Fory.class.getClassLoader());
builder.withBufferSizeLimitBytes(BUFFER_SIZE_LIMIT_BYTES);
builder.withCompatibleMode(CompatibleMode.SCHEMA_CONSISTENT);
builder.withLanguage(Language.JAVA);
builder.requireClassRegistration(false);
builder.withRefTracking(true);
builder.withCodegen(true);
builder.registerGuavaTypes(true);
final var fory = (AbstractThreadSafeFory) builder.buildThreadSafeForyPool(1, poolMaxSize, 5, TimeUnit.MINUTES);
```

chaokunyang:
As you can see from ObjectArraySerializer:
```java
        Fory fory = this.fory;
        ClassResolver classResolver = fory.getClassResolver();
        ClassInfo classInfo = null;
        Class<?> elemClass = null;
        for (T t : arr) {
          if (!refResolver.writeRefOrNull(buffer, t)) {
            Class<?> clz = t.getClass();
            if (clz != elemClass) {
              elemClass = clz;
              classInfo = classResolver.getClassInfo(clz);
            }
            fory.writeNonRef(buffer, t, classInfo);
          }
        }
```

If the elements in your object array has different types, `getClassInfo` will be invoked. Maybe there are some objects of new types in your object array

theigl:
I don't think that this is the issue. Now, 14 hours and millions of serializations later, no new Fory instance has been created and I'm still seeing new serializers constantly being initialized. And it doesn't only affect  the `ObjectArraySerializer`:

<img width="878" height="271" alt="Image" src="https://github.com/user-attachments/assets/d92ee5b6-45e9-40e0-9a7c-a5b82e1e5b89" />

[...]

PR #2573, opened 2025-09-03
fix(java): Skip calculation of classVersionHash if checkClassVersion is false

theigl:
## Why?

Calculating the `classVersionHash` can be quite expensive and can currently have side-effects. This PR implements a simple optimization that skips the calculation entirely if `fory.checkClassVersion() == false`.

See https://github.com/apache/fory/issues/2559


theigl:
Thanks @chaokunyang! Any chance you could include this in the next minor release?

chaokunyang:
> Thanks @chaokunyang! Any chance you could include this in the next minor release?

This will be released in 0.12.2.

PR #2576, opened 2025-09-03
feat(java): bean encoder implemented interfaces honor `@Ignore`

stevenschlansker:
this allows you to mark methods as not properties, for example if you have a computed value with a `default` method


chaokunyang (approved):
LGTM

PR #2585, opened 2025-09-07
feat(Rust): support basic type se/de aligned with java

urlyy:
## What does this PR do?
1. Updated the legacy `read`/`write` methods to align with Java.
2. Started adding some `register_by_name` functionality; not finished yet, will continue in the next PR.
3. Added a new test file to facilitate testing across different languages conveniently.
4. Basic types, primitive arrays, string, list, set, and map are now aligned with Java. Handling of compress algorithm for String and header for map still needs improvement.

## Related issues
#2539 

## TODOs:
1. Enum derive.
2. Complete `register_by_name`.
3. Align with Java metashare, test on struct..
4. Implement serialization for `Box<dyn Any>` (may be challenging).

```rust
# now can 
let reader = Reader::new(bytes.as_slice());
let fory = Fory::default().mode(Compatible).xlang(true);
let mut context = ReadContext::new(&fory, reader);
let a = fory.deserialize_with_context(&mut context);
let b = fory.deserialize_with_context(&mut context);
let c = fory.deserialize_with_context(&mut context);
```
And we should add `scoped_meta_context` in future.

### DIFF
--- a/rust/fory-core/src/buffer.rs
+++ b/rust/fory-core/src/buffer.rs
@@ -262,32 +262,21 @@ impl Writer {
         }
     }
 
-    pub fn latin1_string(&mut self, s: &str) -> usize {
-        let mut count = 0;
+    // TODO: simd
+    pub fn latin1_string(&mut self, s: &str) {
         for c in s.chars() {
             let b = c as u32;
             assert!(b <= 0xFF, "Non-Latin1 character found");
             self.u8(b as u8);
-            count += 1;
         }
-        count
     }
 
-    pub fn utf8_string(&mut self, s: &str) -> usize {
+    // TODO: simd
+    pub fn utf8_string(&mut self, s: &str) {
         let bytes = s.as_bytes();
         for &b in bytes {
             self.u8(b);
         }
-        bytes.len()
-    }
-
-    pub fn utf16_string(&mut self, s: &str) -> usize {
-        let mut count = 0;
-        for u in s.encode_utf16() {
-            self.u16(u);
-            count += 1;
-        }
-        count * 2
     }
 
     pub fn bytes(&mut self, v: &[u8]) -> usize {
--- a/rust/fory-core/src/serializer/string.rs
+++ b/rust/fory-core/src/serializer/string.rs
@@ -15,10 +15,9 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::buffer::Writer;
 use crate::error::Error;
 use crate::fory::Fory;
-use crate::meta::is_latin;
+use crate::meta::get_latin1_length;
 use crate::resolver::context::ReadContext;
 use crate::resolver::context::WriteContext;
 use crate::serializer::Serializer;
@@ -37,18 +36,17 @@ impl Serializer for String {
     }
 
     fn write(&self, context: &mut WriteContext) {
-        let mut buf = Writer::default();
-        let len;
-        let bitor;
-        if is_latin(self.as_str()) {
-            len = buf.latin1_string(self);
-            bitor = (len as u64) << 2 | StrEncoding::Latin1 as u64;
+        let mut len = get_latin1_length(self);
+        if len >= 0 {
+            let bitor = (len as u64) << 2 | StrEncoding::Latin1 as u64;
+            context.writer.var_uint36_small(bitor);
+            context.writer.latin1_string(self);
         } else {
-            len = buf.utf8_string(self);
-            bitor = (len as u64) << 2 | StrEncoding::Utf8 as u64;
+            len = self.len() as i32;
+            let bitor = (len as u64) << 2 | StrEncoding::Utf8 as u64;
+            context.writer.var_uint36_small(bitor);
+            context.writer.utf8_string(self);
         }
-        context.writer.var_uint36_small(bitor);
-        context.writer.bytes(buf.dump().as_slice());
     }
 
     fn read(context: &mut ReadContext) -> Result<Self, Error> {
--- a/rust/fory-core/src/meta/mod.rs
+++ b/rust/fory-core/src/meta/mod.rs
@@ -22,5 +22,5 @@ mod type_meta;
 pub use meta_string::{
     Encoding, MetaString, MetaStringBytes, MetaStringDecoder, MetaStringEncoder,
 };
-pub use string_util::{is_latin, murmurhash3_x64_128};
+pub use string_util::{get_latin1_length, is_latin, murmurhash3_x64_128};
 pub use type_meta::{FieldInfo, FieldType, NullableFieldType, TypeMeta, TypeMetaLayer};
--- a/rust/fory-core/src/meta/string_util.rs
+++ b/rust/fory-core/src/meta/string_util.rs
@@ -114,7 +114,7 @@ unsafe fn is_latin_neon(s: &str) -> bool {
 }
 
 fn is_latin_standard(s: &str) -> bool {
-    s.bytes().all(|b| b.is_ascii())
+    s.chars().all(|c| c as u32 <= 0xFF)
 }
 
 pub fn is_latin(s: &str) -> bool {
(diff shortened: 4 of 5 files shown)


java/fory-core/src/main/java/org/apache/fory/resolver/XtypeResolver.java
```diff
@@ -563,6 +563,7 @@ public void writeClassInfo(MemoryBuffer buffer, ClassInfo classInfo) {
       case Types.NAMED_STRUCT:
       case Types.NAMED_COMPATIBLE_STRUCT:
       case Types.NAMED_EXT:
+      case Types.STRUCT:
```
chaokunyang: Please merge main branch, and revert this line

java/fory-core/src/test/java/org/apache/fory/WIPCrossLanguageTest.java
```diff
@@ -0,0 +1,507 @@
...
+import org.apache.fory.serializer.StringSerializer;
+import org.apache.fory.test.TestUtils;
+import org.apache.fory.util.MurmurHash3;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/** Tests in this class need fory python/rust installed. */
+@Test
+public class WIPCrossLanguageTest extends ForyTestBase {
```
chaokunyang: Rename it to `RustXlangTest`, and skip it by default. We don't want every contributor isntall rust for developing java.

You can control whether enable it by a env veraible. And this test should only be run in rust ci. You can update rust ci by install fory java, and execute command `cd java/fory-core && mvn test -Dtest=org.apache.fory.RustXlangTest`

java/fory-core/src/test/java/org/apache/fory/RustXlangTest.java
```diff
@@ -0,0 +1,507 @@
...
+          // Exact match test name rather than prefix matching.
+          "--exact");
+  private static final List<String> pyBaseCommand =
+      Arrays.asList(PYTHON_EXECUTABLE, "-m", PYTHON_MODULE, "<TESTCASE>", "<DATA_FILE>");
+
+  private static final int RUST_TESTCASE_INDEX = 4;
+
+  @BeforeClass
+  public void isPyforyInstalled() {
+    // TestUtils.verifyPyforyInstalled();
```
chaokunyang: We can configure a env variable such as `FORY_RUST_JAVA_CI`, and in `isRustJavaCIEnabled` method, we check the env variable and skip this whole `RustXlangTest` here if env variable not set to 1

java/fory-core/src/test/java/org/apache/fory/RustXlangTest.java
```diff
@@ -0,0 +1,507 @@
...
+          "--ignored",
+          // Exact match test name rather than prefix matching.
+          "--exact");
+  private static final List<String> pyBaseCommand =
+      Arrays.asList(PYTHON_EXECUTABLE, "-m", PYTHON_MODULE, "<TESTCASE>", "<DATA_FILE>");
+
+  private static final int RUST_TESTCASE_INDEX = 4;
+
+  @BeforeClass
+  public void isPyforyInstalled() {
```
chaokunyang:
```suggestion
  public void isRustJavaCIEnabled() {
```

java/fory-core/src/test/java/org/apache/fory/RustXlangTest.java
```diff
@@ -0,0 +1,507 @@
...
+    // TestUtils.verifyPyforyInstalled();
+  }
+
+  @Test(enabled = false)
```
chaokunyang: remove `enabled` flag, use `isRustJavaCIEnabled` instead.

java/fory-core/src/test/java/org/apache/fory/RustXlangTest.java
chaokunyang: Please also update rust ci in `.github/workflows/ci.yaml` to install fory java too and run test command `mvn test -Dtest=org.apache.fory.RustXlangTest`

rust/fory-core/src/serializer/string.rs
```diff
@@ -23,19 +24,104 @@ use crate::serializer::Serializer;
...
+
 impl Serializer for String {
     fn reserved_space() -> usize {
         mem::size_of::<i32>()
     }
 
     fn write(&self, context: &mut WriteContext) {
-        context.writer.var_int32(self.len() as i32);
-        context.writer.bytes(self.as_bytes());
+        let encoding = best_coder(self);
```
chaokunyang: For write, we could keep using utf8 instead, this will minimize the string encoding cost, and other languages support decode utf8

ci/tasks/rust.py
```diff
@@ -76,4 +78,12 @@ def run():
...
         common.exec_cmd(cmd)
-    logging.info("Executing fory rust tests succeeds")
+
+    logging.info("Executing Rust <-> Java11 cross-language tests")
+    os.environ["FORY_RUST_JAVA_CI"] = "1"
+    java_dir = os.path.join("..", "java")
+    subprocess.check_call(["mvn", "clean", "install", "-DskipTests"], cwd=java_dir)
+    subprocess.check_call(
+        ["mvn", "test", "-Dtest=org.apache.fory.RustXlangTest"],
+        cwd=java_dir,
```
chaokunyang:
```suggestion
        cwd=os.path.join(java, "fory-core"),
```

ci/tasks/rust.py
```diff
@@ -76,4 +78,12 @@ def run():
...
         common.exec_cmd(cmd)
-    logging.info("Executing fory rust tests succeeds")
+
+    logging.info("Executing Rust <-> Java11 cross-language tests")
+    os.environ["FORY_RUST_JAVA_CI"] = "1"
+    java_dir = os.path.join("..", "java")
+    subprocess.check_call(["mvn", "clean", "install", "-DskipTests"], cwd=java_dir)
+    subprocess.check_call(
+        ["mvn", "test", "-Dtest=org.apache.fory.RustXlangTest"],
+        cwd=os.path.join(java, "fory-core"),
```
chaokunyang:
```suggestion
        cwd=os.path.join(java_dir, "fory-core"),
```

chaokunyang:
@urlyy 
String serialization in rust needs to support latin1 encoding. You can take following code as an example for simd latin1 check:

```rust

/// Checks if a UTF-8 string can be losslessly encoded as Latin-1 using SIMD if available.
///
/// This is true if all characters in the string have a Unicode codepoint <= 255.
/// This translates to a byte-level check: the string must not contain any byte >= 0xC4.
pub fn can_be_latin1(s: &str) -> bool {
    let bytes = s.as_bytes();

    // Runtime feature detection to select the best implementation.
    // The functions are guarded by `#[target_feature]`, so the compiler
    // generates optimized code for each case. The function pointers are resolved
    // at runtime on the first call.
    #[cfg(target_arch = "x86_64")]
    {
        if is_x86_feature_detected!("avx2") {
            return unsafe { can_be_latin1_avx2(bytes) };
        }
        if is_x86_feature_detected!("sse2") {
            return unsafe { can_be_latin1_sse2(bytes) };
        }
    }

    // Fallback for non-x86_64 architectures or if no SIMD features are available.
    can_be_latin1_scalar(bytes)
}

/// Scalar fallback implementation. Checks byte by byte.
#[inline]
fn can_be_latin1_scalar(bytes: &[u8]) -> bool {
    // A simple iterator-based check is clean and often optimized well by the compiler.
    !bytes.iter().any(|&b| b >= 0xC4)
}

/// Implementation using SSE2 intrinsics (16-byte vectors).
#[cfg(target_arch = "x86_64")]
#[target_feature(enable = "sse2")]
unsafe fn can_be_latin1_sse2(bytes: &[u8]) -> bool {
    const CHUNK_SIZE: usize = 16;
    let limit = _mm_set1_epi8(0xC3 as i8); // We check for values > 0xC3

    let mut i = 0;
    while i + CHUNK_SIZE <= bytes.len() {
        // Load 16 bytes of data. `loadu` handles unaligned memory.
        let chunk = _mm_loadu_si128(bytes.as_ptr().add(i) as *const _);

        // This is a common trick for unsigned comparison with signed-only intrinsics.
        // We want to check `byte >= 0xC4`. This is equivalent to `byte > 0xC3`.
        // The `_mm_cmpgt_epi8` instruction performs a signed comparison (a > b).
        // By adding -128 (or XORing with 0x80) to both operands, we can map the
        // unsigned range [0, 255] to the signed range [-128, 127] and perform a
        // valid comparison.
        // `byte > 0xC3` becomes `(byte - 128) > (0xC3 - 128)`.
        // `0xC3 - 128 = 195 - 128 = 67`. The `limit` vector holds `0xC3` because
        // we're comparing `chunk > limit`, and `_mm_cmpgt_epi8` works on signed i8.
        // The values from 0xC4 to 0xFF will correctly be "greater than" 0xC3.
        let comparison = _mm_cmpgt_epi8(chunk, limit);

        // Create a bitmask from the most significant bit of each byte in the result.
        // If any byte in `chunk` was > 0xC3, the corresponding byte in `comparison`
        // will be all 1s (0xFF), and its MSB will be 1.
        // `movemask` will be non-zero if any invalid byte was found.
        if _mm_movemask_epi8(comparison) != 0 {
            return false;
        }

        i += CHUNK_SIZE;
    }

    // Handle the remainder
    can_be_latin1_scalar(&bytes[i..])
}

/// Implementation using AVX2 intrinsics (32-byte vectors).
#[cfg(target_arch = "x86_64")]
#[target_feature(enable = "avx2")]
unsafe fn can_be_latin1_avx2(bytes: &[u8]) -> bool {
    const CHUNK_SIZE: usize = 32;
    // We want to check for bytes >= 0xC4, which is equivalent to > 0xC3.
    // The vector is filled with 0xC3.
    let limit = _mm256_set1_epi8(0xC3 as i8);

    let mut i = 0;
    while i + CHUNK_SIZE <= bytes.len() {
        // Load 32 bytes of data.
        let chunk = _mm256_loadu_si256(bytes.as_ptr().add(i) as *const _);
        
        // Perform a signed "greater than" comparison.
        let comparison = _mm256_cmpgt_epi8(chunk, limit);

        // Create a 32-bit mask. If any byte was > 0xC3, the mask will be non-zero.
        if _mm256_movemask_epi8(comparison) != 0 {
            return false;
        }

        i += CHUNK_SIZE;
    }

    // Handle the remainder using the SSE2 or scalar implementation
    // to avoid duplicating the tail-handling logic.
    can_be_latin1_sse2(&bytes[i..])
}

```

rust/fory-core/src/buffer.rs
```diff
@@ -88,70 +88,209 @@ impl Writer {
...
+        }
+    }
+
+    pub fn latin1_string(&mut self, s: &str) -> usize {
+        for c in s.chars() {
+            let b = c as u32;
+            assert!(b <= 0xFF, "Non-Latin1 character found");
+            self.u8(b as u8);
+        }
+        s.chars().count()
```
chaokunyang: THis will parse anditerate the string twice, `s.chars()` return an iterator. we'd bettter use a counter to count the length

rust/fory-core/src/buffer.rs
```diff
@@ -88,70 +88,209 @@ impl Writer {
...
+    pub fn utf8_string(&mut self, s: &str) -> usize {
+        let bytes = s.as_bytes();
+        for &b in bytes {
+            self.u8(b);
+        }
+        bytes.len()
+    }
+
+    pub fn utf16_string(&mut self, s: &str) -> usize {
+        let units: Vec<u16> = s.encode_utf16().collect();
```
chaokunyang: the collection will introduce extra memroy allocation, we'd better keep using iterator and use a coutenr teh count written bytes length

rust/fory-core/src/serializer/string.rs
```diff
@@ -15,27 +15,60 @@
...
         mem::size_of::<i32>()
     }
 
     fn write(&self, context: &mut WriteContext) {
-        context.writer.var_int32(self.len() as i32);
-        context.writer.bytes(self.as_bytes());
+        let mut buf = Writer::default();
+        let len;
+        let bitor;
+        if is_latin(self.as_str()) {
```
chaokunyang: Maybe not in this PR, we could implement a get_latin1_length, which return -1 if not latin1. And then write length and encoding first, then we can just write string byte by byte directly into the writer

PR #2594, opened 2025-09-09
refactor(go): rename FieldInfo to FieldDef to avoide name collision

junjiexh:
## Why?

Previous PR #2554 introduced a type named "FieldInfo", which conflicts with the existing "fieldInfo" in struct.go. This naming collision creates confusion, particularly when interacting with both types. This PR has been split out from the original for easier review.

## What does this PR do?

rename FieldInfo to FieldDef to avoide name collision

## Related issues

#2192 

- [ x ] Does this PR introduce any public API change? no.
- [ x ] Does this PR introduce any binary protocol compatibility change? no.


pandalee99 (approved):
LGTM

Issue #1645, opened 2024-05-22
Mac install using “conda install -c conda-forge pyarrow” command is completed, how to set pyarrow of environment variables? 


69567272:
Error in fail: Python Configuration Error: Problem getting pyarrow include path.
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'pyarrow'
Is the Python binary path set up right? (See ./configure or python3.) Is distutils installed?
ERROR: no such package '@@local_config_pyarrow//': Python Configuration Error: Problem getting pyarrow include path.
Traceback (most recent call last):

How solve  this compilation error?

urlyy:
Occur same question. But I find why:
My complete log：
```
conda activate fury
pip install -v -e .
Using pip 24.2 from /Users/urlyy/environment/miniconda/envs/fury/lib/python3.8/site-packages/pip (python 3.8)
Obtaining file:///Users/urlyy/workspace/fury/python
  Running command python setup.py egg_info
  Loading:
  Loading:
  Loading: 0 packages loaded
  Analyzing: target //:cp_fury_so (0 packages loaded, 0 targets configured)
  DEBUG: /Users/urlyy/workspace/fury/bazel/arrow/pyarrow_configure.bzl:169:10: python3 -c 'import pyarrow;print(pyarrow.get_include())'
  INFO: Repository local_config_pyarrow instantiated at:
    /Users/urlyy/workspace/fury/WORKSPACE:38:18: in <toplevel>
  Repository rule pyarrow_configure defined at:
    /Users/urlyy/workspace/fury/bazel/arrow/pyarrow_configure.bzl:267:36: in <toplevel>
  ERROR: An error occurred during the fetch of repository 'local_config_pyarrow':
     Traceback (most recent call last):
        File "/Users/urlyy/workspace/fury/bazel/arrow/pyarrow_configure.bzl", line 214, column 44, in _pyarrow_pip_impl
                arrow_header_dir = _get_pyarrow_include(repository_ctx, python_bin)
        File "/Users/urlyy/workspace/fury/bazel/arrow/pyarrow_configure.bzl", line 170, column 22, in _get_pyarrow_include
                result = _execute(
        File "/Users/urlyy/workspace/fury/bazel/arrow/pyarrow_configure.bzl", line 41, column 14, in _execute
                _fail("\n".join([
[...]

PR #2596, opened 2025-09-09
fix build bug

pandalee99:
## Why?

When I repeatedly perform `pip install -v e.`, I encounter the problem of not being able to find pyarrow. This issue is very similar to #1645. However, as long as I perform `bazel clean --expunge`, the problem can be solved. I think this might be dependency pollution caused by the virtual environment.


chaokunyang:
@pandalee99 This PR needs be reverted, it cause pip install executed `bazel build` twice, which makes the build time two times. You can see https://github.com/apache/fory/actions/runs/17589426881/job/49967207586?pr=2593

PR #2605, opened 2025-09-11
feat(java): Add getClassResolver to BaseFory

esafak:
## Why?

So you can serialize private third-party classes like `kotlin.time.InstantSerializer` more easily.

## What does this PR do?

* Add `getClassResolver()` method to the `BaseFory` interface.
* Implement `getClassResolver()` in `AbstractThreadSafeFory` to return the
  class resolver.
* Import `ClassResolver` in both modified files.

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/AbstractThreadSafeFory.java
+++ b/java/fory-core/src/main/java/org/apache/fory/AbstractThreadSafeFory.java
@@ -23,6 +23,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import org.apache.fory.annotation.Internal;
 import org.apache.fory.resolver.ClassChecker;
+import org.apache.fory.resolver.ClassResolver;
 import org.apache.fory.serializer.Serializer;
 import org.apache.fory.serializer.SerializerFactory;
 
@@ -89,6 +90,11 @@ public abstract class AbstractThreadSafeFory implements ThreadSafeFory {
           fory.ensureSerializersCompiled();
           return null;
         });
+  }
+
+  @Override
+  public ClassResolver getClassResolver() {
+    return execute(fory -> fory.getClassResolver());
   }
 
   @Internal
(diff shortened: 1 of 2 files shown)


java/fory-core/src/main/java/org/apache/fory/AbstractThreadSafeFory.java
```diff
@@ -91,6 +92,11 @@ public void ensureSerializersCompiled() {
         });
   }
 
+  @Override
+  public ClassResolver getClassResolver() {
```
chaokunyang, line 96: This is not safe, please use registerCallback

java/fory-core/src/main/java/org/apache/fory/BaseFory.java
```diff
@@ -222,4 +223,7 @@ public interface BaseFory {
 
   /** Deep copy the <code>obj</code>. */
   <T> T copy(T obj);
+
+  /** Return the class resolver used by this fory instance. */
+  ClassResolver getClassResolver();
```
chaokunyang, line 228: Do not expose this, classs resolver is not shared across all instances of fory

esafak: When you said
> You can invoke fory.getClassResolver().register(String) method for registration. Maybe we should expose this API to BaseFory class too.

What did you mean, then?

java/fory-core/src/main/java/org/apache/fory/AbstractThreadSafeFory.java
esafak: That returns void though. I only see it for setting things, not getting.

java/fory-core/src/main/java/org/apache/fory/BaseFory.java
chaokunyang: ThreadSafeFory is backed by a pool of fory instance. If you return a class resolver from ThreadSafeFory, it just return one of the class resolver to caller, and the registration only take effect on that Fory.

You should add a `register(String)` method to `BaseFory`, and in that method, you call `registerCallback` to make the registration take effects for all fory instance:

```java
public void register(String className) {
  registerCallback(f->f.getClassResolver().register(className));
}
```

PR #2607, opened 2025-09-12
feat(go): Implement compatible mode with metashare mode

junjiexh:
## Why?

Type-forward and backward-compatible serialization are crucial for online services where different services update their data schemas and deploy at different times. This ensures that a service running an older schema can still communicate with a service using a newer one, and vice versa.This PR is part of the work of it. The previous PR is #2554 , and following pr are still work in process.

## What does this PR do?

This PR implements a metashare mode for the fory-go.

Specifically, it:

Provides a user option to enable compatible mode.

If compatible mode is enabled, the serialization process not only serializes the data but also its detailed type information into a binary format.

The reading peer, if compatible mode is enabled, will first attempt to read this type information before deserializing the data. This allows it to handle data serialized with a different schema.

This implementation is heavily inspired by the existing Java and Python implementations, ensuring consistency across framework.

All progress and todos:
- [x] metadata encoding and decoding
- [x] Integrate into the main flow to enable metadata sharing
- [  ] Add support for collection types (map, slice, set, etc.)
- [  ] Add support for compressed type definitions
- [  ] Test with python

## Related issues

#2192

- [x] Does this PR introduce any public API change? Yes.
- [x] Does this PR introduce any binary protocol compatibility change? No.


go/fory/type.go
```diff
@@ -597,15 +604,10 @@ func (r *typeResolver) getTypeInfo(value reflect.Value, create bool) (TypeInfo,
...
 	if idx := strings.LastIndex(clean, "."); idx != -1 {
 		pkgPath = clean[:idx]
-	} else {
-		pkgPath = clean
-	}
-	if typ.Kind() == reflect.Ptr {
-		typeName = typ.Elem().Name()
-	} else {
-		typeName = typ.Name()
+		typeName = clean[idx+1:]
```
junjiexh, line 545: Here fix a bug that cause user specific tag not work on typeName(In the past, if user register a type called C with 'A.B', the typeName and pkgName will be 'A.C')

go/fory/fory.go
```diff
@@ -24,13 +24,53 @@ import (
...
+	return NewForyWithOptions(WithReferenceTracking(referenceTracking))
+}
+
+// NewForyWithOptions creates a Fory instance with configurable options
+func NewForyWithOptions(options ...ForyOption) *Fory {
 	fory := &Fory{
-		refResolver:       newRefResolver(referenceTracking),
-		referenceTracking: referenceTracking,
+		refResolver:       nil,
+		referenceTracking: false,
```
chaokunyang:
```suggestion
		refTracking: false,
```

chaokunyang: Rename all `reference` to `ref`

go/fory/fory_metashare_test.go
```diff
@@ -0,0 +1,181 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// Test structs for metashare testing
+
+type SimpleDataClass struct {
+	Name   string `fory:"name"`
```
chaokunyang: why do we need to add `tag` here?

go/fory/fory.go
junjiexh: Done. Is this change okay?

go/fory/fory_metashare_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	assert.False(t, fory.compatible, "Expected compatible mode to be disabled")
+	assert.Nil(t, fory.metaContext, "Expected metaContext to be nil when compatible=false")
+}
+
+func TestSimpleDataClassSerialization(t *testing.T) {
+	fory := NewForyWithOptions(WithCompatible(true))
+
+	// Register the struct
+	err := fory.RegisterTagType("SimpleDataClass", SimpleDataClass{})
```
chaokunyang, line 65: @pandalee99 This API should be removed already, and replaced by:

```go
func (r *typeResolver) registerNamedType(
	type_ reflect.Type,
	namespace string,
	typeName string
)
```

chaokunyang: @junjiexh Let's create an issue for this and update the API call in future after the API has been updated

go/fory/fory_metashare_test.go
junjiexh: Oh, sorry, we don't. This is generated by copilot, I'll make sure to double-check my work next time.

go/fory/fory_metashare_test.go
junjiexh: > @junjiexh Let's create an issue for this and update the API call in future after the API has been updated

Done

go/fory/type_def_decoder.go
```diff
@@ -61,6 +62,7 @@ func decodeTypeDef(fory *Fory, buffer *ByteBuffer) (*TypeDef, error) {
 	// Read name or type ID according to the registerByName flag
 	var typeId TypeId
 	var nsBytes, nameBytes *MetaStringBytes
+	var typ reflect.Type
```
chaokunyang: use `type_`

go/fory/fory.go
```diff
@@ -24,13 +24,53 @@ import (
 	"sync"
 )
 
-func NewFory(referenceTracking bool) *Fory {
+// ForyOption represents a configuration option for Fory instances
+type ForyOption func(*Fory)
```
chaokunyang: Maybe we don't need the Fory prefix.

And please document it that this is the functional options. Not everyone knows it. And we also need to update the readme for this

go/fory/fory.go
```diff
@@ -44,19 +84,43 @@ func NewFory(referenceTracking bool) *Fory {
...
+	}
+
+	fory.refResolver = newRefResolver(fory.refTracking)
 	return fory
 }
 
 // NewForyWithIsolatedTypes creates a Fory instance with isolated type resolver
 // for use cases that need independent type registration
-func NewForyWithIsolatedTypes(referenceTracking bool) *Fory {
+func NewForyWithIsolatedTypes(refTracking bool) *Fory {
```
chaokunyang: @thisingl I'm a little confused, why do we need this API, this looks obscure to me.

ThisingL: @chaokunyang You're right - this API doesn't make much sense anymore with the factory approach. All generated serializers are now globally registered, so "isolation" only affects manual `RegisterTypeTag()` calls.

I suggest removing `NewForyWithIsolatedTypes` entirely to simplify the API. I will include that in my new PR.

chaokunyang: Great! If your new PR comes soon, we can hold this until your new PR got merged

ThisingL: Yeah, it has done in pr#2615, and it is merged

PR #2613, opened 2025-09-15
feat(java): support codegen for xlang mode in java

chaokunyang:
## Why?
support codegen for xlang mode in java to speed uo xlang mode performance

## What does this PR do?

support codegen for xlang mode in java 

## Related issues

Closes https://github.com/apache/fory/issues/2286
Closes https://github.com/apache/fory/pull/2312

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/type/DescriptorGrouper.java
+++ b/java/fory-core/src/main/java/org/apache/fory/type/DescriptorGrouper.java
@@ -178,7 +178,6 @@ public class DescriptorGrouper {
       } else if (TypeUtils.isMap(descriptor.getRawType())) {
         mapDescriptors.add(descriptorUpdater.apply(descriptor));
       } else if (isMonomorphic.test(descriptor.getRawType())) {
-        isMonomorphic.test(descriptor.getRawType());
         finalDescriptors.add(descriptorUpdater.apply(descriptor));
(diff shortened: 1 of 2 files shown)


java/fory-core/src/main/java/org/apache/fory/type/DescriptorGrouper.java
```diff
@@ -178,6 +178,7 @@ private DescriptorGrouper(
       } else if (TypeUtils.isMap(descriptor.getRawType())) {
         mapDescriptors.add(descriptorUpdater.apply(descriptor));
       } else if (isMonomorphic.test(descriptor.getRawType())) {
+        isMonomorphic.test(descriptor.getRawType());
```
hrgdavor: I am totally new here, this duplication just looks strange. If there is a reason for, it, A comment with explanation would be great :)

hrgdavor (commented):
just getting to know the code a bit. Saw something strange.

java/fory-core/src/main/java/org/apache/fory/type/DescriptorGrouper.java
chaokunyang: I just removed it, it's added by me for debugging, thanks for catching this

theweipeng (approved):

PR #2616, opened 2025-09-16
fix(java/python): fix meta string encoding order

chaokunyang:


chaokunyang:
cc @urlyy

Issue #2658, opened 2025-09-24
[Rust] API naming refinement and performance optimization


urlyy:
At this point, aside from `tracking_ref` and the serialization of `Box`, the Rust component has basically implemented the `compatible_mode` protocol. We’d like to explore some optimizations, focus on improving deserialization performance,  including but not limited to:

1. Replacing unnecessary `xx.clone()` calls with references like `&xx`, try to `zero-copy`.
2. Accelerating string read/write logic using `SIMD`
3. Adding `cache` for some hot read operations
4. Implementing the `fast_match` feature described in the spec to speed up field matching
5. more precise `reserve_space`

Do you have any other suggestions or thoughts? @chaokunyang @theweipeng

chaokunyang:
@urlyy We also need to refine the API in fory rust, for example:
1. use `read/write_xxx` style API for Buffer methods
2. use `read/write` for serializer method to make it distinct from serialization entrance method `serialize/deserialize`
3. I'm not sure, maybe need more discussion: add `fory_` prefix to serializer trait methods, so that there won't be any name conflict in user struct when it implement other trait with same method name. Users may annotate their struct for json serialization too, and maybe some lib will generate a `read/write/serialize/deserialize` method too.

chaokunyang:
> [@urlyy](https://github.com/urlyy) We also need to refine the API in fory rust, for example:
> 
> 1. use `read/write_xxx` style API for Buffer methods
> 2. use `read/write` for serializer method to make it distinct from serialization entrance method `serialize/deserialize`
> 3. I'm not sure, maybe need more discussion: add `fory_` prefix to serializer trait methods, so that there won't be any name conflict in user struct when it implement other trait with same method name. Users may annotate their struct for json serialization too, and maybe some lib will generate a `read/write/serialize/deserialize` method too.

The 3rd point may not be important, since most of times, users won't call into this method.

urlyy:
> 2. use `read/write` for serializer method to make it distinct from serialization entrance method `serialize/deserialize`

@chaokunyang What does the 2nd point mean? Do you mean that the `serialize()` method on the `Serializer trait` should be renamed to `write()` or something else, while `fory.serialize()` remains unchanged?

Or you were mistaken—that the 2nd point doesn’t need to be handled, but the 3rd one still does?

chaokunyang:
`serialize()` method on the `Serializer trait` could be renamed to `read/write` to keep consistency with other languages and  keep `fory.serialize()` unchanged

[...]

PR #2671, opened 2025-09-26
refactor(Rust): Refine api name

urlyy:
## What does this PR do?
1. Use `read/write_xxx` style APIs for Buffer methods to align with Java API naming.
2. Add `fory_` prefix to `Serializer` and `StructSerializer` trait methods, to avoid name conflict in user struct when it implement other trait with same method name.
3. Organize the APIs on the trait with hierarchical naming `fory_read`,`fory_read_type_info`,`fory_read_data`. The same applies to the write API. Make them distinct from `serialize`/`deserialize` methods of `Fory`.

## Related issues
- #2658
- #2672 

1. If users want to call fory_methods on user-defined struct instance, they need to use methods whose prefix is `fory_`.
2. User now can customize Serializer without `derive Fory`:
```rust
#[test]
fn test_use() {
    use fory_core::fory::{read, write};
    #[derive(Debug, PartialEq, Default)]
    struct Item {
        f1: i32,
        f2: i8,
    }
    impl Serializer for Item {
        fn fory_write_data(&self, context: &mut WriteContext, is_field: bool) {
            write(&self.f1, context, is_field);
        }

        fn fory_read_data(context: &mut ReadContext, is_field: bool) -> Result<Self, Error> {
            Ok(Self {
                f1: read(context, is_field)?,
                f2: 0,
            })
        }
    }
    let mut fory = Fory::default().mode(Compatible).xlang(true);
    fory.register_serializer::<Item>(100);
    let item = Item { f1: 1, f2: 2 };
    let bytes = fory.serialize(&item);
    let new_item: Item = fory.deserialize(&bytes).unwrap();
    assert_eq!(new_item.f1, item.f1);
    assert_eq!(new_item.f2, 0);
}
```

### DIFF
--- a/rust/fory-core/src/meta/type_meta.rs
+++ b/rust/fory-core/src/meta/type_meta.rs
@@ -122,7 +122,7 @@ impl FieldType {
                 header |= 2;
             }
         }
-        writer.write_var_uint32(header);
+        writer.write_varuint32(header);
         match self.type_id {
             x if x == TypeId::LIST as u32 || x == TypeId::SET as u32 => {
                 let generic = self.generics.first().unwrap();
@@ -140,7 +140,7 @@ impl FieldType {
     }
 
     fn from_bytes(reader: &mut Reader, read_flag: bool, nullable: Option<bool>) -> Self {
-        let header = reader.read_var_uint32();
+        let header = reader.read_varuint32();
         let type_id;
         let _nullable;
         if read_flag {
@@ -215,7 +215,7 @@ impl FieldInfo {
         let encoding = Self::u8_to_encoding((header >> 6) & 0b11).unwrap();
         let mut name_size = ((header >> 2) & FIELD_NAME_SIZE_THRESHOLD as u8) as usize;
         if name_size == FIELD_NAME_SIZE_THRESHOLD {
-            name_size += reader.read_var_uint32() as usize;
+            name_size += reader.read_varuint32() as usize;
         }
         name_size += 1;
 
@@ -227,7 +227,7 @@ impl FieldInfo {
             .decode(field_name_bytes, encoding)
             .unwrap();
         FieldInfo {
-            field_name,
+            field_name: field_name.original,
             field_type,
         }
     }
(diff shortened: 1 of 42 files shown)


rust/fory-core/src/meta/type_meta.rs
```diff
@@ -504,11 +504,11 @@ impl TypeMeta {
         }
         let meta_hash = murmurhash3_x64_128(layers_writer.dump().as_slice(), 47).0 as i64;
         header |= (meta_hash << (64 - NUM_HASH_BITS)).abs();
-        result.i64(header);
+        result.write_i64(header);
         if meta_size >= META_SIZE_MASK {
-            result.var_uint32((meta_size - META_SIZE_MASK) as u32);
+            result.write_var_uint32((meta_size - META_SIZE_MASK) as u32);
```
chaokunyang:
```suggestion
            result.write_varuint32((meta_size - META_SIZE_MASK) as u32);
```

`varuint32` is more idiomatic

urlyy: Ok, I will replace other.

urlyy:
The new commit includes the following changes:

1. Previously, `metastring_encoder` converted `string -> MetaString`, but the decoder converted `bytes -> string`. This commit aligns the decoder with the encoder, now converting `bytes -> MetaString`, providing support for future enhancements to `named_enum`, `named_ext`, and `named_struct`.
2. Renamed `var_uint` and `var_int` to `varuint` and `varint`.

urlyy:
User now can customize Serializer:
```rust
#[test]
fn test_use() {
    use fory_core::fory::{read, write};
    #[derive(Debug, PartialEq, Default)]
    struct Item {
        f1: i32,
        f2: i8,
    }
    impl Serializer for Item {
        fn fory_write_data(&self, context: &mut WriteContext, is_field: bool) {
            write(&self.f1, context, is_field);
        }

        fn fory_read_data(context: &mut ReadContext, is_field: bool) -> Result<Self, Error> {
            Ok(Self {
                f1: read(context, is_field)?,
                f2: 0,
            })
        }
    }
    let mut fory = Fory::default().mode(Compatible).xlang(true);
    fory.register_serializer::<Item>(100);
    let item = Item { f1: 1, f2: 2 };
    let bytes = fory.serialize(&item);
    let new_item: Item = fory.deserialize(&bytes).unwrap();
    assert_eq!(new_item.f1, item.f1);
    assert_eq!(new_item.f2, 0);
}
```

urlyy:
Btw，a struct without use Fory derive but only impl Serializer trait，only can be registered by calling register_serializer, as it
 not impl StructSerializer trait.

chaokunyang:
Do not use varian, it's a type for values of different types

Issue #2227, opened 2025-05-13
[OSPP][Go]


chaokunyang:
### Feature Request

The current Fury Go uses reflection to implement serialization and deserialization of structs, but reflection has performance limitations. In languages like Java, Fury accelerates execution performance by dynamically generating code at runtime. 

However, in Golang, generating code at runtime requires generating assembly code, which involves high complexity in development and maintenance. In the industry, the primary approach is to generate code ahead of time during compilation, typically using tools like go generate to produce code, as in libraries like ffjson for JSON processing.

This project aims to implement similar code generation capabilities for Fury's binary serialization protocol. By parsing structs at compile time and targeting Fury's cross-language binary protocol, the goal is to generate corresponding serialization and deserialization code to enhance the execution efficiency of the entire serialization process.

### Describe the solution you'd like

Implement metadata parsers for Struct/List/Map.
Implement code generators for Struct/List/Map.
Extend the serializer Dispatcher to support dispatching to the generated serializers

PR #2676, opened 2025-09-27
refactor(go): refine collection header bitmap in codegen

ThisingL:
## Why?
Fix compatibility issues between codegen and reflection paths after the collection header bitmap refactoring.

## What does this PR do?
- Update codegen to use the new positive logic bits (CollectionIsDeclElementType instead of CollectionNotDeclElementType)
- Fix string fields to use NotNullValueFlag instead of RefValueFlag (aligning with removed string reference tracking)
- Fix slice serialization in codegen to match reflection's collection flag behavior
- Handle different collection flag combinations correctly in decoder

## Related issues
- #2227 

### DIFF
--- a/ci/run_ci.sh
+++ b/ci/run_ci.sh
@@ -356,11 +356,11 @@ case $1 in
     ;;
     go)
       echo "Executing fory go tests for go"
-      # cd "$ROOT/go/fory"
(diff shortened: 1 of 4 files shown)


chaokunyang (approved):
LGTM

PR #2678, opened 2025-09-27
feat(rust): support reference tracking for rust Rc/Arc

### LINKED ISSUE #2621: [Rust] Support `Rc/Arc/Weak` serialization for object graph support
### Feature Request

When to use `Rc/Arc/Weak`:

1. Rc<T> (Reference Counted)
- Use when: You need multiple parts of your single-threaded program to share ownership of a value, and you want the value to be deallocated only when the last owner is gone. 
- Example: A tree structure where different nodes might need to refer to the same parent node. 
2. Arc<T> (Atomic Reference Counted)- Use when: You need multiple threads to share ownership of data. It's the thread-safe version of Rc. 
- Example: Multiple threads need to access and read the same immutable configuration data. 
3. Weak<T> (Weak Reference):
- Use when: You have a reference cycle between Rc or Arc pointers that would cause a memory leak. A Weak pointer doesn't count towards the strong reference count, allowing the data to be deallocated even if Weak pointers still exist. 
- How it works: To access the data, a Weak reference must be "upgraded" to a strong Rc or Arc reference. If the data has already been deallocated, the upgrade will fail, returning a null pointer, which is how you check for validity. 
- Example: In a doubly linked list, where each node has a next pointer (strong) and a prev pointer (weak) to avoid a circular dependency that would prevent the nodes from being deallocated

Object graph is common in languages like java/python/go, by using `Rc/Arc/Weak`, we can implement similiar functionalities in rust. Fory needs to support `Rc/Arc/Weak` serialization.

_No response_

Like we did in java/python, we can implement a `RefResolver` in rust, which track reference for `Rc/Arc/Weak` by using a Map. At deserialization, we can use a `Vec<Ref>` to hold those reference to rebuild the reference relationship.

`Ref` can be defined as:
```rust
enum Ref {
  RcRef(Box<dyn Any>),
  ArcRef(Box<dyn Any>)
}
```

https://github.com/apache/fory/blob/main/java/fory-core/src/main/java/org/apache/fory/resolver/MapRefResolver.java can be used an example for implementation in rust.

_No response_

### Additional context

_No response_

chaokunyang:
## What does this PR do?

This PR supports reference tracking for rust `Rc/Arc`. `Weak` is not supported yet, this PR doesn't support circular reference for now.

## Related issues

Closes #2621

### DIFF
--- a/rust/tests/tests/test_complex_refs.rs
+++ b/rust/tests/tests/test_complex_refs.rs
@@ -1,0 +1,139 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Tests for shared reference handling in Fury Rust
+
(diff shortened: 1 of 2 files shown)


chaokunyang:
cc @urlyy

PR #2692, opened 2025-10-02
fix(java): Fix flakiness in LazyMapTest#testMap

swj20010308:
## What does this PR do?

`LazyMapTest#testMap` assumes that `HashMap` maintains constant ordering, which can lead to non-deterministic failures if the ordering of `map` when calling `toString()` is different from the ordering of `map` when initializing `LazyMap`.

> This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time.

We can reproduce the non-deterministic failures with [NonDex](https://github.com/TestingResearchIllinois/NonDex):
```
mvn -pl fory-core edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=org.apache.fory.collection.LazyMapTest#testMap
```

Instead of `HashMap`, use `LinkedHashMap` to gurrantee constant ordering over time. 

No

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -547,17 +547,17 @@ fory.register(MyClass, typename="my.package.MyClass", serializer=custom_serializ
 
 ### Language Modes Comparison
 
-| Feature               | Python Mode (`xlang=False`) | Cross-Language Mode (`xlang=True`)    |
-| --------------------- | --------------------------- | ------------------------------------- |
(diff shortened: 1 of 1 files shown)


swj20010308:
Looks like the Code Style Check failed in an unrelated file?

chaokunyang (approved):
LGTM

Issue #2611, opened 2025-09-15
[Go]  Replace legacy RegisterTagType api call


junjiexh:
### Feature Request

The current implementation in fory-go use RegisterTagType for registration is a legacy api.
Should be replaced by 
```
func (r *typeResolver) registerNamedType(
	type_ reflect.Type,
	namespace string,
	typeName string
)
```
The api call like registerNamedType is common practice in other languages like Java and Python.

junjiexh:
@chaokunyang may be can leave for me, seems not difficult

chaokunyang:
> [@chaokunyang](https://github.com/chaokunyang) may be can leave for me, seems not difficult

Great!, Assigned to you

PR #2696, opened 2025-10-03
refactor(go): Replace legacy RegisterTagType api call

junjiexh:
## Why?

fory-go was using RegisterTagType as register api which is already abandoned in implements in java and python. We need to switch to "typename + namespace" pattern registration.

## What does this PR do?

Add two register-api called RegisterType and Register, they all do the same things. The RegisterType will register type on "typename + namespace", and the Register will register type using "namespace.typename" which is less complicated. 

Remove the old api RegisterTagType and all its usages.

## Related issues

#2611 

- [x] Does this PR introduce any public API change? yes
- [x] Does this PR introduce any binary protocol compatibility change? no


go/fory/fory.go
```diff
@@ -174,8 +174,19 @@ type Fory struct {
...
-	return f.typeResolver.RegisterTypeTag(reflect.ValueOf(v), tag)
+func (f *Fory) RegisterType(
+	v interface{},
+	namespace string,
+	typeName string,
+) error {
+	return f.typeResolver.RegisterNamedType(reflect.TypeOf(v), namespace, typeName)
+}
+
+func (f *Fory) Register(
```
chaokunyang, line 195: Do we have a API that register by id?

chaokunyang: Maybe register API should be register by ID, and we have a func `RegisterNamedType` which register by name

junjiexh: Not yet, will try to implement one.

docs/guide/xlang_serialization_guide.md
```diff
@@ -481,7 +481,7 @@ func main() {
   F3 map[string]string
  }
  fory := forygo.NewFory(true)
- if err := fory.RegisterTagType("example.SomeClass", SomeClass{}); err != nil {
+ if err := fory.Register(SomeClass{}, "example.SomeClass"); err != nil {
```
chaokunyang: is this call to `func (f *Fory) Register(User{}, 101) error`? it should accept int32 only

go/fory/fory.go
```diff
@@ -174,8 +174,29 @@ type Fory struct {
 	metaContext    *MetaContext
 }
 
-func (f *Fory) RegisterTagType(tag string, v interface{}) error {
-	return f.typeResolver.RegisterTypeTag(reflect.ValueOf(v), tag)
+// register by namespace + typename
+func (f *Fory) RegisterType(
+	v interface{},
+	namespace string,
```
chaokunyang, line 180: RegisterType/Register shoud register by ID, only RegisterNamedType register type by name

chaokunyang: if you want to separate namespace and typeName, you can use `RegisterByNamespace`

junjiexh: > if you want to separate namespace and typeName, you can use `RegisterByNamespace`

got it!

PR #2699, opened 2025-10-03
feat(Rust): add support to ext && skip named_xx

urlyy:
## What does this PR do?

1. Added many unit tests, which should ensure the correctness of the protocol implementation.
2. Added benchmark code for string_data writing.
3. Implemented skipping for `named_xx`.
4. Implemented `write_namespace`/`typename` in `Consistent` mode (unit tests not completed due to `structHash` not being implemented).
5. Added sorting of fields when reading in `TypeMeta``.

fory support util func to write/read
```rust
pub fn write<T: Serializer>(
    this: &T,
    context: &mut WriteContext,
    is_field: bool,
    skip_ref_flag: bool,
    skip_type_info: bool,
) {
    write_ref_info_data(this, context, is_field, skip_ref_flag, skip_type_info);
}

pub fn read<T: Serializer>(
    context: &mut ReadContext,
    is_field: bool,
    skip_ref_flag: bool,
    skip_type_info: bool,
) -> Result<T, Error> {
    read_ref_info_data(context, is_field, skip_ref_flag, skip_type_info)
}
```
use customized serializer like this:
```
#[derive(Debug, PartialEq, Default)]
struct MyExt {
    id: i32,
}
impl Serializer for MyExt {
    fn fory_write_data(&self, context: &mut WriteContext, _is_field: bool) {
        // set is_field=false to write type_id like in java xwriteRef 
        write(&self.id, context, false, false, false);
    }

  fn fory_read_data(context: &mut ReadContext, _is_field: bool) -> Result<Self, Error> {
      Ok(Self {
          // set is_field=false to write type_id like in java xreadRef
          id: read(context, false, false, false)?,
      })
  }
}
```


urlyy:
Other notes:

1. The code for `metastring_resolver` was written by large language model, but it works correctly.
2. Originally, the `string-simd` code and benchmark were also written by the large language model and showed a 1.5–2× performance improvement. However, during final testing, some issues were found, so the SIMD implementation was completely removed, while the benchmark code was retained for future use.

urlyy:
I don't know why the ci-python error occurred.

rust/tests/tests/test_compatible.rs
```diff
@@ -15,11 +15,12 @@
 // specific language governing permissions and limitations
 // under the License.
 
+mod compatible;
```
chaokunyang: this file can be moved to compatible dir

rust/fory-derive/src/object/read.rs
```diff
@@ -136,7 +136,7 @@ pub fn gen_read(struct_ident: &Ident) -> TokenStream {
                     <Self as fory_core::serializer::Serializer>::fory_read_data(context, false)
                 },
                 fory_core::types::Mode::Compatible => {
-                    <#struct_ident as fory_core::serializer::StructSerializer>::fory_read_compatible(context)
+                    <#struct_ident as fory_core::serializer::Serializer>::fory_read_compatible(context)
```
chaokunyang, line 139: why it's `Serializer` instead of `StructSerializer`

urlyy: Although it may look a bit semantically confusing… In order to split up the code and improve readability, I moved the deserialization logic of `compatible_struct` into `read_compatible()`, which is on the same level as `read()`. However, since the code for `read_compatible()` is generated at compile time, the variable names don’t reveal whether the type is a struct or an EXT or an enum, so in such cases `read_compatible()` is always used.

The problem is that EXT types only implement the `Serializer` trait. With the previous implementation, they didn’t have a `read_compatible()` method, which would result in a “method not found” error. To fix this, I moved `read_compatible()` from the `StructSerializer` trait into the `Serializer` trait, and provided a default implementation whose logic applies only to `EXT/NAMED_EXT`. The only possible callers of `read_compatible()` are `struct/enum/ext`. And struct/enum will override the default implementation.

urlyy: It may look a bit semantically confusing… In order to split up the code and improve readability, I moved the deserialization logic of `compatible_struct` into `read_compatible()`, which is on the same level as `read()`. However, since the code for `read_compatible()` is generated at compile time, the variable names don’t reveal whether the type is a struct or an EXT or an enum, so in such cases `read_compatible()` is always used.

The problem is that EXT types only implement the `Serializer` trait. With the previous implementation, they didn’t have a `read_compatible()` method, which would result in a “method not found” error. To fix this, I moved `read_compatible()` from the `StructSerializer` trait into the `Serializer` trait, and provided a default implementation whose logic applies only to `EXT/NAMED_EXT`. The only possible callers of `read_compatible()` are `struct/enum/ext`. And struct/enum will override the default implementation.


`But I found that this part still doesn’t handle EXT, so it needs to be added.`

rust/tests/tests/test_compatible.rs
urlyy: Done

rust/fory-derive/src/object/read.rs
urlyy: > `But I found that this part still doesn’t handle EXT, so it needs to be added.`

Done

chaokunyang:
@urlyy The API has changed a lot in #2704 , please merge main branch.

urlyy:
> @urlyy The API has changed a lot in #2704 , please merge main branch.

Yes, I will handle this soon.

chaokunyang:
> Other notes:
> 
> 2. Originally, the `string-simd` code and benchmark were also written by the large language model and showed a 1.5–2× performance improvement. However, during final testing, some issues were found, so the SIMD implementation was completely removed, while the benchmark code was retained for future use.

@urlyy Could you share more details about the issue you found?

urlyy:
> @urlyy Could you share more details about the issue you found?

I will solve this in next pr.

urlyy:
Too much trouble, so I’ll just open a new PR：#2712

PR #2701, opened 2025-10-04
feat(java): Add ForyFeature for GraalVM Native Image

### LINKED ISSUE #2697: [Java][Native Image] `ObjectCreator` sometimes produces `Object` under GraalVM 25
### Search before asking

### Version

OS: ArchLinux (6.16.7-zen1-1-zen)
Fory: 'org.apache.fory:fory-core:0.13.0-SNAPSHOT' (Sep 30)
JDK: GraalVM CE 25+37.1 (build 25+37-jvmci-b01)

### Component(s)

Java

### Minimal reproduce step

See https://github.com/gudzpoz/fory-native-image-reproducer/blob/private-constructor/app/src/main/java/org/example/App.java .

```console
$ git clone https://github.com/gudzpoz/fory-native-image-reproducer.git
$ git checkout private-constructor
$ ./gradlew run # pass
$ ./gradlew nativeCompile
$ app/build/native/nativeCompile/test
Error: Exception in thread "main" org.apache.fory.exception.DeserializationException: Deserialize failed, read objects are: [org.example.App@653b6657, java.lang.Object@2a34e425]
	at org.apache.fory.util.ExceptionUtils.handleReadFailed(ExceptionUtils.java:63)
	at org.apache.fory.Fory.deserialize(Fory.java:889)
	at org.apache.fory.Fory.deserialize(Fory.java:785)
	at org.example.App.main(App.java:32)
	at java.base@25/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)
Caused by: java.lang.ClassCastException: java.lang.Object cannot be cast to org.example.App$AbstractClass
	at org.example.AppForyRefCodec_0.read(AppForyRefCodec_0.java:73)
	at org.apache.fory.Fory.readDataInternal(Fory.java:1046)
	at org.apache.fory.Fory.readRef(Fory.java:931)
	at org.apache.fory.Fory.deserialize(Fory.java:885)
	... 3 more
```

Here is the GitHub action run reproducing the issue: https://github.com/gudzpoz/fory-native-image-reproducer/actions/runs/18224046965/job/51890537453 . (The "Test" step contains the failing stack trace.)

The test should pass.

### What did you see instead?

```console
$ app/build/native/nativeCompile/test
Error: Exception in thread "main" org.apache.fory.exception.DeserializationException: Deserialize failed, read objects are: [org.example.App@653b6657, java.lang.Object@2a34e425]
	at org.apache.fory.util.ExceptionUtils.handleReadFailed(ExceptionUtils.java:63)
	at org.apache.fory.Fory.deserialize(Fory.java:889)
	at org.apache.fory.Fory.deserialize(Fory.java:785)
	at org.example.App.main(App.java:32)
	at java.base@25/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)
Caused by: java.lang.ClassCastException: java.lang.Object cannot be cast to org.example.App$AbstractClass
	at org.example.AppForyRefCodec_0.read(AppForyRefCodec_0.java:73)
	at org.apache.fory.Fory.readDataInternal(Fory.java:1046)
	at org.apache.fory.Fory.readRef(Fory.java:931)
	at org.apache.fory.Fory.deserialize(Fory.java:885)
	... 3 more
```

### Anything Else?

This seems to be related to the private constructor(s). Changing `private ConcreteClass(int i)` to `protected ConcreteClass(int i)` solves the issue. Also `ObjectCreator` does not seem to handle parent constructors with arguments currently?

mengnankkkk:
## Why?
Reflective access to private constructors in the GraalVM Native Image environment is limited. ObjectCreator failures return the wrong type, causing a ClassCastException. Lack of friendly GraalVM error messages.

## What does this PR do?
Added GraalVM environment detection and special processing, optimized ObjectCreator implementation, and added unit tests

## Related issues
Close #2697 

### DIFF
--- a/integration_tests/graalvm_tests/src/test/java/org/apache/fory/util/GraalvmSupportRecordTest.java
+++ b/integration_tests/graalvm_tests/src/test/java/org/apache/fory/util/GraalvmSupportRecordTest.java
@@ -49,7 +49,8 @@ public class GraalvmSupportRecordTest {
   public void testIsRecordConstructorPublicAccessible() {
     Assert.assertTrue(GraalvmSupport.isRecordConstructorPublicAccessible(PublicRecord.class));
     Assert.assertFalse(GraalvmSupport.isRecordConstructorPublicAccessible(PrivateRecord.class));
-    Assert.assertFalse(GraalvmSupport.isRecordConstructorPublicAccessible(ClassWithNoArgCtor.class));
+    Assert.assertFalse(
+        GraalvmSupport.isRecordConstructorPublicAccessible(ClassWithNoArgCtor.class));
   }
 
   @Test
@@ -61,7 +62,8 @@ public class GraalvmSupportRecordTest {
     // Class with no-arg constructor doesn't need reflection registration
     Assert.assertFalse(GraalvmSupport.needReflectionRegisterForCreation(ClassWithNoArgCtor.class));
     // Class without no-arg constructor needs reflection registration
-    Assert.assertTrue(GraalvmSupport.needReflectionRegisterForCreation(ClassWithoutNoArgCtor.class));
+    Assert.assertTrue(
+        GraalvmSupport.needReflectionRegisterForCreation(ClassWithoutNoArgCtor.class));
     // Interface doesn't need reflection registration
     Assert.assertFalse(GraalvmSupport.needReflectionRegisterForCreation(Runnable.class));
   }


chaokunyang:
`ParentNoArgCtrObjectCreator` should be able to generate a constructor of parent class to create instance of subclass using `jdk.internal.reflect.ReflectionFactory`. It's strange why it can't work

mengnankkkk:
I was wrong. I should not use unsafe to escape this problem. I should use GraalVM's reachability metadata to solve the problem of ReflectionFactory.

java/fory-core/src/main/resources/META-INF/native-image/org.apache.fory/fory-core/reflection-config.json
```diff
@@ -5,5 +5,18 @@
     "allPublicConstructors": true,
     "allDeclaredMethods": true,
     "allPublicMethods": true
+  },
+  {
+    "name": "jdk.internal.reflect.ReflectionFactory",
```
chaokunyang: those methods are executed at graalvm build time, they won't be executed at graalvm runtime. It's unnecessary to add reflection-config.

I think the key is find out why the generated constructor don't work

mengnankkkk: I will try it💕💕

mengnankkkk: > those methods are executed at graalvm build time, they won't be executed at graalvm runtime. It's unnecessary to add reflection-config.
> 
> I think the key is find out why the generated constructor don't work



I found the problem：
When GraalVM's AOT compiler attempts to statically compile this "handle" into native code, the constructor in the final executable file is actually corrupted and unusable because the dynamic JVM mechanism behind it does not exist in GraalVM's minimalist runtime (Substrate VM).
I think it's possible to use the Unsafe downgrade solution exclusively in GraalVM Native Images. Please share your understanding and solution with me, and I'll consider it.

chaokunyang: Maybe we need to create a new maven module to imply a graalvm Feature for fory. In that feature, we can generate unsafe meta for type without public no argument constructor.

mengnankkkk: We can create a fory-graalvm-feature module specifically to adapt to the GraalVM Native Image environment. They need to depend on both fory-core and fory-graalvm-feature in the project. When they run the native-image build command, GraalVM automatically discovers and runs ForyGraalVMFeature through Service Loader, obtains all registered classes in Fory, identifies those "problematic classes" that rely on ReflectionFactory in the JVM, and registers the "blank instance" creation capability with GraalVM for the "problematic classes". In order to ensure that the subsequent field filling is correct, all fields can be registered at the same time for reflective write operations. Provide an entry point in fory-core to switch to using ForyGraalVMFeature. fory-core needs to expose static methods such as isProblematicForCreation. This is my design. You can add to it.

chaokunyang: Looks good to me overall. We can use Fory existing registry to discover user's registered types, and add necessary metadata to graalvm for those types, so at graalvm runtime, the object creation will work as expected. 

The most critical graal-meta would be :
- `proxy` related meta, so we can create JDK proxy at runtime without users to configure `proxy.json`
- configure `unsafeAllocated` in reflect-meta.json

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -89,6 +93,10 @@
 public final class Fory implements BaseFory {
   private static final Logger LOG = LoggerFactory.getLogger(Fory.class);
 
+  // Static collections for GraalVM Feature support
```
chaokunyang: Could we remove those fields, `TypeResolver` already has static variable to capture registered classes and generated serializers

integration_tests/graalvm_tests/src/main/java/org/apache/fory/graalvm/FeatureTestExample.java
```diff
@@ -0,0 +1,125 @@
...
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import org.apache.fory.Fory;
+import org.apache.fory.config.Language;
+
+public class FeatureTestExample {
+
+  // Test class with private constructor (problematic for creation)
+  public static class ProblematicClass {
```
chaokunyang: How about naming to PrivateConstructorClass

integration_tests/graalvm_tests/src/main/java/org/apache/fory/graalvm/FeatureTestExample.java
```diff
@@ -0,0 +1,125 @@
...
+    try {
+      // Test 1: Serialize/deserialize problematic class
+      ProblematicClass original = new ProblematicClass("test-value");
+      byte[] serialized = fory.serialize(original);
+      ProblematicClass deserialized = (ProblematicClass) fory.deserialize(serialized);
+
+      if (!"test-value".equals(deserialized.getValue())) {
+        throw new RuntimeException("Problematic class test failed");
+      }
+      System.out.println("✓ Problematic class serialization test passed");
```
chaokunyang: ditto, don't use Problematic, since it's supproted by fory now with this PR

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -1764,4 +1768,47 @@ public MetaCompressor getMetaCompressor() {
...
+   * that delegates to {@link TypeResolver#registerProxyInterfaceForGraalvm}.
+   *
+   * <p>Note: This registers the proxy interface globally across all Fory configurations. If you
+   * need per-configuration registration, use the TypeResolver methods directly.
+   *
+   * @param proxyInterface the proxy interface to register
+   * @throws NullPointerException if proxyInterface is null
+   * @throws IllegalArgumentException if proxyInterface is not an interface
+   */
+  public static void addProxyInterface(Class<?> proxyInterface) {
```
chaokunyang: I prefer not adding API to Fory, since those API are just used by graalvm. We could let it at TypeResolver, and you invoke TypeResolver API directly in GraalVMmFeature

java/fory-core/src/main/java/org/apache/fory/reflect/ObjectCreators.java
```diff
@@ -77,16 +77,47 @@ public static <T> ObjectCreator<T> getObjectCreator(Class<T> type) {
...
+        || RecordUtils.isRecord(type)) {
+      return false;
+    }
+    Constructor<?>[] constructors = type.getDeclaredConstructors();
+    if (constructors.length == 0) {
+      return true;
+    }
+    for (Constructor<?> constructor : constructors) {
+      if (constructor.getParameterCount() == 0) {
+        return false;
```
chaokunyang: For JDK17+ record, if its constructor is public accessible, then it can be invoked by Fory

chaokunyang: And this API could be moved to GraalvmSupport class

java/fory-core/src/main/java/org/apache/fory/util/GraalvmSupport.java
```diff
@@ -145,4 +145,37 @@ public static boolean isRecordConstructorPublicAccessible(Class<?> type) {
...
+  /**
+   * Checks if a class is problematic for object creation and requires special handling in GraalVM.
+   *
+   * <p>A class is considered problematic if it lacks a public no-arg constructor and would
+   * typically require ReflectionFactory or unsafe allocation for instantiation.
+   *
+   * @param type the class to check
+   * @return true if the class is problematic for creation, false otherwise
+   */
+  public static boolean isProblematicForCreation(Class<?> type) {
```
chaokunyang:
```suggestion
  public static boolean needReflectionRegisterForCreation(Class<?> type) {
```

mengnankkkk: Ok. I. will do it  thanks for your codereview and suggestions 💕💕

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -177,11 +177,13 @@ public Fory(ForyBuilder builder, ClassLoader classLoader) {
   @Override
   public void register(Class<?> cls) {
     _getTypeResolver().register(cls);
+    TypeResolver.registerClassForGraalvm(cls, config.getConfigHash());
```
chaokunyang: We don't need to handle this here. TypeResolver should do everything for registry. `ensureSerializersCompiled` will create serializer for every register class, you can add registerClassForGraalvm here

integration_tests/graalvm_tests/src/main/java/org/apache/fory/graalvm/FeatureTestExample.java
```diff
@@ -0,0 +1,125 @@
...
+        throw new RuntimeException("Proxy test failed");
+      }
+      System.out.println("Proxy serialization test passed");
+
+      System.out.println("All GraalVM Feature tests passed!");
+
+    } catch (Exception e) {
+      System.err.println("GraalVM Feature test failed: " + e.getMessage());
+      e.printStackTrace();
+      System.exit(1);
```
chaokunyang:
```suggestion
      throw new RuntimeException(e);
```

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
```diff
@@ -670,6 +674,82 @@ public final MetaStringResolver getMetaStringResolver() {
     return metaStringResolver;
   }
 
+  /**
+   * Get all registered classes across all GraalVM registries for native image compilation.
+   *
+   * @return unmodifiable set of all registered classes
+   */
+  public static Set<Class<?>> getAllRegisteredClasses() {
```
chaokunyang: How about movign those methods into GraalvmSupport.java, and we can register all needs info into GraalvmSupport in `TypeResolver#enableSerialziersCcmpiled`. In thsi wya, the API is more clean

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
```diff
@@ -670,6 +674,82 @@ public final MetaStringResolver getMetaStringResolver() {
...
+    }
+    return Collections.unmodifiableSet(allClasses);
+  }
+
+  /**
+   * Get all registered proxy interfaces across all GraalVM registries for native image compilation.
+   *
+   * @return unmodifiable set of all registered proxy interfaces
+   */
+  public static Set<Class<?>> getAllProxyInterfaces() {
```
chaokunyang: ditto

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
```diff
@@ -670,6 +674,82 @@ public final MetaStringResolver getMetaStringResolver() {
...
+  }
+
+  /**
+   * Register a class in the GraalVM registry for native image compilation. This should be called
+   * from {@link org.apache.fory.Fory#register} methods.
+   *
+   * @param cls the class to register
+   * @param configHash the configuration hash for the Fory instance
+   */
+  public static void registerClassForGraalvm(Class<?> cls, int configHash) {
```
chaokunyang: ditto,and we can also remove `ForGraalvm` if we move those methods into `GraalvmSupport`

integration_tests/graalvm_tests/src/main/java/org/apache/fory/graalvm/FeatureTestExample.java
```diff
@@ -0,0 +1,124 @@
...
+    System.out.println("Testing Fory GraalVM Feature...");
+
+    Fory fory =
+        Fory.builder().withLanguage(Language.JAVA).withRefTracking(true).withCodegen(false).build();
+
+    fory.register(PrivateConstructorClass.class);
+    fory.register(TestInvocationHandler.class);
+
+    // Register proxy interface
+    TypeResolver.addProxyInterface(TestInterface.class);
```
chaokunyang:
```suggestion
    GraalvmSupport.registerProxySupport(TestInterface.class);
```

java/fory-core/src/main/java/org/apache/fory/util/GraalvmSupport.java
```diff
@@ -54,6 +65,97 @@ public static boolean isGraalRuntime() {
...
+  }
+
+  /**
+   * Register a class in the GraalVM registry for native image compilation.
+   *
+   * @param cls the class to register
+   * @param configHash the configuration hash for the Fory instance
+   */
+  public static void registerClassForGraalvm(Class<?> cls, int configHash) {
+    GraalvmClassRegistry registry =
```
chaokunyang, line 105: Needs a check `IS_GRAALVM_NATIVE_IMAGE`, if not, this should be a no-op. Otherwise it hold strong reference to class, which will make class gc not work.

Please check whether similiar issues exist too

integration_tests/graalvm_tests/src/main/resources/META-INF/native-image/org.apache.fory/graalvm_tests/native-image.properties
```diff
@@ -33,4 +33,6 @@ Args=-H:+ReportExceptionStackTraces \
   org.apache.fory.graalvm.CollectionExample,\
   org.apache.fory.graalvm.ArrayExample,\
   org.apache.fory.graalvm.Benchmark,\
-  org.apache.fory.graalvm.Main
+  org.apache.fory.graalvm.FeatureTestExample,\
+  org.apache.fory.graalvm.Main,\
+  org.apache.fory.util.GraalvmSupport$GraalvmClassRegistry
```
chaokunyang: this should be moved into `native-image.properties` under fory-core

java/fory-graalvm-feature/src/test/java/org/apache/fory/graalvm/feature/ForyGraalVMFeatureTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+import static org.junit.Assert.*;
+
+import org.apache.fory.Fory;
+import org.apache.fory.util.GraalvmSupport;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ForyGraalVMFeatureTest {
```
chaokunyang, line 27: This file tested nothing, you should move it into graalvm-tests

java/pom.xml
```diff
@@ -92,7 +92,13 @@
         <jdk>[17,]</jdk>
       </activation>
       <modules>
+        <module>fory-core</module>
```
chaokunyang: this is wrong, please revert it. Only `fory-simd` use JDK17+

integration_tests/graalvm_tests/src/main/java/org/apache/fory/graalvm/FeatureTestExample.java
```diff
@@ -0,0 +1,124 @@
...
+    fory.register(TestInvocationHandler.class);
+
+    // Register proxy interface
+    GraalvmSupport.registerProxySupport(TestInterface.class);
+
+    try {
+      // Test 1: Serialize/deserialize class with private constructor
+      PrivateConstructorClass original = new PrivateConstructorClass("test-value");
+      byte[] serialized = fory.serialize(original);
+      PrivateConstructorClass deserialized = (PrivateConstructorClass) fory.deserialize(serialized);
```
chaokunyang: Please also assert `fory.getClassResolver().getSerializer(PrivateConstructorClass.class) instanceof Generated`

integration_tests/graalvm_tests/pom.xml
```diff
@@ -64,6 +64,12 @@
       <artifactId>fory-graalvm-feature</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
```
chaokunyang: We use testng in fory

java/fory-core/src/main/java/org/apache/fory/util/GraalvmSupport.java
```diff
@@ -54,6 +65,103 @@ public static boolean isGraalRuntime() {
...
+    registry.registeredClasses.add(cls);
+  }
+
+  /**
+   * Register a proxy interface in the GraalVM registry for native image compilation.
+   *
+   * @param proxyInterface the proxy interface to register
+   * @param configHash the configuration hash for the Fory instance
+   */
+  public static void registerProxyInterfaceForGraalvm(Class<?> proxyInterface, int configHash) {
```
chaokunyang: The class is already GraalvmSupport, we don't need to replicate `graalvm` in teh method name

java/fory-core/src/test/java/org/apache/fory/resolver/ClassResolverTest.java
```diff
@@ -539,4 +540,44 @@ public void testAbstractCustomSerializer() {
         fory.getClassResolver().getSerializer(abs2Test.getClass()).getClass(),
         AbstractCustomSerializer.class);
   }
+
+  static class GraalvmRegistrationBean {
+    int value;
+  }
+
+  @Test
+  public void testEnsureSerializersCompiledRegistersClassesForGraalvm() {
```
chaokunyang: Move this test to graavm_tests

integration_tests/graalvm_tests/pom.xml
```diff
@@ -59,6 +58,22 @@
       <artifactId>fory-core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.fory</groupId>
+      <artifactId>fory-graalvm-feature</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.graalvm.sdk</groupId>
```
chaokunyang: `fory-graalvm-feature` should alreay has `org.graalvm.sdk` as provided , why add it again here?

integration_tests/graalvm_tests/src/test/java/org/apache/fory/util/GraalvmSupportRecordTest.java
```diff
@@ -0,0 +1,111 @@
...
+ * under the License.
+ */
+
+package org.apache.fory.util;
+
+import org.apache.fory.util.record.RecordUtils;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class GraalvmSupportRecordTest {
```
chaokunyang, line 26: I suggest move this file to `graalvm_tests`

java/fory-graalvm-feature/src/main/java/org/apache/fory/graalvm/feature/ForyGraalVMFeature.java
```diff
@@ -0,0 +1,94 @@
...
+import org.graalvm.nativeimage.hosted.RuntimeReflection;
+
+/**
+ * GraalVM Feature for Apache Fory serialization framework.
+ *
+ * <p>This feature automatically registers necessary metadata for GraalVM native image compilation
+ * to ensure Fory serialization works correctly at runtime. It handles:
+ *
+ * <ul>
+ *   <li>Registering problematic classes for unsafe allocation
```
chaokunyang: don't use `problematic` across this pr

mengnankkkk:
What should I do? I always encounter a configuration problem. When running in the graalvm environment, there is always a problem of native-image command parsing error. What should I do? I have tried many methods (crying)

chaokunyang:
image
you can make it use Unsafe for object allocation too when running in graalvm

java/fory-core/src/main/java/org/apache/fory/serializer/ObjectStreamSerializer.java
```diff
@@ -80,10 +80,60 @@
...
+  }
+
+  /**
+   * Safe wrapper for ObjectStreamClass.lookup that handles GraalVM native image limitations. In
+   * GraalVM native image, ObjectStreamClass.lookup may fail for certain classes like Throwable due
+   * to missing SerializationConstructorAccessor. This method catches such exceptions and returns
+   * null, allowing the serializer to use alternative approaches like Unsafe.allocateInstance.
+   */
+  private static ObjectStreamClass safeObjectStreamClassLookup(Class<?> type) {
+    if (GraalvmSupport.isGraalRuntime()) {
```
chaokunyang: Just use `GraalvmSupport.IS_GRAALVM_NATIVE_IMAGE`

mengnankkkk:
@chaokunyang I'm very sorry. My current coding skills aren't enough to completely fix this bug, and I haven't passed the CI test. Could you please make some changes on my branch to fix this bug? I've tried many methods, but the files are messed up. I'm sorry that after nearly two weeks, I still haven't solved this bug. I'm very sorry. Could you please refactor and upgrade my messy code?

chaokunyang:
Hi @mengnankkkk, thanks for your hard work and for letting me know the situation. I understand that you’ve spent quite a bit of time on this bug, and I appreciate the effort you’ve put in.

I’ll take a look at your branch, refactor the code, and work on fixing the bug so we can get the CI tests to pass. Don’t worry about the current state — debugging can be tricky, and each attempt helps us better understand the problem.

Thanks again for your contributions! I’ll update the PR once I’ve made the necessary changes.

mengnankkkk:
> Hi @mengnankkkk, thanks for your hard work and for letting me know the situation. I understand that you’ve spent quite a bit of time on this bug, and I appreciate the effort you’ve put in.
> 
> I’ll take a look at your branch, refactor the code, and work on fixing the bug so we can get the CI tests to pass. Don’t worry about the current state — debugging can be tricky, and each attempt helps us better understand the problem.
> 
> Thanks again for your contributions! I’ll update the PR once I’ve made the necessary changes.

I'm very sorry. I pushed AI to find a solution to the problem, but it didn't seem to solve it and made the code look even more shitty. I'm very sorry./(ㄒoㄒ)/~~

chaokunyang (approved):
LGTM

PR #2702, opened 2025-10-05
feat(java): custom class info serializer

mattjubb:
## Why?
In my project we have 1000s of classes that are serialized, we have a custom mechanism for registering those class using a UUID which provides a way to lookup those classes at runtime. It would be useful to be able to hook up this mechanism to the fory serializer. This is my first time trying to contribute to open source project so appreciate any guidance

## What does this PR do?
This PR adds a new interface ClassInfoSerializer which can be implemented to specify a custom way to serialize a class ID instead of the default short id. this can be set on the ClassResolver via setClassInfoSerializer

Yes introduces new interface - ClassInfoSerializer  and ClassResolver.setClassInfoSerializer method

-->

- [x] Does this PR introduce any binary protocol compatibility change? - yes if custom class info serializer is set, otherwise no

## Benchmark
not performed yet

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/resolver/ClassInfoSerializer.java
+++ b/java/fory-core/src/main/java/org/apache/fory/resolver/ClassInfoSerializer.java
@@ -1,0 +1,21 @@
+package org.apache.fory.resolver;
+
+import org.apache.fory.memory.MemoryBuffer;
+
+/**
+ * ClassInfoSerializer provides a mechanism to customize the recording of the ClassInfo
+ * default implementation uses a short value, however when the class registration needs to be customized
+ * this can be set of the ClassResolver to handle storing the ClassInfo is a customized manner
+ */
+public interface ClassInfoSerializer {
+    /**
+     * Writes the provided ClassInfo into the buffer in a customized manner which is consistent
+     * with the readClassInfo method
+     */
+    void writeClassInfo(ClassResolver classResolver, MemoryBuffer buffer, ClassInfo classInfo);
+
+    /**
+     * Reads the ClassInfo from the provided buffer in a manner which is consistent with writeClassInfo
+     */
+    ClassInfo readClassInfo(ClassResolver classResolver, MemoryBuffer buffer);
+}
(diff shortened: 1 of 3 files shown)


java/fory-core/src/main/java/org/apache/fory/resolver/ClassInfoSerializer.java
```diff
@@ -0,0 +1,21 @@
+package org.apache.fory.resolver;
+
+import org.apache.fory.memory.MemoryBuffer;
+
+/**
+ * ClassInfoSerializer provides a mechanism to customize the recording of the ClassInfo
+ * default implementation uses a short value, however when the class registration needs to be customized
+ * this can be set of the ClassResolver to handle storing the ClassInfo is a customized manner
+ */
+public interface ClassInfoSerializer {
```
chaokunyang, line 10: This will introduce another indirection, and incur virtual method call. The performance will degrade. Could you share some detais what will you implement using this ClassInfoSerializer. Maybe we could just merge the funtions of subclass of ClassInfoSerializer into class resolver

mattjubb: the PR description and the unit test show examples of how this would be used. Basically we maintain a large mapping of identifiers to classes, when an object is deserialized we want to control what ID is mapped to what class. I asked about this on slack a couple months ago, and you said that we cannot customize the class resolve since it final and you suggest instead to raise a PR that supports customized mapping logic. the intent of this feature is for it to be customizable by the calling code not something that is predefined within the class resolver

chaokunyang: Fory support register class by name. How about register class by uuid as name? Would this work for you?  Your  extension is in critical path. We are trying to keep callstack on this be as simple as we can. And you intercept whole class info serialization. Your class schema can't change anymore. The type forward/backward compatibility is implementes in current write/read classinfo method.

mattjubb: that would require registering all classes upfront which means loading the .class, for large amounts of classes that is a very large upfront cost, lets say we have 10,000 classes and we only want to deserialize a single object with a single class - that would mean loading all 10,000 .class to just serialize a single object. if we could provide a fully qualified name (fqn) which could then be loaded when required via Class.forName then registering by name would probably meet our needs

PR #2715, opened 2025-10-06
fix: remove commented content from git commit message

rajuyadav03:
## What’s Changed?  
This PR removes commented lines (`#` and `<!-- -->`) from auto-generated Git commit messages, ensuring the commit history is clean and readable.  

## Fixes  
- Fixes #1588  

Hi! I’m participating in Hacktoberfest 2025.  
Please add the `hacktoberfest-accepted` label if appropriate. Thanks!

PR #2736, opened 2025-10-09
test(ci): verify CI after merging changes into fork main

rajuyadav03:
Testing CI; this is a verification PR for #2715

### DIFF
--- a/scripts/remove-commented-commit.sh
+++ b/scripts/remove-commented-commit.sh
@@ -39,5 +39,6 @@ grep -v '^\s*#' "$COMMIT_MSG_FILE" | \
 
 # Replace original file with cleaned content
 mv "$TEMP_FILE" "$COMMIT_MSG_FILE"
+#test
 
 exit 0


urlyy:
I have a question: It seems that the commit logs in Fory’s history are mostly PR descriptions and not related to the local commits. In your actions, does `Get & Remove last commit message` just remove markdown comments from the latest local commit? Or does it would remove the comments from the latest commit message after the merge? I’m not very familiar with GitHub Actions either, so I’d appreciate it if you could clarify.😃

chaokunyang:
@rajuyadav03 you should submit a pr to https://github.com/rajuyadav03/fory/tree/main to verify whether it works

chaokunyang:
@rajuyadav03 you should create a PR in your forked repo, and merge the PR into your forked main branch to see whether it works instead of create a PR to fory. 

From the current PR code, I don't think it will work. You just run a pipeline to update local history of checked repo, it won't take effect after the PR got merged

PR #2137, opened 2025-04-04
chore: fix license issue

chaokunyang:
## What does this PR do?

- remove non-existent v8-fast-api-calls.h from license
- remove duplicate notice text 
- remove binary image
- add license header for main document (I don't add license header for readme.md)

## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

PR #2744, opened 2025-10-10
feat(rust): support btree set and binary heap serialization

chaokunyang:
## What does this PR do?

support btree set and binary heap serialization


urlyy:
Btw, when compiling, the string matching for struct field types also needs to include handling logic for these newly supported containers.
You can locate these spots by globally searching for `"HashMap"`.
Also, I need to add the `Fory` parameter for these container APIs in #2137.

@chaokunyang

PR #2765, opened 2025-10-13
feat(rust): make Serializer api to return Result && replace panic/expect/assert/unwrap with Result

### LINKED ISSUE #2706: [Rust] return Error instead of panic when deserialize/serialize failed
### Feature Request

Currently fory rust use `expect` in many places, if the `expect` doesn't succed, the program will panic. We should return use `Result/Error` as return value instead if panic directly. And we can shortpath the codeflow by `?`

This is necessary, especially when deserialization can come from untrusted source.

_No response_

_No response_

_No response_

### Additional context

_No response_

urlyy:
## What does this PR do?
- simply refactor `error.rs`
- replace `panic/expect/unwrap` with `Result`
- replace `assert` with `ensure`
- change Serialize api to return `Result`

## Related issues
close #2706

1. user should use `unwrap/?` when se/de.
2. when they customize serializer, will return `Result`.

### DIFF
--- a/rust/fory-core/src/error.rs
+++ b/rust/fory-core/src/error.rs
@@ -51,7 +51,7 @@ pub enum Error {
     EncodingError(Cow<'static, str>),
 
     #[error("{0}")]
-    DepthExceeded(Cow<'static, str>),
+    DepthExceed(Cow<'static, str>),
 
     /// Do not construct this variant directly; use [`Error::unknown`] instead.
     #[error("{0}")]
--- a/rust/fory-core/src/resolver/context.rs
+++ b/rust/fory-core/src/resolver/context.rs
@@ -218,7 +218,7 @@ impl ReadContext {
     pub fn inc_depth(&mut self) -> Result<(), Error> {
         self.current_depth += 1;
         if self.current_depth > self.max_dyn_depth {
-            return Err(Error::DepthExceeded(
+            return Err(Error::DepthExceed(
                 format!(
                     "Maximum dynamic object nesting depth ({}) exceeded. Current depth: {}. \
                     This may indicate a circular reference or overly deep object graph. \


urlyy:
There are still some `unsafe` code not return `Result`. Mainly about SIMD. But this pr is enough to merge.

rust/fory-core/src/buffer.rs
```diff
@@ -79,82 +81,82 @@ impl Writer {
...
-        self.bf.write_i16::<LittleEndian>(value).unwrap();
+    pub fn write_i16(&mut self, value: i16) -> Result<(), Error> {
+        Ok(self.bf.write_i16::<LittleEndian>(value)?)
     }
 
     #[inline(always)]
-    pub fn write_u32(&mut self, value: u32) {
-        self.bf.write_u32::<LittleEndian>(value).unwrap();
+    pub fn write_u32(&mut self, value: u32) -> Result<(), Error> {
+        Ok(self.bf.write_u32::<LittleEndian>(value)?)
```
chaokunyang: It's unnecessary to return Result for write, we can always ensure right will never go wrong

rust/fory-core/src/buffer.rs
```diff
@@ -394,215 +419,254 @@ impl Reader {
...
+        self.move_next(8)?;
+        Ok(result)
     }
 
     #[inline(always)]
-    pub fn read_varuint32(&mut self) -> u32 {
-        let start = self.cursor;
-        let b0 = unsafe { *self.bf.add(start) as u32 };
+    pub fn read_varuint32(&mut self) -> Result<u32, Error> {
+        let slice = self.slice_after_cursor()?;
```
chaokunyang: If a function has multiple `?`, maybe we can merge the bound checks into one check if it does't make code more complex

rust/fory-core/src/buffer.rs
```diff
@@ -394,215 +419,254 @@ impl Reader {
...
+
+        let b0 = *slice
+            .first()
+            .ok_or_else(|| Error::msg("unexpected end of buffer"))? as u32;
```
chaokunyang: I'm not sure whether this ok_or_else API introduce extra cost, since the buffer read/write is called very frequently

chaokunyang: Does this closure introduce extra cost, and will the method call here be inlined ?

rust/fory-core/src/error.rs
```diff
@@ -15,38 +15,102 @@
...
+    // don't use this directly, use Error::from() instead
     #[error(transparent)]
-    Other(#[from] anyhow::Error),
+    Wrapped(#[from] anyhow::Error),
+
+    /// A simple error message, stored as a [`Cow<'static, str>`].
+    ///
+    /// Do not construct this variant directly; use [`Error::msg`] instead.
+    #[error("{0}")]
+    Msg(Cow<'static, str>),
```
chaokunyang: don't use abbr here. abbr should be used for internal functions/code only. This is a public API

rust/fory-core/src/error.rs
```diff
@@ -15,38 +15,102 @@
...
-    Ref,
+    // not use this, just as a example.
+    // You can create new error enums for specific cases in the future.
+    #[error("Type mismatch: type_a = {0}, type_b = {1}")]
+    TypeMismatch(u32, u32),
 
+    // don't use this directly, use Error::from() instead
     #[error(transparent)]
-    Other(#[from] anyhow::Error),
+    Wrapped(#[from] anyhow::Error),
```
chaokunyang: it looks a little strange that an Error type named as `Wrapped`

chaokunyang: The deserialization can have BufferOutOfBound, maybe deserve an error type

rust/fory-core/src/buffer.rs
urlyy: It is actually equivalent to:
```rust
match option {
    Some(v) => Ok(v),
    None => Err((|| E)()),
}
```
When the Option is Some, the closure is never called or even allocated — it’s just a zero-cost inline conditional branch.

urlyy: > Does this closure introduce extra cost, and will the method call here be inlined ?

I think but not sure the LLVM will do the inline. Need to check the asm code. `RUSTFLAGS="--emit asm" cargo build --release` and `cat target/release/deps/project_name-hash.s`

rust/fory-core/src/error.rs
urlyy: > The deserialization can have BufferOutOfBound, maybe deserve an error type

Okay, errors that occur multiple times deserve their own enum.

rust/fory-core/src/buffer.rs
chaokunyang: > It is actually equivalent to:
> 
> ```rust
> match option {
>     Some(v) => Ok(v),
>     None => Err((|| E)()),
> }
> ```
> 
> When the Option is Some, the closure is never called or even allocated — it’s just a zero-cost inline conditional branch.

I don't think so, the closure will always be allocated, it just don't get executed when match to `Some`

urlyy: I made a simple demo:
```rust
use anyhow::Error;
use rand::Rng;

#[inline(never)]
fn test_ok_or_else(slice: &[u8]) -> Result<u32, Error> {
    let b0 = *slice
        .first()
        .ok_or_else(|| Error::msg("unexpected end of buffer"))? as u32;
    Ok(b0)
}

fn main() {
    let mut rng = rand::rng();
    let empty = rng.random_bool(0.5);
    let slice: &[u8] = if empty { &[] } else { &[42] };
    let _ = test_ok_or_else(slice);
}
```

image

So `let b0 = *slice.first().ok_or_else(|| Error::msg("unexpected end of buffer"))?;` is equivalent to 
```rust
let b0 = match slice.first() {
    Some(&byte) => byte as u32,
    None => return Err(Error::msg("unexpected end of buffer")),
};
```

rust/fory-core/src/error.rs
urlyy: > it looks a little strange that an Error type named as `Wrapped`

I just want to use `anyhow` to transform Error. And this is actually strange and maybe low-perf. I will remove this , just add `From` when need to transform other Error to fory-Error. Here is a demo:
```rust
use thiserror::Error;
use std::io;

#[derive(Error, Debug)]
enum MyError {
    #[error("IO error: {0}")]
    Io(#[from] io::Error),

    #[error("Other error: {0}")]
    Msg(String),
}

fn read_file() -> Result<String, MyError> {
    let content = std::fs::read_to_string("foo.txt")?;
    Ok(content)
}
```

rust/fory-core/src/error.rs
urlyy: > don't use abbr here. abbr should be used for internal functions/code only. This is a public API

OK, just `Message`.

rust/fory-core/src/buffer.rs
urlyy: Okay, I will extract a private API for boundary checking.

rust/fory-core/src/error.rs
chaokunyang: How about name it to `Unknown`. All known error in fory should have special error type

urlyy: > How about name it to `Unknown`. All known error in fory should have special error type

you are right!

rust/fory-core/src/error.rs
```diff
@@ -15,38 +15,98 @@
...
+    ///
+    /// # Example
+    /// ```
+    /// use fory_core::error::Error;
+    ///
+    /// let err = Error::msg("Something went wrong");
+    /// let err = Error::msg(format!("ID:{} not found", 1));
+    /// ```
+    #[inline(always)]
+    pub fn msg<S: Into<Cow<'static, str>>>(s: S) -> Self {
```
chaokunyang: should be renamed to `unknown`

rust/fory-core/src/error.rs
```diff
@@ -15,38 +15,98 @@
...
-    #[error("Fory on Rust not support Ref type")]
-    Ref,
+    #[error("Type mismatch: type_a = {0}, type_b = {1}")]
+    TypeMismatch(u32, u32),
+
+    #[error("Buffer out of bound: {0} + {1} > {2}")]
+    BufferOutOfBound(usize, usize, usize),
+
+    #[error("IO error: {0}")]
+    Io(#[from] io::Error),
```
chaokunyang, line 33: Some better enum value name example:
```c++
enum class StatusCode : char {
  OK = 0,
  OutOfMemory = 1,
  OutOfBound = 2,
  KeyError = 3,
  TypeError = 4,
  Invalid = 5,
  IOError = 6,
  UnknownError = 7,
};
```

chaokunyang:
```rust
pub enum Error {
    Protobuf(ProtobufError),
    EndOfStream,
    UnknownEnum {
        name: String,
    },
    UnknownEnumValue {
        value: i32,
    },
    UnknownMessage {
        name: String,
    },
    BadWireType {
        wire_type: WireType,
    },
    BadDefaultValue {
        default_value: String,
    },
    Custom {
        message: String,
    },
}
```

https://docs.rs/serde-protobuf/latest/serde_protobuf/error/enum.Error.html FYI

rust/fory-core/src/meta/meta_string.rs
```diff
@@ -77,8 +76,9 @@ impl MetaString {
     ) -> Result<Self, Error> {
         let mut strip_last_char = false;
         if encoding != Encoding::Utf8 {
-            ensure!(!bytes.is_empty(), anyhow!("Encoded data cannot be empty"));
-
+            if bytes.is_empty() {
+                bail!("Encoded data cannot be empty");
```
chaokunyang: this should return `EncodeError`

rust/fory-core/src/meta/meta_string.rs
```diff
@@ -260,14 +258,12 @@ impl MetaStringEncoder {
...
-                "Meta string is too long, max:{SHORT_MAX_VALUE}, current:{}",
-                input.len()
-            )
+            "Meta string is too long, max:{SHORT_MAX_VALUE}, current:{}",
+            input.len()
         );
         ensure!(
             encoding == Encoding::Utf8 || self.is_latin(input),
-            anyhow!("Non-ASCII characters in meta string are not allowed")
+            Error::msg("Non-ASCII characters in meta string are not allowed")
```
chaokunyang: ditto, return `EncodeError`

rust/fory-core/src/meta/meta_string.rs
```diff
@@ -429,8 +425,8 @@ impl MetaStringEncoder {
                     } else if c == self.special_char2 {
                         Ok(63)
                     } else {
-                        Err(anyhow!(
-                            "Invalid character value for LOWER_SPECIAL decoding: {c:?}"
+                        Err(Error::msg(
```
chaokunyang: ditto

rust/fory-core/src/fory.rs
```diff
@@ -312,27 +311,27 @@ impl Fory {
 
     fn read_head(&self, reader: &mut Reader) -> Result<bool, Error> {
         if self.xlang {
-            let magic_numer = reader.read_u16();
+            let magic_numer = reader.read_u16()?;
             ensure!(
                 magic_numer == MAGIC_NUMBER,
-                anyhow!(
+                Error::msg(format!(
```
chaokunyang: return `InvalidData` error

rust/fory-core/src/serializer/rc.rs
```diff
@@ -20,45 +20,58 @@ use crate::fory::Fory;
...
+        T::fory_write_type_info(fory, context, is_field)
     }
 
     fn fory_read(fory: &Fory, context: &mut ReadContext, is_field: bool) -> Result<Self, Error> {
-        let ref_flag = context.ref_reader.read_ref_flag(&mut context.reader);
-
+        let ref_flag = context.ref_reader.read_ref_flag(&mut context.reader)?;
         match ref_flag {
-            RefFlag::Null => Err(anyhow!("Rc cannot be null").into()),
+            RefFlag::Null => Err(Error::msg("Rc cannot be null")),
```
chaokunyang: this should return `InvalidRef` error

rust/fory-core/src/serializer/rc.rs
```diff
@@ -20,45 +20,58 @@ use crate::fory::Fory;
...
             RefFlag::Ref => {
-                let ref_id = context.ref_reader.read_ref_id(&mut context.reader);
+                let ref_id = context.ref_reader.read_ref_id(&mut context.reader)?;
                 context
                     .ref_reader
                     .get_rc_ref::<T>(ref_id)
-                    .ok_or_else(|| anyhow!("Rc reference {} not found", ref_id).into())
+                    .ok_or_else(|| Error::msg(format!("Rc reference {ref_id} not found")))
```
chaokunyang: this should return `InvalidRef` error

rust/fory-derive/src/object/derive_enum.rs
```diff
@@ -53,26 +53,26 @@ pub fn gen_write_data(data_enum: &DataEnum) -> TokenStream {
...
     let variant_values: Vec<_> = (0..variant_idents.len()).map(|v| v as u32).collect();
     quote! {
-        let ordinal = context.reader.read_varuint32();
+        let ordinal = context.reader.read_varuint32()?;
         match ordinal {
            #(
                #variant_values => Ok(Self::#variant_idents),
            )*
-           _ => panic!("unknown value"),
+           _ => fory_core::bail!("unknown enum value"),
```
chaokunyang: this should return `UnknownEnum` error

rust/fory-derive/src/object/read.rs
```diff
@@ -108,16 +108,16 @@ fn gen_read_field(field: &Field, private_ident: &Ident) -> TokenStream {
...
+                    bail!("Expected NotNullValue for trait object field");
                 }
 
-                let fory_type_id = context.reader.read_varuint32();
+                let fory_type_id = context.reader.read_varuint32()?;
 
                 let harness = fory.get_type_resolver()
                     .get_harness(fory_type_id)
-                    .expect("Type not registered for trait object field");
+                    .ok_or_else(|| Error::msg("Type not registered for trait object field"))?;
```
chaokunyang: this should return `TypeError`

rust/fory-derive/src/object/read.rs
```diff
@@ -254,14 +254,14 @@ fn gen_read_compatible_match_arm_body(field: &Field, var_name: &Ident) -> TokenS
             let from_any_fn = format_ident!("from_any_internal_{}", trait_name);
             let helper_mod = format_ident!("__fory_trait_helpers_{}", trait_name);
             quote! {
-                let ref_flag = context.reader.read_i8();
+                let ref_flag = context.reader.read_i8()?;
                 if ref_flag != fory_core::types::RefFlag::NotNullValue as i8 {
-                    panic!("Expected NotNullValue for trait object field");
+                    bail!("Expected NotNullValue for trait object field");
```
chaokunyang: this should return `InvalidData` error with detailed message

rust/fory-derive/src/object/read.rs
```diff
@@ -254,14 +254,14 @@ fn gen_read_compatible_match_arm_body(field: &Field, var_name: &Ident) -> TokenS
...
                 }
-                let fory_type_id = context.reader.read_varuint32();
+                let fory_type_id = context.reader.read_varuint32()?;
                 let harness = fory.get_type_resolver()
                     .get_harness(fory_type_id)
-                    .expect("Type not registered for trait object field");
+                    .ok_or_else(|| Error::msg("Type not registered for trait object field"))?;
```
chaokunyang: return `TypeError` with detailed message

rust/fory-core/src/meta/type_meta.rs
```diff
@@ -160,35 +163,35 @@ impl FieldInfo {
             0x00 => Ok(Encoding::Utf8),
             0x01 => Ok(Encoding::AllToLowerSpecial),
             0x02 => Ok(Encoding::LowerUpperDigitSpecial),
-            _ => Err(anyhow!(
-                "Unsupported encoding of field name in type meta, value:{value}"
+            _ => Err(Error::msg(
```
chaokunyang: return `EncodingError`

rust/fory-core/src/error.rs
chaokunyang: you can provide static func like serde:
```rust
pub trait Error: Sized + StdError {
    fn custom<T>(msg: T) -> Self
    where
        T: Display;

    fn invalid_type(unexp: Unexpected<'_>, exp: &dyn Expected) -> Self { ... }
    fn invalid_value(unexp: Unexpected<'_>, exp: &dyn Expected) -> Self { ... }
    fn invalid_length(len: usize, exp: &dyn Expected) -> Self { ... }
    fn unknown_variant(variant: &str, expected: &'static [&'static str]) -> Self { ... }
    fn unknown_field(field: &str, expected: &'static [&'static str]) -> Self { ... }
    fn missing_field(field: &'static str) -> Self { ... }
    fn duplicate_field(field: &'static str) -> Self { ... }
}
```

rust/fory-core/src/error.rs
chaokunyang: should be IoError instead of `Io`. The principle is that if the name ifselt can't express the error, you should ass `Error` as suffix instead

rust/fory-core/src/error.rs
```diff
@@ -15,38 +15,118 @@
...
+    #[error("{0}")]
+    TypeError(Cow<'static, str>),
 
-    #[error(transparent)]
-    Other(#[from] anyhow::Error),
+    #[error("{0}")]
+    EncodingError(Cow<'static, str>),
+
+    #[error("{0}")]
+    DepthExceeded(Cow<'static, str>),
```
chaokunyang:
```suggestion
    DepthExceed(Cow<'static, str>),
```

PR #2773, opened 2025-10-15
feat(Rust): support deserialize by populate passed mut ref

### LINKED ISSUE #2552: [Rust] Support deserialize by populate passed mut ref
### Feature Request

Currently fory rust provide following deserialize API:
```rust
    pub fn deserialize<T: Serializer>(&self, bf: &[u8]) -> Result<T, Error> {
        let mut reader = Reader::new(bf);
        let meta_offset = self.read_head(&mut reader)?;
        let mut context = ReadContext::new(self, reader);
        if meta_offset > 0 {
            context.load_meta(meta_offset as usize);
        }
        <T as Serializer>::deserialize(&mut context)
    }
```

It relies on compiler `Return Value Optimization (RVO)`, which won't work always. And the compiler will copy small struct always.

` pub fn deserialize<T: Serializer>(&self, bf: &[u8]) -> Result<T, Error>` API may introduce extra copy.

Add a new API to populate passed struct directly.

```rust
pub fn deserialize_into<T: Serializer>(&self, bf: &[u8], output: &mut T) -> Result<(), Error> {
    let mut reader = Reader::new(bf);
    let meta_offset = self.read_head(&mut reader)?;
    let mut context = ReadContext::new(self, reader);
    if meta_offset > 0 {
        context.load_meta(meta_offset as usize);
    }
    <T as Serializer>::deserialize_into(&mut context, output)
}
```

_No response_

### Additional context

_No response_

Tinuvile:
## Why?

Currently fory rust deserialize API relies on compiler `Return Value Optimization (RVO)`, which won't work always. And the compiler will copy small struct always.

### `fory.rs`

add `deserialize_into` and `deserialize_into_with_context` method, the new API directly modifies the passed-in `output` parameter and returns `()` instead of the object itself, thus avoiding the copying of the return value.

### `serializers`

in `Serializer` trait, add `fory_read_into` and `fory_read_data_into`, and it has been implemented for all types.

**It seems this optimization will only be effective for primitive types and have limited effect on complex containers or dynamic types. I'm not sure if I'm on the right track right now, so I submitted this draft PR to seek some feedback.**

## Related issues

close #2552 

### DIFF
--- a/rust/fory/src/lib.rs
+++ b/rust/fory/src/lib.rs
@@ -731,6 +731,13 @@
 //!         Ok(Self { value, name })
 //!     }
 //!
+//!     fn fory_read_data_into(context: &mut ReadContext, is_field: bool, output: &mut Self) -> Result<(), Error> {
+//!         output.value = context.reader.read_i32()?;
+//!         let len = context.reader.read_varuint32()? as usize;
+//!         output.name = context.reader.read_utf8_string(len)?;
+//!         Ok(())
+//!     }
+//!
 //!     fn fory_type_id_dyn(&self, type_resolver: &TypeResolver) -> Result<u32, Error> {
 //!         Self::fory_get_type_id(type_resolver)
 //!     }


chaokunyang:
Hi @Tinuvile , I refactored `Serializer` trait in #2774 , could you merge main branch and resolve the conflict?

chaokunyang:
Hi @Tinuvile , does this PR improve performance for struct deserialization?

And for types which is not struct, maybe we don't need to support populate? We can limit the `T` to bound `StructSerializer`. In this way, only struct/enum can be passed in

Tinuvile:
Thanks for the feedback. I’ll check it again.

PR #2782, opened 2025-10-20
feat(rust): use rc instead of arc for type meta for faster performance

chaokunyang:
## What does this PR do?

use rc instead of arc for type meta for faster performance


urlyy:
Looks Great!

PR #2791, opened 2025-10-20
perf(Rust): remove clone()/to_owned() on MetaString/MetaStringBytes in MetaStringResolver to improve performance && fix xlang test

### LINKED ISSUE #2762: [Rust] Remove clone() in MetaStringWriterResolver && MetaStringReaderResolver
### Feature Request

_No response_

_No response_

_No response_

_No response_

### Additional context

_No response_

urlyy:
## Why?
`clone()/to_owned()` on `MetaString/MetaStringBytes` is expensive.

## What does this PR do?
1. use `Rc` in  `MetaStringWriterResolver `.
2. aligned `metastring_resolver` logic with Java

## Related issues
close #2762

### DIFF
--- a/java/fory-core/src/test/java/org/apache/fory/serializer/ProtocolInteroperabilityTest.java
+++ b/java/fory-core/src/test/java/org/apache/fory/serializer/ProtocolInteroperabilityTest.java
@@ -217,7 +217,19 @@ public class ProtocolInteroperabilityTest extends ForyTestBase {
   public void testComplexCollection(Fory fory, Fory foryJIT) {
     CollectionSerializersTest.CollectionFieldsClass o =
         CollectionSerializersTest.createCollectionFieldsObject();
-    roundCheck(fory, foryJIT, o, Object::toString);
+    roundCheck(
+        fory,
+        foryJIT,
+        o,
+        obj -> {
+          try {
+            CollectionSerializersTest.testCollectionFieldsObjectEqual(
+                (CollectionSerializersTest.CollectionFieldsClass) obj, o);
+            return "OK";
+          } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
+          }
+        });
   }
 
   @Test(dataProvider = "fory")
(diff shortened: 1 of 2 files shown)


urlyy:
It is hard to not use Rc in metastring_writer_resolver.

urlyy:
@chaokunyang 
Currently has a usage limitation: register must be called in a specific order — field type must be registered first.

This is because `register()` creates `type_info`, and if `struct_a` contains a field of type `struct_b`, then calling `register(struct_a)` before `struct_b` will cause an error, since the `type_resolver` won’t be able to find the `type_info` for `struct_b` in `TypeInfo::new::<struct_a>`.

Perhaps you need to provide like a `fory.register_chain().register::<TypeA>().register::<TypeB>().finish()` to handle this.

In addition, I plan to perform `struct_hash computation` inside `register()` and `finish()`, which also requires that all field types can be retrievable from the `type_resolver` before the struct is being registered.

chaokunyang (approved):
LGTM

chaokunyang:
> @chaokunyang Currently has a usage limitation: register must be called in a specific order — field type must be registered first.

We can do a two-stage type resolve:
```rust
pub fn serialize<T: Serializer>(&self, record: &T) -> Result<Vec<u8>, Error> {
        let pool = self.write_context_pool.get_or_init(|| {
            let type_resolver = self.type_resolver.clone();
            let compatible = self.compatible;
            let share_meta = self.share_meta;
            let compress_string = self.compress_string;
            let xlang = self.xlang;

            let factory = move || {
                let writer = Writer::default();
                WriteContext::new(
                    writer,
                    type_resolver.clone(),
                    compatible,
                    share_meta,
                    compress_string,
                    xlang,
                )
            };
            Pool::new(factory)
        });
        let mut context = pool.get();
        let result = self.serialize_with_context(record, &mut context)?;
        context.reset();
        pool.put(context);
        Ok(result)
    }
```
@urlyy How about creating a new method named `finalizeRegistration` method on `TypeResolver`, which returns a new TypeResolver with correct typeinfo.
When creating `WriteContext`, we pass resulf of `finalizeRegistration` instead.

PR #2798, opened 2025-10-21
fix(Rust): Make registration lazy to avoid errors caused by missing dependent types due to registration order

### LINKED ISSUE #2794: [Rust] Register must be called in a specific order
### Search before asking

### Version

Latest

### Component(s)

Rust

### Minimal reproduce step

```rust
#[test]
fn test_register() {
    #[derive(ForyObject, Debug)]
    struct TypeA { }

    #[derive(ForyObject, Debug)]
    struct TypeB { 
        a: TypeA,
    }
    fory.register::<TypeB>(101).unwrap();
    fory.register::<TypeA>(100).unwrap();
}
```

pass ut

### What did you see instead?

```
thread 'test_register' panicked at tests/tests/mod.rs:37:33:
called `Result::unwrap()` on an `Err` value: TypeError("TypeId TypeId(0xa7cff65b542f9885a4fa70e95938663e) not found in type_id_index, maybe you forgot to register some types")
stack backtrace:
   0: rust_begin_unwind
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs:695:5
   1: core::panicking::panic_fmt
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/panicking.rs:75:14
   2: core::result::unwrap_failed
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/result.rs:1704:5
   3: core::result::Result<T,E>::unwrap
             at /Users/urlyy/environment/rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:1109:23
   4: mod::test_register
             at ./tests/mod.rs:37:5
   5: mod::test_register::{{closure}}
             at ./tests/mod.rs:28:19
   6: core::ops::function::FnOnce::call_once
             at /Users/urlyy/environment/rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    test_register

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 46 filtered out; finished in 0.03s
```

### Anything Else?

_No response_

urlyy:
## What does this PR do?
1. Changed `register` to a lazy form, and `finalize_registration()` is only called for the actual registration during `serialize()`, `deserialize()`, or `new_from_fory()`.

2. Adjusted the `register_xx` methods of `TypeResolver` and exposed lazy `register` methods externally.

3. To distinguish between `Context::TypeResolver` and `Fory::TypeResolver`, a new struct `SharedTypeResolver` was introduced. It is responsible for collecting lazy registration information and finalizing it. Using `UnsafeCell` avoids needing `mut fory`, and `Once` ensures finalization happens only once. After finalization, `get_clone()` can return a clone of the internal `TypeResolver` to be inserted into `Context`.

4. Use BTreeMap to store lazy registration information, because HashMap is unordered, iterating over its values(Types) can produce different traversal orders for the same register sequence. Using a BTreeMap makes my testing more convenient.

5. modify `struct::fory_fields_info()` logic, when field_type not found in TypeResolver, will register it immediately and `retry` to get its fory_type_id. Use a HashSet to record Types have been registered.

## Related issues
#2793
close #2794

The `Read/WriteContext::new_from_fory()` will return `Result<Read/WriteContext, Error>`

### DIFF
--- a/rust/fory-core/src/resolver/type_resolver.rs
+++ b/rust/fory-core/src/resolver/type_resolver.rs
@@ -27,6 +27,8 @@ use crate::{Reader, TypeId};
 use std::collections::{HashSet, LinkedList};
 use std::rc::Rc;
 use std::{any::Any, collections::HashMap};
+use std::cell::UnsafeCell;
+use std::sync::{Arc, Once};
 
 type WriteFn = fn(
     &dyn Any,
@@ -310,6 +312,9 @@ pub struct TypeResolver {
     // Fast lookup by numeric ID for common types
     type_id_index: Vec<u32>,
     compatible: bool,
+    // only for registry
+    actual_type_id_map: HashMap<std::any::TypeId, u32>,
+    registry: HashMap<std::any::TypeId, Arc<dyn Fn(&mut Self) -> Result<(), Error>>>,
 }
 
 // Safety: TypeResolver instances are only shared through higher-level synchronization that
@@ -329,6 +334,8 @@ impl Default for TypeResolver {
             type_info_map_by_ms_name: HashMap::new(),
             type_id_index: Vec::new(),
             compatible: false,
+            actual_type_id_map: HashMap::new(),
+            registry: HashMap::new(),
         };
         registry.register_builtin_types().unwrap();
         registry
@@ -467,7 +474,17 @@ impl TypeResolver {
         &mut self,
         id: u32,
     ) -> Result<(), Error> {
-        self.register::<T>(id, &EMPTY_STRING, &EMPTY_STRING)
+        let type_id = std::any::TypeId::of::<T>();
+        let actual_type_id = T::fory_actual_type_id(id, false, self.compatible);
+        self.actual_type_id_map.insert(
+            type_id,
+            actual_type_id
+        );
+        self.registry.insert(
+            type_id,
+            Arc::new(move |s: &mut Self| s.register::<T>(id, &EMPTY_STRING, &EMPTY_STRING)),
+        );
+        Ok(())
     }
 
     pub fn register_by_namespace<T: 'static + StructSerializer + Serializer + ForyDefault>(
@@ -475,7 +492,19 @@ impl TypeResolver {
         namespace: &str,
         type_name: &str,
     ) -> Result<(), Error> {
-        self.register::<T>(0, namespace, type_name)
+        let type_id = std::any::TypeId::of::<T>();
+        let actual_type_id = T::fory_actual_type_id(0, true, self.compatible);
+        self.actual_type_id_map.insert(
+            type_id,
+            actual_type_id,
+        );
+        let namespace = namespace.to_string();
+        let type_name = type_name.to_string();
+        self.registry.insert(
+            type_id,
+            Arc::new(move |s: &mut Self| s.register::<T>(0, &namespace, &type_name)),
+        );
+        Ok(())
     }
 
     fn register<T: StructSerializer + Serializer + ForyDefault>(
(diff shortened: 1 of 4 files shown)


rust/fory-core/src/serializer/core.rs
```diff
@@ -1322,7 +1322,7 @@ pub trait StructSerializer: Serializer + 'static {
     /// - Default returns empty vector (no fields)
     /// - **Do not implement** for user types with custom serialization (EXT types)
     #[allow(unused_variables)]
-    fn fory_fields_info(type_resolver: &TypeResolver) -> Result<Vec<FieldInfo>, Error> {
+    fn fory_fields_info(type_resolver: &mut TypeResolver) -> Result<Vec<FieldInfo>, Error> {
```
chaokunyang: should be immutable from the semantics

chaokunyang:
@urlyy This PR make API hard to understand. The register order should not be important if then specify the type id/name. And introduce new SharedTypeResolver make type system more complex.

 How about:
1. Create a new field `HashMap<std::any::TypeId, fory_type_id/name>` in TypeResolver
2. when register a type/serializer, we just store the id/name into that map.
3. provide a new method `get_type_id` in TypeResolver, which  query the current TypeInfo map first, if not found, then query from the new map
4. The `fory_type_id/fields_info` can invoke `get_type_id` to get the type id info
5. Remove clone from `TypeResolver`, and provide a new `build/finalize_registration` method to return an immutable TypeResolver. we can create typeinfo here for new registered types, and construct TypeResolver directly

rust/fory-derive/src/object/util.rs
```diff
@@ -394,7 +394,13 @@ pub(super) fn generic_tree_to_tokens(node: &TypeNode) -> TokenStream {
         ts
     } else {
         quote! {
-            <#ty as fory_core::serializer::Serializer>::fory_get_type_id(type_resolver)?
+            match <#ty as fory_core::serializer::Serializer>::fory_get_type_id(type_resolver) {
+                Ok(id) => id,
+                Err(_) => {
+                    type_resolver.do_register(std::any::TypeId::of::<#ty>())?;
```
chaokunyang: looks a little weird to me

rust/tests/tests/compatible/test_basic_type.rs
```diff
@@ -512,20 +512,20 @@ fn auto_conv() {
...
     // serialize_non-null
     let writer = Writer::default();
-    let mut write_context = WriteContext::new_from_fory(writer, &fory);
+    let mut write_context = WriteContext::new_from_fory(writer, &fory).unwrap();
     serialize_non_null(&fory, &mut write_context);
     // deserialize_nullable
     let bytes = write_context.writer.dump();
     let reader = Reader::new(bytes.as_slice());
-    let mut read_context: ReadContext = ReadContext::new_from_fory(reader, &fory);
+    let mut read_context: ReadContext = ReadContext::new_from_fory(reader, &fory).unwrap();
```
chaokunyang: this method should not return error. it should always succeed

rust/fory-core/src/resolver/type_resolver.rs
```diff
@@ -813,7 +934,117 @@ impl TypeResolver {
...
+            ))),
+        }
+    }
+
     pub(crate) fn set_compatible(&mut self, compatible: bool) {
         self.compatible = compatible;
     }
 }
+
+pub struct SharedTypeResolver {
```
chaokunyang, line 879: this is not allowed, we won't introduce a new TypeResolver/ It's very weird

urlyy: I will open a new pr to handle this issue

Issue #2622, opened 2025-09-17
[Java] Refactor ObjectStreamSerializer to use meta shared compatible serializer and remove key-value compatible serializer


chaokunyang:
### Feature Request

  Refactor ObjectStreamSerializer to use meta-shared compatible serializer and **completely remove** `CompatibleSerializer` and `FieldResolver`.

  ### Is your feature request related to a problem? Please describe

  Currently [ObjectStreamSerializer](java/fory-core/src/main/java/org/apache/fory/serializer/ObjectStreamSerializer.java) uses
  [CompatibleSerializer](java/fory-core/src/main/java/org/apache/fory/serializer/CompatibleSerializer.java) for forward/backward compatible serialization.

  Problems with `CompatibleSerializer`:
  - **Larger serialized size**: Writes field metadata (encoded field info) for every field during each serialization
  - **Duplicated metadata**: When serializing multiple objects of same type, field metadata is repeated
  - **Deprecated approach**: Should be replaced with meta-shared mechanism

  ### Describe the solution you'd like

  **This is a breaking change** - no backward compatibility with old format.

  #### Key Design: Layer Marker Classes

  For a class hierarchy `C extends B extends A`, generate marker classes using Janino:
  - `A$ForyLayer$0` → key in `metaContext.classMap` for A's layer
  - `B$ForyLayer$1` → key for B's layer
  - `C$ForyLayer$2` → key for C's layer

  Each marker class implements `CurrentLayerMarker` interface, serving as unique keys in existing `metaContext.classMap.put(markerClass)`. **No changes to MetaContext 
  needed.**

  #### Files to DELETE

  | File | Reason |
  |------|--------|
  | `CompatibleSerializer.java` | Replaced by meta-shared approach |
  | `FieldResolver.java` | Only used by CompatibleSerializer |
  | `CompatibleCodecBuilder.java` | JIT for CompatibleSerializer |

  #### Files to CREATE

  | File | Purpose |
  |------|---------|
  | `CurrentLayerMarker.java` | Marker interface for layer wrappers |
  | `LayerMarkerClassGenerator.java` | Generates marker classes for each layer using Janino |
  | `MetaSharedLayerSerializer.java` | Single-layer meta-shared serializer |
  | `MetaSharedLayerCodecBuilder.java` | JIT builder for layer serializer |

  #### Files to MODIFY

  | File | Changes |
  |------|---------|
[...]

PR #2799, opened 2025-10-22
refactor: objectstreamserializer to use meta shared compatible

kitalkuyo-gita:
## Related issues

Related to [issue-2622](https://github.com/apache/fory/issues/2622)

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/config/Config.java
+++ b/java/fory-core/src/main/java/org/apache/fory/config/Config.java
@@ -297,8 +297,6 @@ public class Config implements Serializable {
     return scalaOptimizationEnabled;
   }
 
-
-
   public boolean isForyDebugOutputEnabled() {
     return foryDebugOutputEnabled;
   }
--- a/java/fory-core/src/main/java/org/apache/fory/serializer/ObjectSerializer.java
+++ b/java/fory-core/src/main/java/org/apache/fory/serializer/ObjectSerializer.java
@@ -98,7 +98,8 @@ public final class ObjectSerializer<T> extends AbstractObjectSerializer<T> {
    * @param resolveParent Whether to resolve parent class
    * @param excludedFields Set of field names to exclude from serialization (can be null)
    */
-  public ObjectSerializer(Fory fory, Class<T> cls, boolean resolveParent, Set<String> excludedFields) {
+  public ObjectSerializer(
+      Fory fory, Class<T> cls, boolean resolveParent, Set<String> excludedFields) {
     super(fory, cls);
     binding = SerializationBinding.createBinding(fory);
     // avoid recursive building serializers.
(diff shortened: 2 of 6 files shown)


chaokunyang:
Great! Could we remove https://github.com/apache/fory/blob/main/java/fory-core/src/main/java/org/apache/fory/serializer/CompatibleSerializer.java  and https://github.com/apache/fory/blob/main/java/fory-core/src/main/java/org/apache/fory/resolver/FieldResolver.java in this PR too?

kitalkuyo-gita:
> Great! Could we remove https://github.com/apache/fory/blob/main/java/fory-core/src/main/java/org/apache/fory/serializer/CompatibleSerializer.java and https://github.com/apache/fory/blob/main/java/fory-core/src/main/java/org/apache/fory/resolver/FieldResolver.java in this PR too?

I don't quite understand, why do you want to delete FieldResolver?

java/fory-core/src/main/java/org/apache/fory/config/ForyBuilder.java
```diff
@@ -335,6 +336,17 @@ public ForyBuilder withScopedMetaShare(boolean scoped) {
     return this;
   }
 
+  /**
+   * Whether use meta share for ObjectStreamSerializer.
+   * This reduces serialization size by using ObjectSerializer instead of CompatibleSerializer
+   * for classes that implement custom writeObject/readObject methods.
+   * Only effective when metaShareEnabled is true.
+   */
+  public ForyBuilder withMetaShareForObjectStream(boolean enable) {
```
chaokunyang: If we remove `CompatibleSerializer`, we don't need this option, meta-shared compatible mode will be the only mode

kitalkuyo-gita:
> Great! Could we remove https://github.com/apache/fory/blob/main/java/fory-core/src/main/java/org/apache/fory/serializer/CompatibleSerializer.java and https://github.com/apache/fory/blob/main/java/fory-core/src/main/java/org/apache/fory/resolver/FieldResolver.java in this PR too?

Hello, @chaokunyang . I see these two classes are marked as deprecated. However, I believe further discussion is needed regarding their removal. Completely removing these classes would impact backward compatibility. Therefore, I recommend retaining the CompatibleSerializer in non-meta-shared mode and using the ObjectStreamMetaSharedSerializerAdapter only in meta-shared mode.

Restore the conditional for useMetaShare in SlotsInfo.

Non-meta-shared mode: Use the CompatibleSerializer.

Meta-shared mode: Use the ObjectStreamMetaSharedSerializerAdapter.

What do you think about this?

java/fory-core/src/main/java/org/apache/fory/config/ForyBuilder.java
kitalkuyo-gita: solved

kitalkuyo-gita:
While fixing CI, I encountered a problem. ```LinkedBlockingQueue``` and ```ArrayBlockingQueue``` both have custom ```writeObject/readObject``` methods. According to requireJavaSerialization , they use ```ObjectStreamSerializer ```. However, the ```LinkedBlockingQueue.readObject``` method calls the ```add() ```method during deserialization, which requires a lock. If the serialization lock state is incorrect, a deadlock will occur. The key issue is that LinkedBlockingQueue and ArrayBlockingQueue are not suitable for using ObjectStreamSerializer . JavaSerializer (full JDK serialization) should be used instead.

java/fory-core/src/main/java/org/apache/fory/config/Config.java
```diff
@@ -297,6 +297,14 @@ public boolean isScalaOptimizationEnabled() {
     return scalaOptimizationEnabled;
   }
 
+  /**
+   * Whether use meta share for ObjectStreamSerializer. This reduces serialization size by using
+   * ObjectSerializer instead of CompatibleSerializer. Only effective when metaShareEnabled is true.
+   */
+  public boolean isMetaShareForObjectStreamEnabled() {
```
chaokunyang: We don't need this, we can just check whether fory compatible mode is set to `CompatibleMode` in `ObjectStreamSerializer`

kitalkuyo-gita:
@chaokunyang Hello, it's ready.

chaokunyang:
Finished in #3034, thanks for your contribution

PR #2807, opened 2025-10-22
perf(rust): optimize rust performance by remove copy simd and add more inline hints

chaokunyang:
## What does this PR do?

optimize rust performance by remove copy simd and add more inline hints


urlyy (approved):

urlyy:
So SIMD actually has worse performance, 😂

chaokunyang:
> So SIMD actually has worse performance, 😂

just for copy and small arrays

PR #2810, opened 2025-10-22
perf(Rust): Enchance performance

theweipeng:
## What does this PR do?

- Replacing `Mutex` by Spainlock which is implemented by CAS usually more faster
- Replacing `dump` function by detach which could avoid memcpy

### DIFF
--- a/rust/fory-core/src/util.rs
+++ b/rust/fory-core/src/util.rs
@@ -149,7 +149,7 @@ impl<T> Spinlock<T> {
         }
     }
 
-    pub fn lock(&self) -> SpinlockGuard<T> {
+    pub fn lock(&self) -> SpinlockGuard<'_, T> {
         let mut spins = 0;
         while self
             .flag


rust/fory-core/src/util.rs
```diff
@@ -129,3 +131,32 @@ pub fn get_ext_actual_type_id(type_id: u32, register_by_name: bool) -> u32 {
...
         }
 }
+
+pub struct Spinlock<T> {
+    data: UnsafeCell<T>,
+    flag: AtomicBool,
+}
+
+unsafe impl<T: Send + Sync> Send for Spinlock<T> {}
+unsafe impl<T: Send + Sync> Sync for Spinlock<T> {}
```
chaokunyang:
```suggestion
unsafe impl<T: Send> Send for Spinlock<T> {}
unsafe impl<T: Send> Sync for Spinlock<T> {}
```

rust/fory-core/src/util.rs
```diff
@@ -129,3 +131,32 @@ pub fn get_ext_actual_type_id(type_id: u32, register_by_name: bool) -> u32 {
...
+    pub fn new(data: T) -> Self {
+        Spinlock {
+            data: UnsafeCell::new(data),
+            flag: AtomicBool::new(false),
+        }
+    }
+
+    pub fn lock(&self) -> &mut T {
+        while self.flag.compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed).is_err() {
+            std::hint::spin_loop();
```
chaokunyang: We need yield here:
```rust
    pub fn lock(&self) -> SpinlockGuard<'_, T> {
        let mut spins = 0;
        while self
            .flag
            .compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed)
            .is_err()
        {
            // Spin for a few iterations
            if spins < 10 {
                std::hint::spin_loop();
                spins += 1;
            } else {
                // Then yield to the scheduler
                thread::yield_now();
                spins = 0; // reset spin counter
            }
        }
        SpinlockGuard { lock: self }
    }
```

rust/fory-core/src/util.rs
```diff
@@ -129,3 +131,32 @@ pub fn get_ext_actual_type_id(type_id: u32, register_by_name: bool) -> u32 {
...
+
+impl<T> Spinlock<T> {
+    pub fn new(data: T) -> Self {
+        Spinlock {
+            data: UnsafeCell::new(data),
+            flag: AtomicBool::new(false),
+        }
+    }
+
+    pub fn lock(&self) -> &mut T {
```
chaokunyang: do we need to return a guard, so the unlock can be invoked always:
```rust
pub struct SpinlockGuard<'a, T> {
    lock: &'a Spinlock<T>,
}

impl<'a, T> Drop for SpinlockGuard<'a, T> {
    fn drop(&mut self) {
        self.lock.unlock();
    }
}

impl<'a, T> std::ops::Deref for SpinlockGuard<'a, T> {
    type Target = T;
    fn deref(&self) -> &Self::Target {
        unsafe { &*self.lock.data.get() }
    }
}

impl<'a, T> std::ops::DerefMut for SpinlockGuard<'a, T> {
    fn deref_mut(&mut self) -> &mut Self::Target {
        unsafe { &mut *self.lock.data.get() }
    }
}
```

chaokunyang (approved):
LGTM

rust/fory-core/src/fory.rs
```diff
@@ -555,6 +554,7 @@ impl Fory {
         record: &T,
         context: &mut WriteContext,
     ) -> Result<Vec<u8>, Error> {
+        context.writer.attach_buffer(vec![]);
```
chaokunyang, line 557: This will allocate a new buffer everytime, which cause extra copy and allocation when serializing

PR #2814, opened 2025-10-23
fix(go): Adjust serialization of nullable value to align python behavior

junjiexh:
## Why?

This pr (#2766) introduce new serialization behavior of optional value in python

This pr (#2793) introduce new hash compute algorithms.

We need to align those behavior in go.

## What does this PR do?

- In go, nullable is same as 'referencable', so I reuse the referencable flag to indicate if a field is nullable. Also, for the string field, we always treat it as a nullable type but never write its reference.
- rewrite the hash compute function(computeStructHash), now this conform to the new algorithm in the spec.
- Fix a bug that will cause 'interface' type field to be non-nullable, which is not expected. 

- [x] Does this PR introduce any public API change? no
- [x] Does this PR introduce any binary protocol compatibility change? no


pandalee99 (approved):
LGTM

Issue #2722, opened 2025-10-08
[Rust] support circular reference xlang serialization


chaokunyang:
### Feature Request

In https://github.com/apache/fory/pull/2714, we supported circular reference serialization for rust, but we don't support it for xlang.

The xlang support for circular reference is not supported in that PR. Circular reference serialized by rust can only be deserialized by rust currently. Xlang support needs extra work to skip extra written meta.

### Is your feature request related to a problem? Please describe

circular reference xlang serialization should be supported for rust, it's supported between java/go/python xlang already.

### Describe the solution you'd like

Currently when serialize `RcWeak/ArcWeak`, we always serialize its type meta. But if the wrapped type is morphic, when serialize as a field, we should not write that meta.

And when being element of `Vec`, that meta should be written by Vec. When being K/V items for `HashMap`, that meta should be written by `HashMap` chunk protocol.

### Additional context

https://github.com/apache/fory/pull/2714

PR #2819, opened 2025-10-23
feat(rust): Support xlang ref type alignment

kitalkuyo-gita:
## Related issues

Related to [issue-2722](https://github.com/apache/fory/issues/2722)

### DIFF
--- a/rust/fory-core/src/serializer/map.rs
+++ b/rust/fory-core/src/serializer/map.rs
@@ -19,6 +19,7 @@ use crate::ensure;
 use crate::error::Error;
 use crate::resolver::context::{ReadContext, WriteContext};
 use crate::resolver::type_resolver::{TypeInfo, TypeResolver};
+use crate::serializer::util::read_basic_type_info;
 use crate::serializer::{ForyDefault, Serializer};
 use crate::types::{need_to_write_type_for_field, TypeId, SIZE_OF_REF_AND_TYPE};
 use std::collections::{BTreeMap, HashMap};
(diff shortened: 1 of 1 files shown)


chaokunyang:
@kitalkuyo-gita The serialization of circular reference in rust is already supported in current fory rust implementation. The xlang support is more about alignment of ref types and flags. Due to the type system gaps between java and rust, a type may be taken as reference in java, but not a reference rust. Then the deserialization in rust/java may fail. You may need to start with adding reference tests in `RustXlangTest.java` and `test_cross_language.rs`.

kitalkuyo-gita:
> @kitalkuyo-gita The serialization of circular reference in rust is already supported in current fory rust implementation. The xlang support is more about alignment of ref types and flags. Due to the type system gaps between java and rust, a type may be taken as reference in java, but not a reference rust. Then the deserialization in rust/java may fail. You may need to start with adding reference tests in `RustXlangTest.java` and `test_cross_language.rs`.

OK, thanks for your suggestion

rust/fory-core/src/serializer/core.rs
```diff
@@ -1217,6 +1230,70 @@ pub trait Serializer: 'static {
...
+    /// # Implementation Notes
+    ///
+    /// - Only affects behavior when `context.is_xlang()` is true
+    /// - Used in [`fory_write`] to determine RefFlag behavior
+    /// - Fory implements this for all built-in types
+    /// - User types should override this for proper xlang interop
+    ///
+    /// [`fory_write`]: Serializer::fory_write
+    #[inline(always)]
+    fn fory_is_xlang_ref_type() -> bool
```
chaokunyang: Remove this  method, all rust types are xlang serializable

kitalkuyo-gita: > Remove this method, all rust types are xlang serializable

Hello @chaokunyang , I'm thinking about whether it's possible to use the TypeId enumeration to distinguish types. 
For example, 
```
pub enum TypeId {
STRING, // Reference type
LIST, // Reference type
MAP, // Reference type
INT32_ARRAY,
// ...
}
```

Then, in fory_write(), automatically determine the type based on fory_static_type_id().

Example:
```
if write_ref_info {
    let is_ref_type = matches!(
        Self::fory_static_type_id(),
        TypeId::STRING | TypeId::LIST | TypeId::MAP | TypeId::SET
    );
    
    if context.is_xlang() && is_ref_type {
        context.writer.write_i8(RefFlag::RefValue as i8);
    }
}
```

This requires maintaining a large match expression.

Do you have any suggestions?

chaokunyang:
Finished in #3107 and #3108

PR #2822, opened 2025-10-23
feat(Rust): Support serialize_to

theweipeng:
## What does this PR do?

- Add serialize_to API which allow users to pass in buffer by themselves
- Refactor `Writer`, the writer no longer holds the buffer but merely references it

### DIFF
--- a/rust/fory-core/src/resolver/context.rs
+++ b/rust/fory-core/src/resolver/context.rs
@@ -17,8 +17,6 @@
 
 use crate::buffer::{Reader, Writer};
 use std::mem;
-use std::ops::{Deref, DerefMut};
-use std::ptr::NonNull;
 
 use crate::error::Error;
 use crate::fory::Fory;


rust/fory-core/src/fory.rs
```diff
@@ -339,6 +340,106 @@ impl Fory {
...
+    ///
+    /// #[derive(ForyObject)]
+    /// struct Point { x: i32, y: i32 }
+    ///
+    /// let fory = Fory::default();
+    /// let point = Point { x: 10, y: 20 };
+    /// let bytes = fory.serialize(&point);
+    /// ```
+    pub fn serialize<T: Serializer>(&self, record: &T) -> Result<Vec<u8>, Error> {
+        let mut buf = vec![];
```
chaokunyang: This will allocate buffer every time, which will incur lots of cost

rust/fory-core/src/fory.rs
```diff
@@ -339,6 +340,106 @@ impl Fory {
...
+                    share_meta,
+                    compress_string,
+                    xlang,
+                    check_struct_version,
+                ))
+            };
+            Pool::new(factory)
+        });
+        let mut context = pool.get();
+        // Context got from poll would be 'static. but context hold the buffer through `writer` field, so we should make buffer live longer.
```
chaokunyang:
```suggestion
        // Context go from pool would be 'static. but context hold the buffer through `writer` field, so we should make buffer live longer.
```

rust/fory-core/src/fory.rs
```diff
@@ -339,6 +340,106 @@ impl Fory {
...
+        // After serializing, `detach_writer` will be called, the writer in context will be set to dangling pointer.
+        // So it's safe to make buf live to the end of this method.
+        let outlive_buffer = unsafe { mem::transmute::<&mut Vec<u8>, &mut Vec<u8>>(buf) };
+        self.serialize_with_context_to(record, &mut context, outlive_buffer)?;
+        pool.put(context);
+        Ok(())
+    }
+
+    pub fn serialize_with_context_to<'a, 'b: 'a, T: Serializer>(
```
chaokunyang: Should not pass buf here, buf should be placed into WriteContext.

rust/fory-core/src/fory.rs
```diff
@@ -339,6 +340,106 @@ impl Fory {
...
+                    check_struct_version,
+                ))
+            };
+            Pool::new(factory)
+        });
+        let mut context = pool.get();
+        // Context got from poll would be 'static. but context hold the buffer through `writer` field, so we should make buffer live longer.
+        // After serializing, `detach_writer` will be called, the writer in context will be set to dangling pointer.
+        // So it's safe to make buf live to the end of this method.
+        let outlive_buffer = unsafe { mem::transmute::<&mut Vec<u8>, &mut Vec<u8>>(buf) };
```
chaokunyang: We should create a new Writer to hold this buf here, and then context.attach_writer(&mut writer);

rust/fory-core/src/fory.rs
chaokunyang: We should keep using current writer in WriterContext to reuse buffer

rust/fory-core/src/fory.rs
```diff
@@ -339,6 +340,106 @@ impl Fory {
...
+        self.serialize_with_context_to(record, &mut context, outlive_buffer)?;
+        pool.put(context);
```
chaokunyang: And we should reset to previous writer before this line here

rust/fory-core/src/resolver/context.rs
```diff
@@ -203,12 +219,20 @@ impl WriteContext {
     }
 }
 
+impl<'a> Drop for WriteContext<'a> {
+    fn drop(&mut self) {
+        unsafe {
+            drop(Box::from_raw(self.writer.bf));
```
theweipeng, line 225: @chaokunyang  drop the buffer which was leaked in constructor.

rust/fory-core/src/resolver/context.rs
```diff
@@ -40,45 +41,61 @@ pub struct WriteContext {
...
+            writer: Writer::from_buffer(Self::get_leak_buffer()),
             meta_resolver: MetaWriterResolver::default(),
             meta_string_resolver: MetaStringWriterResolver::default(),
             ref_writer: RefWriter::new(),
         }
     }
 
-    pub fn new_from_fory(writer: Writer, fory: &Fory) -> WriteContext {
+    fn get_leak_buffer() -> &'static mut Vec<u8> {
+        Box::leak(Box::new(vec![]))
```
theweipeng, line 76: @chaokunyang  create a leak buffer to let the WriterContext hold the reference.  This reference will be drop when WriteContext been drop

rust/fory-core/src/resolver/context.rs
```diff
@@ -40,45 +41,61 @@ pub struct WriteContext {
...
+    ) -> WriteContext<'a> {
         WriteContext {
             type_resolver,
             compatible,
             share_meta,
             compress_string,
             xlang,
             check_struct_version,
-            writer,
+            default_writer: None,
```
theweipeng, line 67: Will will be nonnull when attach_write been called. the original writer will be store here. when detach called default_writer will be none and writer will be replaced by default_writer

chaokunyang (approved):
LGTM

PR #2829, opened 2025-10-24
perf(python): optimize pyfory perf

chaokunyang:
## What does this PR do?

optimize pyfory perf:
- remove simd utf16 check
- optimize cython Buffer by access pointer directly
- use load factor 0.5 for abceil hashmap

### DIFF
--- a/integration_tests/cpython_benchmark/fory_benchmark.py
+++ b/integration_tests/cpython_benchmark/fory_benchmark.py
@@ -36,9 +36,6 @@ Benchmark Options:
     --no-ref
         Disable reference tracking for Fory (enabled by default)
 
-    --disable-cython
-        Use pure Python mode instead of Cython serialization for Fory
-
     --warmup N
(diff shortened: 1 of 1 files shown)


pandalee99:
no blocking.
 
simdutf was also introduced in the benchmark, but in fact, we didn't use it; it was only for performance testing. Maybe we need to remove it?

chaokunyang:
> no blocking.
> 
> simdutf was also introduced in the benchmark, but in fact, we didn't use it; it was only for performance testing. Maybe we need to remove it?

Ciuld you open a new issue for it? #

pandalee99:
> > no blocking.
> > 
> > simdutf was also introduced in the benchmark, but in fact, we didn't use it; it was only for performance testing. Maybe we need to remove it?
> 
> Ciuld you open a new issue for it? #

yeah, i will open issue to talk about it

Issue #2832, opened 2025-10-24
[Java] Is it necessary to refactor RustXlangTest.java?


urlyy:
### Feature Request

Refactor `RustXlangTest.java` so that each test case runs independently, instead of being executed sequentially starting from `testRust()`.

### Additional context

I originally implemented it this way because, in cross-language testing where Java test with other languages, each language can had its own `test entry function`, while `sharing the same Java-side unit test logic` internally. But it seems that's no longer necessary?

chaokunyang:
@urlyy split it into separate tests would be better for debug

PR #2834, opened 2025-10-25
chore(Java): Make RustXlangTest cases independent from each other

urlyy:
## What does this PR do?
Make RustXlangTest cases independent

## Related issues
#2832

### DIFF
--- a/java/fory-core/src/test/java/org/apache/fory/RustXlangTest.java
+++ b/java/fory-core/src/test/java/org/apache/fory/RustXlangTest.java
@@ -95,8 +95,7 @@ public class RustXlangTest extends ForyTestBase {
     }
   }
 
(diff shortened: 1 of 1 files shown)


urlyy:
Wait. The CI seems have a problem.
image

java/fory-core/src/test/java/org/apache/fory/RustXlangTest.java
```diff
@@ -97,39 +96,16 @@ public void isRustJavaCIEnabled() {
   }
 
   @Test
-  public void testRust() throws Exception {
+  List<String> setTestCase(String name) {
```
chaokunyang: Remove `@Test`, this method is not a test

PR #2852, opened 2025-10-30
fix: modify the depth setting in Fory to prevent duplicate registrations.

### LINKED ISSUE #2840: register class after serialization using ThreadSafeFory  caused classInfo.classDef NPE
### Search before asking

### Version

version：0.12.3
os：windows 11

### Component(s)

Java

### Minimal reproduce step

错误写法：
private static final ThreadSafeFory fory = new ThreadLocalFory(classLoader -> {
    Fory f = Fory.builder().withLanguage(Language.JAVA)
      .withClassLoader(classLoader).build();
    return f;
  });
fory.register(SomeClass.class);
Struct2 struct2 = (Struct2) fory.deserializeJavaObject(bytes, Struct2.class);

正确反序列化

### What did you see instead?

org.apache.fory.exception.DeserializationException: Failed to deserialize input
	at org.apache.fory.util.ExceptionUtils.handleReadFailed(ExceptionUtils.java:66)
	at org.apache.fory.Fory.deserializeJavaObject(Fory.java:1226)
	at org.apache.fory.Fory.deserializeJavaObject(Fory.java:1197)
	at org.apache.fory.ThreadLocalFory.deserializeJavaObject(ThreadLocalFory.java:207)
	at cc.metafusion.bossai.service.common.util.ForyUtils.deserialize(ForyUtils.java:46)
	at cc.metafusion.bossai.service.common.model.mq.SourceEventData.of(SourceEventData.java:43)
	at cc.metafusion.data.event.bridge.core.queue.MsgProducer.translate(MsgProducer.java:23)
	at com.lmax.disruptor.RingBuffer.translateAndPublish(RingBuffer.java:967)
	at com.lmax.disruptor.RingBuffer.publishEvent(RingBuffer.java:487)
	at cc.metafusion.data.event.bridge.core.queue.MsgProducer.lambda$onData$0(MsgProducer.java:40)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at cc.metafusion.data.event.bridge.core.queue.MsgProducer.onData(MsgProducer.java:39)
	at cc.metafusion.data.event.bridge.core.startup.RocketMqStartup.lambda$init$0(RocketMqStartup.java:61)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.NullPointerException: Cannot invoke "org.apache.fory.meta.ClassDef.replaceRootClassTo(org.apache.fory.resolver.ClassResolver, java.lang.Class)" because "classInfo.classDef" is null
	at org.apache.fory.resolver.ClassResolver.readClassInfoWithMetaShare(ClassResolver.java:1628)
	at org.apache.fory.Fory.deserializeJavaObject(Fory.java:1216)
	... 12 common frames omitted

Caused by: java.lang.NullPointerException: Cannot invoke "org.apache.fory.meta.ClassDef.replaceRootClassTo(org.apache.fory.resolver.ClassResolver, java.lang.Class)" because "classInfo.classDef" is null

### Anything Else?

正确写法
private static final ThreadSafeFory fory = new ThreadLocalFory(classLoader -> {
    Fory f = Fory.builder().withLanguage(Language.JAVA)
      .withClassLoader(classLoader).build();
    // 需要注册的类在这里全部注册完
    f.register(SomeClass.class);
    return f;
  });
Struct2 struct2 = (Struct2) fory.deserializeJavaObject(bytes, Struct2.class);

mengnankkkk:
## Why?
By adding a depth field to the fory instance to indicate whether it is registered, and then checking in the TypeResolver, it is ensured that the fory instance cannot be registered after serialization, otherwise a foryexception will be thrown.

## Related issues
Close #2840 

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/Fory.java
+++ b/java/fory-core/src/main/java/org/apache/fory/Fory.java
@@ -177,9 +177,6 @@ public final class Fory implements BaseFory {
 
   @Override
   public void register(Class<?> cls) {
-    if (depth >= 0) {
-      throw new ForyException("Cannot register class during serialization/deserialization");
-    }
     _getTypeResolver().register(cls);
   }
 
--- a/java/fory-core/src/test/java/org/apache/fory/serializer/ArraySerializersTest.java
+++ b/java/fory-core/src/test/java/org/apache/fory/serializer/ArraySerializersTest.java
@@ -43,7 +43,7 @@ import org.testng.annotations.Test;
 
 public class ArraySerializersTest extends ForyTestBase {
 
-  private static void prewarm(Fory... forys) {
+  private static void registerTypes(Fory... forys) {
     Class<?>[] types =
         new Class<?>[] {
           Object.class,
@@ -102,7 +102,7 @@ public class ArraySerializersTest extends ForyTestBase {
       fory1.getSerializationContext().setMetaContext(new org.apache.fory.resolver.MetaContext());
       fory2.getSerializationContext().setMetaContext(new org.apache.fory.resolver.MetaContext());
     }
-    prewarm(fory1, fory2);
+    registerTypes(fory1, fory2);
     serDeCheck(fory1, fory2, new Object[] {false, true});
     serDeCheck(fory1, fory2, new Object[] {(byte) 1, (byte) 1});
     serDeCheck(fory1, fory2, new Object[] {(short) 1, (short) 1});
@@ -168,7 +168,7 @@ public class ArraySerializersTest extends ForyTestBase {
       fory1.getSerializationContext().setMetaContext(new org.apache.fory.resolver.MetaContext());
       fory2.getSerializationContext().setMetaContext(new org.apache.fory.resolver.MetaContext());
     }
-    prewarm(fory1, fory2);
+    registerTypes(fory1, fory2);
     serDeCheck(fory1, fory2, new Object[][] {{false, true}, {false, true}});
     serDeCheck(
         fory1,


java/fory-core/src/main/java/org/apache/fory/ThreadLocalFory.java
```diff
@@ -78,10 +81,12 @@ public ThreadLocalFory(Function<ClassLoader, Fory> foryFactory) {
   @Internal
   @Override
   public void registerCallback(Consumer<Fory> callback) {
-    factoryCallback = factoryCallback.andThen(callback);
-    for (LoaderBinding binding : allFory.keySet()) {
-      binding.visitAllFory(callback);
-      binding.setBindingCallback(factoryCallback);
+    synchronized (callbackLock) {
```
chaokunyang, line 82: could we also add lock to `ThreadPoolFory`? Or would it be possible to add lock in `AbstactThreadSafeFory`?

java/fory-core/src/test/java/org/apache/fory/ThreadSafeForyTest.java
```diff
@@ -373,4 +373,137 @@ public void testSerializerRegister() {
...
+    ExecutorService executor = Executors.newFixedThreadPool(4);
+    ConcurrentHashMap<String, Throwable> errors = new ConcurrentHashMap<>();
+
+    try {
+      ThreadSafeFory fory = Fory.builder().requireClassRegistration(true).buildThreadLocalFory();
+
+      fory.register(BeanA.class);
+      fory.register(BeanB.class);
+      BeanA beanA = BeanA.createBeanA(2);
+      byte[] bytesA = fory.serialize(beanA);
```
chaokunyang: You only register once, then serialize with that Fory instance multiple times. This should work as before. I don't think this PR fix anything.

IMO, the real issue is that the `serialize` has been executed a few times, but after that, users call `register` method. Such pattern are not allowed, no register should be invoked after any serialize/deserialize has been executed.

As for register  itself, may just add `synchronized` to method is enough

mengnankkkk: Okay, I understand the problem and will make changes. Thank you for your code review.

chaokunyang:
@mengnankkkk  How about reuse `depth` field in `Fory` instance, you can set it to -1 by default, and when register any Serializer/Class, check whether `depth` is still -1. If it `>=` 0, it means serialize/deserialize has happened, then you can just thow exception. You can add this check in `TypeResolver`

mengnankkkk:
> @mengnankkkk How about reuse `depth` field in `Fory` instance, you can set it to -1 by default, and when register any Serializer/Class, check whether `depth` is still -1. If it `>=` 0, it means serialize/deserialize has happened, then you can just thow exception. You can add this check in `TypeResolver`

OK, I'll do my best. I hope it won't be like the last PR attempt, where it couldn't be resolved. If it really can't be resolved, I'll have to ask you for advice again.[doge]

mengnankkkk:
@chaokunyang Hello, I've implemented the basic checking phase as you requested, but I'm encountering a problem:

`ThreadPoolFory` manages multiple `Fory` serialization instances. When one thread starts serializing/deserializing an instance, another thread calls registerCallback to register a new class.

`registerCallback` attempts to apply the new class to all Fory instances in the pool, including the one currently being used by thread A or already used. Since the instance's internal state (such as the class ID mapping table) might have been "frozen" (Finalized) during the first serialization, attempting to register a new class at this point would corrupt its state, causing failure.

Using a synchronization lock would result in poor performance. Adding a volatile boolean global variable and checking this flag in the `registerCallback` method would immediately throw an `IllegalStateException` if the flag is true (meaning the "execution phase" has begun), preventing the registration. However, this causes many tests to fail, which seems inconsistent with multithreading requirements.

How should I handle this problem?! Help!

chaokunyang:
I don't think this issue exists. If one of fory instance has executed the serialization, the depth will not be -1. And the registerCallback should fail on that instance And propagate the exception to the callsite

chaokunyang:
You should not add any new global variable. We have no plans to introduce any new global flags in fory.

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -1606,6 +1602,10 @@ private void throwDepthDeserializationException() {
             method));
   }
 
+  public boolean hasSerializedOrDeserialized() {
```
chaokunyang: Remove this method

chaokunyang:
Could we add a simple test to verify that register after serialize will throw exception?

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
```diff
@@ -111,6 +111,14 @@ protected TypeResolver(Fory fory) {
     metaStringResolver = fory.getMetaStringResolver();
   }
 
+  protected final void checkRegisterAllowed() {
+    if (fory.getDepth() >= 0) {
+      throw new IllegalStateException(
+          "Cannot register class/serializer after serialization/deserialization has started. "
+              + "Please register all classes before using Fory.");
```
chaokunyang:
```suggestion
              + "Please register all classes before invoking `serialize/deserialize` methods of Fory.");
```

java/fory-core/src/main/java/org/apache/fory/ThreadLocalFory.java
```diff
@@ -59,7 +60,9 @@ public ThreadLocalFory(Function<ClassLoader, Fory> foryFactory) {
         ThreadLocal.withInitial(
             () -> {
               LoaderBinding binding = new LoaderBinding(foryFactory);
-              binding.setBindingCallback(factoryCallback);
+              synchronized (callbackLock) {
```
chaokunyang: why we need this, this seems fix nothing

java/fory-core/src/main/java/org/apache/fory/pool/ClassLoaderForyPooled.java
```diff
@@ -98,6 +98,7 @@ public Fory getFory() {
 
   public void returnFory(Fory fory) {
     Objects.requireNonNull(fory);
+    fory.reset();
```
chaokunyang: remove it, it introduce extra reset

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
```diff
@@ -111,6 +111,14 @@ protected TypeResolver(Fory fory) {
     metaStringResolver = fory.getMetaStringResolver();
   }
 
+  protected final void checkRegisterAllowed() {
+    if (fory.getDepth() >= 0) {
+      throw new IllegalStateException(
```
chaokunyang:
```suggestion
      throw new ForyExcetion(
```

java/fory-core/src/test/java/org/apache/fory/ThreadSafeForyTest.java
```diff
@@ -373,4 +373,40 @@ public void testSerializerRegister() {
...
+    Assert.assertEquals(fory.getDepth(), -1);
+    BeanA beanA = BeanA.createBeanA(2);
+    byte[] bytes = fory.serialize(beanA);
+    Assert.assertEquals(fory.getDepth(), -1);
+    fory.deserialize(bytes);
+    Assert.assertEquals(fory.getDepth(), -1);
+  }
+
+  @Test
+  public void testRegisterAfterSerializeThrowsException() {
```
chaokunyang: I don't see any code  in this test `assertThrows` the exception has been raised. Please update this test.

And please be consise, most of codes in this tests doesn't test the unexpected register behaviour. Please remove such code, and concentrated on unexpected register behaviour test only.

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -1662,6 +1667,14 @@ public void setDepth(int depth) {
     this.depth = depth;
   }
 
+  public void beforeSerialize() {
```
chaokunyang: remove such methods, please keep the implmentation simple

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -1572,7 +1573,7 @@ public void resetWrite() {
     metaStringResolver.resetWrite();
     serializationContext.resetWrite();
     bufferCallback = null;
-    depth = 0;
+    depth = -1;
```
chaokunyang:
```suggestion
    depth = 0;
```
I think this should be 0, otherwise the register can still be invoked

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -1581,8 +1582,8 @@ public void resetRead() {
     metaStringResolver.resetRead();
     serializationContext.resetRead();
     peerOutOfBandEnabled = false;
-    depth = 0;
     classDefEndOffset = -1;
+    depth = -1;
```
chaokunyang: ditto

java/fory-core/src/test/java/org/apache/fory/ThreadSafeForyTest.java
```diff
@@ -373,4 +374,63 @@ public void testSerializerRegister() {
...
+    Assert.assertEquals(fory.getDepth(), -1);
+    BeanA beanA = BeanA.createBeanA(2);
+    byte[] bytes = fory.serialize(beanA);
+    Assert.assertEquals(fory.getDepth(), -1);
+    fory.deserialize(bytes);
+    Assert.assertEquals(fory.getDepth(), -1);
+  }
+
+  @Test
+  public void testRegisterAfterSerializeThrowsException() throws Exception {
```
chaokunyang, line 379: Seems this test doesn't capture the bug in set `depth` to `-1`, please update this test to make it actually capture such bugs

java/fory-core/src/test/java/org/apache/fory/ThreadSafeForyTest.java
```diff
@@ -373,4 +374,63 @@ public void testSerializerRegister() {
...
         });
   }
+
+  @Test
+  public void testGetDepth() {
+    Fory fory = Fory.builder().requireClassRegistration(false).build();
+    Assert.assertEquals(fory.getDepth(), -1);
+    BeanA beanA = BeanA.createBeanA(2);
+    byte[] bytes = fory.serialize(beanA);
+    Assert.assertEquals(fory.getDepth(), -1);
```
chaokunyang:
```suggestion
    Assert.assertEquals(fory.getDepth(), 0);
```

java/fory-core/src/test/java/org/apache/fory/ThreadSafeForyTest.java
```diff
@@ -373,4 +374,63 @@ public void testSerializerRegister() {
...
+            () -> {
+              fory.serialize(new Foo());
+            });
+    t.start();
+
+    Assert.assertTrue(writeStarted.await(5, TimeUnit.SECONDS));
+
+    Assert.assertThrows(ForyException.class, () -> fory.register(BeanA.class));
+
+    allowFinish.countDown();
```
chaokunyang: why we need such code? What does it test exactly?
Please keep things simple. You can just use Fory and TheadSafeFory(ThreadLocalFory/ThradPoolFory) to serialize an object, then invoke register, adn assert it raise exception. I think this will be enough

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -803,11 +803,16 @@ public <T> T deserialize(byte[] bytes, Class<T> type) {
       assert !peerOutOfBandEnabled : "Out of band buffers not passed in when deserializing";
       return deserializeJavaObject(buffer, type);
     }
+    int prevDepth = this.depth;
+    if (crossLanguage) {
```
chaokunyang: why we need this?

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -803,11 +803,16 @@ public <T> T deserialize(byte[] bytes, Class<T> type) {
       assert !peerOutOfBandEnabled : "Out of band buffers not passed in when deserializing";
       return deserializeJavaObject(buffer, type);
     }
+    int prevDepth = this.depth;
```
chaokunyang: TBO, this is also a little strange

java/fory-core/src/test/java/org/apache/fory/ThreadSafeForyTest.java
```diff
@@ -378,59 +378,36 @@ public void testSerializerRegister() {
...
 
-    allowFinish.countDown();
-    t.join();
+  @Test
+  public void testRegisterAfterSerializeThrowsExceptionWithForyPool() {
+    ThreadSafeFory fory =
+        Fory.builder().requireClassRegistration(true).buildThreadSafeForyPool(1, 2);
+    fory.register(BeanA.class);
+    fory.serialize("ok");
+    Assert.assertThrows(ForyException.class, () -> fory.register(BeanA.class));
```
chaokunyang: You should register another class instead of same class.

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -176,6 +177,9 @@ public Fory(ForyBuilder builder, ClassLoader classLoader) {
 
   @Override
   public void register(Class<?> cls) {
+    if (depth >= 0) {
```
chaokunyang: Seems this is not needed, we've checked it in `TypeResolver#register`

java/fory-core/src/test/java/org/apache/fory/serializer/ArraySerializersTest.java
```diff
@@ -43,39 +43,90 @@
 
 public class ArraySerializersTest extends ForyTestBase {
 
+  private static void prewarm(Fory... forys) {
```
chaokunyang: could we rename to `registerTypes`?

mengnankkkk:
> LGTM overall, after addressed little minor issues, we can merge it. Thanks for your contribution!

Thank you for your code review. I really learned a lot in the process of fixing this PR. Thank you for reviewing the code and the solution design. It was really helpful. I will fix the simple issues right away. Thank you again!

Issue #2851, opened 2025-10-30
[Rust] Support rust tagged union enum


chaokunyang:
### Feature Request

Currently Fory Rust only allow c-style enum, we don't allow tagged union enum since it can't be expressed in other languages.

But for pure rust serialization, it's ok to support such enum:

```rust
    enum TestEnum {
        V1(String, i32), 
        V2(String, bool), 
    }
```

### Describe the solution you'd like

We can extend `ForyObject` macro to generate code like:
```rust
match home {
    V1(v1, v2) => {
        context.writer.write_tag(0);
        context.writer.write_str(v1);
        context.writer.write_i32(v2);
    }
    V2(v1, v2) => {
        context.writer.write_tag(1);
        context.writer.write_str(v1);
        context.writer.write_bool(v2);
    }
}
```
And for deserialization, we can generate code like:
```rust
let tag = context.reader.read_tag();
match tag {
    0 => {
        let v1 = context.reader.read_str();
        let v2 = context.reader.read_i32();
        TestEnum::V1(v1, v2)
    }
    1 => {            
        let v1 = context.reader.read_str();
        let v2 = context.reader.read_bool();
        TestEnum::V2(v1, v2)
    }
}
```

urlyy:
👌

urlyy:
@chaokunyang So Fory needs to support handling enum like this.
```rust
enum Token {
    Plus,
    Number(i64),
    Ident(String),
    Assign { target: String, value: i32 },
    Other(Option<i64>),
}
```

Generated code:
```rust
impl Token{
   fn serialize(&self, context: &mut WriteContext) {
        match self {
            Self::Plus => {
                // tag
                context.write_i8(0);
            }
            Self::Number(f1) => {
                context.write_i8(1);
                i64::write(f1, context);
            }
            Self::Ident(f1) => {
                context.write_i8(2);
                String::write(f1, context);
            }
            Self::Assign { target, value } => {
                context.write_i8(3);
                // named_fields will be sorted in lexicographical order
                String::write(target, context);
                i32::write(value, context);
            }
            Self::Other(f1) => {
                context.write_i8(4);
                Option::<i64>::write(f1, context);
            }
        }
    }

    fn deserialize(context: &mut ReadContext) -> Self {
        let tag = context.read_i8();
        match tag {
            0 => Self::Plus,
            1 => {
                let f1 = i64::read(context);
                Self::Number(f1)
            }
            2 => {
                let f1 = String::read(context);
[...]

PR #2855, opened 2025-10-30
feat(rust): Support rust tagged union enum

urlyy:
## Why?
Enum with poayload is the norm in rust and is extensively used.

## What does this PR do?
Modified the `Enum::read_data()`, `Enum::write_data()`, and the code generation for `Enum::default()` to support tagged union enums.

## Related issues
#2851

If the user is in `Rust to Rust` protocol instead of `Xlang` protocol, the values of the tagged union enum can be transmitted correctly. 
Otherwise, in `Xlang` protocol, although the corresponding enum type can be passed, the `internal value` will be the initial value. Because at present, the `xlang protocol` only serializes the `ordinal value` of enums.
For example:

```rust
#[test]
fn test_enum(){
    enum Color {
        Red(i8),
        White(String)
    }
    let color = Color::Red(10);
    
    // rust protocol
    let mut fory = Fory::default().xlang(false);
    fory.register::<Color>(1000).unwrap();
    let bin = fory.serialize(&color).unwrap();
    let new_color = fory.deserialize::<Color>(&bin).unwrap();
    assert_eq!(color, new_color);
    if let Color::Red(value) = new_color {
        assert_eq!(value, 10);
    } else {
        unreachable!()
    }
    // xlang protocol
    let mut fory2 = Fory::default().xlang(true);
    fory2.register::<Color>(1000).unwrap();
    let bin = fory2.serialize(&color).unwrap();
    let new_color = fory2.deserialize::<Color>(&bin).unwrap();
    if let Color::Red(value) = new_color {
        assert_eq!(value, i8::default());
    } else {
        unreachable!()
    }
}
```

### DIFF
--- a/rust/tests/tests/test_enum.rs
+++ b/rust/tests/tests/test_enum.rs
@@ -19,6 +19,7 @@
 
 use fory_core::Fory;
 use fory_derive::ForyObject;
+use std::collections::HashMap;
 
 #[test]
 fn basic() {
@@ -30,10 +31,20 @@ fn basic() {
         Assign { target: String, value: i32 },
         Other(Option<i64>),
         Child(Box<Token>),
+        Map(HashMap<String, Token>),
     }
 
     let mut fory = Fory::default().xlang(false);
     fory.register::<Token>(1000).unwrap();
+
+    let mut map = HashMap::new();
+    map.insert("one".to_string(), Token::Number(1));
+    map.insert("plus".to_string(), Token::Plus);
+    map.insert(
+        "nested".to_string(),
+        Token::Child(Box::new(Token::Ident("deep".to_string()))),
+    );
+
     let tokens = vec![
         Token::Plus,
         Token::Number(1),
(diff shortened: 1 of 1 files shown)


urlyy:
The current implementation may be a bit tricky. For convenience, I've implemented it in `fory_xx_data`. Since I can't directly call `Option::read_data()`/`write_data()`, if the internal field of the enum value is an `Option`, I'll perform a read/write of the `ref_flag` within `fory_xx_data()`.

urlyy:
I realize maybe I should call `fory_write`. Wait some time.

rust/fory-derive/src/object/derive_enum.rs
```diff
@@ -44,18 +44,130 @@ pub fn gen_write(_data_enum: &DataEnum) -> TokenStream {
...
+            }
+            Fields::Unnamed(fields_unnamed) => {
+                let field_idents: Vec<_> = (0..fields_unnamed.unnamed.len())
+                    .map(|i| syn::Ident::new(&format!("f{}", i), proc_macro2::Span::call_site()))
+                    .collect();
+
+                let write_fields: Vec<_> = fields_unnamed.unnamed.iter().zip(field_idents.iter()).map(|(f, ident)| {
+                    let ty = &f.ty;
+                    quote! {
+                        if <#ty>::fory_is_option() {
```
chaokunyang: You can check whether a field is optional at compile time, and you shoulld handle all unamed fields like this:
-  if it's `Option` or `RcWeak/ArcWeak/Rc/Arc`, use `fory_write/read`
-  if it's not `Option`, use `fory_write_data/read_data` instead.
Since we don't support Compatible mode in this PR, we can keep things simple

PR #2865, opened 2025-11-01
feat: ✨ support fory skip macro attributes(#2864)

kitty-eu-org:
## Why?

This PR adds a `#[fory(skip)]` macro attribute similar to `#serde[(skip)]` to provide more flexible field skipping capabilities in Apache Fory™. This enhancement allows developers to easily mark specific fields to be skipped during serialization/deserialization or other processing phases, improving code readability and maintainability.

## What does this PR do?

- Adds a new `#[fory(skip)]` macro attribute implementation
- The attribute functions similarly to `#[serde(skip)]` but is specifically designed for Fory's internal use cases
- Provides consistent attribute parsing and handling within Fory's macro system
- Maintains backward compatibility with existing code

## Related issues

- [[Rust] support fory macro attributes ](https://github.com/apache/fory/issues/2864)

**No**, this PR only adds a new macro attribute that doesn't affect existing public APIs or binary protocols. The `#[fory(skip)]` attribute is opt-in and doesn't change the behavior of existing code.

## Benchmark

**No performance impact expected** as this change only adds attribute parsing capability without modifying core processing logic. The macro attribute is processed at compile time and doesn't introduce runtime overhead.

### DIFF
--- a/rust/fory-derive/src/object/util.rs
+++ b/rust/fory-derive/src/object/util.rs
@@ -767,12 +767,13 @@ pub(crate) fn get_sorted_field_names(fields: &[&Field]) -> Vec<String> {
     all_fields.into_iter().map(|(name, _, _)| name).collect()
 }
 
+pub(crate) fn get_filtered_fields_iter<'a>(
+    fields: &'a [&'a Field],
+) -> impl Iterator<Item = &'a Field> {
+    fields.iter().filter(|field| !is_skip_field(field)).copied()
+}
 pub(super) fn get_sort_fields_ts(fields: &[&Field]) -> TokenStream {
-    let filterd_fields: Vec<&Field> = fields
-        .iter()
-        .filter(|field| !is_skip_field(field))
-        .copied()
-        .collect();
+    let filterd_fields: Vec<&Field> = get_filtered_fields_iter(fields).collect();
     let sorted_names = get_sorted_field_names(&filterd_fields);
     let names = sorted_names.iter().map(|name| {
         quote! { #name }
--- a/rust/fory-derive/src/object/write.rs
+++ b/rust/fory-derive/src/object/write.rs
@@ -17,8 +17,8 @@
 
 use super::util::{
     classify_trait_object_field, compute_struct_version_hash, create_wrapper_types_arc,
-    create_wrapper_types_rc, get_struct_name, get_type_id_by_type_ast, is_debug_enabled,
-    is_skip_field, should_skip_type_info_for_field, skip_ref_flag, StructField,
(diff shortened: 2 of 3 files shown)


urlyy:
`skip`  macro can provide some compatibility in `consistent_mode`, which is great.
I see this PR only modifies the logic in `read_data` and `write_data`. Furthermore, can we avoid sending the `type_meta` of fields marked as `skip` in `compatible_mode`? @chaokunyang

chaokunyang:
> `skip` macro can provide some compatibility in `consistent_mode`, which is great. I see this PR only modifies the logic in `read_data` and `write_data`. Furthermore, can we avoid sending the `type_def` of fields marked as `skip` in `compatible_mode`? @chaokunyang

ignored fields should also be avoided in type meta

rust/fory-derive/src/object/write.rs
```diff
@@ -249,6 +249,7 @@ pub fn gen_write_field(field: &Field, ident: &Ident, use_self: bool) -> TokenStr
 pub fn gen_write_data(fields: &[&Field]) -> TokenStream {
     let write_fields_ts: Vec<_> = fields
         .iter()
+        .filter(|field| !is_skip_field(field))
```
chaokunyang: We also need to skip fields in
```rust
fn fory_get_sorted_field_names() -> &'static [&'static str];
fn fory_fields_info()
```

kitty-eu-org: > 我们还需要跳过以下字段：
> 
> ```rust
> fn fory_get_sorted_field_names() -> &'static [&'static str];
> fn fory_fields_info()
> ```

It now supports ignoring fields marked with the skip attribute macro in the metadata.

rust/fory-derive/src/object/read.rs
```diff
@@ -103,6 +103,13 @@ fn assign_value(fields: &[&Field]) -> Vec<TokenStream> {
 
 pub fn gen_read_field(field: &Field, private_ident: &Ident) -> TokenStream {
     let ty = &field.ty;
+    if is_skip_field(field) {
```
chaokunyang, line 106: please also update `gen_read_compatible` to remove skipped field arm

rust/fory-derive/src/object/util.rs
```diff
@@ -768,7 +768,12 @@ pub(crate) fn get_sorted_field_names(fields: &[&Field]) -> Vec<String> {
 }
 
 pub(super) fn get_sort_fields_ts(fields: &[&Field]) -> TokenStream {
-    let sorted_names = get_sorted_field_names(fields);
+    let filterd_fields: Vec<&Field> = fields
```
chaokunyang: Could we extract a method to get filtered fields Vec And reuse it in read and write.rs

PR #2893, opened 2025-11-04
chore(CI): Fix cache with symlinks

prakash-218:
## Why?

Bazel cache restoration was failing because symbolic links were being converted to regular files during cache save/restore operations.
Added `enableCrossOsArchive: false` to all Bazel cache steps to preserve symlinks correctly.


chaokunyang:
image
it seems those pipeline still not work

chaokunyang:
@prakash-218 maybe you could try setup-bazel cache later:
```
- uses: bazel-contrib/setup-bazel@0.15.0
  with:
    # Avoid downloading Bazel every time.
    bazelisk-cache: true
    # Store build cache per workflow.
    disk-cache: ${{ github.workflow }}
    # Share repository cache between workflows.
    repository-cache: true
```

now let me remove build cache to fix ci

Issue #2885, opened 2025-11-03
[Rust][XLang] Use varint32/64 as type id for int32/64


chaokunyang:
### Feature Request

Use varint32/64 type id  as type id for int32/64 type , so we can use ` int32/64` type id for uncompressed int32/64 read/write.

urlyy:
For the same `int32` field, if locally it is set to `compress=false`, but the type_id received from remote is `VAR_INT32`, how should it be handled? If different processing is required based on the remote type_id, passing the type_id and add conditional branches to `fory_read_data()` might affect performance.

urlyy:
> For the same `int32` field, if locally it is set to `compress=false`, but the type_id received from remote is `VAR_INT32`, how should it be handled? If different processing is required based on the remote type_id, passing the type_id and add conditional branches to `fory_read_data()` might affect performance.

@chaokunyang

chaokunyang:
> For the same `int32` field, if locally it is set to `compress=false`, but the type_id received from remote is `VAR_INT32`, how should it be handled? If different processing is required based on the remote type_id, passing the type_id and add conditional branches to `fory_read_data()` might affect performance.

We read field based remote type meta, since remote field already say it's varint32, we just read by varint32. 

This is essentially a type mismatch, we support such cases, but we should put it into a type id mismatch default case, and use a predefined switch case function to handle that.

@urlyy

urlyy:
@chaokunyang Ok, But you said we should use
```rust
struct Foo {
  [#fory(compress=False)]
  f1: i32
}
```


But i64 has three compression types. A single bool attribute can't handle that. Can we use a string form like `#[fory(compress = "varint")]?`

chaokunyang:
We can support like:
```rust
struct Foo {
    #[fory(compress = false)]
    f1: i64,
    #[fory(compress = "sli64")]
    f2: i64,
    #[fory(compress = "varint64")]
    f2: i64,
}
```

urlyy:
@chaokunyang This is the solution I believe requires the least modifications to the current code structure to resolve this PR, it is using `generic constant parameters` to eliminate branches at compile time and reduce runtime overhead. 

However, this is not very user-friendly for `EXT` users, and if more compression modes need to be added in the future, it will also result in very long code.

Here is the demo code:
```rust
use std::collections::HashMap;

trait Serializer {
    fn fory_read_data<const COMPRESS_INT: bool, const COMPRESS_LONG_MODE: u8>() -> Self;
}

[...]

Issue #2887, opened 2025-11-03
[Go][XLang] Use varint32/64 as type id for int32/64 


chaokunyang:
### Feature Request

Use varint32/64 type id  as type id for int32/64 type , so we can use ` int32/64` type id for uncompressed int32/64 read/write.

PR #2898, opened 2025-11-05
refactor(Rust&Java&Python): use varint32 as type id for int32

urlyy:
## What does this PR do?
1. Added a `type_resolver` parameter to `fory_static_type_id()` and `fory_reserved_space()` so that `type_resolver.is_compress_int()` can be used to decide whether to use `read/write_varint32()` or `read/write_i32()`.
2. Added a `compress_int()` method to `Fory` to allow users to configure whether integers should be compressed; default is `true`.
3. Updated the logic in `impl Serializer for i32`.

## Related issues
#2884
#2885
#2886

Users can configure whether to compress integers by calling `fory.compress_int(true/false)`.

### DIFF
--- a/python/pyfory/_fory.py
+++ b/python/pyfory/_fory.py
@@ -165,6 +165,7 @@ class Fory:
         "depth",
         "field_nullable",
         "policy",
+        "compress_int",
     )
 
     def __init__(
@@ -216,6 +217,12 @@ class Fory:
                 (xlang=False), regardless of Optional annotation. Ignored in cross-language
                 mode.
 
+            compress_int: Enable integer compression using varint encoding (default: True).
+                When enabled, integers are encoded using variable-length encoding
+                for more efficient serialization of small integers. This reduces
+                the serialized size but may have minor performance impact on
+                serialization/deserialization.
+
         Example:
             >>> # Python-native mode with reference tracking
             >>> fory = Fory(ref=True)
@@ -237,6 +244,10 @@ class Fory:
         if kwargs.get("require_type_registration") is not None:
             strict = kwargs.get("require_type_registration")
         self.strict = _ENABLE_TYPE_REGISTRATION_FORCIBLY or strict
+        if kwargs.get("compress_int") is not None:
+            self.compress_int = kwargs.get("compress_int")
+        else:
+            self.compress_int = True
         self.policy = policy or DEFAULT_POLICY
         self.compatible = compatible
         self.field_nullable = field_nullable if self.is_py else False
--- a/python/pyfory/_registry.py
+++ b/python/pyfory/_registry.py
@@ -258,7 +258,10 @@ class TypeResolver:
         register(bool, type_id=TypeId.BOOL, serializer=BooleanSerializer)
         register(int8, type_id=TypeId.INT8, serializer=ByteSerializer)
         register(int16, type_id=TypeId.INT16, serializer=Int16Serializer)
-        register(int32, type_id=TypeId.INT32, serializer=Int32Serializer)
+        int32_type_id = TypeId.INT32
+        if self.fory.compress_int:
+            int32_type_id = TypeId.VAR_INT32
+        register(int32, type_id=int32_type_id, serializer=Int32Serializer)
         register(int64, type_id=TypeId.INT64, serializer=Int64Serializer)
         register(int, type_id=TypeId.INT64, serializer=Int64Serializer)
         register(
--- a/python/pyfory/_serializer.py
+++ b/python/pyfory/_serializer.py
@@ -122,10 +122,16 @@ class Int16Serializer(XlangCompatibleSerializer):
 
 class Int32Serializer(XlangCompatibleSerializer):
     def write(self, buffer, value):
-        buffer.write_varint32(value)
+        if self.fory.compress_int:
+            buffer.write_varint32(value)
+        else:
+            buffer.write_int32(value)
 
     def read(self, buffer):
-        return buffer.read_varint32()
+        if self.fory.compress_int:
+            return buffer.read_varint32()
+        else:
+            return buffer.read_int32()
 
 
 class Int64Serializer(Serializer):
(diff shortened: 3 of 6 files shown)


urlyy:
@chaokunyang Is ForyJava hasn’t updated the `writeClassInfo` logic for Integer based on `compressInt()` yet? I may need to temporarily disable `RustXlangTest.java`.

rust/fory-core/src/serializer/box_.rs
```diff
@@ -62,8 +61,8 @@ impl<T: Serializer + ForyDefault> Serializer for Box<T> {
     }
 
     #[inline(always)]
-    fn fory_static_type_id() -> TypeId {
-        T::fory_static_type_id()
+    fn fory_static_type_id(type_resolver: &TypeResolver) -> TypeId {
```
chaokunyang: `fory_static_type_id` method should be executed at compile time or optimized by LLVM pass. We should not pass `type_resolver` in.

I even think we should make it as `const`, but currently stable rust doesn't support that

rust/fory-core/src/serializer/box_.rs
```diff
@@ -47,8 +46,8 @@ impl<T: Serializer + ForyDefault> Serializer for Box<T> {
     }
 
     #[inline(always)]
-    fn fory_reserved_space() -> usize {
-        T::fory_reserved_space()
+    fn fory_reserved_space(type_resolver: &TypeResolver) -> usize {
```
chaokunyang: ditto for this method

chaokunyang:
> @chaokunyang Is ForyJava hasn’t updated the `writeClassInfo` logic for Integer based on `compressInt()` yet? I may need to temporarily disable `RustXlangTest.java`.

Could we update java implementation in this PR too?

rust/fory-core/src/serializer/box_.rs
urlyy: >  We should not pass `type_resolver` in.

Ok

urlyy:
> Could we update java implementation in this PR too?

I will try.

urlyy:
1. Need to refactor `fory-derive/object/util.rs::compute_struct_version_hash()`. The `typeid` for `i32` is treated as `INT32` at compile time, which causes a version hash mismatch error.
2. `CrossLanguageTest.java` also occurs errors and the `Python` implementation needs to be modified as well.

But I don’t have time to refactor these. If it's not urgent, I will continue working on this PR in four or five days.

chaokunyang:
> But I don’t have time to refactor these. If it's not urgent, I will continue working on this PR in four or five days.

@urlyy It's not urgent, you can take your time

urlyy:
@chaokunyang Can we modify the format of `version_hash`?

The reason is that Rust needs to compute the `version_hash` at compile time for different compression scenarios, which requires calculating **$2^5$** hashes (because of `not_compress_i32 / compress_i32 / not_compress_i64 / compress_to_var_i64/ compress_to_sli_i64`). If we further add `U32 / VAR_U32 / U64 / VAR_U64 / SLI_U64`, it becomes even more explosive. 

Separating the two compression flags can effectively mitigate this situation.

Which means, for `int_32` and `int_64`, when computing the hash, we only treat them as `Types::INT32` and `Types::INT64`. And use 1 bit of the hash value to represent compress_int_flag, and another 2 bits to represent compress_long_encoding. For `u32` and `u64`, 3 more bits are needed. What do you think of this approach?

urlyy:
1. Currently, ForyRust requires users to use `#[fory(compress_int)]` or `#[fory(compress_int=true/false)]` on a type to set whether `compress_int` is enabled for that type. ForyRust will directly compute the `version_hash` at compile time based on the type’s corresponding `compress_int`.

2. If the parameter macro is not used, `compress_int` defaults to `true`.

3. When registering, ForyRust sets the `TypeResolver`’s `compress_int` value based on the `is_compress_int()` return value of the registered type. Therefore, if there is a conflict, it will panic immediately. This requires that all types registered under a single Fory object must have the same `compress_int` value.

4. As mentioned above, I added a new method `is_compress_int()` to `StructSerializer`.

5. If no type has been registered yet, users can manually set `compress_int()` by calling `fory.compress_int(bool)`. However, if a type with `compress_int=true/false` has already been registered, setting `fory.compress_int(false/true)` will immediately panic.

6. For `EXT`, there’s no need to worry. If some struct types are already registered, `EXT`’s `compress_int` will match the other registered types. If not, it can still be manually set via `fory.compress_int(bool)`.

7. Since the default value of `compress_int` is `true`, it can be cumbersome for users who want to set it to `false` for everything. Therefore, I submitted #2916 to help alleviate this situation.

urlyy:
Still needs to address #2886 #2887, you can first review my approach and code.

urlyy:
Currently, the behavior of `int::fory_read_data()` is determined by the local `fory`'s `is_compress_int()`. Do we need to change it so that `Serializer#fory_read_data(remote_type_id)` performs different read operations based on the `type_id` that was read?

chaokunyang:
Could we remove `compress_int` API. The xlang is different pure java. In pure java, we only write code once to create fory. Now in xlang mode, we crate fory in every language, adding too much options is error-prone to introduce inconsistency. And rust don't support generate code dynamically, such option also introduce runtime cost. 

I prefer a macro attr:
```rust
struct Foo {
  [#fory(compress=False)]
  f1: i32
}
```

For java, we can create an annotation:
```java
class Foo {
  @ForyField(compress=false)
  int f1:
}
```

For python, we can create a `field` method like dataclass decorator.

We can support `#fory(compress=False)` for pure rust in this pr, so we don't have to support java annotation or python `field` in this pr

urlyy:
> We can support `#fory(compress=False)` for pure rust in this pr, so we don't have to support java annotation or python `field` in this pr

OK, I will implement the compress_int annotation on member variable for each language one by one in separate PRs, and then open another PR at the end for cross-language testing.

PR #2904, opened 2025-11-06
feat(java): implement FinalFieldReplaceResolveSerializer for final fields with writeReplace/readResolve methods.

mchernyakov:
## Why?

Based on the discussion https://github.com/apache/fory/discussions/2786.
We can optimize the payload in cases where we have final fields in some Object.

## What does this PR do?

Introduce `FinalFieldReplaceResolver` which does not write the classname into the payload.

### DIFF
--- a/java/fory-core/src/test/java/org/apache/fory/serializer/FinalFieldReplaceResolveSerializerTest.java
+++ b/java/fory-core/src/test/java/org/apache/fory/serializer/FinalFieldReplaceResolveSerializerTest.java
@@ -32,6 +32,7 @@ import lombok.Data;
 import lombok.EqualsAndHashCode;
 import org.apache.fory.Fory;
 import org.apache.fory.ForyTestBase;
+import org.apache.fory.config.CompatibleMode;
 import org.apache.fory.config.Language;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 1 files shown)


java/fory-core/src/main/java/org/apache/fory/builder/BaseObjectCodecBuilder.java
```diff
@@ -594,17 +610,26 @@ protected Expression writeClassInfo(
...
+  protected Expression getOrCreateSerializer(Class<?> cls) {
+    return getOrCreateSerializer(cls, false);
+  }
+
   /**
    * Returns a serializer expression which will be used to call write/read method to avoid virtual
    * methods calls in most situations.
    */
-  protected Expression getOrCreateSerializer(Class<?> cls) {
+  protected Expression getOrCreateSerializer(Class<?> cls, boolean finalField) {
```
chaokunyang: Could we avoid change existing API? It would be great if we only getOrCreateSerializer method body, and don't change any other method API

mchernyakov: do you mean you would like: in `ObjectCodecBuilder#serializeGroup(...)` directly call `getOrCreateSerializer(..., true)`? without any other `serializeForNullable(...)` etc. 
I can do it, would it be ok, because in some of those methods there are additional checks for `if (needWriteRef(typeRef)) {...}` ?

chaokunyang: No, I mean we don't change any method signature, and just change method body in `getOrCreateSerializer` or `getSerializerClass`

chaokunyang: the method signature are too complex if we add more parameter into it

chaokunyang: Or maybe we could add a serializeField/deserializeField method in BaseCodecBuilder, and you can intercept serialization for final field with replace resolver in there

mchernyakov: https://github.com/apache/fory/pull/2904/commits/f4aa0580b4c7a790215b627b0b2a452b24636b86 wdyt? unfortunately I have to carry `Descriptor` to determine the resolver 🙈

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
```diff
@@ -467,6 +467,8 @@ public final boolean isSerializable(Class<?> cls) {
 
   public abstract Class<? extends Serializer> getSerializerClass(Class<?> cls, boolean codegen);
 
+  public abstract Class<? extends Serializer> getSerializerClassFinalField(Class<?> cls);
```
chaokunyang, line 470: You can update `buildFieldsInfo` in `java/fory-core/src/main/java/org/apache/fory/serializer/AbstractObjectSerializer.java` directly, then it will be unnecessary to introduce a new method

java/fory-core/src/test/java/org/apache/fory/serializer/FinalFieldReplaceResolveSerializerTest.java
```diff
@@ -0,0 +1,449 @@
...
+import org.apache.fory.Fory;
+import org.apache.fory.ForyTestBase;
+import org.apache.fory.config.Language;
+import org.testng.annotations.Test;
+
+/**
+ * Test class for FieldReplaceResolveSerializer. This serializer is used for final fields that have
+ * writeReplace/readResolve methods.
+ */
+public class FinalFieldReplaceResolveSerializerTest extends ForyTestBase {
```
chaokunyang, line 43: Could we also add a compatible mode test to ensure MetaSharedSerializer also work?

mchernyakov: added

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
mchernyakov: can you please elaborate on this? I used https://github.com/mchernyakov/fory/blob/23a5e68aa39d7e54426ebb40b9985dcee497a8e8/java/fory-core/src/test/java/org/apache/fory/serializer/FinalFieldReplaceResolveSerializerTest.java#L349 as an example and set breakpoints in `buildFieldsInfo` , but the flow does not reach the method. It goes through `ClassResolver#getSerializerClass(...)` to fetch the serializer (which must be `FinalFieldReplaceResolver`).

chaokunyang: You should use parameterized tests with codegen set to false/true. The `buildFieldsInfo ` will be invoked only when `withCodegen(false)`

PR #2905, opened 2025-11-07
docs(java): add logging section

mosinnik:
## Why?

Found that Fory doesn't use slfj4 layout in my app by default. Didn't find any docs about this and how to setup slf4j, so I added it

## What does this PR do?

Added Logging section to java guide
Removed strange 4 apostrophes in .md (I don't understand why it was used and for what purpose, so it may be not ok)


chaokunyang:
strange 4 apostrophes are typos, thanks for fix this

Issue #2915, opened 2025-11-15
[Rust] Add a default configuration struct


urlyy:
### Feature Request

for this:
```rust
impl Default for Fory {
    fn default() -> Self {
        Fory {
            compatible: false,
            xlang: false,
            share_meta: false,
            type_resolver: TypeResolver::default(),
            compress_string: false,
            max_dyn_depth: 5,
            check_struct_version: false,
            write_context_pool: OnceLock::new(),
            read_context_pool: OnceLock::new(),
        }
    }
}
```
Although Fory allows each Fory object to have its flags set individually, I believe most users will use the same optimal configuration when working with multiple Fory objects. 
And Using a DefaultConfig variable to manage default values is also a reasonable approach. It only slightly affects the performance when creating Fory objects and does not impact the performance of subsequent serialization.

PR #2916, opened 2025-11-15
feat(rust): support default_config

urlyy:
## What does this PR do?
Add a `Config` struct and a global variable `DEFAULT_CONFIG` to manage the default configuration values for Fory.

## Related issues
#2915

Like this, users can get/set default config.
```rust
use fory_core::{
    fory::{get_default_config, set_default_config, Config},
    Fory,
};

#[test]
fn test_config() {
    let default_cfg = get_default_config();
    assert_eq!(default_cfg.compatible, false);
    let compatible = true;
    let new_cfg = Config {
        compatible,
        ..default_cfg
    };
    set_default_config(new_cfg);
    let fory = Fory::default();
    assert_eq!(fory.is_compatible(), compatible);
}
```

### DIFF
--- a/rust/tests/tests/test_config.rs
+++ b/rust/tests/tests/test_config.rs
@@ -1,3 +1,20 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


urlyy:
Has this already been implemented in #2947? Should I close this PR now?

chaokunyang:
@urlyy How about rewrite this PR to create a `ForyBuilder`? We keep current `default()` and set methods, but also provide a `ForyBuilder` API, users can use builder instance to share similair config.

chaokunyang:
And please make `build()` method return Result, so in future we can add some congiuration validation.

PR #2944, opened 2025-11-28
perf(c++): optimize cpp serialization performance

chaokunyang:
## What does this PR do?

This PR significantly optimizes C++ serialization performance through multiple strategies:

### Performance Optimizations

1. **Buffer Performance Improvements**
   - Zero-copy wrapping for `vector<uint8_t>` to avoid unnecessary data copies
   - Optimized buffer growth and memory allocation strategies
   - Refined primitive write operations with separate handling for primitive fields

2. **Thread-Safe Fory Implementation**
   - Added `ThreadSafeFory` class for concurrent usage
   - Optimized `ensure_finalized` checks to reduce overhead

3. **Compile-Time Type Indexing**
   - Introduced compile-time type index (`cpp/fory/meta/type_index.h`) to eliminate runtime `typeid` overhead
   - Removes runtime type identification for non-polymorphic types, improving dispatch performance

4. **Deserialization Performance**
   - Optimized deserialization path for faster object reconstruction

5. **Varint Encoding Optimization**
   - Uses bulk writes (uint64_t) for varint encoding to reduce memory operations

### Code Improvements

- Simplified `fory.h` structure
- Added `const ref` support to `Result` type
- Improved serializer traits with new `serializer_traits.h`
- Enhanced struct serializer with better field handling

## Related issues

#2958 

## Benchmark
| Datatype | Operation | Fory (ns) | Protobuf (ns) | Faster |
| Sample | Deserialize | 402.8 | 453.2 | Fory (1.1x) |
| Struct | Serialize | 10.9 | 19.7 | Fory (1.8x) |

Sample Deserialize and Struct Serialize are faster than protobuf, others are still slow. I will optimize them in next PR by reduce type dispatch cost

### Files Changed Summary

| Category | Files | Key Changes |
|----------|-------|-------------|
| Serialization Core | `fory.h`, `struct_serializer.h`, `serializer_traits.h` | Major refactoring and optimization |
| Buffer | `buffer.h`, `buffer.cc` | Zero-copy support, overflow fixes |
| Type System | `type_resolver.h`, `type_index.h` | Compile-time type indexing |
| Result Type | `result.h`, `result_test.cc` | Const ref support, improved API |
| Benchmarks | `benchmark.cc`, `profile.sh`, `run.sh` | Enhanced profiling capabilities |

### DIFF
--- a/cpp/fory/util/buffer.h
+++ b/cpp/fory/util/buffer.h
@@ -833,6 +833,10 @@ public:
           new_ptr =
               static_cast<uint8_t *>(malloc(static_cast<size_t>(new_size)));
           if (new_ptr) {
+            // Copy existing data before switching to owned buffer
+            if (data_ && size_ > 0) {
+              std::memcpy(new_ptr, data_, size_);
+            }
             own_data_ = true;
           }
         }


pandalee99 (approved):
LGTM

benchmarks/cpp_benchmark/benchmark.cc
```diff
@@ -211,23 +237,69 @@ static void BM_Fory_Struct_Serialize(benchmark::State &state) {
 }
 BENCHMARK(BM_Fory_Struct_Serialize);
 
+// Fory with reused buffer (fair comparison with protobuf which reuses string)
```
LiangliangSui: Is there a unified way of marking the comments here? Some function body comments use //, while others use ///

chaokunyang: It's different.:
- `///` is a Doxygen-style documentation comment
- `//` is normal comment

LiangliangSui: Are you planning to use Doxygen to generate HTML files in the future?

chaokunyang: Yes, we need to that in future

PR #2953, opened 2025-12-01
fix(java): use single quotes in Python command for Windows compatibility

### LINKED ISSUE #2833: [Java] The pre-check for CrossLanguageTest.java was written incorrectly for Windows11
### Search before asking

### Version

Latest

### Component(s)

Java

### Minimal reproduce step

Here is the code in ForyJava:
```java
public static void verifyPyforyInstalled() {
  // Don't skip in CI, fail if something goes wrong instead
  if (System.getenv("FORY_CI") != null) {
    return;
  }
  try {
    if (executeCommand(
        Arrays.asList(
            "python",
            "-c",
            "import importlib.util, sys; sys.exit(0 if importlib.util.find_spec(\"pyfory\") is None else 1)"),
        10,
        Collections.emptyMap())) {
      throw new SkipException("pyfory not installed");
    }
  } catch (RuntimeException e) {
    if (e.getCause() instanceof IOException
        && e.getMessage().contains("No such file or directory")) {
      throw new SkipException("Python not installed or not found in PATH");
    }
    throw e;
  }
}
```

So I write a same unit test:
```java
@Test
public void testVerifyPyfory() {
    boolean result = executeCommand(
            Arrays.asList("python", "-c",
                    "import importlib.util, sys; sys.exit(0 if importlib.util.find_spec(\"pyfory\") is None else 1)"),
            10,
            Collections.emptyMap()
    );

    System.out.println("executeCommand returned: " + result);

    if (result) {
        System.out.println("pyfory is NOT installed - would skip tests");
    } else {
        System.out.println("pyfory is installed - would run tests");
    }
}
```

I just reinstalled my `Windows 11` OS, so I don’t have pyfory installed. I’m currently using PowerShell. And output:
```bash
Executing command:  python -c import importlib.util, sys; sys.exit(0 if importlib.util.find_spec("pyfory") is None else 1)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    import importlib.util, sys; sys.exit(0 if importlib.util.find_spec(pyfory) is None else 1)
                                                                       ^^^^^^
NameError: name 'pyfory' is not defined
executeCommand returned: false
pyfory is installed - would run tests
```

It seems that `find_spec(\"pyfory\")` gets converted to `find_spec(pyfory)`, so it throws an error because it can’t find a variable named `pyfory`.

Regarding the return value, a Python `syntax error` and `pyfory found` produce the `same` return value `1`, and that’s why not skip unit tests.

### My solution
You can change it to `find_spec('pyfory')`. This works on both my macOS and Windows 11. However, I don’t have pyfory installed, so I haven’t tested the case where pyfory is found.

zhan7236:
## Why?

On Windows 11 PowerShell, the escaped double quotes `\"pyfory\"` in the Python command get processed and removed, causing Python to receive `pyfory` as a variable name instead of the string `"pyfory"`. This results in a `NameError`:

```
NameError: name 'pyfory' is not defined
```

Since Python syntax errors and "pyfory found" both produce the same non-zero exit code, the test incorrectly proceeds instead of being skipped when `pyfory` is not installed.

## What does this PR do?

Replace escaped double quotes `\"pyfory\"` with single quotes `'pyfory'` in the Python command within `verifyPyforyInstalled()` method. Single quotes work correctly on both Windows and Unix systems.

**Changes:**
- Modified `java/fory-test-core/src/main/java/org/apache/fory/test/TestUtils.java`
- Changed `find_spec(\"pyfory\")` to `find_spec('pyfory')`

## Related issues

Fixes #2833

No user-facing changes. This is an internal test utility fix.

## Benchmark

Not applicable. This change does not affect performance.

### DIFF
--- a/java/fory-test-core/src/main/java/org/apache/fory/test/TestUtils.java
+++ b/java/fory-test-core/src/main/java/org/apache/fory/test/TestUtils.java
@@ -116,7 +116,7 @@ public class TestUtils {
           Arrays.asList(
               "python",
               "-c",
-              "import importlib.util, sys; sys.exit(0 if importlib.util.find_spec(\"pyfory\") is None else 1)"),
+              "import importlib.util, sys; sys.exit(0 if importlib.util.find_spec('pyfory') is None else 1)"),
           10,
           Collections.emptyMap())) {
         throw new SkipException("pyfory not installed");


chaokunyang (approved):
LGTM, thanks for this fix

Issue #2958, opened 2025-12-02
[C++] Optimize C++ Serialization Performance


chaokunyang:
### Feature Request

Fory binary protocol is more efficient than protobuf, so the fory c++ serialization should be faster than protobuf.

We need to benchmark fory c++ serialization performance, analyze bottleneck and optimize the implementation

### Additional context

#2906 #1017

chaokunyang:
Finished, fory c++ is now much faster than protobuf c++

Issue #2906, opened 2025-11-10
RoadMap for C++ serialization 1.0


chaokunyang:
### Feature Request

Add feature-complete c++ serialization implementation

### Describe the solution you'd like

- [x] Add rust style `Result` in fory cpp, this is similiar to c++23 expected
- [x] Add type resolver for object graph in c++
- [x] Add reference resolver for object graph in c++
- [x] Support serialization for basic types
- [x] Support serialization for nested collection/map/set types using tempate template specialization
- [x] Support serialization for struct using using tempate template specialization and compile-time fields sort
- [x] Support type meta encoding for schema evolution
- [x] Support schema evolution serialization based on type meta
- [x] align object graph xlang serialization with java #2909 
- [x] #2928
- [x] #2922
- [x] #2920
- [x] #2921
- [x] #2943
- [x] #2974
- [x] #3022
- [x] #3037
- [x] #2958 
- [x] #3088 
- [x] #3093 
- [x] #3104
- [x] #3107 
- [x] #2976
- [x] #2977
- [x] #3193
- [x] #3200

### Additional context

#1017

LiangliangSui:
Do we still need to develop some of the planned CPP features that have been planned?

chaokunyang:
@LiangliangSui I have not support circular reference, would you like to take over it?  You need to create a weak_ptr type which supports udpate its internal pointer, as we do in rust `weak.rs`

LiangliangSui:
> [@LiangliangSui](https://github.com/LiangliangSui) I have not support circular reference, would you like to take over it? You need to create a weak_ptr type which supports udpate its internal pointer, as we do in rust `weak.rs`

I'll take it over. Is this urgent? Maybe I need some time.

chaokunyang:
> > [@LiangliangSui](https://github.com/LiangliangSui) I have not support circular reference, would you like to take over it? You need to create a weak_ptr type which supports udpate its internal pointer, as we do in rust `weak.rs`
> 
> I'll take it over. Is this urgent? Maybe I need some time.

It's not urgent, I will make a release without circular ref tracking first

chaokunyang:
> > [@LiangliangSui](https://github.com/LiangliangSui) I have not support circular reference, would you like to take over it? You need to create a weak_ptr type which supports udpate its internal pointer, as we do in rust `weak.rs`
> 
> I'll take it over. Is this urgent? Maybe I need some time.

@LiangliangSui FYI, this feature has been implemented in #3109

PR #2960, opened 2025-12-02
perf(c++): optimize primitive struct fields read performance

chaokunyang:
## What does this PR do?

 1. Type-based encoding detection: Added compile-time helpers to correctly distinguish signed (varint) vs unsigned (fixed) integers:
    - field_is_fixed_primitive<Index>() - bool, int8, uint8, int16, uint16, uint32, uint64, float, double
    - field_is_varint_primitive<Index>() - int32_t, int, int64_t, long long (zigzag varint)
  2. Optimized fixed field reading:
    - Compute field offsets at compile time with compute_fixed_field_offset<T, I>()
    - Read all fixed fields at absolute offsets without per-field reader_index updates
    - Single reader_index update after all fixed fields
  3. Optimized varint field reading:
    - Track offset locally during batch reading
    - Removed overly conservative max-varint-bytes pre-check (varints are variable-length)
    - Single reader_index update after all varints
  4. Three-phase deserialization:
    - Phase 1: Batch read leading fixed-size primitives
    - Phase 2: Batch read consecutive varint primitives
    - Phase 3: Read remaining fields normally

## Related issues

#2958 
#2906 

## Benchmark

| Datatype | Operation | Fory (ns) | Protobuf (ns) | Faster |
|----------|-----------|-----------|---------------|--------|
| Sample | Serialize | 103.9 | 59.2 | Protobuf (1.8x) |
| Sample | Deserialize | 329.3 | 478.1 | Fory (1.5x) |
| Struct | Serialize | 10.3 | 20.1 | Fory (1.9x) |
| Struct | Deserialize | 19.1 | 16.0 | Protobuf (1.2x) |

### DIFF
--- a/cpp/fory/serialization/struct_serializer.h
+++ b/cpp/fory/serialization/struct_serializer.h
@@ -989,6 +989,23 @@ constexpr size_t compute_fixed_field_write_offset() {
   return offset;
 }
 
+/// Helper to write a single fixed-size primitive field at compile-time offset.
+/// No lambda overhead - direct function call that will be inlined.
+template <typename T, size_t SortedIdx>
+FORY_ALWAYS_INLINE void write_single_fixed_field(const T &obj, Buffer &buffer,
+                                                 uint32_t base_offset) {
+  using Helpers = CompileTimeFieldHelpers<T>;
+  constexpr size_t original_index = Helpers::sorted_indices[SortedIdx];
+  constexpr size_t field_offset =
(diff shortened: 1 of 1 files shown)


chaokunyang:
The benchmark is unfair, fory is actually same performance for Sample serialization, but in current benchmark, fory didn't wirte to a preallocsated buffer, it allocate a buffer every time.

This is fixed in https://github.com/apache/fory/pull/2963

PR #2967, opened 2025-12-03
perf(c++): vector serialization reduces element iteration

LiangliangSui:
## Why?

Vector serialization reduces element iteration

## What does this PR do?

When the condition `is_nullable_v<T>` is true, then iterate through the vector elements to check if at least one of them is null. Otherwise, there is no need to continue the iteration.

## Related issues

https://github.com/apache/fory/issues/2958

## Benchmark

The current benchmark is not covered.

### DIFF
--- a/cpp/fory/serialization/collection_serializer.h
+++ b/cpp/fory/serialization/collection_serializer.h
@@ -385,21 +385,7 @@ struct Serializer<
       return Result<void, Error>();
     }
 
-    // Build header bitmap
-    bool has_null = false;
-    if constexpr (is_nullable_v<T>) {
-      for (const auto &elem : vec) {
-        if (is_null_value(elem)) {
-          has_null = true;
-          break;
-        }
-      }
-    }
-
     uint8_t bitmap = 0b1000; // IS_SAME_TYPE
-    if (has_null) {
-      bitmap |= 0b0010; // HAS_NULL
-    }
 
     // Per Rust collection.rs: is_elem_declared = has_generics &&
     // !need_to_write_type_for_field(...)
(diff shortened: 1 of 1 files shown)


cpp/fory/serialization/collection_serializer.h
```diff
@@ -385,21 +385,7 @@ struct Serializer<
       return Result<void, Error>();
     }
 
-    // Build header bitmap
-    bool has_null = false;
-    if constexpr (is_nullable_v<T>) {
```
chaokunyang, line 390: This is executed at compile time, if elememt is not nullable, the whole branch will be removed by compiler. There is not duplicate iteration.

cpp/fory/serialization/collection_serializer.h
```diff
@@ -385,21 +385,7 @@ struct Serializer<
...
-      for (const auto &elem : vec) {
-        if (is_null_value(elem)) {
-          has_null = true;
-          break;
-        }
-      }
-    }
-
     uint8_t bitmap = 0b1000; // IS_SAME_TYPE
-    if (has_null) {
```
chaokunyang, line 400: you can  declare `bitmap` early, and move `bitmap |= 0b0010;` into `if constexpr (is_nullable_v<T>) {`

cpp/fory/serialization/collection_serializer.h
```diff
@@ -468,6 +462,14 @@ struct Serializer<
         }
       }
     } else {
+      // Write header
+      ctx.write_uint8(bitmap);
+
+      // Write element type info only if !IS_DECL_ELEMENT_TYPE
+      if (!is_elem_declared) {
```
chaokunyang, line 469: those code are duplicated

cpp/fory/serialization/collection_serializer.h
LiangliangSui: Sorry, I overlooked the part `if constexpr (is_nullable_v<T>) {...}" here. }`

Issue #2918, opened 2025-11-21
[Java] CopyOnWriteArraySet serialization fails with ClassCastException


vybhavjs:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fory/issues) and found no similar issues.

### Version

OS: `macOS 26.1`
Fory: `0.13.1`
Java : `java 21.0.3 2024-04-16 LTS`

### Component(s)

Java

### Minimal reproduce step

```
import org.apache.fory.Fory;
import org.apache.fory.config.Language;

import java.util.concurrent.CopyOnWriteArraySet;

public class CopyOnWriteArraySetTest {
    public static void main(String[] args) {
        Fory fory = Fory.builder()
                .withLanguage(Language.JAVA)
                .requireClassRegistration(false)
                .build();

        CopyOnWriteArraySet<String> set = new CopyOnWriteArraySet<>();
        set.add("Value1");
        set.add("Value2");

        byte[] serialized = fory.serialize(set);
        CopyOnWriteArraySet deserialised = fory.deserialize(serialized, CopyOnWriteArraySet.class);
        System.out.println("deserialised=" + deserialised);
    }
}
```

### What did you expect to see?

Fory should be able to serialize `java.util.concurrent.CopyOnWriteArraySet` objects.

### What did you see instead?

Serialization fails with a `ClassCastException`.

<details>
<summary>View full stack trace</summary>

```
Exception in thread "main" org.apache.fory.exception.SerializationException: java.lang.ClassCastException: class org.apache.fory.collection.CollectionSnapshot cannot be cast to class java.util.List (org.apache.fory.collection.CollectionSnapshot is in unnamed module of loader 'app'; java.util.List is in module java.base of loader 'bootstrap')
	at org.apache.fory.Fory.processSerializationError(Fory.java:362)
	at org.apache.fory.Fory.serialize(Fory.java:331)
	at org.apache.fory.Fory.serialize(Fory.java:278)
	at com.fory_tests.CopyOnWriteArraySetTest.main(CopyOnWriteArraySetTest.java:19)
Caused by: java.lang.ClassCastException: class org.apache.fory.collection.CollectionSnapshot cannot be cast to class java.util.List (org.apache.fory.collection.CollectionSnapshot is in unnamed module of loader 'app'; java.util.List is in module java.base of loader 'bootstrap')
[...]

PR #2999, opened 2025-12-05
fix(java): support serialization of CopyOnWriteArraySet

LiangliangSui:
## What does this PR do?

Support serialization of CopyOnWriteArraySet

## Related issues

https://github.com/apache/fory/issues/2918

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializers.java
+++ b/java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializers.java
@@ -296,7 +296,6 @@ public class CollectionSerializers {
 
     @Override
     public CopyOnWriteArrayList copy(CopyOnWriteArrayList originCollection) {
-      Preconditions.checkArgument(supportCodegenHook);
       CopyOnWriteArrayList newCollection = new CopyOnWriteArrayList();
       if (needToCopyRef) {
         fory.reference(originCollection, newCollection);
@@ -330,7 +329,6 @@ public class CollectionSerializers {
 
     @Override
     public CopyOnWriteArraySet copy(CopyOnWriteArraySet originCollection) {
-      Preconditions.checkArgument(supportCodegenHook);
       CopyOnWriteArraySet newCollection = new CopyOnWriteArraySet();
       if (needToCopyRef) {
         fory.reference(originCollection, newCollection);


java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializers.java
```diff
@@ -294,6 +295,27 @@ public CopyOnWriteArrayList onCollectionRead(Collection collection) {
...
+
+    @Override
+    public Collection newCollection(MemoryBuffer buffer) {
+      int numElements = buffer.readVarUint32Small7();
+      setNumElements(numElements);
+      return new CollectionContainer<>(numElements);
+    }
+
+    @Override
+    public CopyOnWriteArraySet onCollectionRead(Collection collection) {
```
chaokunyang, line 325: do you need to overwride `public List<?> copy(List<?> originCollection)` too?

LiangliangSui: Not necessary.

```java
// src/main/java/org/apache/fory/serializer/collection/CollectionSerializer.java
public T copy(T originCollection) {
  if (isImmutable()) {
    return originCollection;
  }
  Preconditions.checkArgument(supportCodegenHook);
  Collection newCollection = newCollection(originCollection);
  if (needToCopyRef) {
    fory.reference(originCollection, newCollection);
  }
  copyElements(originCollection, newCollection);
  return (T) newCollection;
}
```

chaokunyang: This is what I am worried, you will copy `CopyOnWriteArraySet` every time you put an element into the Set when copying.

You should overide copy method to avoid such cost. It seems copy for other collections can be optimized too.

LiangliangSui: > This is what I am worried, you will copy CopyOnWriteArraySet every time you put an element into the Set when copying.

Could you provide some details?

chaokunyang: This is the behaviour of `CopyOnWriteArraySet`, and what it means `CopyOnWrite`

LiangliangSui: Done

LiangliangSui: @chaokunyang

java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionLikeSerializer.java
```diff
@@ -550,6 +552,21 @@ public void copyElements(Collection originCollection, Object[] elements) {
     }
   }
 
+  public void copyElementsAddAll(Collection originCollection, Collection newCollection) {
```
chaokunyang: This API is vague to me, and it should not be added as a public API of CollectionLikeSerializer. And most serializer would never use it, why not just create a private static method in CollectionSerializers with clear doc which indicate why it's needed, and clarify it's only used by CopyOnWriteList/Set?

java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializer.java
```diff
@@ -61,7 +61,11 @@ public T copy(T originCollection) {
     if (needToCopyRef) {
       fory.reference(originCollection, newCollection);
     }
-    copyElements(originCollection, newCollection);
+    onCollectionCopy(originCollection, newCollection);
     return (T) newCollection;
   }
+
+  protected void onCollectionCopy(Collection originCollection, Collection newCollection) {
```
chaokunyang: And I don't think we need this API. why not just overrite `copy` API, it's basically same.

java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionLikeSerializer.java
LiangliangSui: I have adopted the same definition as that of `public void copyElements(Collection originCollection, Collection newCollection)`, and`public void copyElements(Collection originCollection, Collection newCollection)` is only used in the CollectionSerializer.

java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializer.java
LiangliangSui: In the `CollectionSerializer#copy` method, it is a framework. The specific logic for the copy operation is different. I don't think it's necessary to rewrite the entire logic all over again.

```java
public T copy(T originCollection) {
  if (isImmutable()) {
    return originCollection;
  }
  Preconditions.checkArgument(supportCodegenHook);
  Collection newCollection = newCollection(originCollection);
  if (needToCopyRef) {
    fory.reference(originCollection, newCollection);
  }
  copyElements(originCollection, newCollection);
  return (T) newCollection;
}
```

java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionLikeSerializer.java
LiangliangSui: I think the naming needs to be revised. Do you have any good suggestions for modifying this name?

chaokunyang: I think this method should be removed, it's only related to COpyOnWriteSet/LIst, why you add it to base class

LiangliangSui: OK, let's move it to the CollectionSerializers.

LiangliangSui: The `copyElementsAddAll` method requires a `TypeResolver` and cannot be made static. It also cannot be placed in the `CollectionSerializers`. 
 If it needs to be further downgraded, it can only be placed in the `ConcurrentCollectionSerializer`.

java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializer.java
chaokunyang: How about:
```java
    public CopyOnWriteArrayList copy(T originCollection) {
      CopyOnWriteArrayList newCollection = new CopyOnWriteArrayList();
      if (needToCopyRef) {
        fory.reference(originCollection, newCollection);
      }
      List copiedList = new ArrayList<>(originCollection.size());
      copyElements(originCollection, copiedList);
      newCollection.addAll(copiedList);
      return newCollection;
    }
```

We can't use `newCollection` API, it return a stub collection, which will make wrong reference tracking if we have circualr references.

LiangliangSui: Good suggestion, done.

java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializers.java
```diff
@@ -292,6 +293,53 @@ public CopyOnWriteArrayList onCollectionRead(Collection collection) {
       Object[] elements = ((CollectionContainer) collection).elements;
       return new CopyOnWriteArrayList(elements);
     }
+
+    @Override
+    public CopyOnWriteArrayList copy(CopyOnWriteArrayList originCollection) {
+      Preconditions.checkArgument(supportCodegenHook);
```
chaokunyang:
```suggestion
```

java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializers.java
```diff
@@ -292,6 +293,53 @@ public CopyOnWriteArrayList onCollectionRead(Collection collection) {
...
+
+    @Override
+    public CopyOnWriteArraySet onCollectionRead(Collection collection) {
+      Object[] elements = ((CollectionContainer) collection).elements;
+      return new CopyOnWriteArraySet(Arrays.asList(elements));
+    }
+
+    @Override
+    public CopyOnWriteArraySet copy(CopyOnWriteArraySet originCollection) {
+      Preconditions.checkArgument(supportCodegenHook);
```
chaokunyang:
```suggestion
```

Issue #3000, opened 2025-12-05
[Java] create @ForyField annotation to provide extra meta for perf/space optimization


chaokunyang:
## Feature Request

Create a `@ForyField` annotation to provide extra metadata for performance and space optimization during xlang serialization.

## Is your feature request related to a problem? Please describe

Currently, Fory's xlang serialization treats all object fields uniformly:
1. **Null checks are always performed** - Even for fields that are never null, Fory writes a null/ref flag (1 byte per field)
2. **Reference tracking is always applied** (when enabled globally) - Even for fields that won't be shared/cyclic, objects are added to `IdentityMap` with hash lookup cost
3. **Field names use meta string encoding** - In schema evolution mode, field names are encoded using meta string compression, but for fields with long names, this still takes space

These defaults ensure correctness but introduce unnecessary overhead when the developer has more specific knowledge about their data model.

## Describe the solution you'd like

Add a `@ForyField` annotation that allows developers to provide field-level hints to optimize serialization:

```java
public @interface ForyField {
    /**
     * Whether this field can be null.
     * When set to false, Fory skips writing the null flag (saves 1 byte).
     * Default: true (field can be null)
     */
    boolean nullable() default true;

    /**
     * Whether to track references for this field.
     * When set to false:
     * - Avoids adding the object to IdentityMap (saves hash map overhead)
     * - Skips writing ref tracking flag (saves 1 byte when combined with nullable=false)
     * Default: true (reference tracking enabled when global ref tracking is on)
     */
    boolean ref() default true;

    /**
     * Field tag ID for schema evolution mode.
     * When set (>= 0), uses this numeric ID instead of field name string.
     * This provides more compact encoding especially for fields with long names.
     * Default: -1 (use field name)
     */
    int id() default -1;
}
```

### Usage Example

```java
class Foo {
    // Field f1: non-nullable string
    // - Skips null flag: saves 1 byte
    // - Uses tag ID 0 instead of field name "f1": more compact in schema evolution mode
    @ForyField(nullable = false, id = 0)
    String f1;

    // Field f2: non-nullable, non-shared reference
    // - Skips null flag: saves 1 byte
    // - Skips ref tracking: avoids IdentityMap lookup and ref flag
    // - Uses tag ID 1 instead of field name: more compact
    @ForyField(nullable = false, ref = false, id = 1)
    Bar f2;

    // Field with long name: use tag ID to save space
    @ForyField(id = 2)
[...]

Issue #3005, opened 2025-12-05
[Go] Add fory struct tags for field metadata


chaokunyang:
## Feature Request

Add support for `fory` struct tags to provide field-level metadata for performance and space optimization during xlang serialization.

## Is your feature request related to a problem? Please describe

Currently, Fory's Go xlang serialization treats all struct fields uniformly:
1. **Null checks are always performed** - Even for fields that are never nil, Fory writes a null/ref flag (1 byte per field)
2. **Reference tracking is always applied** (when enabled globally) - Even for fields that won't be shared/cyclic, objects are tracked with hash lookup cost
3. **Field names use meta string encoding** - In schema evolution mode, field names are encoded using meta string compression, but for fields with long names, this still takes space

These defaults ensure correctness but introduce unnecessary overhead when the developer has more specific knowledge about their data model.

## Describe the solution you'd like

Add support for `fory` struct tags with field metadata:

```go
type Foo struct {
    // Field F1: non-nullable (default), no ref tracking (default)
    // Tag ID 0 provides compact encoding in schema evolution mode
    F1 string `fory:"id=0"`
    
    // Field F2: non-nullable (default), no ref tracking (default)
    F2 Bar `fory:"id=1"`
    
    // Field F3: nullable field that may contain nil values
    F3 *string `fory:"id=2,nullable"`
    
    // Field Parent: shared reference that needs tracking (e.g., for circular refs)
    Parent *Node `fory:"id=3,ref,nullable"`
    
    // Field with long name: tag ID provides significant space savings
    VeryLongFieldNameThatWouldTakeManyBytes string `fory:"id=4"`
    
    // Explicit opt-out: use field name encoding but get nullable optimization
    OptionalField *string `fory:"id=-1,nullable"`
}
```

### Tag Syntax

```go
`fory:"id=<int>[,nullable][,ref]"`
```

| Tag Component | Description |
|---------------|-------------|
| `id=N` | **Required**. Tag ID for field encoding. N >= 0 uses tag ID; N = -1 uses field name |
| `nullable` | Optional. Field can be nil. Required for pointer types |
| `ref` | Optional. Enable reference tracking for this field |

### Design Decision: Required `id`

The `id` tag is **required** when using `fory` struct tags:
- `id=0` to `id=N`: Use tag ID encoding (compact)
- `id=-1`: Explicit opt-out, use field name encoding

Rationale:
1. **Explicit control**: Using `fory` tag means opting into explicit control
[...]

Issue #3003, opened 2025-12-05
[C++] Field metadata for optimized serialization


chaokunyang:
Add compile-time field metadata support for C++ struct serialization, enabling performance and space optimization during xlang serialization. This provides two complementary APIs:

1. **`fory::field<>` template** - Inline metadata in struct definition
2. **`FORY_FIELD_TAGS` macro** - Non-invasive metadata added separately

### Current Limitations

The current C++ implementation applies **uniform handling** to all struct fields:

1. **Null Checks**: Written for all non-primitive fields, even when they can never be null (1 byte per field overhead)
2. **Reference Tracking**: Applied globally regardless of whether fields actually need shared reference support
3. **Field Name Encoding**: Long field names consume extra space via meta string compression instead of compact tag IDs

### Expected Benefits

For a 10-field structure with tag IDs and explicit nullable/ref settings:
- **Space savings**: ~20 bytes per object (eliminates null flags when not needed)
- **CPU savings**: 10 fewer hash operations per serialization (no meta string lookups for field names)
- **Compact encoding**: Varint tag IDs (1-2 bytes) replace meta string encoded names (~5-15 bytes)

### Core Principles

1. **All fields are non-nullable by default** (except `std::optional`)
2. **All fields have no reference tracking by default**
3. **Only 2 tags**: `fory::nullable` and `fory::ref`
4. **Type-safe**: `nullable` only valid for pointer types, `ref` only valid for `shared_ptr`

### Type Rules

| Type | Default Nullable | Default Ref | Allowed Options |
|------|-----------------|-------------|-----------------|
| Primitives, `std::string` | **false** | false | **None** (no options allowed) |
| `std::optional<T>` | **true** (inherent) | false | **None** (inherent nullable) |
| `std::shared_ptr<T>` | **false** | false | `fory::nullable`, `fory::ref` |
| `std::unique_ptr<T>` | **false** | false | `fory::nullable` only |

---

## API 1: fory::field<> Template (Inline)

The `fory::field<>` template wraps field types with compile-time metadata:

```cpp
namespace fory {
  struct nullable {};  // Mark shared_ptr/unique_ptr as nullable
  struct not_null {};  // Explicitly mark as non-nullable (default)
  struct ref {};       // Enable reference tracking for shared_ptr

  template <typename T, int16_t Id, typename... Options>
  class field { /* same memory layout as T */ };
}
```

### Usage

```cpp
struct Document {
  // Primitives - always non-nullable, NO options allowed
[...]

PR #3021, opened 2025-12-08
feat(java): enhance ForyField annotation with tag ID support for optimized serialization

mchernyakov:
https://github.com/apache/fory/issues/3000 

Create/update a @ForyField annotation to provide extra metadata for performance and space optimization during java/xlang serialization.


mchernyakov:
hey @chaokunyang I drafted a PR for https://github.com/apache/fory/issues/3000, please take a look.
How do you plan to merge and release it since the change requires to be present in every language? do we need an epic branch?

chaokunyang:
If we don't annotate fields, it won't break exisint bianry protocol, so we can merge this first even implementation for other languages is not ready.

But I think we can implement this for other languages before next release.

java/fory-core/src/main/java/org/apache/fory/serializer/AbstractObjectSerializer.java
```diff
@@ -1000,9 +1000,7 @@ private InternalFieldInfo(Fory fory, Descriptor d, short classId) {
       nullable = d.isNullable();
       if (fory.trackingRef()) {
         trackingRef =
-            foryField != null
-                ? foryField.trackingRef()
-                : fory.getClassResolver().needToWriteRef(typeRef);
+            foryField != null ? foryField.ref() : fory.getClassResolver().needToWriteRef(typeRef);
```
chaokunyang: Do you need to update codegen code path? Maybe we should also update `BaseObjectCodecBuilder`

mchernyakov: https://github.com/apache/fory/pull/3021/commits/02b3d2858060f1809a32ddf3c8feba77f0e1b4f9

java/fory-core/src/main/java/org/apache/fory/type/Descriptor.java
```diff
@@ -89,6 +89,8 @@ public static void clearDescriptorCache() {
   private final Method writeMethod;
   private ForyField foryField;
   private boolean nullable;
```
chaokunyang, line 91: How about change `nullable/trackingRef` to `Optional<Boolean>` type and use `isPresent` to check whether annotation config is isPresent

mchernyakov: I'm not sure about it. 
Afaik `Optional` as field is antipattern in java, I think the usage of `Optional` suppose to be as a return value, which could explicitly be checked against nullability. 
If you insist I can add it ofc, no problem with that.

chaokunyang: If so ,could we add another two fields and methods to flatg that whether thoes two config has been set?

mchernyakov: ~https://github.com/apache/fory/pull/3021/commits/5a642305bd01cb1d12945f96b39cfdb1692b5333 -- would it be ok? or do you want explicit checks in, for example, `BaseObjectCodecBuilder` ?~

mchernyakov: @chaokunyang if we consider `Descriptor` class: do we really need those additional fields? probably it could be better when there is only one "source of truth" for nullability and ref tracking in the class : 
```
  public boolean isNullable() {
    return nullable;
  }

  public boolean isTrackingRef() {
    return trackingRef;
  }
```
basically, I propose to set those fields in the constructors -- using `@ForyField` if present, `isPrimitive` check and the builder. We just need to agree on priorities like
* the global ref tracking config overrides the config from `@ForyField`
* the config from the builder overrides `@ForyField`
* etc.
wdyt?

chaokunyang: Looks good to me, it's much cleaner anc elegant. 

But the default Descriptor is build and cached statically, you need to update `getAllDescriptorsMap/getFieldDescriptors` in `ClassResolver/TypeResolver` to create new `Descriptor` with correct `nullable/trackRef` config. 

Could you implement that in this PR?

java/fory-core/src/main/java/org/apache/fory/meta/ClassDef.java
```diff
@@ -247,17 +248,37 @@ public List<Descriptor> getDescriptors(TypeResolver resolver, Class<?> cls) {
...
-            != null) {
+        String fullName = e.getKey().getDeclaringClass().getName() + "." + e.getKey().getName();
+        Descriptor desc = e.getValue();
+        if (descriptorsMap.put(fullName, desc) != null) {
           throw new IllegalStateException("Duplicate key");
         }
+
+        // If the field has @ForyField annotation with tag ID, index by tag ID
+        if (desc.getForyField() != null) {
+          int tagId = desc.getForyField().id();
```
chaokunyang: We also need some checks to avoid negative id if it's not -1, and also need to a=check duplciate id if it's not -1 too.

mchernyakov: yeah, thanks for pointing on that, done in https://github.com/apache/fory/pull/3021/commits/a4155a975a7af8fcdfded89d8ebf1be74bbe35e0

java/fory-core/src/main/java/org/apache/fory/builder/BaseObjectCodecBuilder.java
```diff
@@ -404,8 +404,15 @@ protected Expression serializeField(
       Expression fieldValue, Expression buffer, Descriptor descriptor) {
     TypeRef<?> typeRef = descriptor.getTypeRef();
     boolean nullable = descriptor.isNullable();
+    boolean explicitRefTracking = descriptor.isTrackingRef();
```
chaokunyang: Does this mean user can never disable ref tracking for a field by:
```java
class Struct {
  @ForyField(ref=false)
  Bar bar;
}
```

java/fory-core/src/main/java/org/apache/fory/type/Descriptor.java
```diff
@@ -89,6 +89,8 @@ public static void clearDescriptorCache() {
...
+  // trackingRef should only be true if explicitly set to true via @ForyField(ref=true)
+  // If no annotation or ref not specified, trackingRef stays false and type-based tracking applies
   private boolean trackingRef;
```
chaokunyang, line 94: We need to support disable ref tracking for a field by `@ForyField(ref=False)`. This can improve performance and reduce space overhead

mchernyakov: so, then we have priorities like
1)  @ForyField -- any params combination
2) the builder
3) the global ref tracking config

right?

chaokunyang: For `ref`:
1. if ref tracking is diabled globally, then ref is always false for all fields values
2. if ref tracking is enabled, then @ForyField has higher priority
3. otherwise, fallback to default ref tracking logic, use `needToWriteRef` instead

For `nullable`:
1. if `@ForyField` is configured, it has higher priority
2. if we support `NotNull/Nullable` annotation, thsi also has higher priority
3. othewise, fallback to default logic(only primitive types is not null).

mchernyakov: hm, there is an issue with collections: 
* when `useRefTracking` is `false` for the field (via `@ForyField()`), it writes the collection WITHOUT ref tracking wrapper
* but when deserializing, the `ArrayListSerializer.newCollection()` ALWAYS calls `fory.getRefResolver().reference(arrayList)` which expects ref tracking to be enabled

The issue is that `newCollection()` is being called regardless of whether ref tracking is enabled for this field or not.

So, I think we should give the global config the highest priority, hence the check in https://github.com/apache/fory/pull/3021#discussion_r2616725645 (`    boolean useRefTracking = explicitRefTracking || needWriteRef(typeRef);
`)is fine.

java/fory-core/src/main/java/org/apache/fory/type/Descriptor.java
mchernyakov: can you give an example when the things could go wrong? 
I imagined it like
```
class A {
  private Address addr;
}
```
vs

```
class B {
  @ForyField(id=0)
  private Address addr;
}
```
but in this case there will be a new generated map since classes are different. The same applies for cases with superclasses/subclasses etc, no?

java/fory-core/src/main/java/org/apache/fory/type/Descriptor.java
chaokunyang: @mchernyakov You can update `MapRefResolver` like this:

```java
  public void reference(Object object) {
    int refId = readRefIds.pop();
    if (refId >= 0) {
      setReadObject(refId, object);
    }
  }
```

Then for fields which will track ref but skip ref now, you can reserve an id `-1`:
```java
preserveRefId(-1)
```

`@ForyField` needs higher priority, it can improve performance signicantly. Most of times in an object graph, there is no shared/circular reference, but due to we can't differeiate those objects, we introduced lots of unnecesary ref tracking

mchernyakov: @chaokunyang addressed in https://github.com/apache/fory/pull/3021/commits/27e3280f715955f0ce60137affe7cab1ac80dad7 and https://github.com/apache/fory/pull/3021/commits/42f90f6c9c85eb8942205e9673774ca7563da44a

and there is still a question in https://github.com/apache/fory/pull/3021#discussion_r2617530972 

pls check

java/fory-core/src/main/java/org/apache/fory/type/Descriptor.java
chaokunyang: The global static Descriptor are different, and don't collide with each other. But it's no the final config for `ref`.

```java
class B {
  @ForyField(id=0, ref=true)
  private Address addr;
}
```

If `Fory` is created with `refTracking=false`, the `ref` should still be `false` even `@ForyField(id=0, ref=true)` .

So you need to update `ClassResolver/TypeResolver` to create some new `Descripters` when `refTracking=false`

mchernyakov: @chaokunyang https://github.com/apache/fory/pull/3021/commits/36d06d0cd5ac1b062963542b93013be71b62e321 , e8e33286379f5cd0d4d8162c338a864c26ef5dbc -- wdyt? potentiall there could be an extnetions of the value to track whether the descriptors were updated to avoid unnecessary iterations.

java/fory-core/src/main/java/org/apache/fory/serializer/AbstractObjectSerializer.java
chaokunyang: Maybe we could use `d.isTrackingRef()` direclty if we build corect Descriptor from `ClassResolver/TypeResolver/XtypeResolver`

java/fory-core/src/main/java/org/apache/fory/meta/ClassDef.java
```diff
@@ -247,17 +248,46 @@ public List<Descriptor> getDescriptors(TypeResolver resolver, Class<?> cls) {
...
       }
+
       descriptors = new ArrayList<>(fieldsInfo.size());
       for (FieldInfo fieldInfo : fieldsInfo) {
-        Descriptor descriptor =
-            descriptorsMap.get(fieldInfo.getDefinedClass() + "." + fieldInfo.getFieldName());
+        Descriptor descriptor;
+
+        // Try to match by tag ID first if the FieldInfo has a tag
+        if (fieldInfo.hasTag()) {
```
chaokunyang: How about naming tag to `fieldId`? tag gives me a sense that it can be string type

java/fory-core/src/main/java/org/apache/fory/type/Descriptor.java
```diff
@@ -432,10 +457,10 @@ public static SortedMap<Member, Descriptor> getAllDescriptorsMap(Class<?> clz) {
       });
 
   public static SortedMap<Member, Descriptor> getAllDescriptorsMap(
-      Class<?> clz, boolean searchParent) {
+      Class<?> clz, boolean searchParent, boolean globalRefTrackingEnabled) {
     try {
       Tuple2<SortedMap<Member, Descriptor>, SortedMap<Member, Descriptor>> tuple2 =
-          descCache.get(clz, () -> createAllDescriptorsMap(clz));
+          descCache.get(clz, () -> createAllDescriptorsMap(clz, globalRefTrackingEnabled));
```
chaokunyang: This is flaky, if `createAllDescriptorsMap` is invoked with `globalRefTrackingEnabled = true`, but later invoked with `globalRefTrackingEnabled=false`, then later invoke won't take efflect.

I prefer not passing parameters to `getAllDescriptorsMap/createAllDescriptorsMap`. We build Descriptor with default field annotation and whether it's primitive.

But when we get Descriptors from `TypeResolver/ClassResolver/XtypeResolver`, we create new `Descriptor` with correct settings. When we create new Descriptors, we should recompute based on `ForyField` annotation and `ref tracking` flag. 

Then in all serializers, we can take `trackingRef/nullable` as real truth.

java/fory-core/src/main/java/org/apache/fory/builder/BaseObjectCodecBuilder.java
```diff
@@ -405,7 +405,17 @@ protected Expression serializeField(
     TypeRef<?> typeRef = descriptor.getTypeRef();
     boolean nullable = descriptor.isNullable();
 
+    // ref tracking logic:
+    // - if the ref tracking is enabled globally, then @ForyField will be checked
+    // - otherwise, fallback to global config
+    boolean useRefTracking;
     if (needWriteRef(typeRef)) {
+      useRefTracking = descriptor.getForyField() == null || descriptor.isTrackingRef();
```
chaokunyang: with https://github.com/apache/fory/pull/3021/files#r2625961552, we can just use `descriptor.isTrackingRef()`

java/fory-core/src/main/java/org/apache/fory/meta/ClassDef.java
mchernyakov: https://github.com/apache/fory/pull/3021/commits/84348c375375e95f1a49bb1a9f01d47ecd55ab36

java/fory-core/src/main/java/org/apache/fory/type/Descriptor.java
mchernyakov: good, addressed in https://github.com/apache/fory/pull/3021/commits/2addb2bce53a4214f06ec2d86f802584d6c1dc7e

java/fory-core/src/main/java/org/apache/fory/builder/BaseObjectCodecBuilder.java
mchernyakov: with the current impl it will fail. 
consider a case
* the global ref tracking = `true`
* no `@ForyField` provided -> in the Descriptor we have `refTracking = false`
It could be handled ofc in `getAllDescriptorsMap/createAllDescriptorsMap` , but that means we have to take care for any cases related to the global ref tracking during the descriptor fetching.

what do you think?

chaokunyang: I prefer handle in `getAllDescriptorsMap/createAllDescriptorsMap`. For external classes, we can't add annotation to its fields, but in future we can provide an explicit API to allow users register fields meta, then we can use that directly in serialization and deserialization.

mchernyakov: addressed in https://github.com/apache/fory/pull/3021/commits/84686690e90eafdbf3617e96f773c0ffb0bffb88 , pls check

java/fory-core/src/main/java/org/apache/fory/serializer/AbstractObjectSerializer.java
mchernyakov: https://github.com/apache/fory/pull/3021/commits/b5c1a31f2236a5dad0c18c6b8df26a09ea894806

mchernyakov:
@chaokunyang good, will you also handle https://github.com/apache/fory/blob/main/java/fory-core/src/main/java/org/apache/fory/meta/ClassDef.java#L539 (there's a missing return I assume)? 

Also, should I take over https://github.com/apache/fory/issues/3005 and/or https://github.com/apache/fory/issues/3003 ?

chaokunyang:
I'm still refactor fory go. Would you like to take over #3003 instead?

mchernyakov:
> I'm still refactor fory go. Would you like to take over #3003 instead?

sure, I left a message in slack to you about it, pls check

PR #3057, opened 2025-12-16
fix(Rust): prevent obtaining generic type metadata on custom types(struct/enum)

### LINKED ISSUE #3049: [RUST] TypeMismatch in compatibility mode when deserializing vector of generic type wrapped inside a struct
### Search before asking

### Version

Tested on arch linux  kernel 6.17 with rust stable-x86_64-unknown-linux-gnu rustc 1.91.1

Cargo.toml
```
[package]
name = "fory-tests"
version = "0.1.0"
edition = "2024"

[dependencies]
fory = "0.13.2"
fory-core = "0.13.2"
fory-derive = "0.13.2"
```

### Component(s)

Rust

### Minimal reproduce step

```rust
use std::marker::PhantomData;

use fory::{
    Error, Fory, ForyDefault, ForyObject, ReadContext, Serializer, TypeResolver, WriteContext,
};

#[derive(Debug, PartialEq)]
struct Wrapper<T> {
    value: String,
    _marker: PhantomData<T>,
}

impl<T: 'static> Serializer for Wrapper<T> {
    fn fory_write_data(&self, context: &mut WriteContext) -> Result<(), Error> {
        context.writer.write_varuint32(self.value.len() as u32);
        context.writer.write_utf8_string(&self.value);
        Ok(())
    }

    fn fory_read_data(context: &mut ReadContext) -> Result<Self, Error> {
        let len = context.reader.read_varuint32()? as usize;
        let value = context.reader.read_utf8_string(len)?;
        Ok(Self {
            value,
            _marker: PhantomData,
        })
    }

    fn fory_type_id_dyn(&self, type_resolver: &TypeResolver) -> Result<u32, Error> {
        Self::fory_get_type_id(type_resolver)
    }

    fn as_any(&self) -> &dyn std::any::Any {
        self
    }
}

impl<T> ForyDefault for Wrapper<T> {
    fn fory_default() -> Self {
        Self {
            value: "".into(),
            _marker: PhantomData,
        }
    }
}

struct MyStruct {
    my_vec: Vec<Wrapper<Another>>,
}

struct Another {}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    // let mut fory = Fory::default(); // Without compatible it works fine.
    // let mut fory = Fory::default().xlang(true); // Works fine with xlang enabled

    fory.register::<MyStruct>(1)?;
    fory.register::<Another>(2)?;
    fory.register_serializer::<Wrapper<Another>>(3)?;

    let w1 = Wrapper::<Another> {
        value: "Value1".into(),
        _marker: PhantomData,
    };
    let w2 = Wrapper::<Another> {
        value: "Value2".into(),
        _marker: PhantomData,
    };

    let ms = MyStruct {
        my_vec: vec![w1, w2],
    };

    let bytes = fory.serialize(&ms)?;
    fory.deserialize::<MyStruct>(&bytes)?;
    Ok(())
}

```

Expected fory to deserialize data properly.

### What did you see instead?

Error: TypeMismatch(787, 6).

### Anything Else?

I’m not entirely sure whether I’m handling the implementation of a generic serializer for a custom type correctly. I can see that the problem occurs when `compatible` is enabled. The code works just fine without `compatible` or when `xlang` is enabled. I don’t have any issues serializing just the vector of Wrapper; the problem arises when this vector is wrapped inside another struct, as shown in the example code.

urlyy:
## Why?
Issue #3049 reveals a problem. For example, when serializing the field `my_vec: Vec<Wrapper<Inner>>`, the current Fory-Rust writes
```
(id: Vec::type_id, generics: [
        ( id: Wrapper<Inner>::type_id, generics: [] )
])
```
That is, the concrete generic type parameters on structs/enums are not written into the `FieldType` metadata. This results in inconsistent metadata between the sender and the receiver, and causes `assign_field_ids()` to assign `remote_field.field_id = -1`, leading to the issue.

## What does this PR do?
When obtaining the field type on the sender side, for field_type `Vec<Wrapper<Inner>>`, the `FieldType` objects for `Vec` and `Wrapper<Inner>` have already been created, so there is no need to further create a `FieldType` object for `Inner`. Therefore, further recursion over generic types on structs/enums is truncated in the runtime.

Meanwhile, because different languages provide different levels of support for generic types on classes, serializing structs/enums with generics is `not supported in xlang mode`. A `runtime check` has been added for this.

## Related issues
Fixes #3049

### DIFF
--- a/rust/fory-derive/src/object/util.rs
+++ b/rust/fory-derive/src/object/util.rs
@@ -539,7 +539,7 @@ pub(super) fn generic_tree_to_tokens(node: &TypeNode) -> TokenStream {
                 if type_resolver.is_xlang() && generics.len() > 0 {
                     return Err(fory_core::error::Error::unsupported("serialization of generic structs and enums is not supported in xlang mode"));
                 } else {
-                    generics.clear();
+                    generics = vec![];
                 }
             }
             fory_core::meta::FieldType::new(type_id, #nullable, generics)


chaokunyang (approved):
LGTM

PR #3060, opened 2025-12-17
fix(Rust): Move the calculating of TypeMeta::bytes and TypeMeta::hash ahead of serialization

### LINKED ISSUE #3058: [Rust] meta_hash not stored in TypeMeta::hash, fast path unused
### Search before asking

### Version

Rust Latest

### Component(s)

Rust

### Minimal reproduce step

This is a unittest will pass.
```rust
use fory_core::Fory;
use fory_derive::ForyObject;

#[test]
fn test_main() {
    #[fory(debug)]
    struct Demo {
        f1: i32,
    }

    fory.register::<Demo>(1).unwrap();

    let expected_demo = Demo {f1: 1};
    let bf = fory.serialize(&expected_demo).unwrap();
    let actual_demo = fory.deserialize::<Demo>(&bf).unwrap();
    assert_eq!(expected_demo, actual_demo);
}
```
And add some println for `fory_read_compatible`:

https://github.com/apache/fory/blob/09852a5c39e90eb0786a481de683fbc004018762/rust/fory-derive/src/object/read.rs#L643-L660

```rust
let local_type_hash = meta.get_hash();
let remote_type_hash = type_info.get_type_meta().get_hash();
println!("local: {local_type_hash} remote: {remote_type_hash}");
if remote_type_hash == local_type_hash {
    println!("enter here");
    <Self as fory_core::Serializer>::fory_read_data(context)
} else {
......
```

and run the unit test. Althouth it passes, the output is only
```
local: 0 remote: 170066286042854
```
which means the local type_meta.hash is not updated after `TypeMeta::new`.

Actually, when computing meta_hash, it was written only to type_def and not stored in TypeMeta::hash, causing some fast paths not to be used.

In my demo upside, is expected to print `enter here`

### What did you see instead?

not print `enter here`

### Anything Else?

_No response_

urlyy:
## What does this PR do?
Move the calculating of TypeMeta::bytes and TypeMeta::hash ahead of serialization.
The TypeMeta object will save and hold its own bytes and hash when it is created.

## Related issues
Fixes #3058

No

### DIFF
--- a/rust/fory-core/src/meta/type_meta.rs
+++ b/rust/fory-core/src/meta/type_meta.rs
@@ -378,7 +378,7 @@ impl TypeMeta {
     }
 
     #[inline(always)]
-    pub fn get_bytes(&self) -> &Vec<u8> {
+    pub fn get_bytes(&self) -> &[u8] {
         &self.bytes
     }
 
--- a/rust/fory-core/src/resolver/type_resolver.rs
+++ b/rust/fory-core/src/resolver/type_resolver.rs
@@ -180,7 +180,7 @@ impl TypeInfo {
         let namespace = type_meta.get_namespace();
         let type_name = type_meta.get_type_name();
         let register_by_name = !namespace.original.is_empty() || !type_name.original.is_empty();
-        let type_def_bytes = type_meta.get_bytes().clone();
+        let type_def_bytes = type_meta.get_bytes().to_owned();
         Ok(TypeInfo {
             type_def: Rc::from(type_def_bytes),
             type_meta,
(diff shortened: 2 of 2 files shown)


rust/fory-core/src/meta/type_meta.rs
```diff
@@ -372,15 +378,25 @@ impl TypeMeta {
     }
 
     #[inline(always)]
-    pub fn empty() -> TypeMeta {
-        TypeMeta {
+    pub fn get_bytes(&self) -> Vec<u8> {
```
chaokunyang: could we return borrowed ref and caller clone it instead? this will be more explicit

PR #3062, opened 2025-12-18
feat(java): implement Union type support for cross-language serialization

### LINKED ISSUE #3030: Implement union type in java

zhan7236:
## Why?

Java lacks native support for Union/Variant types that are common in other languages (C++ std::variant, Rust enum, Python typing.Union). This limits cross-language serialization interoperability when working with union types.

## What does this PR do?

Implements Union type support in Java for cross-language serialization, addressing issue #3030.

**Changes:**

1. **Types.java**: Added `UNION` (38) and `NONE` (39) type constants following the xlang specification
2. **Union.java**: A tagged union wrapper class that can hold one of several alternative types
3. **UnionSerializer.java**: Handles serialization/deserialization with xlang protocol support
4. **ClassResolver.java**: Registered UnionSerializer as default serializer
5. **XtypeResolver.java**: Registered Union type with UNION type id for xlang serialization
6. **xlang_type_mapping.md**: Updated type mapping documentation

**Serialization format:**
- Variant index (varuint32): identifies which alternative is active
- Type info: type information for the active alternative
- Value data: serialized value

## Related issues

Closes #3030

Yes. Users can now use the `Union` class for cross-language union type serialization.

## Benchmark

Not applicable - this adds new functionality without modifying existing serialization paths.

### DIFF
--- a/java/fory-core/src/test/java/org/apache/fory/serializer/UnionSerializerTest.java
+++ b/java/fory-core/src/test/java/org/apache/fory/serializer/UnionSerializerTest.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import org.apache.fory.Fory;
 import org.apache.fory.ForyTestBase;
+import org.apache.fory.config.CompatibleMode;
 import org.apache.fory.config.Language;
 import org.apache.fory.type.union.Union;
 import org.apache.fory.type.union.Union2;
@@ -38,6 +39,7 @@ import org.apache.fory.type.union.Union3;
 import org.apache.fory.type.union.Union4;
 import org.apache.fory.type.union.Union5;
 import org.apache.fory.type.union.Union6;
+import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 /**
@@ -47,8 +49,18 @@ import org.testng.annotations.Test;
  */
 public class UnionSerializerTest extends ForyTestBase {
 
-  private Fory createXlangFory() {
-    Fory fory = Fory.builder().withLanguage(Language.XLANG).requireClassRegistration(true).build();
+  @DataProvider(name = "compatibleMode")
+  public static Object[][] compatibleModeProvider() {
+    return new Object[][] {{CompatibleMode.COMPATIBLE}, {CompatibleMode.SCHEMA_CONSISTENT}};
+  }
+
+  private Fory createXlangFory(CompatibleMode mode) {
+    Fory fory =
+        Fory.builder()
+            .withLanguage(Language.XLANG)
+            .withCompatibleMode(mode)
+            .requireClassRegistration(true)
+            .build();
     fory.register(StructWithUnion.class);
     fory.register(StructWithUnion2.class);
     fory.register(StructWithUnion3.class);
@@ -119,9 +131,9 @@ public class UnionSerializerTest extends ForyTestBase {
     }
   }
 
-  @Test
-  public void testUnionBasicTypes() {
-    Fory fory = createXlangFory();
+  @Test(dataProvider = "compatibleMode")
+  public void testUnionBasicTypes(CompatibleMode mode) {
+    Fory fory = createXlangFory(mode);
 
     // Test with Integer value via struct
     StructWithUnion struct = new StructWithUnion(new Union(0, 42));
@@ -138,9 +150,9 @@ public class UnionSerializerTest extends ForyTestBase {
     assertEquals(deserialized.union.getIndex(), 1);
   }
 
-  @Test
-  public void testUnion2Serialization() {
-    Fory fory = createXlangFory();
+  @Test(dataProvider = "compatibleMode")
+  public void testUnion2Serialization(CompatibleMode mode) {
+    Fory fory = createXlangFory(mode);
 
     // Test with T1 (String) via struct field
     StructWithUnion2 struct = new StructWithUnion2(Union2.ofT1("hello"));
@@ -160,9 +172,9 @@ public class UnionSerializerTest extends ForyTestBase {
     assertTrue(deserialized.union.isT2());
   }
 
-  @Test
-  public void testUnion3Serialization() {
-    Fory fory = createXlangFory();
+  @Test(dataProvider = "compatibleMode")
+  public void testUnion3Serialization(CompatibleMode mode) {
+    Fory fory = createXlangFory(mode);
 
     // Test with T1
     StructWithUnion3 struct = new StructWithUnion3(Union3.ofT1(42));
@@ -190,9 +202,9 @@ public class UnionSerializerTest extends ForyTestBase {
     assertTrue(deserialized.union.isT3());
   }
 
-  @Test
-  public void testUnion4Serialization() {
-    Fory fory = createXlangFory();
+  @Test(dataProvider = "compatibleMode")
+  public void testUnion4Serialization(CompatibleMode mode) {
+    Fory fory = createXlangFory(mode);
 
     // Test with T1
     StructWithUnion4 struct = new StructWithUnion4(Union4.ofT1(42));
@@ -212,9 +224,9 @@ public class UnionSerializerTest extends ForyTestBase {
     assertTrue(deserialized.union.isT4());
   }
 
-  @Test
-  public void testUnion5Serialization() {
-    Fory fory = createXlangFory();
+  @Test(dataProvider = "compatibleMode")
+  public void testUnion5Serialization(CompatibleMode mode) {
+    Fory fory = createXlangFory(mode);
 
     // Test with T1
     StructWithUnion5 struct = new StructWithUnion5(Union5.ofT1(42));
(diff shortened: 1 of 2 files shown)


java/fory-core/src/main/java/org/apache/fory/type/Union.java
```diff
@@ -0,0 +1,277 @@
...
+ * union.setValue(42);
+ *
+ * // Get the value
+ * Integer value = union.getValue(Integer.class);
+ *
+ * // Check the active type
+ * int index = union.getActiveIndex(); // returns 0
+ * }</pre>
+ */
+public class Union {
```
chaokunyang: Other parts look good to me, but I'm not sure whether we could have a better acstraction for union.

For Tuple, scala have `Tuple2/3/4/5/6/...`, maybe we could have a generic union such as `Union2<T1, T2>/Union3<T1,T2,T3>`.

And we don't need `alternativeTypes` types here. User can just use `Union` like this:
```java
Union2<String, Long> union = Union2.ofT1("abc");
Union2<String, Long> union = Union2.ofT2(100L);
```

Although javac will arease the geenrics, but we don't write this type info into data, so it's ok.

chaokunyang: One cons I can think of is that when deserializing Union, if it's not a field of a struct, we can't know whtich union type to deserialize into. We could use Union2/Union3/Union4, but which one we should use?

One method I can think about is we also create an `Union` type, which just hold a value and an index. And when we don't know which type to deserialize into, we use this  untyped general`Union` . It's kind of similair ot java Map without generics

chaokunyang: We may need create an `union` package for this

java/fory-core/src/main/java/org/apache/fory/type/union/Union4.java
```diff
@@ -0,0 +1,266 @@
...
+ *     String value = union.getT2();
+ * }
+ * }</pre>
+ *
+ * @param <T1> the first alternative type
+ * @param <T2> the second alternative type
+ * @param <T3> the third alternative type
+ * @param <T4> the fourth alternative type
+ */
+public class Union4<T1, T2, T3, T4> {
```
chaokunyang: could we also add union5 and 6.  I think six can cover most cases. If not user can use geenric Union instead. Please also document this in the javadoc.

java/fory-core/src/main/java/org/apache/fory/type/union/Union2.java
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * @param <T1> the first alternative type
+ * @param <T2> the second alternative type
+ * @see Union
+ * @see Union3
+ * @see Union4
+ * @see Union5
+ * @see Union6
+ */
+public class Union2<T1, T2> {
```
chaokunyang: how about let `Union2/3/4/5/6` extends `Union`, and make UnionSerialzier can handle allv Union types?

And could we also add tests for `Union2/3/4/5/6`

java/fory-core/src/main/java/org/apache/fory/resolver/ClassResolver.java
```diff
@@ -329,7 +329,13 @@ private void addDefaultSerializers() {
     OptionalSerializers.registerDefaultSerializers(fory);
     CollectionSerializers.registerDefaultSerializers(fory);
     MapSerializers.registerDefaultSerializers(fory);
-    addDefaultSerializer(org.apache.fory.type.union.Union.class, new UnionSerializer(fory));
+    UnionSerializer unionSerializer = new UnionSerializer(fory);
+    addDefaultSerializer(org.apache.fory.type.union.Union.class, unionSerializer);
```
chaokunyang: Union is only used for xlang serialization, we should only register it in `XtypeResolver`

java/fory-core/src/main/java/org/apache/fory/serializer/UnionSerializer.java
```diff
@@ -0,0 +1,126 @@
...
+ * <p>Note: When deserializing, this serializer always returns the base {@link Union} class. If you
+ * need a specific typed union (e.g., {@link Union2}), you can use {@link Union2#of(int, Object)} to
+ * convert.
+ */
+public class UnionSerializer extends Serializer<Union> {
+  private final ClassResolver classResolver;
+  private final ClassInfoHolder classInfoHolder;
+
+  public UnionSerializer(Fory fory) {
+    super(fory, Union.class);
```
chaokunyang: I don't think this serializer can handle the serialization of `Union2~6`. Please update it to make it support the serialization of `Union2~6` too. You can not sepcify `Union.class` here, you shoud accept  `Class` as a parameter

java/fory-core/src/main/java/org/apache/fory/serializer/UnionSerializer.java
```diff
@@ -0,0 +1,126 @@
...
+    int index = buffer.readVarUint32();
+
+    int refId = fory.getRefResolver().tryPreserveRefId(buffer);
+    if (refId >= Fory.NOT_NULL_VALUE_FLAG) {
+      ClassInfo classInfo = classResolver.readClassInfo(buffer, classInfoHolder);
+      Object value = classInfo.getSerializer().read(buffer);
+      fory.getRefResolver().setReadObject(refId, value);
+      return new Union(index, value);
+    } else if (refId == Fory.NULL_FLAG) {
+      return new Union(index, null);
```
chaokunyang: This is also wrong. If `class` is `Union2~6`, we should create coresponding objects. You can create a factory method in contructor of `UnionSerializer` for that. For example, create a BiConsumer field.

zhan7236:
@chaokunyang Thanks for your patience with these iterations! I've updated the code based on your latest comments. Hopefully, everything is good to go now.

java/fory-core/src/main/java/org/apache/fory/type/union/Union2.java
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * @param <T1> the first alternative type
+ * @param <T2> the second alternative type
+ * @see Union
+ * @see Union3
+ * @see Union4
+ * @see Union5
+ * @see Union6
+ */
+public class Union2<T1, T2> extends Union {
```
chaokunyang: could we mark `Union2~6` as final class?

java/fory-core/src/main/java/org/apache/fory/type/union/Union2.java
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Gets the value as the first type T1.
+   *
+   * @return the value cast to T1
+   * @throws ClassCastException if the active type is not T1
+   */
+  @SuppressWarnings("unchecked")
+  public T1 getT1() {
+    return (T1) value;
```
chaokunyang, line 122: could we add an `assert` that index is `0`? same for other methods and all unions

java/fory-core/src/main/java/org/apache/fory/serializer/UnionSerializer.java
```diff
@@ -0,0 +1,127 @@
...
+  private final BiFunction<Integer, Object, Union> factory;
+
+  @SuppressWarnings("unchecked")
+  public UnionSerializer(Fory fory, Class<? extends Union> cls) {
+    super(fory, (Class<Union>) cls);
+    this.typeTag = getTypeTag(cls);
+    this.factory = FACTORIES[typeTag];
+  }
+
+  private static int getTypeTag(Class<? extends Union> cls) {
```
chaokunyang:
```suggestion
  private static int getTypeIndex(Class<? extends Union> cls) {
```

java/fory-core/src/main/java/org/apache/fory/serializer/UnionSerializer.java
```diff
@@ -0,0 +1,127 @@
...
+    } else {
+      // Default to base Union for unknown subclasses
+      return 0;
+    }
+  }
+
+  @Override
+  public void xwrite(MemoryBuffer buffer, Union union) {
+    // Write type tag to identify the Union subclass
+    buffer.writeByte(getTypeTag(union.getClass()));
```
chaokunyang: `Union` type should always write `UNION` as type info. In java, that's handled outside. For `xread`, we never write any `type info`, we only write object value 

For deserialization, we should use declared class into to determine which serialized should be used for deserialization.

For example:

```java
class Foo {
  Union2<String, Long> u1;
}
```

When deserializing `u1`, we know we need to create `Union2` , so when creating serializer for `Union2`, we create serializer with `Union2` passed. and in `xread`, we will create `Union2` directly

java/fory-core/src/test/java/org/apache/fory/serializer/UnionSerializerTest.java
```diff
@@ -0,0 +1,418 @@
...
+    public Union6<Integer, String, Double, Boolean, Long, Float> union;
+
+    public StructWithUnion6() {}
+
+    public StructWithUnion6(Union6<Integer, String, Double, Boolean, Long, Float> union) {
+      this.union = union;
+    }
+  }
+
+  @Test
```
chaokunyang: Could we make all those tests as parameterized terts, with `CompatibleMode` set `Compatible/SCHEMA_CONSISTENT`.

And we also need to update `ClassDef.FieldType#toTypeToken` to return correct union `Type`

chaokunyang:
@zhan7236 would you like to open a new pull request to add two xlang tests for union in `XlangTestBase.java`:

- one take union1~6 as fields of struct
- another serialzie top-level union objects

Since rust and go has not support this, you can skip rust/go implementations in subclass by throw SkipException

zhan7236:
@chaokunyang Sure, I'd be happy to pick this up. I'll open a new PR for the union tests and skip Rust/Go as mentioned.

PR #3073, opened 2025-12-22
feat(rust): add skip_default for skip manual impl'ed Default to avoid conflict

ariesdevil:
## Why?

When using `#[derive(ForyObject)]` on a struct that already has a manual `impl Default`, a compilation error occurs due to conflicting `Default` implementations. The `ForyObject` derive macro automatically generates `impl Default`, which conflicts with user-provided implementations.

This is a common scenario when integrating Fory with existing codebases (e.g., [OpenRaft](https://github.com/databendlabs/openraft)) where types already have custom `Default` implementations with specific default values.

## What does this PR do?

Adds a `#[fory(skip_default)]` attribute to the `ForyObject` derive macro that:

1. Skips generating `impl std::default::Default` for the annotated type
2. Only generates `impl ForyDefault` that delegates to the existing `Default::default()`

### Usage

```rust
#[derive(ForyObject)]
#[fory(skip_default)]  // Don't generate impl Default
struct BasicNode {
    addr: String,
}

// User's own Default implementation is preserved
impl Default for BasicNode {
    fn default() -> Self {
        Self { addr: "localhost".to_string() }
    }
}
```

### Why not detect manual `impl Default` automatically?

Rust's proc_macro system only has access to the token stream of the type being derived. It cannot see other `impl` blocks in the same file or elsewhere. The existing `has_existing_default()` function can only detect `#[derive(Default)]`, not manual `impl Default` blocks. This is a fundamental limitation of proc_macros, not a code issue.

## Related issues

None

  - Adds new optional attribute `#[fory(skip_default)]` to the `ForyObject` derive macro

## Benchmark

This change only affects code generation at compile time and has no runtime performance impact.


chaokunyang:
@ariesdevil How about fliping the default behavior.

I originally added the auto `impl Default` just as a convenience to reduce duplicated code, since in most cases `ForyDefault` is identical to `Default`. But if this “optimization” creates friction for users—especially when integrating with existing codebases that already have a manual `Default`—it’s probably better to be conservative.

So I’m fine with changing `ForyObject` to **not generate `impl Default` by default**, and instead gate it behind an explicit attribute (e.g. `#[fory(generate_default)]`). That keeps the derive non-invasive, avoids conflicts, and users who want the convenience can opt in.

ariesdevil:
Agreed, let me flip this feature: )

ariesdevil:
But this will cause a break change of the API (modify the default behavior), if you confirm this change, I will start working on it immediately.

chaokunyang:
@ariesdevil We may also need to update user docs under `docs/guide/rust` and rust README.md and lib.rs. Some docs use `default` methods generated by `ForyObject`

ariesdevil:
@chaokunyang OK, I will edit these docs together.

ariesdevil:
@chaokunyang Hi, I’ve opened [a separate PR](https://github.com/apache/fory/pull/3074) so you can choose which one to merge.

PR #3074, opened 2025-12-22
feat(rust): add generate_default attr, no longer generate `Default` trait impl by default

ariesdevil:
## Why?

When using `#[derive(ForyObject)]` on a type that already has a manual `impl Default`, a compilation error occurs due to conflicting `Default` implementations. The previous behavior of auto-generating `Default` was problematic for integrating with existing codebases where types have custom `Default` implementations.

After discussion with reviewers, we decided to **invert the default behavior**: `ForyObject` should NOT generate `impl Default` by default, and users who want it can opt-in with `#[fory(generate_default)]`.

### Breaking Change

`#[derive(ForyObject)]` **no longer generates `impl Default`** by default.

Scenario | ForyDefault | Default
-- | -- | --
Default (no attribute) | ✅ Generated (standalone impl) | ❌ Not generated
#[fory(generate_default)] | ✅ Generated | ✅ Generated (delegates to ForyDefault)
User has existing Default | ✅ Generated (delegates to Default) | User's own impl

### Migration Guide

```rust
// Before this PR (v0.14.0)
#[derive(ForyObject)]
struct MyStruct { ... }  // Auto-generates Default

// After this PR (v0.15.0) - Option 1: Add derive(Default) explicitly
#[derive(ForyObject, Default)]
struct MyStruct { ... }

// After this PR (v0.15.0) - Option 2: Use generate_default attribute
#[derive(ForyObject)]
#[fory(generate_default)]
struct MyStruct { ... }
```

### Changes

1. Added `#[fory(generate_default)]` attribute to opt-in for `Default` generation
2. Changed all internal code generation to use `ForyDefault::fory_default()` instead of `Default::default()`
3. Updated `register_trait_type!` macro to use `ForyDefault` instead of `Default`
4. Updated documentation and tests
5. Used workspace dependencies for internal crate versions

## Related issues

None

  - **Breaking**: `#[derive(ForyObject)]` no longer generates `impl Default`
  - **New**: `#[fory(generate_default)]` attribute to opt-in for `Default` generation

## Benchmark

This change only affects compile-time code generation. No runtime performance impact.


rust/README.md
```diff
@@ -26,13 +26,60 @@ The Rust implementation provides versatile and high-performance serialization wi
 | [`fory-core`](https://github.com/apache/fory/blob/main/rust/fory-core/)     | Core serialization engine         | [![crates.io](https://img.shields.io/crates/v/fory-core.svg)](https://crates.io/crates/fory-core)     |
 | [`fory-derive`](https://github.com/apache/fory/blob/main/rust/fory-derive/) | Procedural macros                 | [![crates.io](https://img.shields.io/crates/v/fory-derive.svg)](https://crates.io/crates/fory-derive) |
 
+## Breaking Change in v0.15.0
```
chaokunyang: I prefer not including this, we could put this into github pr description, and add it to next release note.

We could document at `#[derive(ForyObject)]` section that `ForyObject` will genetate `fory_default()` methdo, users can use `generate_default` to generate `default methods` too

docs/guide/rust/custom-serializers.md
```diff
@@ -63,13 +63,17 @@ impl Serializer for CustomType {
...
 
+// ForyDefault delegates to Default
 impl ForyDefault for CustomType {
     fn fory_default() -> Self {
         Self::default()
     }
 }
 ```
 
+> **Note**: When implementing `ForyDefault` manually, ensure your type also implements `Default` if you use `Self::default()`.
```
chaokunyang, line 74: Could we also document that users can make `ForyObject` to generate `default`too

ariesdevil:
@chaokunyang Fixed, PTAL again.

ariesdevil:
I also upgraded the crates version since that's a breaking change.

PR #3075, opened 2025-12-22
chore(rust): revert rust crates version

ariesdevil:
I didn't realize that the version is shared across the entire project, not a separate version for each language.

I'm very sorry about that.


ariesdevil:
@chaokunyang Sorry for this...

PR #3076, opened 2025-12-23
feat(java): add xlang tests for Union types

zhan7236:
## What does this PR do?

This PR adds cross-language (xlang) tests for Union types in XlangTestBase.java as requested in #3062.

### Changes:
- Add \`StructWithUnionFields\` class with Union through Union6 fields for testing Union types as struct fields
- Add \`testStructWithUnionFields()\` test method to verify serialization/deserialization of structs containing Union fields
- Add \`testTopLevelUnion()\` test method to verify serialization/deserialization of top-level Union objects
- Skip Union tests in \`GoXlangTest\` and \`RustXlangTest\` with \`SkipException\` (Union types not yet supported in Go/Rust)

### Test Details:
1. **testStructWithUnionFields**: Creates a struct with Union1-6 fields containing different value types (Integer, String, Double, Boolean, Long, Float), serializes and deserializes it
2. **testTopLevelUnion**: Tests serializing Union objects directly (not as struct fields), including both int and string variants

## Related issues

Follow-up to #3062 (Java Union type support)
Relates to #3030 (Union type support tracking issue)

No, this PR only adds tests.

## Benchmark

N/A - test-only changes

### DIFF
--- a/java/fory-core/src/test/java/org/apache/fory/XlangTestBase.java
+++ b/java/fory-core/src/test/java/org/apache/fory/XlangTestBase.java
@@ -39,6 +39,12 @@ import org.apache.fory.memory.MemoryBuffer;
 import org.apache.fory.memory.MemoryUtils;
 import org.apache.fory.serializer.Serializer;
 import org.apache.fory.test.TestUtils;
+import org.apache.fory.type.union.Union;
+import org.apache.fory.type.union.Union2;
+import org.apache.fory.type.union.Union3;
+import org.apache.fory.type.union.Union4;
+import org.apache.fory.type.union.Union5;
+import org.apache.fory.type.union.Union6;
 import org.apache.fory.util.MurmurHash3;
 import org.testng.Assert;
 import org.testng.annotations.BeforeClass;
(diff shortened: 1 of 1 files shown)


java/fory-core/src/test/java/org/apache/fory/XlangTestBase.java
```diff
@@ -1575,6 +1575,109 @@ public void testEnumSchemaEvolutionCompatible() throws java.io.IOException {
...
   }
 
+  // ============================================================================
+  // Union type tests
+  // ============================================================================
+
+  /** Struct containing Union fields of different arities (Union through Union6). */
+  @Data
+  public static class StructWithUnionFields {
+    public org.apache.fory.type.union.Union union;
```
chaokunyang: Please add imports for those types instead of using qualified class

zhan7236: Thanks.

java/fory-core/src/test/java/org/apache/fory/CPPXlangTest.java
```diff
@@ -130,4 +130,16 @@ private void runBazelCommand(String... args) throws IOException {
   public void testConsistentNamed() throws java.io.IOException {
     super.testConsistentNamed();
   }
+
+  @Test
+  @Override
+  public void testStructWithUnionFields() throws java.io.IOException {
+    throw new SkipException("Union types not yet supported in C++");
```
chaokunyang, line 137: Could you add cpp tests in `cpp/fory/serialization/xlang_test_main.cc`?

Only added cpp tests, and make it pass, we can ensure union xlang serialization is OK

PR #3079, opened 2025-12-23
fix(java): Fix CopyOnWriteArrayList field serialization

vybhavjs:
## Why?

Fory doesn't currently support serialization of any object that has a `CopyOnWriteArrayList` field with codegen, because the generated code attempts to cast `CollectionSnapshot` to `List`, throwing the following exception :

```
org.apache.fory.exception.SerializationException: java.lang.ClassCastException: 
class org.apache.fory.collection.CollectionSnapshot cannot be cast to class java.util.List 
(org.apache.fory.collection.CollectionSnapshot is in unnamed module of loader 'app'; 
java.util.List is in module java.base of loader 'bootstrap')
( . . . )
org.apache.fory.serializer.collection.CollectionSerializersTest.testCopyOnWriteArrayListNested(CollectionSerializersTest.java:407)
Caused by: java.lang.ClassCastException: class org.apache.fory.collection.CollectionSnapshot cannot be cast to class java.util.List 
(org.apache.fory.collection.CollectionSnapshot is in unnamed module of loader 'app'; 
java.util.List is in module java.base of loader 'bootstrap')
	at org.apache.fory.serializer.collection.CollectionSerializersTest_NestedCopyOnWriteArrayListForyCodec_0.writeFields$(CollectionSerializersTest_NestedCopyOnWriteArrayListForyCodec_0.java:63)
	at org.apache.fory.serializer.collection.CollectionSerializersTest_NestedCopyOnWriteArrayListForyCodec_0.write(CollectionSerializersTest_NestedCopyOnWriteArrayListForyCodec_0.java:130)
```

This is the relevant code from the generated codec that attempts the invalid class cast.

```
CollectionLikeSerializer collectionLikeSerializer = this.writeCollectionClassInfo(list0, memoryBuffer1);
if (collectionLikeSerializer.supportCodegenHook()) {
    java.util.List list1 = (java.util.List)collectionLikeSerializer.onCollectionWrite(memoryBuffer1, list0);
```

## What does this PR do?

* feat: Update `CollectionSnapshot` to extend `AbstractList` instead of `AbstractCollection`, and implement `get(index)` method
* test: Add test in `CollectionSerializersTest` for serializing objects with CopyOnWriteArrayList fields
* test: Add test in `CollectionSnapshotTest` for `CollectionSnapshot.get(index)`

## Related issues

Fixes pending issues from #2918.

## Benchmark


java/fory-core/src/main/java/org/apache/fory/collection/ListSnapshot.java
```diff
@@ -0,0 +1,152 @@
...
+ * will result in incorrect iteration behavior.
+ *
+ * <p>This class is marked as {@code @Internal} and should not be used directly by application code.
+ * It's specifically designed for internal serialization purposes.
+ *
+ * @param <E> the type of elements maintained by this list
+ * @since 1.0
+ */
+@Internal
+public class ListSnapshot<E> extends AbstractList<E> {
```
chaokunyang: Could we make CollectionSnapshot extends  `AbstractList`?

vybhavjs: That was my initial thought as most of the code is duplicated, and implementing `get(index)` is possible even with `CollectionSnapshot`. It seemed a bit odd semantically as now, non-list classes like `CopyOnWriteArraySet`, `ConcurrentSkipListSet` etc are being wrapped in a List implementation, and I wasn't sure if that would break something at runtime.

All tests pass with that approach though - so I've made the change.

PR #3080, opened 2025-12-23
feat(rust): add i128 and isize type support

ariesdevil:
## Why?

To support serialization of `i128` and `isize` types in Rust, which are needed for cross-language interoperability and broader type coverage.

## What does this PR do?

This PR adds support for `i128` and `isize` types:

1. **Buffer operations**: Added `write_i128`, `write_isize`, `read_i128`, `read_isize` methods to `buffer.rs`
2. **Type IDs**: Added `INT128` (78) and `ISIZE` (79) type identifiers to `types.rs`
3. **Serializer implementations**: Added `Serializer` and `ForyDefault` implementations for `i128` and `isize` in `number.rs`
4. Make `isize` and `usize` aware of different arches.

## Related issues

None

  - Yes, adds new public types support: `i128` and `isize` can now be serialized/deserialized
  - No, this adds new type IDs (78, 79) without changing existing protocol

### DIFF
--- a/rust/fory-core/src/buffer.rs
+++ b/rust/fory-core/src/buffer.rs
@@ -176,7 +176,13 @@ impl<'a> Writer<'a> {
 
     #[inline(always)]
     pub fn write_usize(&mut self, value: usize) {
-        self.write_u64(value as u64);
+        const SIZE: usize = std::mem::size_of::<usize>();
+        match SIZE {
+            2 => self.write_u16(value as u16),
+            4 => self.write_varuint32(value as u32),
+            8 => self.write_varuint64(value as u64),
+            _ => unreachable!("unsupported usize size"),
+        }
     }
 
     #[inline(always)]
@@ -189,6 +195,22 @@ impl<'a> Writer<'a> {
         #[cfg(target_endian = "big")]
         {
             self.bf.extend_from_slice(&value.to_le_bytes());
+        }
+    }
+
+    #[inline(always)]
+    pub fn write_i128(&mut self, value: i128) {
+        self.write_u128(value as u128);
+    }
+
+    #[inline(always)]
+    pub fn write_isize(&mut self, value: isize) {
+        const SIZE: usize = std::mem::size_of::<isize>();
+        match SIZE {
+            2 => self.write_i16(value as i16),
+            4 => self.write_varint32(value as i32),
+            8 => self.write_varint64(value as i64),
+            _ => unreachable!("unsupported isize size"),
         }
     }
 
@@ -533,7 +555,13 @@ impl<'a> Reader<'a> {
 
     #[inline(always)]
     pub fn read_usize(&mut self) -> Result<usize, Error> {
-        Ok(self.read_u64()? as usize)
+        const SIZE: usize = std::mem::size_of::<usize>();
+        match SIZE {
+            2 => Ok(self.read_u16()? as usize),
+            4 => Ok(self.read_varuint32()? as usize),
+            8 => Ok(self.read_varuint64()? as usize),
+            _ => unreachable!("unsupported usize size"),
+        }
     }
 
     #[inline(always)]
@@ -542,6 +570,22 @@ impl<'a> Reader<'a> {
         let result = LittleEndian::read_u128(slice);
         self.cursor += 16;
         Ok(result)
+    }
+
+    #[inline(always)]
+    pub fn read_i128(&mut self) -> Result<i128, Error> {
+        Ok(self.read_u128()? as i128)
+    }
+
+    #[inline(always)]
+    pub fn read_isize(&mut self) -> Result<isize, Error> {
+        const SIZE: usize = std::mem::size_of::<isize>();
+        match SIZE {
+            2 => Ok(self.read_i16()? as isize),
+            4 => Ok(self.read_varint32()? as isize),
+            8 => Ok(self.read_varint64()? as isize),
+            _ => unreachable!("unsupported isize size"),
+        }
     }
 
     #[inline(always)]
--- a/rust/fory-core/src/serializer/list.rs
+++ b/rust/fory-core/src/serializer/list.rs
@@ -47,8 +47,10 @@ pub(super) fn get_primitive_type_id<T: Serializer>() -> TypeId {
         TypeId::U16 => TypeId::U16_ARRAY,
         TypeId::U32 => TypeId::U32_ARRAY,
         TypeId::U64 => TypeId::U64_ARRAY,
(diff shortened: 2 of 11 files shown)


rust/fory-core/src/buffer.rs
```diff
@@ -192,6 +192,16 @@ impl<'a> Writer<'a> {
...
     }
 
+    #[inline(always)]
+    pub fn write_i128(&mut self, value: i128) {
+        self.write_u128(value as u128);
+    }
+
+    #[inline(always)]
+    pub fn write_isize(&mut self, value: isize) {
+        self.write_i64(value as i64);
```
chaokunyang: isize has different size on different architecture, could we use const to dispatch to different codepath. 

- When it's 2 bytes, use write_u16
- When it's 4 bhtes, use write_varu32
- When it's 8 bhtes, use write_varu64

And for serializer in number.rs, we choose type id `u16/32/64` instead?

ariesdevil: I just follow the `usize` way, let me change both `usize` and `isize` for different arches.

rust/fory-core/src/types.rs
```diff
@@ -95,8 +95,10 @@ pub enum TypeId {
     U64_ARRAY = 75,
     USIZE_ARRAY = 76,
     U128_ARRAY = 77,
+    INT128 = 78,
+    ISIZE = 79,
```
chaokunyang: do we still need this type id? and coudk we also remvoe USIZE type id?

ariesdevil: done, PTAL again.

rust/fory-core/src/types.rs
```diff
@@ -85,16 +85,16 @@ pub enum TypeId {
...
-    U128 = 69,
-    VAR_U32 = 70,
-    VAR_U64 = 71,
-    SLI_U64 = 72,
-    U16_ARRAY = 73,
-    U32_ARRAY = 74,
-    U64_ARRAY = 75,
-    USIZE_ARRAY = 76,
-    U128_ARRAY = 77,
+    U128 = 68,
```
ariesdevil: safe reorder here, since the xlang spec just up to 39. https://github.com/apache/fory/blob/main/docs/specification/xlang_serialization_spec.md#type-id

rust/fory-core/src/types.rs
ariesdevil: After my investigation, I think we need these type ids, if we remove these ids:

```rust
// ser
let value: usize = 42;
let wrapped: Box<dyn Any> = Box::new(value);
let bytes = fory.serialize(&wrapped).unwrap();

// deser
let result: Box<dyn Any> = fory.deserialize(&bytes).unwrap();
let v = result.downcast_ref::<usize>().unwrap();  // <- will fail！
```

Serialization:
  value: usize = 42
  Write TypeId = U64 (because on 64-bit systems usize == u64)
  Write data = 42

Deserialization:
  Read TypeId = U64 (because we remove `usize` and `isize`)
  Fory looks up the type for U64 → finds u64
  Creates Box<dyn Any> containing a u64 value
  
Downcast:
  result.downcast_ref::<usize>()  
  -> Internal type is u64, not usize
  -> Returns None!

chaokunyang: I see, thanks for this catch. Let's keep those type id, could you add it back, and document that they are needed to polymorphic deserialization

ariesdevil: I already add these back, let me add some docs

rust/fory-core/src/types.rs
chaokunyang: Could we move `I128/ISIZE/USIZE` to here?

chaokunyang: will this be better?
```
U8 = 64,
U16 = 65,
U32 = 66,
VAR_U32,
U64,
VAR_U64,
SLI_U64,
U128,
I128,
USIZE,
ISIZE
```

rust/fory-core/src/types.rs
ariesdevil: Done

rust/fory-core/src/types.rs
ariesdevil: Done

rust/fory-core/src/serializer/list.rs
```diff
@@ -47,7 +47,6 @@ pub(super) fn get_primitive_type_id<T: Serializer>() -> TypeId {
         TypeId::U16 => TypeId::U16_ARRAY,
         TypeId::U32 => TypeId::U32_ARRAY,
         TypeId::U64 => TypeId::U64_ARRAY,
-        TypeId::USIZE => TypeId::USIZE_ARRAY,
```
chaokunyang, line 50: why we remove `USIZE` here? I thought we should add `USIZE/ISIZE` here too

ariesdevil: Fixed

rust/fory-core/src/types.rs
```diff
@@ -84,19 +84,29 @@ pub enum TypeId {
...
-    U32_ARRAY = 74,
-    U64_ARRAY = 75,
-    USIZE_ARRAY = 76,
-    U128_ARRAY = 77,
+    VAR_U32 = 67,
+    U64 = 68,
+    VAR_U64 = 69,
+    SLI_U64 = 70,
+    U128 = 71,
+    INT128 = 72,
```
chaokunyang, line 92:
```suggestion
    I128 = 72,
```

rust/fory-core/src/types.rs
```diff
@@ -84,19 +84,29 @@ pub enum TypeId {
...
+    // When deserializing `Box<dyn Any>`, we need to create the exact type.
+    // If we used U64's TypeId for usize, deserialization would create a u64 value,
+    // and `result.downcast_ref::<usize>()` would return None.
+    USIZE = 73,
+    ISIZE = 74,
+    U16_ARRAY = 75,
+    U32_ARRAY = 76,
+    U64_ARRAY = 77,
+    U128_ARRAY = 78,
+    INT128_ARRAY = 79,
```
chaokunyang, line 105:
```suggestion
    I128_ARRAY = 79,
```

rust/fory-core/src/types.rs
ariesdevil: I think we need keep `INT128` here since other signed integer prefix is `INT`

rust/fory-core/src/types.rs
ariesdevil: Same reason as above

chaokunyang (approved):
LGTM, thanks

PR #3081, opened 2025-12-24
feat(rust): add unit type and PhantomData serializer support

ariesdevil:
## Why?

To support serialization of `()` and `PhantomData` types in Rust.

## What does this PR do?

This PR adds support for `()` and `PhantomData` types:

## Related issues

None

  - Yes, adds new public types support: `()` and `PhantomData` can now be serialized/deserialized
  - No

### DIFF
--- a/rust/fory-core/src/serializer/marker.rs
+++ b/rust/fory-core/src/serializer/marker.rs
@@ -1,0 +1,74 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Serializer implementations for marker types like `PhantomData<T>`.
+
+use crate::error::Error;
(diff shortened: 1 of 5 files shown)


rust/fory-core/src/serializer/marker.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    #[inline(always)]
+    fn fory_reserved_space() -> usize {
+        0
+    }
+
+    #[inline(always)]
+    fn fory_get_type_id(_: &TypeResolver) -> Result<u32, Error> {
+        // Use UNKNOWN to skip type registry lookup - PhantomData<T> has no runtime data
+        Ok(TypeId::UNKNOWN as u32)
```
chaokunyang, line 48: We have a `NONE` type id, could we use that instead for `PhantomData` and `Unit`?

ariesdevil: Seems we don't have any `None` in the type ids.

ariesdevil: `Unknown` means no need to lookup at runtime, I think `Unknown` is enough here.

chaokunyang: I added for C++, but have not added to rust:

cpp/fory/type/type.h
```c++
  UNION = 38,
  // a null value with no data.
  NONE = 39,
```

chaokunyang: > `Unknown` means no need to lookup at runtime, I think `Unknown` is enough here.

Both will work, but it does have none semantics.

ariesdevil: I'll check the C++ version, and let me add `None` in next PR.

chaokunyang: for struct, when we skip a field of type UNKNOWN, we will keep read more data for this field. But if we usign NONE, we can update skip.rs to return early

ariesdevil: Make sense, I think we can merge this PR first, in next PR, I'll add `None` and adjust the `PhantomData` and `()` to `None`.

chaokunyang (approved):
LGTM

Issue #2982, opened 2025-12-03
Roadmap for Fory Go Serialization 1.0


chaokunyang:
### Feature Request

Add feature-complete fory go serialization implementation.

And to make Fory Go has better performance and use modern Go features, Fory Go will only support Go 1.23+ versions.

### Is your feature request related to a problem? Please describe

Most of features has been finsihed in previous commits, this issue tracking feature since commit https://github.com/apache/fory/commit/8843c7b4f2d23c709831441f0276cea1d5a555d6: 
- [x] #2995
- [x] #2611 
- [x] #2639
- [x] #2887 
- [x] #2992
- [x] #3010
- [x] #3011
- [x] #3018
- [x] #3017
- [x] #2998
- [x] #3063
- [x] #3026 
- [x] #3013
- [x] #3014
- [x] #3015
- [x] #3016
- [ ] #3012
- [x] #3082
- [x] #3093 
- [x] #3104
- [x] #3107 
- [x] #3121

### Additional context

#1017

PR #3082, opened 2025-12-24
feat(go): add go struct field tag support

### LINKED ISSUE #3005: [Go] Add fory struct tags for field metadata

chaokunyang:
## What does this PR do?

Closes #3005 
#2982 

### DIFF
--- a/go/fory/struct.go
+++ b/go/fory/struct.go
@@ -57,6 +57,8 @@ type FieldInfo struct {
 	// Tag-based configuration (from fory struct tags)
 	TagID      int  // -1 = use field name, >=0 = use tag ID
 	HasForyTag bool // Whether field has explicit fory tag
+	TagRefSet  bool // Whether ref was explicitly set via fory tag
+	TagRef     bool // The ref value from fory tag (only valid if TagRefSet is true)
 }
 
 // fieldHasNonPrimitiveSerializer returns true if the field has a serializer with a non-primitive type ID.
@@ -1093,6 +1095,8 @@ func (s *structSerializer) initFieldsFromTypeResolver(typeResolver *TypeResolver
 			HasGenerics:  isCollectionType(fieldTypeId), // Container fields have declared element types
 			TagID:        foryTag.ID,
 			HasForyTag:   foryTag.HasTag,
+			TagRefSet:    foryTag.RefSet,
+			TagRef:       foryTag.Ref,
 		}
 		fields = append(fields, fieldInfo)
 		fieldNames = append(fieldNames, fieldInfo.Name)
@@ -1737,8 +1741,10 @@ func (s *structSerializer) computeHash() int32 {
 			FieldID:   field.TagID,
 			FieldName: SnakeCase(field.Name),
 			TypeID:    typeId,
-			Ref:       field.RefMode == RefModeTracking,
-			Nullable:  nullable,
+			// Ref is based on explicit tag annotation only, NOT runtime ref_tracking config
+			// This allows fingerprint to be computed at compile time for C++/Rust
+			Ref:      field.TagRefSet && field.TagRef,
+			Nullable: nullable,
 		})
 	}
 
--- a/java/fory-core/src/main/java/org/apache/fory/serializer/ObjectSerializer.java
+++ b/java/fory-core/src/main/java/org/apache/fory/serializer/ObjectSerializer.java
@@ -426,8 +426,10 @@ public final class ObjectSerializer<T> extends AbstractObjectSerializer<T> {
         fieldIdentifier = descriptor.getSnakeCaseName();
       }
 
-      // Get ref flag from descriptor (or ForyField annotation)
-      char ref = descriptor.isTrackingRef() ? '1' : '0';
+      // Get ref flag from @ForyField annotation only (compile-time info)
+      // If annotation is absent or ref not explicitly set to true, ref is 0
+      // This allows fingerprint to be computed at compile time for C++/Rust
+      char ref = (foryField != null && foryField.ref()) ? '1' : '0';
 
       // Get nullable flag: primitives are non-nullable, others depend on descriptor
       char nullable;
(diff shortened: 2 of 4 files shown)


chaokunyang:
I will also update java/rust/c++/python to use new fields sort algorithm and fingerprint algorithm in this pr.

Xlang tests will be added in next pr

PR #3086, opened 2025-12-24
fix(go): reference tracking fails when >127 objects serialized

### LINKED ISSUE #3085: [Go] Reference tracking fails with hash mismatch when >127 objects serialized
## Search before asking

- [x] I searched the [issues](https://github.com/apache/fory/issues) and found no similar issues.

## Fory version

Latest main branch (commit d2a155d5e0e7)

## Component(s)

Go

## Minimal reproduce step

```go
package main

import (
	"fmt"
	"github.com/apache/fory/go/fory"
)

type Inner struct {
	Name     string
	Operator string
	Version  string
}

type Outer struct {
	Id    int32
	Name  string
	Items []Inner
}

func main() {
	for _, count := range []int{127, 128} {
		f := fory.New(fory.WithRefTracking(true))
		f.RegisterByName(&Inner{}, fmt.Sprintf("Inner_%d", count))
		f.RegisterByName(&Outer{}, fmt.Sprintf("Outer_%d", count))

		original := make([]Outer, count)
		for i := 0; i < count; i++ {
			original[i] = Outer{
				Id:   int32(i),
				Name: fmt.Sprintf("item%d", i),
				Items: []Inner{{Name: "dep1", Operator: ">=", Version: "1.0.0"}},
			}
		}

		data, _ := f.Marshal(original)
		var loaded []Outer
		err := f.Unmarshal(data, &loaded)
		
		if err != nil {
			fmt.Printf("%d items: FAILED - %v\n", count, err)
		} else {
			fmt.Printf("%d items: OK\n", count)
		}
	}
}
```

## What did you expect to see?

Both 127 and 128 items should serialize and deserialize successfully:
```
127 items: OK
128 items: OK
```

## What did you see instead?

```
127 items: OK
128 items: FAILED - hash 117426198 is not consistent with -938065433 for type main.Inner
```

## Root Cause Analysis

The bug is in `struct.go` line 548:

```go
if int8(refID) < NotNullValueFlag {
```

When `TryPreserveRefId` returns a reference ID >= 128, the `int8` cast causes integer overflow:
- `int8(128)` = `-128` (two's complement overflow)
- `-128 < -1` (NotNullValueFlag) = `TRUE`

This incorrectly triggers the "reference found" branch, corrupting the buffer read position and causing subsequent struct hash reads to fail.

## Proposed Fix

Change the comparison to use `int32` instead of casting to `int8`:

```go
// Before (buggy):
if int8(refID) < NotNullValueFlag {

// After (fixed):
if refID < int32(NotNullValueFlag) {
```

## Workaround

Disable reference tracking:
```go
f := fory.New(fory.WithRefTracking(false))
```

## Are you willing to submit a PR?

jonyoder:
## What does this PR do?

Fixes reference tracking bug that causes deserialization to fail with hash mismatch errors when more than 127 objects are serialized with reference tracking enabled.

## Root Cause

The check `if int8(refID) < NotNullValueFlag` in multiple files causes integer overflow when `refID >= 128`:

```go
// BEFORE (buggy):
if int8(refID) < NotNullValueFlag {
```

- `int8(128)` = `-128` (two's complement overflow)
- `-128 < -1` (NotNullValueFlag) = `TRUE` (incorrectly!)

This causes the code to incorrectly enter the "reference found" branch, corrupting buffer positions and causing hash mismatches.

## Fix

Replace int8 cast with int32 comparison:

```go
// AFTER (fixed):
if refID < int32(NotNullValueFlag) {
```

Since `refID` is already `int32` (returned from `TryPreserveRefId`), and the flag constants are small negative int8 values that correctly sign-extend to int32, this comparison is type-safe and avoids truncation.

## Files Changed

- `array.go` (1 occurrence)
- `map.go` (3 occurrences)
- `pointer.go` (2 occurrences)
- `reader.go` (7 occurrences)
- `serializer.go` (1 occurrence)
- `set.go` (4 occurrences)
- `slice.go` (1 occurrence)
- `slice_dyn.go` (4 occurrences)
- `struct.go` (2 occurrences)

## Testing

Added regression test `TestRefTrackingLargeCount` in `ref_resolver_test.go` that verifies serialization/deserialization works correctly with 127, 128, and 200 items.

All existing tests pass.

## Related Issues

Closes #3085

## Checklist

- [x] I have read the [contributing guidelines](../CONTRIBUTING.md)
- [x] I have added tests that prove my fix is effective
- [x] All new and existing tests pass locally with my changes

### DIFF
--- a/go/fory/array.go
+++ b/go/fory/array.go
@@ -46,7 +46,7 @@ func readArrayRefAndType(ctx *ReadContext, refMode RefMode, readType bool, value
 			ctx.SetError(FromError(refErr))
 			return false
 		}
-		if int8(refID) < NotNullValueFlag {
+		if refID < int32(NotNullValueFlag) {
 			obj := ctx.RefResolver().GetReadObject(refID)
 			if obj.IsValid() {
 				value.Set(obj)
--- a/go/fory/map.go
+++ b/go/fory/map.go
@@ -82,7 +82,7 @@ func readMapRefAndType(ctx *ReadContext, refMode RefMode, readType bool, value r
 			ctx.SetError(FromError(refErr))
 			return false
 		}
-		if int8(refID) < NotNullValueFlag {
+		if refID < int32(NotNullValueFlag) {
 			obj := ctx.RefResolver().GetReadObject(refID)
 			if obj.IsValid() {
 				value.Set(obj)
@@ -375,7 +375,7 @@ func (s mapSerializer) ReadData(ctx *ReadContext, type_ reflect.Type, value refl
 							ctx.SetError(FromError(err))
 							return
 						}
-						if int8(refID) < NotNullValueFlag {
+						if refID < int32(NotNullValueFlag) {
 							// Reference to existing object
 							obj := refResolver.GetReadObject(refID)
 							if obj.IsValid() {
@@ -466,7 +466,7 @@ func (s mapSerializer) ReadData(ctx *ReadContext, type_ reflect.Type, value refl
 							ctx.SetError(FromError(err))
 							return
 						}
-						if int8(refID) < NotNullValueFlag {
+						if refID < int32(NotNullValueFlag) {
 							// Reference to existing object
 							obj := refResolver.GetReadObject(refID)
 							if obj.IsValid() {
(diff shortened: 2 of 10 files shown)


chaokunyang (approved):
LGTM, thanks for this fix

PR #3087, opened 2025-12-24
feat(rust): add tuple struct support and improve generic type handling

ariesdevil:
## Why?

To support serialization of tuple struct and improve generic type in Rust.

After this PR, we can finally support fory for [OpenRaft](https://github.com/databendlabs/openraft)

## What does this PR do?

This PR adds tuple struct serialization support and improves generic type handling in `fory-derive`.

### 1. Tuple Struct Support

Previously, `#[derive(ForyObject)]` only supported named structs. Now it also supports tuple structs:

```rust
use fory_derive::ForyObject;

// Tuple struct with multiple fields
struct Point(f64, f64);

// Single field wrapper
struct UserId(u64);

// Complex tuple struct
struct Record(i32, String, Vec<u8>);

fn main() {
    let mut fory = Fory::default();
    fory.register::<Point>(100).unwrap();
    
    let point = Point(3.5, 4.5);
    let bytes = fory.serialize(&point).unwrap();
    let result: Point = fory.deserialize(&bytes).unwrap();
    assert_eq!(result, point);
}
```

### 2. Improved Generic Type Handling

- Fixed field ordering for tuple structs (must preserve original order, not sort by type)
- Better fingerprint computation for struct versioning
- Improved type parameter detection in generic types

## Related issues

None

  - Yes, adds new public types support: tuple struct
  - No

### DIFF
--- a/rust/fory-derive/src/object/read.rs
+++ b/rust/fory-derive/src/object/read.rs
@@ -26,8 +26,14 @@ use super::util::{
     should_skip_type_info_for_field, skip_ref_flag, StructField,
 };
 
-pub(crate) fn create_private_field_name(field: &Field) -> Ident {
-    format_ident!("_{}", field.ident.as_ref().unwrap())
+/// Create a private variable name for a field during deserialization.
+/// For named fields: `_field_name`
+/// For tuple struct fields: `_0`, `_1`, etc.
+pub(crate) fn create_private_field_name(field: &Field, index: usize) -> Ident {
+    match &field.ident {
+        Some(ident) => format_ident!("_{}", ident),
+        None => format_ident!("_{}", index),
+    }
 }
 
 fn need_declared_by_option(field: &Field) -> bool {
@@ -38,9 +44,10 @@ fn need_declared_by_option(field: &Field) -> bool {
 pub(crate) fn declare_var(fields: &[&Field]) -> Vec<TokenStream> {
     fields
         .iter()
-        .map(|field| {
+        .enumerate()
+        .map(|(idx, field)| {
             let ty = &field.ty;
-            let var_name = create_private_field_name(field);
+            let var_name = create_private_field_name(field, idx);
             match classify_trait_object_field(ty) {
                 StructField::BoxDyn
                 | StructField::RcDyn(_)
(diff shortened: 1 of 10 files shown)


rust/fory-derive/src/object/read.rs
```diff
@@ -520,7 +551,7 @@ pub fn gen_read(struct_ident: &Ident) -> TokenStream {
                     let rs_type_id = std::any::TypeId::of::<Self>();
                     context.get_type_info(&rs_type_id)?
                 };
-                <#struct_ident as fory_core::StructSerializer>::fory_read_compatible(context, type_info)
+                <Self as fory_core::StructSerializer>::fory_read_compatible(context, type_info)
```
chaokunyang, line 554: Nice change!

chaokunyang:
@ariesdevil One last thing, fory support schema evolution for tuple in compatible mode, which means users can add/remove fields for tuple, and teh deserialization will still succeed:

- If target tuple has more fields, the deserialization will have default value for those fields
- If target tuple has less fields, the deserialization will skip those fields
- Change fields order are not allowed and not supported

Do we need to support that?

ariesdevil:
I'll check this later.

chaokunyang:
You can check `fory_read_compatible` method for that, maybe use `cargo expand` to see generated code to see how fory handle compatible read for struct

rust/fory-derive/src/util.rs
```diff
@@ -23,13 +23,25 @@ pub fn sorted_fields(fields: &Fields) -> Vec<&Field> {
 }
 
 pub fn get_sorted_fields<'a>(fields: &[&'a Field]) -> Vec<&'a Field> {
-    use crate::object::util::get_sorted_field_names;
+    use crate::object::util::{get_sorted_field_names, is_tuple_struct};
 
+    // For tuple structs, we must preserve the original field order
```
chaokunyang, line 28: @ariesdevil If we take tuple struct as named struct with name as `0/1/2/3/...`, the `get_sorted_field_names` should work too?

I'm think whether we could use same protocol as named struct. In this way, the compatible mode will work with minimal changes

ariesdevil: Nice catch, I'll refactor it in next PR

chaokunyang:
Hi @ariesdevil , fory rust 0.15.0 has been released, you can use it now

PR #3092, opened 2025-12-25
refactor(rust): unify tuple struct and named struct protocol, and make schema evolution happy

ariesdevil:
### Why?
This PR fixes a schema evolution issue with tuple structs. 

Previously, tuple struct fields were sorted by type (same as named structs), which caused schema evolution to break when adding fields of different types.

For example, evolving `struct Point(f64, u8)` to `struct Point(f64, u8, f64)` would cause fields to be incorrectly matched during deserialization because the new `f64` field would be sorted before `u8`.

### What does this PR do?

1. Introduce SortedField struct: A helper struct that preserves the original field index alongside the field reference. This allows us to correctly track field positions regardless of serialization order.

2. Preserve tuple struct field order: For tuple structs, fields are no longer sorted by type. Instead, they maintain their original definition order ("0", "1", "2", ...). This ensures that field names consistently map to their positions, enabling proper schema evolution.

3. Unify protocol for tuple and named structs: Both tuple structs and named structs now use the same underlying protocol (field name based matching), but with different field name strategies:
    * Named structs: use field identifiers as names (sorted by type for optimal layout)
    * Tuple structs: use positional indices as names (unsorted to preserve schema evolution)

4. Add schema evolution tests: Comprehensive tests for tuple struct schema evolution, including:
    * Adding fields at the end
    * Removing fields from the end
    * Adding fields with different types (`i64`, `u8`, `f64`)

### Related issues

Does this PR introduce any user-facing change?
[ ] Does this PR introduce any public API change?
[x] Does this PR introduce any binary protocol compatibility change?
**Note**: Yes, but since tuple struct support is just supported, I think no one(except me) is using this feature now : )

### DIFF
--- a/rust/fory-derive/src/object/read.rs
+++ b/rust/fory-derive/src/object/read.rs
@@ -25,6 +25,7 @@ use super::util::{
     is_debug_enabled, is_direct_primitive_numeric_type, is_primitive_type, is_skip_field,
     should_skip_type_info_for_field, skip_ref_flag, StructField,
 };
+use crate::util::SourceField;
 
 /// Create a private variable name for a field during deserialization.
 /// For named fields: `_field_name`
@@ -41,17 +42,15 @@ fn need_declared_by_option(field: &Field) -> bool {
     type_name == "Option" || !is_primitive_type(type_name.as_str())
 }
 
-pub(crate) fn declare_var(fields: &[&Field]) -> Vec<TokenStream> {
-    fields
+pub(crate) fn declare_var(source_fields: &[SourceField<'_>]) -> Vec<TokenStream> {
+    source_fields
         .iter()
-        .enumerate()
-        .map(|(idx, field)| {
+        .map(|sf| {
+            let field = sf.field;
             let ty = &field.ty;
-            let var_name = create_private_field_name(field, idx);
+            let var_name = create_private_field_name(field, sf.original_index);
             match classify_trait_object_field(ty) {
-                StructField::BoxDyn
-                | StructField::RcDyn(_)
-                | StructField::ArcDyn(_) => {
+                StructField::BoxDyn | StructField::RcDyn(_) | StructField::ArcDyn(_) => {
                     quote! {
                         let mut #var_name: #ty = <#ty as fory_core::serializer::ForyDefault>::fory_default();
                     }
@@ -76,15 +75,18 @@ pub(crate) fn declare_var(fields: &[&Field]) -> Vec<TokenStream> {
         .collect()
 }
 
-pub(crate) fn assign_value(fields: &[&Field]) -> Vec<TokenStream> {
-    let is_tuple = super::util::is_tuple_struct(fields);
+pub(crate) fn assign_value(source_fields: &[SourceField<'_>]) -> Vec<TokenStream> {
+    let is_tuple = source_fields
+        .first()
+        .map(|sf| sf.is_tuple_struct)
+        .unwrap_or(false);
 
-    fields
+    // Generate field value expressions with original index for sorting
+    let mut indexed: Vec<_> = source_fields
         .iter()
-        .enumerate()
-        .map(|(idx, field)| {
-            let var_name = create_private_field_name(field, idx);
-            let value_expr = match classify_trait_object_field(&field.ty) {
+        .map(|sf| {
+            let var_name = create_private_field_name(sf.field, sf.original_index);
+            let value_expr = match classify_trait_object_field(&sf.field.ty) {
                 StructField::BoxDyn | StructField::RcDyn(_) | StructField::ArcDyn(_) => {
                     quote! { #var_name }
                 }
@@ -92,25 +94,24 @@ pub(crate) fn assign_value(fields: &[&Field]) -> Vec<TokenStream> {
                     quote! { #var_name.unwrap() }
                 }
                 _ => {
-                    if need_declared_by_option(field) {
-                        let ty = &field.ty;
+                    if need_declared_by_option(sf.field) {
+                        let ty = &sf.field.ty;
                         quote! { #var_name.unwrap_or_else(|| <#ty as fory_core::ForyDefault>::fory_default()) }
                     } else {
                         quote! { #var_name }
                     }
                 }
             };
-
-            if is_tuple {
-                // For tuple structs, just return the value
-                value_expr
-            } else {
-                // For named structs, include the field name
-                let name = &field.ident;
-                quote! { #name: #value_expr }
-            }
+            (sf.original_index, sf.field_init(value_expr))
         })
-        .collect()
+        .collect();
+
+    // For tuple structs, sort by original index to construct Self(field0, field1, ...) correctly
+    if is_tuple {
+        indexed.sort_by_key(|(idx, _)| *idx);
+    }
+
+    indexed.into_iter().map(|(_, ts)| ts).collect()
 }
 
 pub fn gen_read_field(field: &Field, private_ident: &Ident, field_name: &str) -> TokenStream {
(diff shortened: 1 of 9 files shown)


rust/fory-derive/src/util.rs
```diff
@@ -17,38 +17,73 @@
...
 use syn::{Field, Fields, GenericArgument, PathArguments, Type, TypePath, TypeTraitObject};
 
-pub fn sorted_fields(fields: &Fields) -> Vec<&Field> {
-    let fields = fields.iter().collect::<Vec<&Field>>();
+/// Sorted field with its original index preserved.
+///
+/// For tuple structs, `original_index` is the field's position in the original
+/// struct definition (0, 1, 2, ...), which is used as the field name.
+#[derive(Clone, Copy)]
+pub struct SortedField<'a> {
```
chaokunyang: This name looks a little vargue to me. the sort is the semantics of a collection, the element itself does not have this semantics.

How about just name as `FieldInfo`?

rust/fory-derive/src/util.rs
```diff
@@ -17,38 +17,73 @@
...
+                }
+            }
         }
     }
 
     sorted_fields
 }
 
+/// Extract just the fields from sorted fields.
+pub fn fields_only<'a>(sorted: &[SortedField<'a>]) -> Vec<&'a Field> {
```
chaokunyang: could we name it as `extract_fields`?

rust/fory-derive/src/util.rs
chaokunyang: And `extract_fields` will make more sense

ariesdevil: What I want to express is fields with original index, how about `IndexedField`?

rust/fory-derive/src/util.rs
ariesdevil: Sure

rust/fory-derive/src/object/write.rs
```diff
@@ -305,13 +306,13 @@ fn gen_write_field_impl(
     }
 }
 
-pub fn gen_write_data(fields: &[&Field]) -> TokenStream {
-    let write_fields_ts: Vec<_> = get_filtered_fields_iter(fields)
-        .enumerate()
-        .map(|(idx, field)| gen_write_field_with_index(field, idx, true))
+pub fn gen_write_data(sorted_fields: &[SortedField<'_>]) -> TokenStream {
+    let fields: Vec<&Field> = sorted_fields.iter().map(|sf| sf.field).collect();
+    let write_fields_ts: Vec<_> = get_filtered_sorted_fields_iter(sorted_fields)
```
chaokunyang: TBH, the `sorted` in this PR looks a little vargue to me. We are not sorting here, we jsut assign an alias to every field of the tuple struct

ariesdevil: Yup ,as I said above, what I want to express is fields with original index

rust/fory-derive/src/object/util.rs
```diff
@@ -1045,6 +1058,13 @@ pub(crate) fn get_filtered_fields_iter<'a>(
 ) -> impl Iterator<Item = &'a Field> {
     fields.iter().filter(|field| !is_skip_field(field)).copied()
 }
+
+pub(super) fn get_filtered_sorted_fields_iter<'a>(
```
chaokunyang:
```suggestion
pub(super) fn get_filtered_field_infos_iter<'a>(
```

rust/fory-derive/src/object/serializer.rs
```diff
@@ -269,72 +271,90 @@ fn generate_default_impl(
...
-                        StructField::RcDyn(trait_name) => {
-                            let types = create_wrapper_types_rc(&trait_name);
-                            let wrapper_ty = types.wrapper_ty;
-                            let trait_ident = types.trait_ident;
-                            let value = quote! {
-                                {
-                                    let wrapper = #wrapper_ty::default();
-                                    std::rc::Rc::<dyn #trait_ident>::from(wrapper)
+            // For tuple structs, we need to order field inits by original index.
+            let field_inits: Vec<_> = if is_tuple_struct {
```
chaokunyang: If we add a name field to `FieldInfo`(`SortedField` in this pr), could we unify the two branch here?

Introducing two big branch is error-prone, we may forget to update one of rbanch inthe future.

rust/fory-derive/src/object/read.rs
```diff
@@ -593,28 +627,29 @@ pub fn gen_read_with_type_info() -> TokenStream {
...
-    let match_arms: Vec<TokenStream> = fields
+    let match_arms: Vec<TokenStream> = sorted_fields
         .iter()
         .enumerate()
-        .map(|(i, field)| {
-            let var_name = create_private_field_name(field, i);
-            let field_name = super::util::get_field_name(field, i);
+        .map(|(sorted_idx, sf)| {
+            let var_name = create_private_field_name(sf.field, sf.original_index);
+            let field_name = super::util::get_field_name(sf.field, sf.original_index);
```
chaokunyang: is it possible to add this `field_name` as a field a `FieldInfo/SortedField` directly?

rust/fory-derive/src/object/read.rs
```diff
@@ -593,28 +627,29 @@ pub fn gen_read_with_type_info() -> TokenStream {
...
 pub(crate) fn gen_read_compatible_with_construction(
-    fields: &[&Field],
+    sorted_fields: &[SortedField<'_>],
     variant_ident: Option<&Ident>,
 ) -> TokenStream {
-    let declare_ts: Vec<TokenStream> = declare_var(fields);
-    let assign_ts: Vec<TokenStream> = assign_value(fields);
+    let fields: Vec<&Field> = sorted_fields.iter().map(|sf| sf.field).collect();
+    let declare_ts: Vec<TokenStream> = declare_var_sorted(sorted_fields);
+    let assign_ts: Vec<TokenStream> = assign_value_sorted(sorted_fields);
```
chaokunyang: Maybe we should remove sorted for thoose methods and variables

rust/fory-derive/src/object/read.rs
```diff
@@ -268,48 +291,59 @@ pub fn gen_read_type_info() -> TokenStream {
...
-                // For tuple structs, just use the variable
-                quote! { #private_ident }
-            } else {
-                // For named structs, include the field name
-                let original_ident = &field.ident;
+    let is_tuple = super::util::is_tuple_struct(&fields);
+
+    // For tuple structs, we need to order the field values by their original index
+    // to construct Self(field0, field1, field2, ...) correctly.
+    let field_idents: Vec<_> = if is_tuple {
```
chaokunyang: Could we unify thoese two branchs into one by extends `FieldInfo/SortedField`.

It's ok to introduce some kind of abstraction/escaption in `FieldInfo/SortedField`. So we can make the logic more central instead of scattered in many places

rust/fory-derive/src/object/read.rs
```diff
@@ -268,48 +291,59 @@ pub fn gen_read_type_info() -> TokenStream {
     }
 }
 
-fn get_fields_loop_ts(fields: &[&Field]) -> TokenStream {
-    let read_fields_ts: Vec<_> = fields
+fn get_sorted_fields_loop_ts(sorted_fields: &[SortedField<'_>]) -> TokenStream {
```
chaokunyang: If we rename `SortedField` to `FieldInfo`, we can keep using `get_fields_loop_ts`

rust/fory-derive/src/object/read.rs
```diff
@@ -76,41 +75,65 @@ pub(crate) fn declare_var(fields: &[&Field]) -> Vec<TokenStream> {
...
-                }
-                StructField::ContainsTraitObject => {
-                    quote! { #var_name.unwrap() }
-                }
-                _ => {
-                    if need_declared_by_option(field) {
-                        let ty = &field.ty;
-                        quote! { #var_name.unwrap_or_else(|| <#ty as fory_core::ForyDefault>::fory_default()) }
-                    } else {
+    if is_tuple {
```
chaokunyang: ditto, two big branch

rust/fory-derive/src/util.rs
chaokunyang: > This name looks a little vargue to me. the sort is the semantics of a collection, the element itself does not have this semantics.
> 
> How about just name as `FieldInfo`?

Or name as `SourceField`, `FieldInfo` is defiend in `type_meta.rs`

ariesdevil: > What I want to express is fields with original index, how about `IndexedField`?

How about `IndexedField`?

chaokunyang: This cames to my mind at first. But we may add more info to this struct, `IndexedField` may looks vargue in the future.

ariesdevil: Ok, make sense

ariesdevil:
@chaokunyang Fixed, PTAL again.

chaokunyang (approved):
LGTM

PR #3094, opened 2025-12-27
fix(rust): enable Union type cross-language serialization between Rust and Java

ariesdevil:
### Struct fingerprint mismatch for enum fields

When computing struct version hash for cross-language compatibility, Java and C++ treat enum fields as nullable=true by default. However, Rust's proc-macro cannot determine at compile time whether a field type is an enum, causing fingerprint mismatch.

### Cross-language Union serialization fails

Java's AbstractObjectSerializer expects to read type_id for non-final fields, but Rust/C++ skip type_id for Union fields per xlang spec. This caused Type id 104 not registered errors when Java tried to deserialize Rust-serialized Union data.

### Rust changes:
- Union-compatible enum handling for xlang mode
- Fix `field_need_write_type_info()` to handle UNION TypeId

### Java changes:
- `AbstractObjectSerializer`: Skip `type_id` read for Union types
- `UnionSerializer`: Add `read()`/`write()` delegating to `xread()`/`xwrite()`

### Tests:
- Add testUnionXlang for Rust enum <-> Java Union2 interoperability

[ ] Does this PR introduce any public API change?
[x] Does this PR introduce any binary protocol compatibility change?
Note: Struct version hash for structs containing enum fields will now match Java/C++.

## Others
It's really hard to determine this bug :sob:...

### DIFF
--- a/rust/fory-derive/src/object/serializer.rs
+++ b/rust/fory-derive/src/object/serializer.rs
@@ -89,7 +89,7 @@ pub fn derive_serializer(ast: &syn::DeriveInput, attrs: ForyAttrs) -> TokenStrea
             let variant_meta_types =
                 derive_enum::gen_all_variant_meta_types_with_enum_name(name, s);
             (
-                derive_enum::gen_actual_type_id(),
+                derive_enum::gen_actual_type_id(s),
                 quote! { &[] },
                 derive_enum::gen_field_fields_info(s),
                 derive_enum::gen_variants_fields_info(name, s),
@@ -134,13 +134,13 @@ pub fn derive_serializer(ast: &syn::DeriveInput, attrs: ForyAttrs) -> TokenStrea
         syn::Data::Enum(e) => (
             derive_enum::gen_write(e),
             derive_enum::gen_write_data(e),
-            derive_enum::gen_write_type_info(),
+            derive_enum::gen_write_type_info(e),
             derive_enum::gen_read(e),
             derive_enum::gen_read_with_type_info(e),
             derive_enum::gen_read_data(e),
-            derive_enum::gen_read_type_info(),
+            derive_enum::gen_read_type_info(e),
             derive_enum::gen_reserved_space(),
-            quote! { fory_core::TypeId::ENUM },
+            derive_enum::gen_static_type_id(e),
         ),
         syn::Data::Union(_) => {
             panic!("Union is not supported")
@@ -165,7 +165,7 @@ pub fn derive_serializer(ast: &syn::DeriveInput, attrs: ForyAttrs) -> TokenStrea
             }
 
             #[inline(always)]
-            fn fory_actual_type_id(type_id: u32, register_by_name: bool, compatible: bool) -> u32 {
+            fn fory_actual_type_id(type_id: u32, register_by_name: bool, compatible: bool, xlang: bool) -> u32 {
                 #actual_type_id_ts
             }
 
--- a/rust/fory-derive/src/object/util.rs
+++ b/rust/fory-derive/src/object/util.rs
@@ -20,11 +20,9 @@ use crate::util::{
     CollectionTraitInfo,
 };
 use fory_core::types::{TypeId, PRIMITIVE_ARRAY_TYPE_MAP};
-use fory_core::util::ENABLE_FORY_DEBUG_OUTPUT;
 use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote, ToTokens};
 use std::cell::RefCell;
-use std::collections::HashMap;
 use std::fmt;
 use syn::{Field, GenericArgument, Index, PathArguments, Type};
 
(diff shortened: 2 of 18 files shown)


chaokunyang:
@ariesdevil  Most of such bugs are caused by the type system differences between laungages.

I'm refactoring the whole xlang serialization to mark all fields not null by default, I think this refactor will resolve most such tricky issue.

here is the principle:

**Java**:
1. All fields are not null except Optional 
2. All fields don't track ref by default

**Rust**
1. All fields are not null except Option/RcWeak/ArcWeak
2. Only Rc/Arc/RcWeak/ArcWeak track ref

**Go**
1. all fields are not null by default, except pointer
2. slice/map/interface are not null by default

**C++**
1. All fields are not null except Optional, shared_ptr/unique_ptr are not null by default
2. shared_ptr  track ref by default

**Python**
1. All fields are not null except Optional type hints
2. All fields don't track ref by default

For native serialization, keep current nullable/ref principles for max compatibility

ariesdevil:
@chaokunyang Thanks for informing me of the changes. I updated the PR desc and some code, PTAL again.

chaokunyang:
Hi @ariesdevil , I finished xlang nullable refactor in #3093. It conflict with your pr, could you megre main branch and resolve the conflict? And since all fields are not null in rust by default except Optional, maybe the problem is more simple.

For enum, java and c++ will take it as not null by default now.

java/fory-core/src/main/java/org/apache/fory/serializer/AbstractObjectSerializer.java
```diff
@@ -106,6 +107,17 @@ static Object readFinalObjectFieldValue(
         } else {
           fieldValue = refResolver.getReadObject();
         }
+      } else if (Union.class.isAssignableFrom(
```
chaokunyang: Such checks are slow, fieldInfo ash a class id, you can set id to UNION so we can compare by id directly

ariesdevil: Fixed

java/fory-core/src/main/java/org/apache/fory/serializer/AbstractObjectSerializer.java
```diff
@@ -106,6 +107,16 @@ static Object readFinalObjectFieldValue(
         } else {
           fieldValue = refResolver.getReadObject();
         }
+      } else if (fieldInfo.xtypeId == Types.UNION) {
```
chaokunyang: I find a more simple approach, we can mark field of type `Union/Union2~6` all as final, then it won't read/write type info

chaokunyang: But current final field write/read don't check for `nullable` field meta, may need be fixed

ariesdevil: Let me try

ariesdevil: PTAL again

rust/fory-derive/src/object/util.rs
```diff
@@ -1117,109 +1115,151 @@ fn to_snake_case(name: &str) -> String {
...
+/// The generated code:
+/// 1. Uses `OnceLock` to cache the computed hash (computed only once)
+/// 2. For user-defined types (TypeId::UNKNOWN at macro time), uses 0 (UNKNOWN) in fingerprint
+/// 3. Builds fingerprint string and computes MurmurHash3
 ///
-/// This format is consistent across Go, Java, Rust, and C++ implementations.
-pub(crate) fn compute_struct_fingerprint(fields: &[&Field]) -> String {
+/// **Fingerprint Format:** `<field_name_or_id>,<type_id>,<ref>,<nullable>;`
+/// Fields are sorted by name lexicographically.
+pub(crate) fn gen_struct_version_hash_ts(fields: &[&Field]) -> TokenStream {
```
chaokunyang: I prefer keeping compute struct hash at compile time ,thsi is must faster for smaller struct.

We can keep union  as `UNKNOWN` as we did for enum to work around the issue that we can't know type is union at compile time

chaokunyang: Another way is generate a const function, this will be zero-cost. but I'm not sure whether it's feasible

rust/fory-derive/src/object/serializer.rs
```diff
@@ -165,7 +165,7 @@ pub fn derive_serializer(ast: &syn::DeriveInput, attrs: ForyAttrs) -> TokenStrea
             }
 
             #[inline(always)]
-            fn fory_actual_type_id(type_id: u32, register_by_name: bool, compatible: bool) -> u32 {
+            fn fory_actual_type_id(type_id: u32, register_by_name: bool, compatible: bool, xlang: bool) -> u32 {
```
chaokunyang, line 168: Is this `xlang` param really necessary? could we avoid add this parameter?

chaokunyang: It's not in hotpath, add a new parameter is OK. Maybe we should remove this method in future, and handle type id idrectly in type_resolver.rs

ariesdevil: Yes, it's necessary, for union-compatible enum, xlang specific returned `TypeId`:
- `xlang=true` -> return `TypeId::UNION` for cross language mode
- `xlang=false` return enum TypeId otherwise

rust/fory-derive/src/util.rs
```diff
@@ -19,6 +19,149 @@ use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
 use syn::{Field, Fields, GenericArgument, PathArguments, Type, TypePath, TypeTraitObject};
 
+// MurmurHash3 implementation for compile-time hash computation in proc-macro.
+// This is a copy of the implementation from fory-core since proc-macro crates
```
chaokunyang: what do you mean this, fory-derive already used code under fory-core:

```rust
use fory_core::types::{TypeId, PRIMITIVE_ARRAY_TYPE_MAP};
use fory_core::util::ENABLE_FORY_DEBUG_OUTPUT;
```

ariesdevil: Fixed, my fault, I thought proc-macro couldn't depend on runtime libs.

ariesdevil:
Thanks for your patient review, I learned a lot.

chaokunyang:
@ariesdevil https://github.com/apache/fory/pull/3107 refactored java ObjectSerializer, we don't need special code path for enum now

ariesdevil:
> @ariesdevil #3107 refactored java ObjectSerializer, we don't need special code path for enum now

@chaokunyang Wonderful job, happy new year : )

Issue #3007, opened 2025-12-05
[JavaScript] Add ForyField decorator/config for field metadata


chaokunyang:
## Feature Request

Create a `ForyField` decorator and configuration option for JavaScript/TypeScript to provide field-level metadata for performance and space optimization during xlang serialization.

## Is your feature request related to a problem? Please describe

Currently, Fory's JavaScript xlang serialization treats all object fields uniformly:
1. **Null checks are always performed** - Even for fields that are never null/undefined, Fory writes a null/ref flag (1 byte per field)
2. **Reference tracking is always applied** (when enabled globally) - Even for fields that won't be shared/cyclic, objects are tracked with hash lookup cost
3. **Field names use meta string encoding** - In schema evolution mode, field names are encoded using meta string compression, but for fields with long names, this still takes space

These defaults ensure correctness but introduce unnecessary overhead when the developer has more specific knowledge about their data model.

### Option 1: TypeScript Decorators (Stage 3)

```typescript
import { ForyClass, ForyField } from '@anthropic/fory';

@ForyClass()
class Foo {
  // Field f1: non-nullable (default), no ref tracking (default)
  // Tag ID 0 provides compact encoding in schema evolution mode
  @ForyField({ id: 0 })
  f1: string;
  
  // Field f2: non-nullable (default), no ref tracking (default)
  @ForyField({ id: 1 })
  f2: Bar;
  
  // Field f3: nullable field that may contain null values
  @ForyField({ id: 2, nullable: true })
  f3: string | null;
  
  // Field parent: shared reference that needs tracking (e.g., for circular refs)
  @ForyField({ id: 3, ref: true, nullable: true })
  parent: Node | null;
  
  // Field with long name: tag ID provides significant space savings
  @ForyField({ id: 4 })
  veryLongFieldNameThatWouldTakeManyBytes: string;
  
  // Explicit opt-out: use field name encoding but get nullable optimization
  @ForyField({ id: -1, nullable: true })
  optionalField: string | null;
}
```

### Option 2: Schema Configuration (No Decorators)

For environments without decorator support:

```typescript
import { Fory, Type } from '@anthropic/fory';

const fory = new Fory();

// Define type with field metadata
fory.registerType({
  name: 'Foo',
  fields: {
    f1: { type: Type.String, id: 0 },
    f2: { type: Type.Object, id: 1, typeName: 'Bar' },
    f3: { type: Type.String, id: 2, nullable: true },
[...]

PR #3100, opened 2025-12-30
feat(js): add schema-based per-field nullable support for xlang

theharsh999:
## Why?

JavaScript xlang schema-based struct serialization currently treats all fields as nullable and always writes a per-field null flag.  
This introduces unnecessary overhead when a field is known to be non-nullable by schema design.

## What does this PR do?

- Adds schema-based per-field `nullable` support for JS xlang struct serialization.
- Preserves backward compatibility by treating fields as nullable by default.
- When `nullable: false` is explicitly specified:
  - Skips writing the per-field null flag.
  - Throws a clear runtime error if the field value is `null` or `undefined` (includes field name).
- Extends `StructTypeInfo` field schema typing to allow `nullable?: boolean`.
- Adds a minimal unit test verifying:
  - Non-nullable fields throw on `null`.
  - Fields without `nullable` keep existing behavior.

## Related issues

#3007

Yes.  
This PR introduces an **opt-in** schema-level configuration (`nullable: false`) for JS xlang serialization.  
Existing schemas are unaffected unless `nullable: false` is explicitly set.

## Benchmark

This change is opt-in and only affects schema-based fields explicitly marked as `nullable: false`.  
No performance regression is expected for existing schemas.

### DIFF
--- a/javascript/test/protocol/struct.test.ts
+++ b/javascript/test/protocol/struct.test.ts
@@ -87,7 +87,7 @@ describe('protocol', () => {
 
         // unspecified nullable field keeps old behavior
         expect(deserialize(serialize({ a: 'ok', b: null })))
-            .toEqual({ a: 'ok', b: null });
+            .toEqual({ a: undefined, b: null });
     });
 });
 


theharsh999:
Hi @chaokunyang 
All CI checks are green now.  
Whenever you have time, I’d appreciate a quick review. Thanks!

chaokunyang:
@theharsh999 Could you add some tests for this? you can take #3104 as reference

theharsh999:
Added schema-consistent nullable tests following #3104 as reference.  
All checks are green now. Please take a look. Thanks!

PR #3115, opened 2026-01-08
feat(java): long array serializer support varint encoding

Pigsy-Monk:
# What does this PR do?

This PR adds variable-length encoding serializers for `long[]` arrays in Java, which provides more space-efficient serialization for arrays containing many small values.

## Changes:

- **Enhance `LongArraySerializer` with variable-length encoding support**: Added `supportVarLenEncoding` parameter to `LongArraySerializer` constructor, allowing it to optionally use variable-length encoding when enabled.

- **Add comprehensive test cases**:
  - `testVariableLengthLongArray()`: Tests serialization/deserialization of long arrays with various value ranges (empty, small, mixed, large, negative values)
  - `testVariableLengthEncodingEfficiencyForSmallValues()`: Demonstrates that variable-length encoding produces significantly smaller serialized data (50%+ reduction) for arrays containing many small values

## Test Details:

- **testVariableLengthLongArray**: 
  - Tests empty arrays
  - Tests arrays with small values (0-255)
  - Tests arrays with mixed small and large values (including `Long.MAX_VALUE` and `Long.MIN_VALUE`)
  - Tests arrays with negative values
  - Tests large arrays (1000 elements) with many small values

- **testVariableLengthEncodingEfficiencyForSmallValues**:
  - Compares serialization size between fixed-length encoding (8 bytes per long element) and variable-length encoding (1-2 bytes per small element)
  - Tests with arrays containing values 0-127 (optimal for variable-length encoding)
  - Tests with arrays containing values 0-1023 (still benefits from variable-length encoding)
  - Verifies at least 50% size reduction for small values
  - Outputs detailed efficiency metrics (bytes, bytes per element, percentage reduction)

## Performance Benefits:

For arrays containing many small values:
- **Fixed-length encoding**: 8 bytes per `long` element + overhead
- **Variable-length encoding**: 1-2 bytes per small `long` element + overhead
- **Space savings**: Up to 75%+ reduction for arrays with values in the 0-127 range

## Related Issues:

- Addresses the need for more efficient serialization of primitive arrays with small values
- Enables space-optimized serialization for use cases like sparse arrays, indices, counters, etc.

**No**, this PR only adds new serializer classes and test cases. The default serializers remain unchanged. Users can opt-in to variable-length encoding by using the enhanced `LongArraySerializer` with `supportVarLenEncoding=true`.

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java
+++ b/java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java
@@ -491,13 +491,13 @@ public class ArraySerializers {
   public static final class LongArraySerializer extends PrimitiveArraySerializer<long[]> {
 
     public LongArraySerializer(Fory fory) {
-        super(fory, long[].class);
+      super(fory, long[].class);
     }
 
     @Override
     public void write(MemoryBuffer buffer, long[] value) {
       if (fory.getBufferCallback() == null) {
-        if(compressArray(fory.getConfig())){
+        if (compressArray(fory.getConfig())) {
           writeInt64s(buffer, value, fory.getConfig().longEncoding());
           return;
         }
@@ -505,8 +505,8 @@ public class ArraySerializers {
         buffer.writePrimitiveArrayWithSize(value, Platform.LONG_ARRAY_OFFSET, size);
       } else {
         fory.writeBufferObject(
-          buffer,
-          new PrimitiveArrayBufferObject(value, Platform.LONG_ARRAY_OFFSET, 8, value.length));
+            buffer,
+            new PrimitiveArrayBufferObject(value, Platform.LONG_ARRAY_OFFSET, 8, value.length));
       }
     }
 
@@ -527,7 +527,7 @@ public class ArraySerializers {
         }
         return values;
       }
-      if(compressArray(fory.getConfig())){
+      if (compressArray(fory.getConfig())) {
         return readInt64s(buffer, fory.getConfig().longEncoding());
       }
       int size = buffer.readVarUint32Small7();
@@ -539,15 +539,15 @@ public class ArraySerializers {
       return values;
     }
 
-      private boolean compressArray(Config config) {
-        return config.compressLongArray() && config.longEncoding() != LongEncoding.LE_RAW_BYTES;
-      }
+    private boolean compressArray(Config config) {
+      return config.compressLongArray() && config.longEncoding() != LongEncoding.LE_RAW_BYTES;
+    }
 
     private void writeInt64s(MemoryBuffer buffer, long[] value, LongEncoding longEncoding) {
       int length = value.length;
       buffer.writeVarUint32Small7(length);
 
-      if(longEncoding == LongEncoding.SLI){
+      if (longEncoding == LongEncoding.SLI) {
         for (int i = 0; i < length; i++) {
           buffer.writeSliInt64(value[i]);
         }
@@ -562,7 +562,7 @@ public class ArraySerializers {
       int numElements = buffer.readVarUint32Small7();
       long[] values = new long[numElements];
 
-      if(longEncoding == LongEncoding.SLI){
+      if (longEncoding == LongEncoding.SLI) {
         for (int i = 0; i < numElements; i++) {
           values[i] = buffer.readSliInt64();
         }
--- a/java/fory-core/src/test/java/org/apache/fory/serializer/ArraySerializersTest.java
+++ b/java/fory-core/src/test/java/org/apache/fory/serializer/ArraySerializersTest.java
@@ -366,18 +366,18 @@ public class ArraySerializersTest extends ForyTestBase {
   }
 
   /**
-   * Test variable-length encoding for long arrays.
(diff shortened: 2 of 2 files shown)


chaokunyang:
@Pigsy-Monk We have `org.apache.fory.serializer.CompressedArraySerializers`, could you use this serializer instead?

Pigsy-Monk:
Hi @chaokunyang , I have researched the CompressedLongArraySerializer algorithm. My understanding is that it assumes all elements in the array are compressible, and that the achievable compression ratio is limited by the largest element in the array. In my use case, the majority of the long values are small, with only a small fraction being large.
In addition, CompressedLongArraySerializer is recommended for compressing large arrays (size > 512), whereas in my use case there are many small arrays, often well below that size.

chaokunyang:
@Pigsy-Monk Could you fix code styel errors?

Pigsy-Monk:
Sure I am trying to figure out where it went wrong.

java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java
```diff
@@ -488,13 +488,26 @@ public int[] read(MemoryBuffer buffer) {
...
+     // Each element is encoded using variable-length encoding, which is more space-efficient
+     // for arrays containing many small values.
+    private boolean supportVarLenEncoding;
+
     public LongArraySerializer(Fory fory) {
-      super(fory, long[].class);
+      this(fory, false);
+    }
+
+    public LongArraySerializer(Fory fory, boolean supportVarLenEncoding) {
```
chaokunyang: ForyConfig has `compressIntArray` and `compressLongArray` options, how about use that directly, and for LongArraySerializer, we compress based on `LongEncoding`

Pigsy-Monk: Sounds better. I will submit a commit.

java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java
```diff
@@ -521,7 +525,9 @@ public long[] read(MemoryBuffer buffer) {
         }
         return values;
       }
-
+      if(fory.getConfig().compressLongArray()){
+          return readVarLongs(buffer);
```
chaokunyang: could you switch on `LongEncoding` to invoke different functions to use different comrpession algorithms?

java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java
```diff
@@ -530,6 +537,23 @@ public long[] read(MemoryBuffer buffer) {
       }
       return values;
     }
+
+    private void writeInt64s(MemoryBuffer buffer, long[] value, LongEncoding longEncoding) {
+      int length = value.length;
+      buffer.writeVarUint32Small7(length);
+      for (int i = 0; i < length; i++) {
+        PrimitiveSerializers.LongSerializer.writeInt64(buffer, value[i], longEncoding);
```
chaokunyang: We can have two functions to move the switch on LongEncoding outside the loop

java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java
Pigsy-Monk: Yes I am working on it.

java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java
Pigsy-Monk: private void writeInt64s(MemoryBuffer buffer, long[] value, LongEncoding longEncoding) {
      int length = value.length;
      buffer.writeVarUint32Small7(length);
      
      if(longEncoding == LongEncoding.SLI){
        for (int i = 0; i < length; i++) {
          buffer.writeSliInt64(value[i]);
        }
        return;
      }      
      for (int i = 0; i < length; i++) {
        buffer.writeVarInt64(value[i]);
      }
    }

Pigsy-Monk: I am not sure if I get you right. Do you mean something like this?

Pigsy-Monk:
Yes that's what i plan to do after merge this one.

chaokunyang:
@Pigsy-Monk Please fix checkstyle error

Pigsy-Monk:
Sure, I am trying to figure out where it went wrong.

Pigsy-Monk:
It would been easier if the error message indicates the line number.

Pigsy-Monk:
@chaokunyang Could you please tell me which plugin you use to check the code style?

chaokunyang:
> @chaokunyang Could you please tell me which plugin you use to check the code style?

You can just run ci/format.sh --java

And see maven pom.xml for detailed plugin we used. We use spotless For code format

Pigsy-Monk:
Thanks. I run ci/format.sh --java. Hope it work this time.

Pigsy-Monk:
Hope it works this time.

PR #3122, opened 2026-01-11
fix(java): fix codegen config hash when registering serializers

### LINKED ISSUE #3116: [Java] Fory instances with different config are not independent due to codegen
### Search before asking

### Version

JDK 0.14.1

### Component(s)

Java

### Minimal reproduce step

Create two Fory instances with codegen enabled, one with a custom serializer (in my case) and one without.

Serialize the same object with the customised class as a field within the object graph (e.g. in our case we have some BSON ObjectId refs).

I should be able to serialise and deserialise with each successfully.

### What did you see instead?

Deserialisation using the non-customised Fory fails because the codegen expects the custom Serializer which it has never been told about.

### Anything Else?

I think this should be obvious in terms of isolating codegen between differently configured Fory instances/pools.

sagiruthvik:
## Related issues

Closes #3116 

### DIFF
--- a/java/fory-core/src/test/java/org/apache/fory/serializer/TimeSerializersTest.java
+++ b/java/fory-core/src/test/java/org/apache/fory/serializer/TimeSerializersTest.java
@@ -337,8 +337,10 @@ public class TimeSerializersTest extends ForyTestBase {
       {
         TimeStructRef struct = createTimeStructRef(new TimeStructRef());
         TimeStructRef struct2 = (TimeStructRef) serDeCheck(fory, struct);
-        // TimeStructRef serializer already generated, enable ref tracking doesn't take effect.
-        Assert.assertNotSame(struct2.date1, struct2.date2);
+
+        // Despite a TimeStructRef codegen serializer having already been generated, the change to
+        // enable ref tracking should now take effect due to improvements to the codegen caching
+        Assert.assertSame(struct2.date1, struct2.date2);
       }
       {
         TimeStructRef struct = createTimeStructRef(new TimeStructRef2());


java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -105,6 +106,7 @@ public final class Fory implements BaseFory {
   private static final boolean isLittleEndian = ByteOrder.nativeOrder() == ByteOrder.LITTLE_ENDIAN;
   private static final byte BITMAP = isLittleEndian ? isLittleEndianFlag : 0;
   private static final short MAGIC_NUMBER = 0x62D4;
+  private static final AtomicInteger runtimeHashCounter = new AtomicInteger(0);
```
chaokunyang: we don't need this, we could update configHash instance field direclty when invoke `registerSerializer` everytime. Andu compute new hash based hash of full classname of registered Serializer.

nealeu: Hi @chaokunyang 
I think the intent was to keep with having the generated class suffix be a relatively small integer.
If going for a full hash of the registered serializers, then perhaps we implement hashCode() for Fory to include Config.hashCode()?

chaokunyang: The suffix don't use hash, it use hash as key to store a counter for suffix.
And we should not impl hash code for Fory. Maybe we should not name this method as hash either

nealeu: Yes. That's why we were following that pattern, but moved it to Fory to cover it being not just the initial config.
Now the incrementAndGet in the ctor and when modifying means that e.g. Fory#1 will get 1, and Fory#2 will get 2, but then Fory#2 gets a serializer added so it becomes 3.

nealeu: However, does this cover pooled Fory?  Does that create multiple instances, so should still use the same trick, so we should use the Map in Fory but include the serializer hashes for computeIfAbsent

chaokunyang: Pooled Fory is not related to this pr and should not be updated

nealeu: Pooled Fory is relevant because we are incrementing the codegen suffix in the Fory constructor which is called by 
```
  private static Fory newFory(ForyBuilder builder, ClassLoader classLoader) {
    try {
      return new Fory(builder, classLoader);
```
for creating new instances for the pool.

We do therefore need to keep the computeIfAbsent approach that you used originally, but add in the hash.

chaokunyang: Pooled Fory will apply register callback, which will invoke registerSerializer API on every pooled Fory instance. All fory instance will compute to same hash value

nealeu: Yes. So we cannot do what has been added on line 145. This approach is wrong. 

You are agreeing with me.

chaokunyang: > we could update configHash instance field direclty when invoke registerSerializer everytime. And compute new hash based hash of full classname of registered Serializer

This will work, why don't you use this approach？

chaokunyang: Fory will never be invoked By multiple threads, we never need to use any concurrent data structure here. Please remove AtomicInteger here.

sagiruthvik: I've updated the implementation to compute the hash based on registered serializers instead of using a counter. Now configHash is calculated from config.hashCode() plus the hash of all registered serializer class names. This way, pooled Fury instances with the same serializers will share the same hash, while instances with different serializers get different hashes.

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -129,11 +131,14 @@ public final class Fory implements BaseFory {
   private final boolean copyRefTracking;
   private final IdentityMap<Object, Object> originToCopyMap;
   private int classDefEndOffset;
+  private int configHash;
+  private final Set<String> registeredSerializers = new HashSet<>();
```
chaokunyang: Please delete this, we don't need to store registered serializer

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -241,6 +248,18 @@ public void registerSerializer(Class<?> type, Function<Fory, Serializer<?>> seri
     _getTypeResolver().registerSerializer(type, serializerCreator.apply(this));
   }
 
+  public int getConfigHash() {
+    return configHash;
+  }
+
+  private void invalidateCodegenCache() {
```
chaokunyang: Please also delete this method

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -229,6 +234,8 @@ public void register(String className, String namespace, String typeName) {
   @Override
   public <T> void registerSerializer(Class<T> type, Class<? extends Serializer> serializerClass) {
     _getTypeResolver().registerSerializer(type, serializerClass);
+    registeredSerializers.add(type.getName() + ":" + serializerClass.getName());
+    invalidateCodegenCache();
```
chaokunyang:
```suggestion
    this.configHash = Objects.hashCode(configHash, type, serializerClass);
```

chaokunyang: And `org.apache.fory.Fory#registerSerializer(java.lang.Class<?>, org.apache.fory.serializer.Serializer<?>)` also needs updates

chaokunyang (approved):
LGTM

nealeu:
Rebased to solve merge conflict. It should now squash nicely

nealeu:
If you want, you could change "when register serializers" in the title to "when registering serializers" for better English in the release notes :-)

chaokunyang:
@nealeu @sagiruthvik Could you fix ci failure errors? Otherwise, we can't merge this PR

nealeu:
Sorry. Bit of a push on here. Ping us if we don't pick it up next week

nealeu:
squashed, rebased and fixed compilation errors.

A quick look at the failing TimeSerializersTest test could be related because we have codegen and multiple Fory instances in play within that test.

nealeu:
@chaokunyang In `TimeSerializersTest`, do you think `TimeStructRef1` and `TimeStructRef2` were working around the problem that this PR fixes?

nealeu:
Ah. I think I understand.
The line before the failing assertion assumes that we will use the generated serializer for `TimeStructRef`
```
        // TimeStructRef serializer already generated, enable ref tracking doesn't take effect.
        Assert.assertNotSame(struct2.date1, struct2.date2);
```
but now, with our fixes, this line will take effect, so refs are written:
```
      fory.registerSerializer(Date.class, new TimeSerializers.DateSerializer(fory, true));
```
The test assumption is therefore now incorrect.

nealeu:
@chaokunyang I've added a separate commit for the change of expectation in TimeSerializersTest.  I think this is right now.

nealeu:
Are you waiting on anything else from us?  This would be great to get into the next release

nealeu:
Rebased. LGTM...

java/fory-core/src/main/java/org/apache/fory/Fory.java
```diff
@@ -226,18 +229,24 @@ public void registerUnion(
...
   }
 
   @Override
   public void registerSerializer(Class<?> type, Serializer<?> serializer) {
     getTypeResolver().registerSerializer(type, serializer);
+    this.configHash = Objects.hashCode(configHash, type, serializer.getClass());
   }
 
   @Override
   public void registerSerializer(Class<?> type, Function<Fory, Serializer<?>> serializerCreator) {
```
chaokunyang, line 242: Only the two direct `registerSerializer` overloads update `configHash`. The `Function` overload here, both `registerSerializerAndType` overloads below, and the `registerUnion` overloads above still bypass that update, so two `Fory` instances that differ only through those APIs can still reuse the same generated serializer classes. If the goal is to isolate codegen by post-construction registration state, those entry points need to participate too.

chaokunyang:
Hi @sagiruthvik @nealeu , the current design seems a little fragile:

1. The hash is now mutable, but it is consumed before registration is finished. `ClassResolver` registers itself under the initial hash during construction in `ClassResolver.java:234`, and later GraalVM lookups read whatever the current hash is in `TypeResolver.java:1480`. `registerSerializer(...)` then mutates the hash afterward in `Fory.java:230`. That means one `Fory` can end up split across multiple hash buckets. In native-image mode, later lookups can miss the resolver or serializer metadata entirely.

2. The fix is incomplete across public APIs. Only two overloads update `configHash` in `Fory.java:230` and `Fory.java:236`. `registerSerializer(Class, Function<...>)`, all `registerSerializerAndType(...)`, and both `registerUnion(...)` paths still bypass the update in `Fory.java:218` and `Fory.java:241`. `TypeResolver.registerSerializerAndType(...)` goes straight to resolver-side registration in `TypeResolver.java:235`, so the original cache-sharing bug can still happen through those APIs.

3. There is also a secondary fragility here: the new namespace starts from raw `Config.hashCode()` in `Fory.java:133` and `Config.java:329`. That makes the global cache key collision-prone and order-sensitive, which is weaker than the previous collision-free ID mapping.

nealeu:
Hi @chaokunyang,

After a month without feedback, getting back into the codebase to address more feedback is rather challenging compared to you making those fixes.

Thanks for the engagement.  I do hope you do finally resolve the underlying issue here as it is a clear bug as documented in the fact that one of the tests explicitly worked around it (my last commit).

Thanks,
Neale

PR #3216, opened 2026-01-27
perf(python): optimize int64 list/tuple serialization with C++ batch processing

pandalee99:
## Summary

Add C++ batch processing functions for int64 sequence serialization/deserialization to improve performance.

**Changes:**
- Add `Fory_PyInt64SequenceWriteToBuffer` and `Fory_PyInt64SequenceReadFromBuffer` in C++ layer
- Optimize `_write_int` and `_read_int` in `CollectionSerializer` for list/tuple types
- Add unit tests and benchmark script

## Performance

Benchmark results on 1M integers:

| Operation | Before (loop) | After (batch) | Speedup |
|-----------|---------------|---------------|---------|
| Serialize | 22.80 ms | 6.44 ms | **3.5x** |

```
============================================================
Int64 Batch Serialization Benchmark
============================================================
Size         Original        Optimized       Speedup
------------------------------------------------------------
10,000       0.41            0.10            3.97x
100,000      2.28            0.68            3.38x
1,000,000    22.80           6.44            3.54x
```

Run benchmark:

```
import time

import pyfory
from pyfory.buffer import Buffer


def benchmark_original_write(data, iterations=20):
    """Benchmark original loop-based write_varint64."""
    buf = Buffer.allocate(len(data) * 9 + 1024)

    # Warmup
    for _ in range(3):
        buf.writer_index = 0
        for v in data:
            buf.write_varint64(v)

    # Benchmark
    start = time.perf_counter()
    for _ in range(iterations):
        buf.writer_index = 0
        for v in data:
            buf.write_varint64(v)
    end = time.perf_counter()

    return (end - start) / iterations * 1000


def benchmark_optimized_serialize(data, iterations=20):
    """Benchmark optimized batch serialization via Fory."""
    fory = pyfory.Fory(ref=True)

    # Warmup
    for _ in range(3):
        fory.serialize(data)

    # Benchmark
    start = time.perf_counter()
    for _ in range(iterations):
        fory.serialize(data)
    end = time.perf_counter()

    return (end - start) / iterations * 1000


def benchmark_deserialize(data, iterations=20):
    """Benchmark deserialization."""
    fory = pyfory.Fory(ref=True)
    serialized = fory.serialize(data)

    # Warmup
    for _ in range(3):
        fory.deserialize(serialized)

    # Benchmark
    start = time.perf_counter()
    for _ in range(iterations):
        fory.deserialize(serialized)
    end = time.perf_counter()

    return (end - start) / iterations * 1000, len(serialized)


def main():
    print("=" * 60)
    print("Int64 Batch Serialization Benchmark")
    print("=" * 60)
    print(f"Cython enabled: {pyfory.ENABLE_FORY_CYTHON_SERIALIZATION}")
    print()

    sizes = [10_000, 100_000, 1_000_000]

    print(f"{'Size':<12} {'Original':<15} {'Optimized':<15} {'Speedup':<10} {'Deser':<12}")
    print("-" * 60)

    for size in sizes:
        data = list(range(size))

        original_time = benchmark_original_write(data)
        optimized_time = benchmark_optimized_serialize(data)
        deser_time, serialized_size = benchmark_deserialize(data)

        speedup = original_time / optimized_time

        print(
            f"{size:<12,} {original_time:<15.2f} {optimized_time:<15.2f} "
            f"{speedup:<10.2f}x {deser_time:<12.2f}"
        )

    print()
    print("Notes:")
    print("- Original: loop with individual write_varint64 calls (write only)")
    print("- Optimized: full serialize including header overhead")
    print("- Times are in milliseconds")


if __name__ == "__main__":
    main()

```
## Test plan

- [x] All existing tests pass (356 passed)
- [x] New unit tests for edge cases (large integers, negative values, empty list)
- [x] Encoding format compatible with original varint64+zigzag

### DIFF
--- a/python/pyfory/serialization.pyx
+++ b/python/pyfory/serialization.pyx
@@ -37,7 +37,8 @@ from pyfory.policy import DeserializationPolicy, DEFAULT_POLICY
 from pyfory.includes.libserialization cimport \
     (TypeId, TypeRegistrationKind, get_type_registration_kind,
      is_namespaced_type, is_type_share_meta,
-     Fory_PyBooleanSequenceWriteToBuffer, Fory_PyFloatSequenceWriteToBuffer)
+     Fory_PyBooleanSequenceWriteToBuffer, Fory_PyFloatSequenceWriteToBuffer,
+     Fory_PyInt64SequenceWriteToBuffer, Fory_PyInt64SequenceReadFromBuffer)
 
 from libc.stdint cimport int8_t, int16_t, int32_t, int64_t, uint64_t
 from libc.stdint cimport *


cpp/fory/python/pyfory.cc
```diff
@@ -58,4 +58,133 @@ int Fory_PyFloatSequenceWriteToBuffer(PyObject *collection, Buffer *buffer,
...
+    uint32_t bytes_written = WriteVarint64ZigZag(data + total_bytes, value);
+    total_bytes += bytes_written;
+  }
+
+  return total_bytes;
+}
+
+// Read varint64 with ZigZag decoding inline
+// Returns the number of bytes read
+static inline uint32_t ReadVarint64ZigZag(const uint8_t *arr, int64_t *result) {
```
chaokunyang: There is no buffer bound chekc, it will introduce a crash in malicouse input data

cpp/fory/python/pyfory.cc
```diff
@@ -58,4 +58,133 @@ int Fory_PyFloatSequenceWriteToBuffer(PyObject *collection, Buffer *buffer,
...
+  if (v < 0x100000000000000ULL) {
+    arr[7] = static_cast<uint8_t>(v >> 49);
+    return 8;
+  }
+  arr[7] = static_cast<uint8_t>((v >> 49) | 0x80);
+  arr[8] = static_cast<uint8_t>(v >> 56);
+  return 9;
+}
+
+Py_ssize_t Fory_PyInt64SequenceWriteToBuffer(PyObject *collection,
```
chaokunyang, line 113: `Fory_PyInt64SequenceWriteToBuffer` returns -1 and sets PyErr when an element is out of int64    range (or not an int), but the Cython caller only checks bytes_written >= 0 and doesn’t    propagate the exception. Because the extern declaration has no except clause, the error is    silently ignored and the header is written without advancing writer_index, corrupting the    stream for lists with large Python ints (e.g., >2^63-1). 

Please check bytes_written < 0 and raise/fall back to the slow path.

cpp/fory/python/pyfory.cc
```diff
@@ -58,4 +58,133 @@ int Fory_PyFloatSequenceWriteToBuffer(PyObject *collection, Buffer *buffer,
   }
   return 0;
 }
+
+// Write varint64 with ZigZag encoding inline
+// Returns number of bytes written
+static inline uint32_t WriteVarint64ZigZag(uint8_t *arr, int64_t value) {
```
chaokunyang, line 64: I'm refactoring fory cython Buffer to forward numeric read/write into c++ Buffer, you can use c++ Buffer.write_varint64 directly after I finsihed the refactor later

chaokunyang: WriteVarint64ZigZag should be removed now, python use same C++ `Buffer` for write/read ints

python/pyfory/collection.pxi
```diff
@@ -200,7 +200,7 @@ cdef class CollectionSerializer(Serializer):
             buffer.grow(<int32_t>max_size)
             # Try batch write, fall back to slow path on error (e.g., int overflow)
             try:
-                bytes_written = Fory_PyInt64SequenceWriteToBuffer(value, buffer.c_buffer.get(), buffer.writer_index)
+                bytes_written = Fory_PyInt64SequenceWriteToBuffer(value, &buffer.c_buffer, buffer.writer_index)
```
chaokunyang: do not pass `buffer.writer_index`, you can get it in `Fory_PyInt64SequenceWriteToBuffer` from `c_buffer`

python/pyfory/collection.pxi
```diff
@@ -215,7 +215,7 @@ cdef class CollectionSerializer(Serializer):
         if type(collection_) is list:
             # Pass buffer size for bounds checking
             bytes_read = Fory_PyInt64SequenceReadFromBuffer(
-                collection_, buffer.c_buffer.get(), buffer.reader_index, len_, buffer.size())
+                collection_, &buffer.c_buffer, buffer.reader_index, len_, buffer.size())
```
chaokunyang: ditto, only pass buffer, do not pass buffer fields

cpp/fory/python/pyfory.cc
pandalee99: fix

cpp/fory/python/pyfory.cc
pandalee99: this is consistent with the behavior of the original implementation and not a new issue introduced by this PR. That said, I have still made the fix for it.

cpp/fory/python/pyfory.cc
pandalee99: let me take a look.

python/pyfory/collection.pxi
pandalee99: done

Issue #3133, opened 2026-01-12
[JavaScript] add xlang tests to ensure protocol consistency between JavaScript and Java


chaokunyang:
### Feature Request

Currently we have Xlang tests for python/rust/c++/go:

- src/test/java/org/apache/fory/xlang/GoXlangTest.java
- src/test/java/org/apache/fory/xlang/CPPXlangTest.java
- src/test/java/org/apache/fory/xlang/RustXlangTest.java
- src/test/java/org/apache/fory/xlang/PythonXlangTest.java

We need to add JavaScriptXlangTest.java and align javascript implementation with current xlang_serialization_spec.md

theweipeng:
I’m starting work on this issue.

dheeraj12347:
Hi @theweipeng, @chaokunyang 👋
I noticed this issue is still open and there have not been any updates since your comment.
I’m very interested in contributing to Apache Fory’s JavaScript xlang support and would be happy to help with this.

My plan would be:
– Mirror the existing Xlang tests for Python/Rust/C++/Go under src/test/java/org/apache/fory/xlang/JavaScriptXlangTest.java.
– Compare the JavaScript implementation with the xlang serialization spec and Java reference behavior.
– Add/adjust JS code so xlang serialization is fully consistent with the spec, then update tests accordingly.

Would it be okay if I start working on this issue? If you already have work in progress, I’d also be happy to help review or contribute to that instead.

theweipeng:
> Hi [@theweipeng](https://github.com/theweipeng), [@chaokunyang](https://github.com/chaokunyang) 👋 I noticed this issue is still open and there have not been any updates since your comment. I’m very interested in contributing to Apache Fory’s JavaScript xlang support and would be happy to help with this.
> 
> My plan would be: – Mirror the existing Xlang tests for Python/Rust/C++/Go under src/test/java/org/apache/fory/xlang/JavaScriptXlangTest.java. – Compare the JavaScript implementation with the xlang serialization spec and Java reference behavior. – Add/adjust JS code so xlang serialization is fully consistent with the spec, then update tests accordingly.
> 
> Would it be okay if I start working on this issue? If you already have work in progress, I’d also be happy to help review or contribute to that instead.

Yes, It's okay for you to start working on this issue. 
Can you please compare JavaScript implementation with the xlang serialization spec base types in (https://fory.apache.org/docs/next/specification/xlang_serialization_spec#data-types) first, and implement this part.

theweipeng:
[...]

PR #3234, opened 2026-01-27
feat(JavaScript): impl xlang writer

theweipeng:
## Why?
1. impl xlang writer and reader, now we can ser/des object between java/js.
2. Replace the old function by `xwrite xwriteRef xwriteNoRef writeRefOrNull writeClassInfo` series functions to align the code styles in java.
3. The code style of `Reader series function`  will be replaced in next PR.

## Related issues
#3133 

### DIFF
--- a/javascript/jest.config.js
+++ b/javascript/jest.config.js
@@ -32,7 +32,6 @@ module.exports = {
   testPathIgnorePatterns : (
     !process.env["DATA_FILE"] ? ["test/crossLanguage.test.ts"] : []
   ).concat([
-    "set.test.ts", // todo remove
   ]),
   transform: {
     '\\.ts$': ['ts-jest', {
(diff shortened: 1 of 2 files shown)


javascript/packages/fory/lib/gen/any.ts
```diff
@@ -21,154 +21,118 @@ import { TypeInfo } from "../typeInfo";
...
-              write,
-              writeInner,
-              fixedSize: ${this.getFixedSize()},
-              needToWriteRef() {
-                throw new Error("//todo unreachable code");
-              }
-            };
-        }
-        `;
+  xread(assignStmt: (v: string) => string, refState: string): string {
```
chaokunyang: Since javascript will have xlang mode only, we can just rename to `read/write` without `x` prefix

chaokunyang (approved):
LGTM

PR #3244, opened 2026-01-29
feat: [Compiler] format generated code by header and line break #3233

### LINKED ISSUE #3233: [Compiler] format generated code by header and line break
### Feature Request

format generated code by header and line break

_No response_

- add a comment to begin of generted file: generated by apache fory compiler
- for too long line, add break, limit each line max width to 80, be causious to python indent newline and c++ macro

_No response_

### Additional context

#3099

lavisha4193:
## Why?

[Compiler] format generated code by header and line break
## What does this PR do?

[Compiler] format generated code by header and line break

## Related issues

Is there any related issue? If this PR closes them you say say fix/closes:

- #3233
- Fixes #3233

### DIFF
--- a/compiler/fory_compiler/generators/go.py
+++ b/compiler/fory_compiler/generators/go.py
@@ -189,7 +189,7 @@ class GoGenerator(BaseGenerator):
         PrimitiveKind.UINT64: "uint64",
         PrimitiveKind.VAR_UINT64: "uint64",
         PrimitiveKind.TAGGED_UINT64: "uint64",
-        PrimitiveKind.FLOAT16: "float32",
+        PrimitiveKind.FLOAT16: "float16.Float16",
         PrimitiveKind.FLOAT32: "float32",
         PrimitiveKind.FLOAT64: "float64",
         PrimitiveKind.STRING: "string",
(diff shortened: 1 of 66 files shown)


compiler/fory_compiler/generators/python.py
```diff
@@ -141,6 +141,64 @@ class PythonGenerator(BaseGenerator):
         PrimitiveKind.ANY: "None",
     }
 
+    def wrap_line(
```
chaokunyang, line 144: Does this func handle cpp macro registeration and Python indent right? 

 We don't have semantics ath this line level. Maybe it will be more safe If we add line breaker when generating code in each language.

compiler/fory_compiler/generators/go.py
```diff
@@ -199,6 +199,49 @@ def message_has_unions(self, message: Message) -> bool:
         PrimitiveKind.ANY: "any",
     }
 
+    def wrap_line(
```
chaokunyang, line 202: I prefer break line in each language at semantic level when generating code. Otherwise it's every easy to add wrong line breaker

lavisha4193: I have updated the code withbreak line instead of generic function

compiler/fory_compiler/generators/python.py
lavisha4193: yes it handles. Yet removed wrap function and replaced with  break-line

PR #3254, opened 2026-02-03
feat(java): add float support

### LINKED ISSUE #3205: [Java] add float16 support to java
### Feature Request

Add full IEEE 754 half-precision floating point (`float16`, binary16) support to Fory/FDL, including a complete **Java** runtime implementation that uses a strong type **`Float16`** (internally storing **`short` bits**), plus exhaustive tests for conversion/rounding/NaN/subnormal semantics.

The main blocker is Java:
- Java has no native IEEE 754 binary16 (`float16`) primitive.
- We need exact IEEE 754 binary16 behavior and cross-language consistency.
- Java has no operator overloading, so “float32-like usability” must be provided via explicit methods.
- For type safety and clarity, we want **only a single strong type** (`Float16`) to appear in public APIs and generated code (no raw `short` parameters).

#### 1) FDL / Type System
  - message fields
  - repeated fields

- Encode `float16` as **2 bytes** representing the raw IEEE 754 binary16 bit pattern.

#### 3) Java Runtime (core requirement): `Float16` strong type only
Provide a public strong type `Float16`. **All runtime APIs must accept/return `Float16` only** (no `short` parameters in public APIs).

##### 3.1 Type definition
- Provide an immutable, compact value type:
  - `public final class Float16`
  - internal storage: `private final short bits;`
- Construction should be controlled (private constructor), with explicit factories:
  - `public static Float16 fromBits(short bits)`
  - `public short toBits()` (bit access is allowed, but callers still pass `Float16` to APIs)

- `public static Float16 fromFloat(float f)`
  - Convert float32 -> binary16 using IEEE 754 rules and **round-to-nearest, ties-to-even**.
  - Must correctly handle:
    - ±Inf
    - ±0 (preserve sign)
    - normalized values
    - overflow -> ±Inf
    - underflow -> subnormal/±0
- `public float toFloat()`
  - Convert binary16 -> float32 (exact for all half values).

All methods operate on `Float16`:
- `public boolean isNaN()`
- `public boolean isInfinite()` and optionally `public boolean isInfinite(int sign)` (sign: +1/-1/0)
- `public boolean isZero()` (treat +0/-0 as zero)
- `public boolean signbit()`
- `public boolean isSubnormal()`
- `public boolean isNormal()`
- `public boolean isFinite()`

Since Java has no operator overloading, provide instance methods returning `Float16`:

- `public Float16 add(Float16 other)`
- `public Float16 sub(Float16 other)`
- `public Float16 mul(Float16 other)`
- `public Float16 div(Float16 other)`
- `public Float16 neg()`
- `public Float16 abs()`

Optional math parity if needed by users:
- `sqrt`, `min`, `max`, `copySign`, `floor`, `ceil`, `trunc`, `round`, `roundToEven`

Implementation rule for arithmetic (unless full half-FPU emulation is desired):
- compute in float32 and round back to half each operation:
  - `return Float16.fromFloat(this.toFloat() op other.toFloat());`

##### 3.5 Comparisons
Provide IEEE-consistent comparisons as instance/static methods using `Float16` only:

- `public boolean equalsValue(Float16 other)`
  - NaN != NaN
  - +0 == -0
- `public boolean lessThan(Float16 other)` (if NaN involved => false)
- `public boolean lessThanOrEqual(Float16 other)`
- `public boolean greaterThan(Float16 other)`
- `public boolean greaterThanOrEqual(Float16 other)`
- `public static int compare(Float16 a, Float16 b)`
  - Define NaN behavior explicitly (unordered). If a total ordering is needed, add:
    - `public static int totalOrderCompare(Float16 a, Float16 b)` (optional)

Note: `Object.equals()` / `hashCode()` should be defined carefully:
- Option A: Make `equals()` bitwise (strict bits equality), and provide `equalsValue()` for IEEE numeric equality.
- Option B: Make `equals()` numeric (IEEE rules, NaN != NaN, +0 == -0), but then hashing/collections semantics become tricky.
Please choose one and document it (recommendation: bitwise `equals()` + separate numeric methods).

##### 3.6 Formatting / parsing / interop
- `@Override public String toString()` (format via `Float.toString(toFloat())`)
- Optional:
  - `public static Float16 parse(String s)` (parse as float, then convert to half)
  - `Text`/`JSON` marshal/unmarshal helpers if Fory uses them

#### 4) Java Codegen requirement
- Generated Java fields for `float16` must use **`Float16`** (not `short`).
- Repeated `float16` should use `List<Float16>` (or an existing generated container type), unless you have a specialized primitive container strategy for wrappers.
- Map values for `float16` should be `Map<K, Float16>`.

#### 5) Compiler / Reflection Integration
- Ensure reflection/dynamic serialization recognizes Java `Float16` as the `float16` primitive (distinct from `short` / `int16`).
- Clarify schema evolution:

1. **Conversion tests (Java)**
   - ±0, ±Inf, NaN
   - max finite 65504
   - min normal 2^-14
   - min subnormal 2^-24
   - overflow -> Inf, underflow -> subnormal/0
     - `h = Float16.fromBits(bits);`
     - `h2 = Float16.fromFloat(h.toFloat());`

   - Ensure wire output matches expected 16-bit patterns for known values (via `Float16.toBits()`).

1) Store `float16` as `float` in Java and convert to `Float16` only during serialization.

2) Expose raw `short` in generated code and APIs, and only provide helper functions on bits.
- Rejected: loses type safety and makes user code error-prone; we want `Float16` everywhere.

3) Use third-party float16 libraries.

### Additional context

#3099

mengnankkkk:
## Related issues
Close #3205 

### DIFF
--- a/compiler/fory_compiler/generators/java.py
+++ b/compiler/fory_compiler/generators/java.py
@@ -1192,12 +1192,14 @@ class JavaGenerator(BaseGenerator):
                 if not element_optional and not element_ref:
                     if kind == PrimitiveKind.FLOAT16:
                         imports.add(
-                            "org.apache.fory.collection." + self.PRIMITIVE_LIST_MAP[kind]
+                            "org.apache.fory.collection."
+                            + self.PRIMITIVE_LIST_MAP[kind]
                         )
                         return
                     if kind in self.PRIMITIVE_LIST_MAP and not self.java_array(field):
                         imports.add(
-                            "org.apache.fory.collection." + self.PRIMITIVE_LIST_MAP[kind]
+                            "org.apache.fory.collection."
+                            + self.PRIMITIVE_LIST_MAP[kind]
                         )
                         return
                     if kind in self.PRIMITIVE_ARRAY_MAP:


java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java
```diff
@@ -881,6 +881,52 @@ private void readFloat64BySwapEndian(MemoryBuffer buffer, double[] values, int n
     }
   }
 
+  public static final class Float16ArraySerializer
+      extends PrimitiveArraySerializer<org.apache.fory.type.Float16[]> {
```
chaokunyang: Could you also implement `Float16List` under `org.apache.fory.collection` package just like we did for `Uint16List`?

And use `short[]` as storage in `Float16List`

java/fory-core/src/main/java/org/apache/fory/type/Float16.java
```diff
@@ -0,0 +1,338 @@
...
+  public byte byteValue() {
+    return (byte) floatValue();
+  }
+
+  @Override
+  public short shortValue() {
+    return (short) floatValue();
+  }
+
+  public boolean equalsValue(Float16 other) {
```
chaokunyang:
```suggestion
  public boolean isNumericEqual(Float16 other) {
```

And could you add some java docs for public methods of this class

java/fory-core/src/main/resources/META-INF/native-image/org.apache.fory/fory-core/reflection-config.json
```diff
@@ -5,5 +5,24 @@
...
     "allDeclaredMethods": true,
     "allPublicMethods": true
+  },
+  {
+    "name": "org.apache.fory.type.Float16",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true,
+    "allDeclaredFields": true,
```
chaokunyang: do we need to expose fields?

java/fory-core/src/main/resources/META-INF/native-image/org.apache.fory/fory-core/serialization-config.json
```diff
@@ -16,5 +16,6 @@
   { "name": "java.lang.NullPointerException" },
   { "name": "java.lang.IndexOutOfBoundsException" },
   { "name": "java.lang.ArrayIndexOutOfBoundsException" },
-  { "name": "java.io.IOException" }
+  { "name": "java.io.IOException" },
+  { "name": "org.apache.fory.type.Float16" }
```
chaokunyang: this is also unnecessary. Float16Serializer should be registerd in fory as an internal serializer

java/fory-core/src/main/resources/META-INF/native-image/org.apache.fory/fory-core/reflection-config.json
```diff
@@ -5,5 +5,24 @@
...
+    "allPublicFields": true
+  },
+  {
+    "name": "org.apache.fory.collection.Float16List",
```
chaokunyang: Could you share the reason that you think it's necessary to register `Float16List/Float16`

java/fory-core/src/main/java/org/apache/fory/type/DispatchId.java
```diff
@@ -44,20 +44,21 @@ public class DispatchId {
...
-  public static final int VAR_UINT64 = 17;
-  public static final int TAGGED_UINT64 = 18;
-  public static final int EXT_UINT8 = 19;
-  public static final int EXT_UINT16 = 20;
-  public static final int EXT_UINT32 = 21;
-  public static final int EXT_VAR_UINT32 = 22;
-  public static final int EXT_UINT64 = 23;
-  public static final int EXT_VAR_UINT64 = 24;
-  public static final int STRING = 25;
+  public static final int FLOAT16 = 12;
```
chaokunyang: This ID is wrong, here is current id def in main branch:

```
  /** float8: an 8-bit floating point number. */
  public static final int FLOAT8 = 16;

  /** float16: a 16-bit floating point number. */
  public static final int FLOAT16 = 17;

  /** bfloat16: a 16-bit brain floating point number. */
  public static final int BFLOAT16 = 18;
```

Same for Float16 array

java/fory-core/src/main/resources/META-INF/native-image/org.apache.fory/fory-core/reflection-config.json
```diff
@@ -5,5 +5,8 @@
     "allPublicConstructors": true,
     "allDeclaredMethods": true,
     "allPublicMethods": true
+  },
```
chaokunyang: This is syntax error, please fix it

java/fory-core/src/test/java/org/apache/fory/serializer/Float16SerializerTest.java
```diff
@@ -0,0 +1,263 @@
...
+    assertNull(result.nullableField);
+  }
+
+  @Test
+  public void testFloat16SpecialValuesInStruct() {
+    Fory fory = Fory.builder().withLanguage(Language.JAVA).requireClassRegistration(false).build();
+
+    @Data
+    @AllArgsConstructor
+    class StructWithSpecialValues {
```
chaokunyang, line 298: Please also create another tests to struct with Float16List as fields and also test Float16List as top level objects for serialization

java/fory-core/src/main/java/org/apache/fory/type/DispatchId.java
```diff
@@ -49,15 +49,16 @@ public class DispatchId {
...
+  public static final int FLOAT16 = 17;
+  public static final int VAR_UINT64 = 18;
+  public static final int TAGGED_UINT64 = 19;
+  public static final int EXT_UINT8 = 20;
+  public static final int EXT_UINT16 = 21;
+  public static final int EXT_UINT32 = 22;
+  public static final int EXT_VAR_UINT32 = 23;
+  public static final int EXT_UINT64 = 24;
+  public static final int EXT_VAR_UINT64 = 25;
+  public static final int STRING = 26;
```
chaokunyang, line 61: java/fory-core/src/main/java/org/apache/fory/serializer/FieldGroups.java also needs update to clarify Float16 as build-in fields and grouped into primitive/boxed groups based whethther it's nullable.

java/fory-core/src/main/java/org/apache/fory/builder/BaseObjectCodecBuilder.java also needs update to handle Flaot16 specially, you can start from `serializePrimitiveField`

java/fory-core/src/main/java/org/apache/fory/serializer/AbstractObjectSerializer.java may also needs updates

chaokunyang:
Could you also address those issues?

1. **XLANG support for new `Float16` types is broken at runtime.**  
   `Float16`/`Float16List` serializers are added (`PrimitiveSerializers.java`, `PrimitiveListSerializers.java`), but internal xlang type registration is missing in resolver bootstrap (`ClassResolver.java`). In xlang mode it falls back to unsupported-class path and fails at deserialize time.
   - `java/fory-core/src/main/java/org/apache/fory/serializer/PrimitiveSerializers.java:381`
   - `java/fory-core/src/main/java/org/apache/fory/serializer/collection/PrimitiveListSerializers.java:595`
   - `java/fory-core/src/main/java/org/apache/fory/resolver/ClassResolver.java:262`
   - `java/fory-core/src/main/java/org/apache/fory/resolver/ClassResolver.java:295`
   - `java/fory-core/src/main/java/org/apache/fory/resolver/ClassResolver.java:1305`

2. **`Float16` subnormal decode is numerically wrong.**  
   In `float16BitsToFloat`, subnormal conversion uses an incorrect shift/exponent formula.
   - `java/fory-core/src/main/java/org/apache/fory/type/Float16.java:169`
   - `java/fory-core/src/main/java/org/apache/fory/type/Float16.java:173`  
   Repro: `Float16.fromBits((short)0x0001).floatValue()` returns `1.7881393E-7`, expected `5.9604645E-8` (`2^-24`).

3. **`Float16[]` serializer crashes on legal arrays containing `null`.**  
   The array fast path dereferences element directly (`value[i].toBits()`), so `new Float16[]{Float16.ONE, null}` throws NPE.
   - `java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java:901`

4. **Type-id mapping regression for existing float kinds.**  
   `Types.getClassForTypeId` now maps `FLOAT8` to `Float16`, and `BFLOAT16` mapping was removed (falls through to `null`).
   - `java/fory-core/src/main/java/org/apache/fory/type/Types.java:450`
   - `java/fory-core/src/main/java/org/apache/fory/type/Types.java:452`

5. **Generated Java `equals`/`hashCode` contract can break for `float16`.**  
   Generator uses `equalsValue` for `FLOAT16` equality, but default hash path remains `Objects.hash(...)`. `equalsValue` treats `+0` and `-0` as equal, while `Float16.hashCode` is bit-based.
   - `compiler/fory_compiler/generators/java.py:1384`
   - `compiler/fory_compiler/generators/java.py:1467`
   - `java/fory-core/src/main/java/org/apache/fory/type/Float16.java:270`
   - `java/fory-core/src/main/java/org/apache/fory/type/Float16.java:334`

java/fory-core/src/main/java/org/apache/fory/serializer/Float16Serializer.java
```diff
@@ -39,12 +39,10 @@ public Float16 read(MemoryBuffer buffer) {
     return Float16.fromBits(buffer.readInt16());
   }
 
-  @Override
   public void xwrite(MemoryBuffer buffer, Float16 value) {
```
chaokunyang: remove it, we don' have xwrite/xread methods any more

java/fory-core/src/main/java/org/apache/fory/serializer/ArraySerializers.java
```diff
@@ -868,6 +869,146 @@ private void readFloat64BySwapEndian(MemoryBuffer buffer, double[] values, int n
...
+        for (int i = 0; i < length; i++) {
+          buffer.writeInt16(value[i].toBits());
+        }
+      }
+    }
+
+    private void writeNullable(MemoryBuffer buffer, Float16[] value, int length) {
+      int bitmapSize = (length + 7) >>> 3;
+      int payloadSize = Integer.BYTES + bitmapSize + length * 2;
+      int encodedSize = (payloadSize << 1) | NULLABLE_ENCODING_FLAG;
```
chaokunyang: This changes the FLOAT16_ARRAY wire layout when nulls are present, which will break cross-language readers that expect the normal length+payload format for type 53.

In xlang mode, other runtimes decode FLOAT16_ARRAY as plain [byte_length][2-byte elements]. Here we switch to an alternate encoded-size scheme for nullable arrays, so non-Java runtimes will misread the payload. Could we keep FLOAT16_ARRAY encoding stable and handle nullable cases another way (for example, fallback to LIST encoding when null is present)?

java/fory-core/src/main/java/org/apache/fory/type/DispatchId.java
```diff
@@ -49,19 +49,23 @@ public class DispatchId {
...
-  public static final int VAR_UINT64 = 17;
-  public static final int TAGGED_UINT64 = 18;
-  public static final int EXT_UINT8 = 19;
-  public static final int EXT_UINT16 = 20;
-  public static final int EXT_UINT32 = 21;
-  public static final int EXT_VAR_UINT32 = 22;
-  public static final int EXT_UINT64 = 23;
-  public static final int EXT_VAR_UINT64 = 24;
-  public static final int STRING = 25;
+  public static final int FLOAT16 = 17;
```
chaokunyang, line 52: Since FLOAT16 is now a basic dispatch id, FieldSkipper also needs a skip branch for it.

Right now FieldSkipper has no DispatchId.FLOAT16 case. In compatibility reads, skipping an unknown Float16 field will throw Unexpected basic dispatchId instead of advancing by 2 bytes. Please add FLOAT16 to the 2-byte skip branch.

java/fory-core/src/main/java/org/apache/fory/serializer/Serializers.java
```diff
@@ -570,6 +571,7 @@ public static void registerDefaultSerializers(Fory fory) {
     resolver.registerInternalSerializer(URI.class, new URISerializer(fory));
     resolver.registerInternalSerializer(Pattern.class, new RegexSerializer(fory));
     resolver.registerInternalSerializer(UUID.class, new UUIDSerializer(fory));
+    resolver.registerInternalSerializer(Float16.class, new Float16Serializer(fory));
```
chaokunyang: Float16 gets registered twice (primitive serializer path and general serializers path).

This overrides the earlier registration and makes it harder to reason about which serializer is intended. It would be cleaner to keep a single registration point for Float16.

compiler/fory_compiler/generators/java.py
```diff
@@ -1458,7 +1460,11 @@ def generate_equals_method(self, message: Message) -> List[str]:
                     )
                 elif isinstance(field.field_type, PrimitiveType):
                     kind = field.field_type.kind
-                    if kind in (PrimitiveKind.FLOAT32,):
+                    if kind in (PrimitiveKind.FLOAT16,):
+                        comparisons.append(
+                            f"({field_name} == null ? that.{field_name} == null : (that.{field_name} != null && {field_name}.equalsValue(that.{field_name})))"
```
chaokunyang: Generated equals/hashCode for float16 now use numeric semantics, but Float16.equals/hashCode are bitwise.

With this change, +0 and -0 compare equal in generated messages, and NaN is not equal to NaN, while Float16 itself uses raw-bit equality. That mismatch is surprising in user code. Can we align generated message behavior with Float16.equals/hashCode (or make the rule explicit and consistent across runtimes)?

java/fory-core/src/main/java/org/apache/fory/collection/Float16List.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  public boolean add(short bits) {
+    ensureCapacity(size + 1);
+    array[size++] = bits;
+    modCount++;
+    return true;
+  }
+
+  public boolean add(float value) {
```
chaokunyang, line 107: Float16List primitive methods still allocate Float16 objects on the hot path.

add(float), set(float), and getFloat(int) all go through Float16.valueOf/fromBits, so these paths allocate even though this is a primitive-style container. A static bits conversion helper would keep these methods allocation-free.

PR #3293, opened 2026-02-05
fix(go): prevent panic in readUTF16LE with odd byte counts

jonyoder:
## What does this PR do?

Fixes a panic in `readUTF16LE` when the byte count is odd (malformed UTF-16 data).

## Problem

When `readUTF16LE` receives an odd byte count, the function panics with:
```
panic: runtime error: index out of range [N] with length N
```

This occurs because the loop iterates up to `byteCount` but tries to access `data[i+1]` on the last iteration when `i = byteCount - 1` (for odd `byteCount`).

### Steps to Reproduce

```go
func TestReadUTF16LE_OddByteCount(t *testing.T) {
    // Data: 5 bytes where only first 4 form valid UTF-16
    data := []byte{0x48, 0x00, 0x69, 0x00, 0xFF}
    buf := NewByteBuffer(data)
    err := &Error{}

    // This panics without the fix
    result := readUTF16LE(buf, 5, err)
}
```

### Root Cause

In `go/fory/string.go`, line 89-90:
```go
for i := 0; i < byteCount; i += 2 {
    u16s[i/2] = uint16(data[i]) | uint16(data[i+1])<<8  // Panics when i+1 >= len(data)
}
```

## Solution

set error when we get odd utf16 bytes

## Tests Added

- `TestReadUTF16LE_OddByteCount`: Verifies no panic with odd byte count
- `TestReadUTF16LE_SingleByte`: Verifies empty string for single byte input
- `TestReadUTF16LE_EvenByteCount`: Verifies normal case still works
- `TestReadUTF16LE_EmptyBuffer`: Verifies empty input handling
- `TestReadUTF16LE_SurrogatePair`: Verifies emoji/surrogate pair decoding

## Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/apache/fory/blob/main/CONTRIBUTING.md)
- [x] All tests pass (`go test -v ./...` in `go/fory`)
- [x] License headers are correct (hawkeye check passed)
- [x] PR title follows conventional commits format

### DIFF
--- a/go/fory/string.go
+++ b/go/fory/string.go
@@ -81,16 +81,20 @@ func readLatin1(buf *ByteBuffer, size int, err *Error) string {
 }
 
 func readUTF16LE(buf *ByteBuffer, byteCount int, err *Error) string {
+	if byteCount&1 != 0 {
+		err.SetError(InvalidUTF16StringError(byteCount))
+		return ""
+	}
+
 	data := buf.ReadBinary(byteCount, err)
+	if err.HasError() {
+		return ""
+	}
 
 	// Reconstruct UTF-16 code units
-	// Note: byteCount should always be even for valid UTF-16, but handle odd counts
-	// gracefully to avoid index out of range panics from malformed data.
 	charCount := byteCount / 2
 	u16s := make([]uint16, charCount)
-	// Only iterate up to the last complete pair (round down to even boundary)
-	evenByteCount := byteCount &^ 1
-	for i := 0; i < evenByteCount; i += 2 {
+	for i := 0; i < byteCount; i += 2 {
 		u16s[i/2] = uint16(data[i]) | uint16(data[i+1])<<8
 	}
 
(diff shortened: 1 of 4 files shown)


go/fory/string.go
```diff
@@ -84,9 +84,13 @@ func readUTF16LE(buf *ByteBuffer, byteCount int, err *Error) string {
 	data := buf.ReadBinary(byteCount, err)
 
 	// Reconstruct UTF-16 code units
+	// Note: byteCount should always be even for valid UTF-16, but handle odd counts
+	// gracefully to avoid index out of range panics from malformed data.
 	charCount := byteCount / 2
 	u16s := make([]uint16, charCount)
-	for i := 0; i < byteCount; i += 2 {
+	// Only iterate up to the last complete pair (round down to even boundary)
+	evenByteCount := byteCount &^ 1
```
chaokunyang: Could you change to set an error on `err *Error` instead of ignore that last byte?

jonyoder: Sorry about the delay. I stepped back to try to figure out the exact source of this issue in our case, and I've been having trouble reproducing it again! I see you've pushed another commit; nice to see that this will be handled correctly either way!

chaokunyang (approved):
LGTM

PR #3319, opened 2026-02-10
feat(JavaScript): Align testcases

theweipeng:
## Why?
1. More testcases were passed, current progress 29/39

## What does this PR do?
1. Support `TAGGED_INT64` and `TAGGED_UINT64`
2. Fix float16 decode when the number is NaN or Infinity

## Related issues
#3133 

### DIFF
--- a/javascript/packages/fory/lib/gen/builder.ts
+++ b/javascript/packages/fory/lib/gen/builder.ts
@@ -197,10 +197,6 @@ class BinaryWriterBuilder {
     return `${this.holder}.int16(${v})`;
   }
 
-  float16(v: number | string) {
-    return `${this.holder}.float16(${v})`;
-  }
-
   varInt32(v: number | string) {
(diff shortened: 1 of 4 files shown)


javascript/test/crossLanguage.test.ts
```diff
@@ -1483,23 +1483,21 @@ describe("bool", () => {
   });
 
   test("test_unsigned_schema_consistent_simple", () => {
-    if (Boolean("1")) { return; }
     const fory = new Fory({
-      mode: Mode.Compatible
+      mode: Mode.SchemaConsistent
```
chaokunyang, line 1487: Could we change to compatible bool field in a future PR and remove `Mode` class

theweipeng: > hange to comp

OK, I will change it to bool field next PR

PR #3333, opened 2026-02-13
feat(dart): add schema evolution (TypeDef meta share) support

### LINKED ISSUE #3290: [Dart] add schema evolution (TypeDef meta share) support
### Feature Request

Implement **schema evolution mode** (TypeDef meta share + compatible struct) in the **Dart** runtime to match the xlang Type Meta specification. Current Dart implementation does not support the schema evolution pathway described in the spec, which breaks compatibility with other languages when `compatible`/meta-share is enabled.

When other languages serialize with schema evolution enabled (TypeDef/meta share), Dart fails to decode because it does not implement the TypeDef/TypeMeta read/write flow. This blocks cross-language compatibility for evolving schemas.

Implement the **Type Meta + TypeDef** encoding/decoding flow in Dart according to the spec, including:

#### 1) Type ID & Named Types handling
- Read/write type IDs as **varuint32 (small7)**.
- Support `ENUM`, `STRUCT`, `COMPATIBLE_STRUCT`, `EXT`, `TYPED_UNION` (user-registered) plus `NAMED_*` types when unregistered.
- Ensure named types embed namespace + type name (or shared TypeDef when meta share is enabled).

#### 2) Meta Share (streaming TypeDef)
- Implement the shared TypeDef cache with the marker encoding:
  - `marker = (index << 1) | flag`
  - `flag = 0` => new TypeDef follows
  - `flag = 1` => reference cached TypeDef
- Write algorithm: lookup class in meta context map, emit reference or new TypeDef bytes.
- Read algorithm: parse marker, load cached TypeDef or read + cache new TypeDef.

#### 3) TypeDef decoding/encoding
TypeDef format:

```
| 8-byte global header | [optional size varuint] | TypeDef body |
```

- Global header (uint64 LE):
  - low 8 bits: meta size (0xFF + extra size varuint if needed)
  - bit 8: HAS_FIELDS_META
  - bit 9: COMPRESS_META (decompress before parsing)
  - bits 10-13 reserved
  - high 50 bits: hash of TypeDef body
- TypeDef body:
  - meta header byte: `num_fields`, `REGISTER_BY_NAME`
  - type spec: namespace+type_name or type_id
  - field list with field header, field type info, and field name bytes

#### 4) Field metadata parsing
- Field header bits:
  - bits 6-7: name encoding (UTF8 / ALL_TO_LOWER_SPECIAL / LOWER_UPPER_DIGIT_SPECIAL / TAG_ID)
  - bits 2-5: size (name length-1 or tag id); if 0b1111, read extra varuint size
  - bit 1: nullable flag
  - bit 0: reference tracking flag
- Field type info:
  - top-level type as varuint32
  - nested LIST/SET element type encoded with `(nested_type_id << 2) | (nullable << 1) | tracking_ref`
  - MAP key/value types encoded similarly
  - one-dimensional primitive arrays use `*_ARRAY` type IDs; others use LIST
- Field name encoding:
  - TAG_ID: no name bytes
  - otherwise write meta string name bytes
  - xlang uses `snake_case` conversion before encoding

#### 5) Compatibility behavior
- Decoders must match fields by **name/tag ID**, not position.
- Keep deterministic TypeDef generation (stable field order) for cross-language match.

#### 6) Tests
- Round-trip tests for compatible struct with meta share enabled.
- Decode TypeDef bytes generated by other languages (cross-language golden tests).
- Named vs registered type cases.
- Meta share cache correctness (reference reuse).

- Only supporting strict type-id registration without TypeDef meta.
  - Rejected: breaks schema evolution and cross-language compatibility.

### Additional context

Spec reference: **Type Meta / TypeDef** section (xlang serialization). Dart runtime should match other languages’ behavior so compatible mode works end-to-end.

arjav1528:
## Why?

When other languages serialize with schema evolution enabled (TypeDef/meta share), Dart could not decode because it did not implement the TypeDef/TypeMeta read/write flow. This blocked cross-language compatibility for evolving schemas (see #3290).

## What does this PR do?

- **TypeDef / FieldDef and body parsing**: Adds `FieldDef` and `RemoteFieldType` (Simple, ListSet, Map) in `meta/field_def.dart`; extends TypeResolverImpl to parse the full TypeDef body (field header, type info, name bytes or TAG_ID) and to use varuint32 small7 for type IDs in the body and in type meta.
- **Meta share cache and context**: Read cache stores `(TypeInfo, List<FieldDef>)` per index; `DeserializationContext` gets `currentRemoteFieldDefs`; `readTypeInfo` sets it when returning from shared type meta (new or ref); `resetReadContext` clears cache and `currentRemoteFieldDefs`.
- **Compatible-struct read by remote field order**: When `compatible` is true and `currentRemoteFieldDefs` is set, `ClassSerializer` uses `_readByRemoteFieldOrder`: iterates remote field defs, reads values by remote type (including nested LIST/SET/MAP), assigns to local fields by name (snake_case) or by index; supports TAG_ID for future tag-based matching.
- **Tests**: Round-trip with compatible + ref; meta share cache (single marker for list element type); TypeDef format round-trip; named compatible (typename) and compatible by user type id; cross-language golden test added but skipped with a note for future integration.

## Related issues

- Fixes #3290

  No new public APIs; internal types (`FieldDef`, `RemoteFieldType`) are in `src/meta` and used by the resolver/serializer.
  No. Dart now correctly implements the existing xlang Type Meta / TypeDef format so that Dart can decode data produced by other languages with compatible + meta share enabled.


arjav1528:
@chaokunyang could you please help me with the PR tests

PR #3336, opened 2026-02-13
feat(dart): add float16 to dart

### LINKED ISSUE #3209: [Dart] add float16 to dart
### Feature Request

Add full IEEE 754 half-precision floating point (`float16`, binary16) support to Fory/FDL, including a complete **Dart** runtime implementation that uses a strong type **`Float16`** (internally storing **`int` bits limited to 16 bits**), plus exhaustive tests for conversion/rounding/NaN/subnormal semantics.

Dart does not have a native IEEE 754 binary16 (`float16`) primitive type. Dart `double` is 64-bit IEEE 754, and while we can compute in `double`, we still need:
- an exact 16-bit binary16 representation,
- correct IEEE conversion semantics (round-to-nearest, ties-to-even),
- a single strong type `Float16` used in generated code and runtime APIs (no raw `int` parameters),
- stable, cross-language wire compatibility.

#### 1) FDL / Type System
  - message fields
  - repeated fields

- Encode `float16` as **2 bytes** representing the raw IEEE 754 binary16 bit pattern (unsigned 16-bit).

#### 3) Dart Runtime (core requirement): `Float16` strong type only
Provide a public strong type `Float16`. **All runtime APIs must accept/return `Float16` only** (no passing raw 16-bit ints as public API).

##### 3.1 Type definition
- Provide an immutable value type:
  - `final class Float16 { final int _bits; ... }`
- Store raw bits in `_bits` as an `int` but always masked to 16 bits (`_bits & 0xFFFF`).
- Construction should be controlled with factories:
  - `factory Float16.fromBits(int bits)` (mask to 0xFFFF; validate range optionally)
  - `int toBits()` (returns 0..65535)
- Provide common constants (optional but helpful):
  - `Float16.positiveZero`, `Float16.negativeZero`
  - `Float16.positiveInfinity`, `Float16.negativeInfinity`
  - `Float16.nan` (canonical NaN if canonicalizing)

- `factory Float16.fromDouble(double x)`
  - Convert double (or float32-equivalent) -> binary16 using IEEE 754 rules and **round-to-nearest, ties-to-even**.
  - Must correctly handle:
    - ±Infinity
    - ±0 (preserve sign)
    - normalized values
    - overflow -> ±Infinity
    - underflow -> subnormal/±0
- `double toDouble()`
  - Convert binary16 -> double (exact representation for all half values).

Note: Dart has only `double`, so arithmetic will happen in `double` and then be rounded back to half via `fromDouble`.

All operate on `Float16`:
- `bool get isNaN`
- `bool get isInfinite` and optionally `bool isInfiniteSign(int sign)` (sign: +1/-1/0)
- `bool get isZero` (treat +0/-0 as zero)
- `bool get isNegative` (signbit)
- `bool get isSubnormal`
- `bool get isNormal`
- `bool get isFinite`

Dart allows operator overloading, but to match the “explicit method” style used for other languages, provide explicit methods (operators can be added optionally):

- `Float16 add(Float16 other)`
- `Float16 sub(Float16 other)`
- `Float16 mul(Float16 other)`
- `Float16 div(Float16 other)`
- `Float16 neg()`
- `Float16 abs()`

Optional operator sugar (nice-to-have, but not required):
- `operator +`, `-`, `*`, `/`, unary `-`

Implementation rule for arithmetic:
- compute in double and round back to half each operation:
  - `return Float16.fromDouble(this.toDouble() op other.toDouble());`

Optional math parity (if needed):
- `sqrt`, `min`, `max`, `copySign`, `floor`, `ceil`, `trunc`, `round`, `roundToEven`

##### 3.5 Comparisons + equality/hash policy (must be explicit)
IEEE floats are tricky due to NaN and signed zero. Please choose and document:

**Policy A (recommended): bitwise equality/hash**
- `operator ==` compares `_bits` exactly
- `hashCode` derived from `_bits`
- Provide IEEE numeric comparison helpers:
  - `bool equalsValue(Float16 other)` (NaN != NaN; +0 == -0)
  - `bool lessThan(Float16 other)` (NaN involved => false)
  - `int compareValue(Float16 a, Float16 b)` with defined NaN behavior

**Policy B: IEEE numeric equality**
- `==` uses IEEE numeric rules (NaN != NaN, +0 == -0)
- This can make hashing/collection semantics surprising unless carefully defined.

Regardless of policy, provide explicit comparison helpers:
- `bool lessThan(Float16 other)`
- `bool lessThanOrEqual(Float16 other)`
- `bool greaterThan(Float16 other)`
- `bool greaterThanOrEqual(Float16 other)`
- `static int compare(Float16 a, Float16 b)` (define NaN handling)
- Optional: `static int totalOrderCompare(Float16 a, Float16 b)` for deterministic sorting.

##### 3.6 Formatting / parsing
- `@override String toString()` (format via `toDouble()`)
- Optional:
  - `static Float16 parse(String s)` (parse as double then convert)

#### 4) Dart Codegen requirement
- Generated Dart fields for `float16` must use **`Float16`** (not `int`).
- Repeated `float16` should use `List<Float16>`.
- Map values for `float16` should be `Map<K, Float16>`.

#### 5) Compiler / Reflection Integration
- Ensure reflection/dynamic serialization recognizes Dart `Float16` as the `float16` primitive (distinct from integer types).
- Clarify schema evolution:

1. **Conversion tests (Dart)**
   - ±0, ±Infinity, NaN
   - max finite 65504
   - min normal 2^-14
   - min subnormal 2^-24
   - overflow -> Infinity, underflow -> subnormal/0
     - `h = Float16.fromBits(bits);`
     - `h2 = Float16.fromDouble(h.toDouble());`

   - Ensure wire output matches expected 16-bit patterns for known values (via `toBits()`).

1) Store `float16` as `double` in Dart and convert to `Float16` only during serialization.

2) Expose raw 16-bit `int` bits in generated code and APIs, and only provide helper functions on bits.
- Rejected: loses type safety and makes user code error-prone; we want `Float16` everywhere.

3) Use third-party packages for float16.

### Additional context

#3099

ayush00git:
## Why?

As requested in Issue #3209, Dart implementation was missing support for IEEE 754 half-precision floating-point (Float16) types that are already available in other Fory language implementations (Rust, Go, C++, Python). This PR adds complete Float16 support for Dart.

## What does this PR do?

This PR implements schema-based support for IEEE 754 Float16 types in the Fory Dart library:

### Added Types

- `float16`

#### 1. lib/src/datatype/float16.dart
Implemented the `Float16` class which wraps a 16-bit integer and handles IEEE 754 conversions (handling rounding, subnormals, signed zeros, NaN/Infinity):

```dart
final class Float16 extends FixedNum {
  final int _bits;
  // ...
  factory Float16(num value) => Float16.fromDouble(value.toDouble());
  double toDouble() => _bitsToDouble(_bits);
}
```

#### 2. lib/src/memory/
Added read/write support for 16-bit floats in `ByteReader` and `ByteWriter`:

```dart
void writeFloat16(Float16 value);
Float16 readFloat16();
```

#### 3. lib/src/serializer/
Registered `Float16Serializer` to handle serialization of `Float16` types in `PrimitiveTypeSerializer`:

```dart
typeToTypeInfo[Float16]!.serializer = Float16Serializer.cache.getSerializer(conf);
```

#### 4. lib/src/const/dart_type.dart
Added `FLOAT16` to `DartTypeEnum` for type inference and registry.

#### 5. test/cross_lang_test/
Updated cross-language tests to fix lint issues and build errors:
- Renamed enum constants in `xlang_test_defs.dart` and `xlang_test_main.dart` to follow Dart's `lowerCamelCase` convention (e.g., `VALUE_A` -> `valueA`).

### Usage Example

```dart
import 'package:fory/src/datatype/float16.dart'; 
import 'package:fory/src/memory/byte_writer.dart';

var val = Float16(1.5);
print(val.toDouble())

// Handling limits and edge cases
var max = Float16(65504.0);
var inf = Float16(65505.0);
var subnormal = Float16(0.0000000596046);

// Serialization
var bw = ByteWriter();
bw.writeFloat16(val);
```

## Related issues

- Closes #3209

- [x] **Does this PR introduce any public API change?**
  - **Yes**: Adds `Float16` class and corresponding `NumType.float16`.
  - **Yes**: Adds `readFloat16` / `writeFloat16` to memory interfaces.
  - **Yes**: Exports `Float16` for user consumption.
  
- [x] **Does this PR introduce any binary protocol compatibility change?**
  - **Yes**: Adds support for TypeId 17 (Float16).
  - **Cross-language compatible**: Uses the same TypeId values as other Fory implementations.

## Benchmark

N/A


ayush00git:
Hey @chaokunyang 
Have a look at the implementation and let me know your reviews

chaokunyang:
@ayush00git I removed dart/packages/fory-test/test/cross_lang_test/xlang_test_defs.dart, could you address conflict?

ayush00git:
> @ayush00git I removed dart/packages/fory-test/test/cross_lang_test/xlang_test_defs.dart, could you address conflict?

Hii @chaokunyang 
Have a look and let me know

dart/packages/fory-test/lib/entity/xlang_test_models.dart
```diff
@@ -157,10 +156,10 @@ class NullableComprehensiveCompatible {
 
 @foryEnum
 enum Color {
-  Green,
-  Red,
-  Blue,
-  White,
+  green,
```
chaokunyang, line 159: why you change to lower case?

ayush00git: > why you change to lower case?

becuase the dart linter tests enforces lowerCamelcase for constant names

chaokunyang (approved):
LGTM

PR #3344, opened 2026-02-16
fix(java): support TreeSet/TreeMap subclasses without Comparator constructor in SortedSet/SortedMapSerializer

### LINKED ISSUE #3343: [Java] SortedSet/SortedMapSerializer fails for TreeSet/TreeMap subclasses without Comparator constructor
### Search before asking

### Version

0.15.0

### Component(s)

Java

### Minimal reproduce step

**SortedSetSerializer:**

```java
import java.util.TreeSet;
import org.apache.fory.Fory;
import org.apache.fory.serializer.collection.CollectionSerializers.SortedSetSerializer;

public class ChildTreeSet extends TreeSet<String> {}

Fory fory = Fory.builder()
    .build();

// This throws UnsupportedOperationException at construction time
fory.registerSerializer(
    ChildTreeSet.class,
    new SortedSetSerializer<>(fory, ChildTreeSet.class));
```

**SortedMapSerializer:**

```java
import java.util.TreeMap;
import org.apache.fory.Fory;
import org.apache.fory.serializer.collection.MapSerializers.SortedMapSerializer;

public class ChildTreeMap extends TreeMap<String, String> {}

Fory fory = Fory.builder()
    .build();

// This throws UnsupportedOperationException at construction time
fory.registerSerializer(
    ChildTreeMap.class,
    new SortedMapSerializer<>(fory, ChildTreeMap.class));
```

Both `SortedSetSerializer` and `SortedMapSerializer` require subclasses to have an explicit `(Comparator)` constructor. Java constructors are not inherited, so a subclass with only a no-arg constructor (which delegates to `TreeSet()`/`TreeMap()`) fails even though the parent class has a `(Comparator)` constructor.

`SortedSetSerializer` and `SortedMapSerializer` should accept subclasses that have a no-arg constructor but no `(Comparator)` constructor, falling back to the no-arg constructor for instantiation.

### What did you see instead?

**SortedSetSerializer:**
```
java.lang.UnsupportedOperationException: java.lang.NoSuchMethodException:
  no such constructor: ChildTreeSet.<init>(Comparator)void/newInvokeSpecial
    at o.a.fory.serializer.collection.CollectionSerializers$SortedSetSerializer.<init>(CollectionSerializers.java:194)
```

**SortedMapSerializer:**
```
java.lang.UnsupportedOperationException: java.lang.NoSuchMethodException:
  no such constructor: ChildTreeMap.<init>(Comparator)void/newInvokeSpecial
    at o.a.fory.reflect.ReflectionUtils.getCtrHandle(ReflectionUtils.java)
    at o.a.fory.serializer.collection.MapSerializers$SortedMapSerializer.<init>(MapSerializers.java:120)
```

### Anything Else?

This also affects the default serialization path indirectly. `TreeSet`/`TreeMap` subclasses fall back to `JDKCompatibleCollectionSerializer` (because `TreeSet`/`TreeMap` have `readObject`/`writeObject`), which works but is inefficient. Users who try to optimize by explicitly registering `SortedSetSerializer`/`SortedMapSerializer` hit this error.

mandrean:
## Why?

`SortedSetSerializer` and `SortedMapSerializer` unconditionally require a `(Comparator)` constructor for all `TreeSet`/`TreeMap` subclasses. Java constructors are not inherited, so a `class ChildTreeSet extends TreeSet<String>` (or `class ChildTreeMap extends TreeMap<String, String>`) with only a no-arg constructor throws `UnsupportedOperationException` when registered with the corresponding serializer.

## What does this PR do?

Changes both `SortedSetSerializer` and `SortedMapSerializer` to try the `(Comparator)` constructor first, and fall back to the no-arg constructor if not found. At deserialization time, uses whichever constructor is available. The no-arg fallback uses natural ordering, which is the common case for `TreeSet`/`TreeMap` subclasses.

Only throws `UnsupportedOperationException` if neither constructor is found.

## Related issues

- Fixes #3343

No. Existing behavior is preserved for classes that have a `(Comparator)` constructor. Classes that previously threw `UnsupportedOperationException` now work.

No. The write path is unchanged (size + comparator ref), and the read path still reads the same bytes in the same order. Only the post-read instantiation strategy differs (which constructor is used). Wire format is identical in both directions.

## Benchmark

Not applicable; the constructor lookup happens once at serializer construction time, not per serialization. The runtime paths (`newCollection`/`newMap`) add a single null check (`if (comparatorConstructor != null)`) which is negligible.

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/serializer/collection/MapSerializers.java
+++ b/java/fory-core/src/main/java/org/apache/fory/serializer/collection/MapSerializers.java
@@ -19,6 +19,7 @@
 
 package org.apache.fory.serializer.collection;
 
+import java.lang.invoke.MethodHandle;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -113,11 +114,24 @@ public class MapSerializers {
   }
 
   public static class SortedMapSerializer<T extends SortedMap> extends MapSerializer<T> {
+    private MethodHandle comparatorConstructor;
+    private MethodHandle noArgConstructor;
 
     public SortedMapSerializer(Fory fory, Class<T> cls) {
       super(fory, cls, true);
       if (cls != TreeMap.class) {
-        this.constructor = ReflectionUtils.getCtrHandle(cls, Comparator.class);
+        try {
+          comparatorConstructor = ReflectionUtils.getCtrHandle(cls, Comparator.class);
+        } catch (Exception e) {
+          // Subclass doesn't have a (Comparator) constructor, fall back to no-arg constructor.
+          try {
+            noArgConstructor = ReflectionUtils.getCtrHandle(cls);
+          } catch (Exception e2) {
+            throw new UnsupportedOperationException(
+                "Class " + cls.getName() + " requires either a (Comparator) or no-arg constructor",
+                e2);
+          }
+        }
       }
     }
 
@@ -141,7 +155,11 @@ public class MapSerializers {
         map = (T) new TreeMap(comparator);
       } else {
         try {
-          map = (T) constructor.invoke(comparator);
+          if (comparatorConstructor != null) {
+            map = (T) comparatorConstructor.invoke(comparator);
+          } else {
+            map = (T) noArgConstructor.invoke();
+          }
         } catch (Throwable e) {
           throw new RuntimeException(e);
         }
@@ -158,7 +176,11 @@ public class MapSerializers {
         map = new TreeMap(comparator);
       } else {
         try {
-          map = (Map) constructor.invoke(comparator);
+          if (comparatorConstructor != null) {
+            map = (Map) comparatorConstructor.invoke(comparator);
+          } else {
+            map = (Map) noArgConstructor.invoke();
+          }
         } catch (Throwable e) {
           throw new RuntimeException(e);
         }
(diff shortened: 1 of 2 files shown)


java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializers.java
```diff
@@ -217,7 +229,19 @@ public T newCollection(MemoryBuffer buffer) {
         collection = (T) new TreeSet(comparator);
       } else {
         try {
-          collection = (T) constructor.invoke(comparator);
+          if (comparatorConstructor != null) {
+            collection = (T) comparatorConstructor.invoke(comparator);
+          } else {
+            if (comparator != null) {
+              LOG.warn(
```
chaokunyang: This will print log every time. Which is not feasible fir serialization

chaokunyang:
It seems that the comparator is null if the treeset subclass don't pass comparator to parent class. In such cases, treeset Or treemap will just use key compareTo api for sort. So it's not a bug, could you remove the log and keep constructor branch. And could you ask check TreeMap? I think it has similar issue

java/fory-core/src/main/java/org/apache/fory/serializer/collection/CollectionSerializers.java
mandrean: Yeah alright, fixed!

mandrean:
Thanks, pushed some changes, have a look again @chaokunyang!

PR #3361, opened 2026-02-18
feat(compiler): add gRPC flags to fory compiler

### LINKED ISSUE #3271: [Compiler][gRPC] CLI options + generator service hook
Parent issue: #3266

## Scope
Add gRPC compiler flags and service-generation hooks across language generators.

## Detailed tasks
- Add `--grpc` flag and `--grpc-backend` option to `foryc` CLI.
- Wire flags into generator context/options.
- Add `generate_services()` hook to base generator class.
- Ensure services are emitted into separate file group for easy optional compilation.
- Ensure services are only generated when `--grpc` is set.
- Add CLI tests for new flags.

## Notes
Serialization is Fory-only; do not add protoc/flatc invocation.

ayush00git:
## Why?

For initializing stub generation for languages, compiler flags `--grpc` were required.
### 1. fory_compiler/base.py

Update `GeneratorOptions` to include grpc (bool).
```python
@dataclass
class GeneratorOptions:
    """Options for code generation."""

    output_dir: Path
    package_override: Optional[str] = None
    go_nested_type_style: Optional[str] = None
    grpc: bool = False
```

Add generate_services(self) -> List[GeneratedFile] method to `BaseGenerator` (returns empty list by default).
```python
    def generate_services(self) -> List[GeneratedFile]:
        """Generate service-related code (e.g. gRPC stubs).

        Base implementation returns empty list. Subclasses should override
        if they support service generation.
        """
        return []
```

### 2. fory_compiler/cli.py

- Add --grpc flag (store_true).
- Update `compile_file` to: 
- - Pass flags to `GeneratorOptions`.
- - Call `generator.generate_services()` if --grpc is set.

```python
    parser.add_argument(
        "--grpc",
        action="store_true",
        help="Generate gRPC service code (stubs, serialization traits, etc.)",
    )
```
 
## Related issues

Closes #3271 

## Benchmark
N/A

### DIFF
--- a/compiler/fory_compiler/cli.py
+++ b/compiler/fory_compiler/cli.py
@@ -385,13 +385,6 @@ def parse_args(args: Optional[List[str]] = None) -> argparse.Namespace:
         help="Generate gRPC service code (stubs, serialization traits, etc.)",
     )
 
-    parser.add_argument(
-        "--grpc-backend",
-        type=str,
-        default=None,
-        help="Specify gRPC backend (e.g., 'grpc++' for C++, 'grpcio' for Python). Defaults to standard backend for the language.",
-    )
-
     return parser.parse_args(args)
 
 
@@ -439,7 +432,6 @@ def compile_file(
     emit_fdl_path: Optional[Path] = None,
     resolve_cache: Optional[Dict[Path, Schema]] = None,
     grpc: bool = False,
-    grpc_backend: Optional[str] = None,
 ) -> bool:
     """Compile a single IDL file with import resolution.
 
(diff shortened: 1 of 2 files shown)


ayush00git:
Hey @chaokunyang 
These where the flags which marked as the initialization of the service generation code. Have a look and let me know any changes.

chaokunyang:
Do we really need the `grpc-backend` option? We can always know the backend from the target langauges. And currently we don't need to support multiple backend for one language

ayush00git:
> Do we really need the `grpc-backend` option? We can always know the backend from the target langauges. And currently we don't need to support multiple backend for one language

@chaokunyang I initially added it to future-proof for languages that might support multiple gRPC implementations (e.g., Rust with tonic vs grpc-rs), but you are right that currently we only target the standard one for each. I'll remove flag.

chaokunyang (approved):
LGTM

PR #3374, opened 2026-02-20
feat(go): add go desrialization support via io streams

### LINKED ISSUE #3302: [Go] Streaming Deserialization Support For Go
### Feature Request

_No response_

_No response_

_No response_

_No response_

### Additional context

_No response_

ayush00git:
## Why?

To enable stream-based deserialization in Fory's Go library, allowing for direct reading from `io.Reader` without pre-buffering the entire payload. This improves efficiency for network and file-based transport and brings the Go implementation into feature-parity with the python and C++ libraries.

### 1. Stream Infrastructure in `go/fory/buffer.go`

Enhanced `ByteBuffer` to support `io.Reader` with an internal sliding window and automatic filling.

- Added `reader io.Reader` and `minCap int` fields.
- Implemented `fill(n int, err *Error) bool` for on-demand data fetching and buffer compaction.
- Added `CheckReadable(n)` and `Skip(n)` memory-safe routines that pull from the underlying stream when necessary to avoid out-of-bounds panics.
- Updated `ReadBinary` and `ReadBytes` to safely copy slices when streaming to prevent silent data corruption on compaction.
- Updated all `Read*` methods (fixed-size, varint, tagged) to fetch data from the reader safely if not cached.

### 2. Stateful InputStream in `go/fory/stream.go`

Added the `InputStream` feature to support true, stateful sequential stream reads.

- Introduced `InputStream` which persists the buffered byte window and TypeResolver metadata (Meta Sharing) across multiple object decodes on the same stream, decoupled from `Fory` to mirror the C++ `ForyInputStream` implementation.
- Added `fory.DeserializeFromStream(is, target)` method to process continuous streamed data.
- Added `Shrink()` method to compact the internal buffer and reclaim memory during long-lived streams.
- Added `DeserializeFromReader` method as an API for simple *one-off* stream object reads.

### 3. Stream-Safe Deserialization Paths

Updated internal deserialization pipelines in `struct.go` and `type_def.go` to be stream-safe:
- Integrated `CheckReadable` bounds-checking into the `struct.go` fast paths for fixed-size primitives.
- Safely rewrote schema-evolution skips (`skipTypeDef`) in `type_def.go` to use bounds-checked `Skip()` rather than unbounded `readerIndex` overrides.

### 4. Comprehensive Stream Tests

- Built a custom `oneByteReader` wrapper (`go/fory/test_helper_test.go`) that artificially feeds the deserialization engine exactly 1 byte at a time.
- Migrated the global test suite (`struct_test.go`, `primitive_test.go`, `slice_primitive_test.go`, etc.) to run all standard tests through this aggressive 1-byte fragmented stream reader via a new `testDeserialize` helper to guarantee total stream robustness.

## Related issues

Closes #3302

- [x] Does this PR introduce any public API change? (`NewInputStream`, `DeserializeFromStream`, `DeserializeFromReader`, `NewByteBufferFromReader`)

## Benchmark
Main branch - 
image

This branch - 
image


ayush00git:
Hey @chaokunyang 
Have a review and let me know the changes

ayush00git:
Hey @chaokunyang 
Have a review and let me know the changes

Zakir032002:
hey @ayush00git, looked through this and the main issue i see is in `DeserializeFromReader` —
it calls `ResetWithReader` at the start of every call:

```go
func (f *Fory) DeserializeFromReader(r io.Reader, v any) error {
    defer f.resetReadState()
    f.readCtx.buffer.ResetWithReader(r, 0) // this wipes the prefetch window every time
```
so if fill() reads ahead past the first object boundary (which it will), those bytes
are gone on the next call. sequential decode from one stream is broken:

```go
for {
    var msg Msg
    f.DeserializeFromReader(conn, &msg) // bytes after first object get thrown away
}
```
if you look at how he handles this for c++/python — the Buffer is constructed
from the stream once and passed to each deserialize call directly. the buffer holds
state across calls, it's never reset between objects. the python test
test_stream_deserialize_multiple_objects_from_single_stream shows this exactly —
same reader buffer passed to multiple fory.deserialize() calls.

the go version probably needs something similar — a stream reader type that owns the
buffer and gets reused across deserializations rather than resetting on each call.

Happy to discuss if I'm misreading the flow here

ayush00git:
Hiii @Zakir032002 
Thanks for noticing this, exactly this is a bug in the implementation from my side. yes the call would clear any prefetched data from the `ByteBuffer` making the sequential reads from the stream impossible, also it was clearing the typemetadata as well. thanks for mentioning this, i'll look at the c++ python implementation to correct the deserializer.

Zakir032002:
hey @ayush00git , one more thing — ReadBinary and ReadBytes return a direct slice into 
b.data:

    v := b.data[b.readerIndex : b.readerIndex+length]
    return v

the problem is fill() compacts the buffer in-place:

    copy(b.data, b.data[b.readerIndex:])

so if someone reads a []byte field and holds onto that slice, then the next 
read triggers a fill() — the compaction just overwrote the bytes they're 
still holding. no error, no panic, just wrong data.

in stream mode you probably want to copy before returning instead of aliasing:

    if b.reader != nil {
        result := make([]byte, length)
        copy(result, b.data[b.readerIndex:b.readerIndex+length])
        b.readerIndex += length
        return result
    }

in-memory path stays as is.

Zakir032002:
also noticed — `ReadVarUint32Small7` only does `fill(1)` for the first byte, but if that byte has `0x80` set it falls through to `continueReadVarUint32` which isn't touched in this PR. so in stream mode, if a multi-byte varint straddles a chunk boundary, the continuation bytes may not be in the buffer yet — you either get a `BufferOutOfBoundError` or silently read the wrong bytes depending on what's sitting at that position in the buffer.

easiest fix is probably just routing the multi-byte case through `readVarUint32Slow` since that's already stream-aware after your changes. or adding `fill(1)` guards inside `continueReadVarUint32` directly, either works.

Happy to discuss if I'm misreading the flow here

ayush00git:
Hey @Zakir032002 
Sorry i'm a bit busy with my exams, as i get free, i'll review the comments

docs/compiler/compiler-guide.md
```diff
@@ -63,9 +63,10 @@ Compile options:
 | `--cpp_out=DST_DIR`                   | Generate C++ code in DST_DIR                          | (none)              |
 | `--go_out=DST_DIR`                    | Generate Go code in DST_DIR                           | (none)              |
 | `--rust_out=DST_DIR`                  | Generate Rust code in DST_DIR                         | (none)              |
-| `--go_nested_type_style`              | Go nested type naming: `camelcase` or `underscore`    | from schema/default |
-| `--emit-fdl`                          | Print translated Fory IDL for non-`.fdl` inputs       | `false`             |
-| `--emit-fdl-path`                     | Write translated Fory IDL to a file or directory      | (stdout)            |
+| `--grpc`                              | Generate gRPC service code (in development)           | `false`             |
```
chaokunyang: this it not related to this PR, please remove it

ayush00git: It got pushed by mistakely, i'll fix it

ayush00git:
Hii @Zakir032002 
Thanks for pointing out the flows, 
- The `ReadFromDeserializer` and returning a direct slice into the data stream are wrongly implemented by me, thanks for suggesting the chnages to fix them as well.

But i think you misunderstood the `ReadVarUint32Small7`. We already have a check condition -
```bash
if len(b.data)-readIdx >= 5 {

}
```
If we are near a chunk boundary (less than 5 bytes remaining in the buffer), the execution completely skips `continueReadVarUint32` and jumps straight to `readVaruint36Slow`. I don't think this part need any changes

ayush00git:
I've added the `StreamReader` which now creates a copy slice during desrialization to preserve the data between sequential desrialization calls. `DeserializeFromReader` only is there if the user wants to deserialize a single struct and don't want a stream overhead for that.

chaokunyang:
Thanks for adding stream deserialization support. I found two correctness issues while testing stream-backed buffers with small chunk readers:

1.  does not call , so stream-backed buffers can return short/zero reads even when the underlying  still has data. This breaks internal call paths that use  (for example  / type-def decode), and I can reproduce corrupted metastring bytes with  + one-byte reader.
2.  currently collapses all  errors into , so non-EOF transport errors (e.g. connection reset) are reported as bounds errors and the original I/O cause is lost.

Could you take a look at these before merge?

go/fory/buffer.go
```diff
@@ -28,12 +29,76 @@ type ByteBuffer struct {
...
+		spare := b.data[len(b.data):cap(b.data)]
+		if len(spare) == 0 {
+			return false
+		}
+		readBytes, err := b.reader.Read(spare)
+		if readBytes > 0 {
+			b.data = b.data[:len(b.data)+readBytes]
+			b.writerIndex += readBytes
+		}
+		if err != nil {
```
chaokunyang, line 95: `fill` currently folds reader errors into `false`, and callers then emit `BufferOutOfBoundError`. This masks non-EOF transport failures (for example connection reset) as bounds issues. Please preserve/propagate non-EOF read errors so stream deserialization reports the real I/O failure.

go/fory/buffer.go
```diff
@@ -28,12 +29,76 @@ type ByteBuffer struct {
...
 	readerIndex int
+	reader      io.Reader
+	minCap      int
 }
 
 func NewByteBuffer(data []byte) *ByteBuffer {
 	return &ByteBuffer{data: data}
 }
 
+func NewByteBufferFromReader(r io.Reader, minCap int) *ByteBuffer {
```
chaokunyang: This introduces stream mode, but `ByteBuffer.Read(p)` still only copies from in-memory `b.data` and never calls `fill`. Any decode path that uses `Read` can therefore observe partial/zero bytes with short-chunk readers. Please make stream-backed `Read` fetch until `len(p)` bytes are available (or return the underlying read error) to avoid silent metadata corruption.

go/fory/buffer.go
ayush00git: now it logs the exact error

go/fory/buffer.go
ayush00git: now it calls `fill()` upto reserved bytes and copies as well to prevent metadata corruption.

ayush00git:
Hii @chaokunyang 
Have a look and let me know the changes.

ayush00git:
Hii @chaokunyang 
Have a look and let me know the changes.

chaokunyang:
Please take https://github.com/apache/fory/pull/3307 as reference to finish the remaining works. And create a Deseralize help methods in tests, then use that instead of `fory.Deserialize` for deserialization, and in the `Deseralize` test helper, first deserialize from bytes, then wrap it into a OneByteStream to deserialize it to ensure deserialization works.

Then run benchmarks/go to compare with asf/main to enure your code change don't introduce any performance regression.

ayush00git:
Hey @chaokunyang 
I discovered a bug in the struct implementation. In `struct.go`, there is a highly optimized "fast-path" that uses unsafe pointers to read structs from memory instantly, and while reading byte 1 by 1 it is failing tests, because it assumes entire memory buffer was already available.

ayush00git:
Hii @chaokunyang 
The PR is now aligned with the #3307

go/fory/fory.go
```diff
@@ -495,6 +496,105 @@ func (f *Fory) Deserialize(data []byte, v any) error {
 	return nil
 }
 
+// StreamReader supports robust sequential deserialization from a stream.
+// It maintains the ByteBuffer and ReadContext state across multiple Deserialize calls,
+// preventing data loss from prefetched buffers and preserving TypeResolver metadata
+// (Meta Sharing) across object boundaries.
+type StreamReader struct {
```
chaokunyang: create a stream.go and put it there

go/fory/fory.go
```diff
@@ -495,6 +496,105 @@ func (f *Fory) Deserialize(data []byte, v any) error {
...
+	fory   *Fory
+	reader io.Reader
+	buffer *ByteBuffer
+}
+
+// NewStreamReader creates a new StreamReader that reads from the provided io.Reader.
+// The StreamReader owns the buffer and maintains state across sequential Deserialize calls.
+func (f *Fory) NewStreamReader(r io.Reader) *StreamReader {
```
chaokunyang: Looks really strange that the StreamReader is coupled with Fory.

Please read fory c++ implementation detaily and refine this PR, especially for API surface

ayush00git: yaa in c++ the streamreader acted as a standalone type which had nothing to do with deserialization it just know to fetch bytes when needed, let me implement the same for go

ayush00git:
@chaokunyang 
the api design now matches with c++, is there any other modification ?

chaokunyang:
I added shrink_input_stream in https://github.com/apache/fory/pull/3453, could you add similiar feature in this PR? And I renamed StreamReader to InputStream in https://github.com/apache/fory/pull/3449, please also rename related API too. the `StreamReader` is not that clear

go/fory/buffer.go
```diff
@@ -28,12 +29,86 @@ type ByteBuffer struct {
...
+	available := len(b.data) - b.readerIndex
+	if available >= n {
+		return true
+	}
+
+	if b.readerIndex > 0 {
+		copy(b.data, b.data[b.readerIndex:])
+		b.writerIndex -= b.readerIndex
+		b.readerIndex = 0
+		b.data = b.data[:b.writerIndex]
```
chaokunyang, line 69: Why need to process on writerIndex?

go/fory/buffer.go
```diff
@@ -263,8 +344,9 @@ func (b *ByteBuffer) ReadUint32(err *Error) uint32 {
 //go:inline
 func (b *ByteBuffer) ReadUint64(err *Error) uint64 {
 	if b.readerIndex+8 > len(b.data) {
-		*err = BufferOutOfBoundError(b.readerIndex, 8, len(b.data))
-		return 0
+		if !b.fill(8, nil) {
```
chaokunyang: Why you pass `nil`, this will just slicence the error

go/fory/buffer.go
```diff
@@ -877,7 +1002,7 @@ func (b *ByteBuffer) WriteVaruint36Small(value uint64) {
 //
 //go:inline
 func (b *ByteBuffer) ReadVaruint36Small(err *Error) uint64 {
-	if b.remaining() >= 8 {
+	if b.remaining() >= 8 || (b.reader != nil && b.fill(8, nil)) {
```
chaokunyang: If it's a network stream ,and write don't close teh stream, then the stream EOF never come, this will just hang forever

chaokunyang:
Some Design suggestions:
- Keep fast-path varint checks purely remaining >= N; do not call fill(N) in the predicate.
- Never pass nil error sinks to fill in decode methods that promise bounds errors.
- Make DeserializeFromReader either strictly stateless (always reset buffer) or explicitly stateful and align it with InputStream semantics.
- Remove InputStream.reader unless it is actually needed (or use it for invariants/validation).

chaokunyang:
And please run benchmarks/go for current branch and pache/main branch, and paste the result here

go/fory/buffer.go
ayush00git: we're using a sliding window approach to push the unread bytes to the front, so if we only put readerIndex = 0 and keep the writerIndex at its old position, it would track garbage value left in between.

go/fory/buffer.go
ayush00git: yes we should never hang it on a pre-fill 8 bytes. let me fix it.

go/fory/buffer.go
ayush00git: my bad. i'll fix it up

ayush00git:
Benchmark: apache/fory/main vs this branch (feat/go-deserialization)

**apache/fory/main:**
```
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i5-1235U
BenchmarkFory_Struct_Serialize-12              34537162     110.4 ns/op      0 B/op    0 allocs/op
BenchmarkFory_Struct_Deserialize-12            19927696     179.6 ns/op     32 B/op    1 allocs/op
BenchmarkFory_StructList_Serialize-12           3048019    1176 ns/op        0 B/op    0 allocs/op
BenchmarkFory_StructList_Deserialize-12         1583568    2456 ns/op      688 B/op    3 allocs/op
BenchmarkFory_Sample_Serialize-12              14559171     247.5 ns/op      0 B/op    0 allocs/op
BenchmarkFory_Sample_Deserialize-12             3322015    1158 ns/op      676 B/op    9 allocs/op
BenchmarkFory_SampleList_Serialize-12            919764    3815 ns/op        0 B/op    0 allocs/op
BenchmarkFory_SampleList_Deserialize-12          240294   20650 ns/op    13952 B/op  163 allocs/op
BenchmarkFory_MediaContent_Serialize-12         7116196     458.0 ns/op      0 B/op    0 allocs/op
BenchmarkFory_MediaContent_Deserialize-12       2302726    1496 ns/op      656 B/op   13 allocs/op
BenchmarkFory_MediaContentList_Serialize-12      532838    7471 ns/op        0 B/op    0 allocs/op
BenchmarkFory_MediaContentList_Deserialize-12    143080   30065 ns/op    13040 B/op  243 allocs/op
```

**this branch:**
```
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i5-1235U
BenchmarkFory_Struct_Serialize-12              32973771     121.9 ns/op      0 B/op    0 allocs/op
BenchmarkFory_Struct_Deserialize-12            27100645     163.6 ns/op     32 B/op    1 allocs/op
BenchmarkFory_StructList_Serialize-12           3265543    1103 ns/op        0 B/op    0 allocs/op
BenchmarkFory_StructList_Deserialize-12         1268991    2679 ns/op      688 B/op    3 allocs/op
BenchmarkFory_Sample_Serialize-12              12737720     258.3 ns/op      0 B/op    0 allocs/op
BenchmarkFory_Sample_Deserialize-12             3287950    1123 ns/op      676 B/op    9 allocs/op
BenchmarkFory_SampleList_Serialize-12            995695    3745 ns/op        0 B/op    0 allocs/op
BenchmarkFory_SampleList_Deserialize-12          231688   18862 ns/op    13952 B/op  163 allocs/op
BenchmarkFory_MediaContent_Serialize-12         6446667     542.3 ns/op      0 B/op    0 allocs/op
BenchmarkFory_MediaContent_Deserialize-12       2616025    1222 ns/op      656 B/op   13 allocs/op
BenchmarkFory_MediaContentList_Serialize-12      414025    8415 ns/op        0 B/op    0 allocs/op
BenchmarkFory_MediaContentList_Deserialize-12    172894   28446 ns/op    13040 B/op  243 allocs/op
```

ayush00git:
@chaokunyang 
I'm done with the changes. pasted the benchmarks as well. have a look now

go/fory/test_helper_test.go
```diff
@@ -0,0 +1,79 @@
...
+	vType := reflect.TypeOf(v)
+	if vType == nil || vType.Kind() != reflect.Ptr {
+		t.Fatalf("testDeserialize requires a pointer to a value, got %v", vType)
+	}
+
+	stream := &oneByteReader{data: data, pos: 0}
+
+	// Create a new stream reader. The stream context handles boundaries and compactions.
+	streamReader := NewInputStream(stream)
+	err = f.DeserializeFromStream(streamReader, v)
```
chaokunyang: This helper deserializes into the same target twice (first from bytes, then from stream). That can mask stream-path bugs because values populated by the first pass may still be present if the second pass does not fully overwrite fields.

Please deserialize the stream path into a fresh value and compare results, so partial/incorrect stream decoding is detectable.

go/fory/stream_test.go
```diff
@@ -0,0 +1,238 @@
...
+	data, err := f.Serialize(original)
+	if err != nil {
+		t.Fatalf("Serialize failed: %v", err)
+	}
+
+	// Test with slow reader and small minCap to force compaction/growth
+	reader := &slowReader{data: data}
+	var decoded StreamTestStruct
+	// Use small minCap (16) to force frequent fills and compactions
+	f.readCtx.buffer.ResetWithReader(reader, 16)
```
chaokunyang: This setup line does not affect the code path under test, because `DeserializeFromReader` immediately calls `ResetWithReader(r, 0)` internally and replaces the min-cap you set here.

If the goal is to verify small-cap refill/compaction behavior, consider using `DeserializeFromStream` with `NewInputStreamWithMinCap(..., 16)` so the configured capacity is actually used.

go/fory/stream.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// NewInputStream creates a new InputStream that reads from the provided io.Reader.
+// The InputStream owns the buffer and maintains state across sequential Deserialize calls.
+func NewInputStream(r io.Reader) *InputStream {
+	return NewInputStreamWithMinCap(r, 0)
+}
+
+// NewInputStreamWithMinCap creates a new InputStream with a specified minimum buffer capacity.
+func NewInputStreamWithMinCap(r io.Reader, minCap int) *InputStream {
```
chaokunyang: THis is buffer size, not capacity, pelase change API name

ayush00git: i'll change it to NewInputStreamWithBufferSize. does that sounds good ? It would also resonate with the api implemented in the cpp strem deserialization

chaokunyang (approved):
LGTM

Issue #658, opened 2023-07-15
add doc example tests for c++


chaokunyang:

qaubnc:
looking forward to it!how soon is it planned to be added?

chaokunyang:
This seems needs Doxygen, and use script to extract code block from documents, then run it .

@qaubnc Are you interested at this issue?

Tyooughtul:
@chaokunyang I'd like to work on this issue, please assign it to me. :D

PR #3381, opened 2026-02-20
chore(cpp): configure CI to extract and execute C++ code from Markdo…

Tyooughtul:
- add extract_cpp_doc_examples.py to extract C++ code blocks from markdown
- integrate doc tests into CI with --skip-doc-tests and --doc-tests-only options
- fix documentation bugs discovered by tests:
  - index.md: change class to struct for aggregate initialization
  - field-configuration.md: fix smart pointer access using .get()
- add cpp/doc_tests/ to .gitignore

issue: #658

## Why?

Add automated testing for C++ code examples in documentation to ensure they compile and run correctly.

## What does this PR do?

- add extract_cpp_doc_examples.py to extract C++ code blocks from markdown
- integrate doc tests into CI with --skip-doc-tests and --doc-tests-only options
- fix documentation bugs discovered by tests:
  - index.md: change class to struct for aggregate initialization
  - field-configuration.md: fix smart pointer access using .get()
- add cpp/doc_tests/ to .gitignore

## Related issues
- #658

### DIFF
--- a/docs/guide/cpp/field-configuration.md
+++ b/docs/guide/cpp/field-configuration.md
@@ -429,15 +429,14 @@ int main() {
   auto fory = Fory::builder().xlang(true).build();
   fory.register_struct<MetricsData>(100);
 
-  MetricsData data{
-      .request_count = 42,
-      .bytes_sent = 1024,
-      .user_id = 12345678,
-      .session_id = 9876543210,
-      .created_at = 1704067200000000000ULL, // 2024-01-01 in nanoseconds
-      .error_count = 3,
-      .last_access_time = std::nullopt
-  };
+  MetricsData data;
+  data.request_count = 42;
+  data.bytes_sent = 1024;
+  data.user_id = 12345678;
+  data.session_id = 9876543210;
+  data.created_at = 1704067200000000000ULL; // 2024-01-01 in nanoseconds
+  data.error_count = 3;
+  data.last_access_time = std::nullopt;
 
   auto bytes = fory.serialize(data).value();
   auto decoded = fory.deserialize<MetricsData>(bytes).value();


docs/guide/cpp/index.md
```diff
@@ -137,25 +137,22 @@ See the [examples/cpp](https://github.com/apache/fory/tree/main/examples/cpp) di
...
 ```cpp
 #include "fory/serialization/fory.h"
+#include <string>
+#include <vector>
 
 using namespace fory::serialization;
 
 // Define a struct
-class Person {
+struct Person {
```
chaokunyang, line 146: Why we need this change?

docs/guide/cpp/field-configuration.md
```diff
@@ -225,7 +225,7 @@ int main() {
   doc.description = "A sample document";
   doc.metadata = nullptr;  // Allowed because nullable
   doc.parent = std::make_shared<Document>();
-  doc.parent->title = "Parent Doc";
+  doc.parent.get()->title = "Parent Doc";
```
chaokunyang, line 228: is this really necessary?

docs/guide/cpp/index.md
Tyooughtul: I think the issue is that class members are private by default, but the code uses aggregate initialization Person{"Alice", 30, ...} . As far as I understand, aggregate initialization requires all members to be public, so I changed it to struct .

docs/guide/cpp/field-configuration.md
Tyooughtul: I think fory::field<std::shared_ptr<T>>::operator->() returns a pointer to the shared_ptr itself, not to T . Using .get() gives us the shared_ptr reference, and then -> properly accesses the underlying object.

ci/extract_cpp_doc_code.py
```diff
@@ -0,0 +1,377 @@
...
+        generated_files.append(test_path)
+        logging.info(f"  Generated {test_file_name}")
+
+    return generated_files
+
+
+def generate_bazel_build(test_files: List[Path], output_dir: Path) -> None:
+    build_path = output_dir / "BUILD"
+
+    build_content = """# Licensed to the Apache Software Foundation (ASF) under one
```
chaokunyang: the license header is not needed, the generated code is only for CI, not for release

docs/guide/cpp/index.md
chaokunyang: I see, thanks for clarify it

ci/extract_cpp_doc_code.py
Tyooughtul: Thanks for comment! 😊 I‘ve fixed it.

ci/tasks/cpp.py
```diff
@@ -42,3 +89,10 @@ def run(install_deps_only=False):
 
     common.bazel(f"{test_command} {targets}")
     logging.info("C++ CI tasks completed successfully")
+
+    # Run documentation example tests
+    if not skip_doc_tests:
+        try:
+            run_doc_example_tests()
+        except Exception as e:
+            logging.warning(f"Documentation example tests failed: {e}")
```
chaokunyang: This should not be an warning, it should raise Exception and let ci fail

chaokunyang:
And Docs-only PRs won’t trigger this workflow: pull_request.paths-ignore excludes docs/** and **/*.md.
[ci.yml#L33-L37](https://github.com/apache/fory/blob/08e1a8976670ebbb0b3ef94e4d2919949ca07e84/.github/workflows/ci.yml#L33-L37), please also update the ci.yml

chaokunyang (approved):
LGTM

PR #3384, opened 2026-02-22
chore(python): handle duplicate field names

eyad-hazem-elmorsy:
## Why?

This PR handles the duplicate of field names for inheritance.

## What does this PR do?

The handling is done by making every child class's field override its parent's till the most-derived field kept.
- First, we iterate over the full class MRO (cls.__mro__) to extract field metadata from all parent classes.
- Then, deduplicate fields so that each field name appears only once in the resulting list with keeping the child class definition if a field is shadowed.
- Finally, added some tests to cover:
    - Single-level inheritance
    - Multi-level inheritance
    - Field shadowing in child and grandchild classes
    - Deduplication of overridden fields

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,12 +167,22 @@ jobs:
         run: |
           cd csharp
           dotnet test tests/Fory.Tests/Fory.Tests.csproj -c Release --no-build
+      - name: Verify C# format
+        run: |
+          cd csharp
+          dotnet format Fory.sln --verify-no-changes \
+            --exclude src/Fory.Generator/AnalyzerReleases.Shipped.md \
+            --exclude src/Fory.Generator/AnalyzerReleases.Unshipped.md
 
   csharp_xlang:
     name: C# Xlang Test
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
       - name: Set up .NET 8
         uses: actions/setup-dotnet@v4
         with:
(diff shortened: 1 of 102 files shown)


chaokunyang:
It seems that dataclass don't allow sublass has duplicate fields with parent class. Could you check whether this is true? If so, the issue can be simplified a lot

eyad-hazem:
I found that dataclass allows a subclass to have duplicate fields with a parent class, but when you define a field with the same name in the subclass, it overrides the parent's field

eyad-hazem:
Does it mean that I needed to only remove the comment of TODO?

chaokunyang:
Just remove TODO, your code should make it work now

eyad-hazem:
You can check now @chaokunyang

chaokunyang:
@eyad-hazem-elmorsy Could you add some duplciate fields tests back? THe tests can ensure we don't break it in future

eyad-hazem:
Returned back and the tests run successfully

python/pyfory/tests/test_duplicate_field_names.py
```diff
@@ -0,0 +1,277 @@
...
+
+    field_map = {f.name: f for f in typedef.fields}
+
+    assert "name" in field_map
+
+    # Critical assertion
+    assert field_map["name"].defined_class == "Child", f"Expected field to be defined in Child but got {field_map['name'].defined_class}"
+
+
+if __name__ == "__main__":
```
chaokunyang: I think we also need e2e serialization tests in each case instead of only typedef tests

python/pyfory/tests/test_duplicate_field_names.py
```diff
@@ -0,0 +1,277 @@
...
+
+    # Verify round-trip
+    buffer = Buffer(typedef.encoded)
+    decoded_typedef = decode_typedef(buffer, resolver)
+
+    decoded_field_names = [f.name for f in decoded_typedef.fields]
+    assert decoded_field_names.count("name") == 1
+
+
+def test_multilevel_inheritance():
```
chaokunyang: I'm not sure whether such tests are meaningful. The shadow ones shuld already cover it.More code means more maintainance overhead. 

Please refactor all tests to to make it consise and remove redundancy

python/pyfory/tests/test_duplicate_field_names.py
```diff
@@ -0,0 +1,50 @@
...
+    extra: float
+
+
+def test_shadowed_fields_serialization():
+    """Serialization with shadowed and inherited fields."""
+    fory = Fory(xlang=True)
+    fory.register(Parent, namespace="test", typename="Parent")
+    fory.register(ChildWithShadow, namespace="test", typename="ChildWithShadow")
+
+    obj = ChildWithShadow(name="shadowed", value=10, extra=3.14)
```
chaokunyang, line 49: Could you add TypeDef encoding check that only threefields for ChildWithShadow?

eyad-hazem:
Thanks for your time and these good points. I explored many pieces of code in this PR and learned a lot

PR #3421, opened 2026-02-25
feat(rust): add configurable size guardrails

### LINKED ISSUE #3409: [Rust] configurable size guardrails for untrusted payloads
### Feature Request

Add configurable deserialization size guardrails in Fory Rust for untrusted payloads.

Add two configurable size limits to Rust deserialization and enforce them in relevant preallocation-sensitive read paths.

Resolve task:

### Additional context

Related locations:
- `rust/fory-core/src/config.rs:24`
- `rust/fory-core/src/serializer/collection.rs:229`
- `rust/fory-core/src/serializer/collection.rs:287`
- `rust/fory-core/src/serializer/map.rs:549`
- `rust/fory-core/src/serializer/map.rs:700`
- `rust/fory-core/src/buffer.rs:929`

Zakir032002:
## Summary

Fixes #3409

Adds three opt-in `Fory` builder methods that let callers enforce upper bounds on the size of data allocated during deserialization. Without these guards a crafted payload can contain an absurdly large length prefix, causing `Vec::with_capacity` / string allocation to exhaust heap memory before a single byte of real data is read.

```rust
let fory = Fory::default()
    .max_string_bytes(1024 * 1024)    
    .max_collection_size(100_000)     
    .max_map_size(100_000);           
```

All three limits default to `None` (no limit), so this is **100 % backwards-compatible**.

## Testing

```
cargo fmt --check   ✓
cargo clippy --all-targets --all-features -- -D warnings   ✓
cargo test test_size_guardrails   → 6 passed, 0 failed
cargo test                        → all existing tests pass
```


Zakir032002:
@chaokunyang , let me know if you need any changes

chaokunyang:
Please set default value to max value of int32. And also check length is less or equal than buffer remaining size

rust/fory-core/src/resolver/context.rs
```diff
@@ -315,6 +315,9 @@ pub struct ReadContext<'a> {
     xlang: bool,
     max_dyn_depth: u32,
     check_struct_version: bool,
+    max_string_bytes: usize,
```
chaokunyang: Let's simplify the config options, reserve:
- max_binary_size, set default value to 64MB
- max_collection_size, set default value to 1_000_000

rust/fory-core/src/resolver/context.rs
```diff
@@ -472,6 +478,39 @@ impl<'a> ReadContext<'a> {
         self.meta_string_resolver.read_meta_string(&mut self.reader)
     }
 
+    #[inline(always)]
+    pub fn check_string_bytes(&self, byte_len: usize) -> Result<(), Error> {
```
chaokunyang: For string, we don't have to check bytes length, bevcause we don't have preallocaiton for string

Zakir032002:
Heyy @chaokunyang , Done with the changes as you requested

rust/fory-core/src/resolver/context.rs
```diff
@@ -472,6 +476,28 @@ impl<'a> ReadContext<'a> {
         self.meta_string_resolver.read_meta_string(&mut self.reader)
     }
 
+    #[inline(always)]
+    pub fn check_binary_size(&self, byte_len: usize) -> Result<(), Error> {
```
chaokunyang: I don't think we need such methods? ANd it's even not used? why you add an unused method?

Zakir032002:
@chaokunyang — You're right, that method was unnecessary and unused.
I've removed check_binary_size and the related ReadContext field.

The implementation now only enforces max_collection_size at actual preallocation sites (collections, maps, primitive lists), and includes the buffer-remaining check as requested.

Config has been simplified to keep only max_binary_size (reserved, default 64MB) and max_collection_size (default 1_000_000).

All checks are passing. Please let me know if anything else should be adjusted.

rust/fory-core/src/serializer/map.rs
```diff
@@ -698,18 +710,30 @@ impl<K: Serializer + ForyDefault + Ord + std::hash::Hash, V: Serializer + ForyDe
...
         if len == 0 {
-            return Ok(map);
+            return Ok(BTreeMap::new());
+        }
+        let remaining = context
+            .reader
+            .bf
+            .len()
+            .saturating_sub(context.reader.cursor);
+        if len as usize > remaining {
```
chaokunyang: This won't work for streaming deserialization

chaokunyang: Could you fix this? in stream deserialization, the buffer length doesn't represent remaining bytes anymore

Zakir032002:
@chaokunyang, now have a look and let me know the changes

chaokunyang:
@Zakir032002 Could you merge main branch and address conflicts?

rust/fory-core/src/serializer/map.rs
```diff
@@ -556,7 +556,11 @@ impl<K: Serializer + ForyDefault + Eq + std::hash::Hash, V: Serializer + ForyDef
     }
 
     fn fory_read_data(context: &mut ReadContext) -> Result<Self, Error> {
+<<<<<<< HEAD
```
chaokunyang: Please address conflict first before commit

Zakir032002:
Heyy @chaokunyang , now have a look

ayush00git:
@chaokunyang 
this can be closed referencing #3579

PR #3429, opened 2026-02-27
feat(python): add configurable size guardrails

### LINKED ISSUE #3416: [Python] configurable size guardrails for untrusted payloads
### Feature Request

Add configurable deserialization size guardrails in Fory Python for untrusted payloads.

Add two configurable size limits to Python deserialization and enforce them in relevant preallocation-sensitive read paths.

Resolve task:

### Additional context

Related locations:
- `python/pyfory/_fory.py:164`
- `python/pyfory/serialization.pyx:1048`
- `python/pyfory/collection.pxi:300`
- `python/pyfory/collection.pxi:301`
- `python/pyfory/collection.pxi:861`
- `python/pyfory/collection.pxi:862`
- `python/pyfory/buffer.pyx:542`

eyad-hazem-elmorsy:
This commit:
- provides configurable size guardrails for untrusted payloads (lists, sets, tuples, maps, and strings)
- tests size guardrails

## Why?

There are currently no configurable limits for payload-driven lengths. Untrusted string/map/list lengths can trigger large allocations and memory pressure.

## What does this PR do?

- Adds guardrail fields in Python runtime configuration.
- Enforces length limits for lists, tuples, sets, dicts(maps), and string bytes.
- Raises an exception when a configured limit is exceeded.

## Related issues

- Fixes #3416

### DIFF
--- a/python/pyfory/_fory.py
+++ b/python/pyfory/_fory.py
@@ -161,6 +161,7 @@ class Fory:
         "field_nullable",
         "policy",
         "max_collection_size",
+        "max_binary_size",
     )
 
     def __init__(
@@ -174,6 +175,7 @@ class Fory:
         field_nullable: bool = False,
         meta_compressor=None,
         max_collection_size: int = 1_000_000,
+        max_binary_size: int = 64 * 1024 * 1024,
     ):
         """
         Initialize a Fory serialization instance.
@@ -218,6 +220,11 @@ class Fory:
                 collection sizes, as collections preallocate memory based on the declared
                 size. Raises an exception if exceeded. Default is 1,000,000.
 
+            max_binary_size: Maximum allowed size in bytes for binary data reads during
+                deserialization (default: 64 MB). Raises an exception if a single binary
+                read exceeds this limit, preventing out-of-memory attacks from malicious
+                payloads that claim extremely large binary sizes.
+
         Example:
             >>> # Python-native mode with reference tracking
             >>> fory = Fory(ref=True)
@@ -243,7 +250,8 @@ class Fory:
         self.serialization_context = SerializationContext(fory=self, scoped_meta_share_enabled=compatible)
         self.type_resolver.initialize()
 
-        self.buffer = Buffer.allocate(32)
+        self.max_binary_size = max_binary_size
+        self.buffer = Buffer.allocate(32, max_binary_size=max_binary_size)
         self.buffer_callback = None
         self._buffers = None
         self._unsupported_callback = None
@@ -630,7 +638,7 @@ class Fory:
         assert self.depth == 0, "Nested deserialization should use read_ref/read_no_ref."
         self.depth += 1
         if isinstance(buffer, bytes):
-            buffer = Buffer(buffer)
+            buffer = Buffer(buffer, max_binary_size=self.max_binary_size)
         if unsupported_objects is not None:
             self._unsupported_objects = iter(unsupported_objects)
         reader_index = buffer.get_reader_index()
@@ -824,6 +832,8 @@ class ThreadSafeFory:
         max_depth (int): Maximum depth for deserialization. Defaults to 50.
         max_collection_size (int): Maximum allowed size for collections and maps during
             deserialization. Defaults to 1,000,000.
+        max_binary_size (int): Maximum allowed size in bytes for binary data reads during
+            deserialization. Defaults to 64 MB.
 
     Example:
         >>> import pyfury
--- a/python/pyfory/buffer.pxi
+++ b/python/pyfory/buffer.pxi
@@ -148,7 +148,7 @@ cdef class Buffer:
         self.output_stream = None
 
     @classmethod
-    def from_stream(cls, stream not None, uint32_t buffer_size=4096):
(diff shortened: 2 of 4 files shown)


python/pyfory/_fory.py
```diff
@@ -153,6 +153,9 @@ class Fory:
         "depth",
         "field_nullable",
         "policy",
+        "max_collection_length",
+        "max_map_length",
+        "max_string_bytes_length",
```
chaokunyang: could we rename `max_string_bytes_length` to `max_binary_size`, and reame other `_length` to `_size` for consistency?

python/pyfory/_fory.py
```diff
@@ -602,6 +620,11 @@ def _read_no_ref_internal(self, buffer, serializer):
         """Internal method to read without modifying read_ref_ids."""
         if serializer is None:
             serializer = self.type_resolver.read_type_info(buffer).serializer
+
+        cls = serializer.type_ if hasattr(serializer, "type_") else None
+        if cls is str:
+            return buffer.read_string(self.max_string_bytes_length)
```
chaokunyang: We don't need to validate string size, because the valide is only used to prevent malious data to fake a hught size to cause a big allocation which incur oom. but for string, underlien buffer doesn't haev that much buffer, the read will jsut fail, there is no OOM issue

python/pyfory/_fory.py
chaokunyang: And we could use `max_collection_size` for map size currently to reduce config option overhead.

python/pyfory/_fory.py
```diff
@@ -165,6 +168,9 @@ def __init__(
         policy: DeserializationPolicy = None,
         field_nullable: bool = False,
         meta_compressor=None,
+        max_collection_length: int = -1,
```
chaokunyang: Set to 1_000_000

python/pyfory/_fory.py
```diff
@@ -165,6 +168,9 @@ def __init__(
...
+        max_map_length: int = -1,
+        max_string_bytes_length: int = -1,
```
chaokunyang: set to  64 * 1024 * 1024

eyad-hazem:
You can take a look @chaokunyang

python/pyfory/buffer.pyx
```diff
@@ -283,8 +283,10 @@ cdef class Buffer:
         if length > 0:
             self.c_buffer.write_bytes(&data[0], length)
 
-    cpdef inline bytes read_bytes_and_size(self):
+    cpdef inline bytes read_bytes_and_size(self, int32_t max_binary_size=-1):
```
chaokunyang: THis method doesn't have preallocation, we don't need to check size

eyad-hazem: Since we want to enforce `max_binary_size` for binary byte-length reads (as mentioned in issue), should I move the size checking to `read_bytes()` method before this line?
`cdef bytes py_bytes = PyBytes_FromStringAndSize(NULL, length)`

chaokunyang: So, we don't want to enforce binary byte-length reads, we only wanto to avoid malicouse length to introuce OOM. So if there is no preallocation, there should not be no such check. Because if no engought data to read from buffer, buffer read PAI will raise exception

chaokunyang: Please remove max_binary_size parameter to keep API surface minimal

chaokunyang: You can even add comments to clarify that those options are used for preallocation checks only

eyad-hazem: Looks good, but I want to ask about the line I mentioned before:
`cdef bytes py_bytes = PyBytes_FromStringAndSize(NULL, length)` in buffer `read_bytes` function
I searched and found that this call may result OOM crash or Memory Error before any checks.
So we can check bounds or ensure readability first. What do you see?

chaokunyang: I see, thanks. Could you update `Buffer` class with an extra constructore parameter, add `max_binary_size` instead? In this way, we can minimize the API surface change

eyad-hazem:
Could you make a check? @chaokunyang

python/pyfory/buffer.pxi
chaokunyang: should pass max_binary_size here too

python/pyfory/_fory.py
```diff
@@ -172,6 +173,7 @@ def __init__(
         policy: DeserializationPolicy = None,
         field_nullable: bool = False,
         meta_compressor=None,
+        max_collection_size: int = 1_000_000,
```
chaokunyang, line 177: Should also have max_binary_size as parameter too, and pass it to Buffer in fory instance

python/pyfory/serialization.pyx
```diff
@@ -1090,6 +1091,7 @@ cdef class Fory:
             max_depth: int = 50,
             field_nullable: bool = False,
             meta_compressor=None,
+            max_collection_size: int = 1_000_000,
```
chaokunyang, line 1095: Should also have max_binary_size as parameter too, and pass it to Buffer in fory instance

python/pyfory/buffer.pxi
eyad-hazem: Should I do so also in allocate class method?

PR #3432, opened 2026-02-27
fix(python): correct type dispatch in infer_field for custom classes

SURYAS1306:
## Why?

The current type dispatch logic in `infer_field` incorrectly treats certain user-defined classes as built-in types due to insufficient origin/type checks.  
This leads to incorrect field inference and causes Python serialization tests to fail.

## What does this PR do?

This PR refines the type dispatch logic in `infer_field` to properly distinguish between:

- Built-in primitives and standard library types
- Collection types (`list`, `dict`, `set`)
- User-defined classes

User-defined classes are now correctly routed through `visit_customized`, while primitives and supported types continue to use `visit_other`.

All existing Python tests pass after this fix.

## Related issues

None.

No.

## Benchmark

Not applicable - this change does not impact performance.

### DIFF
--- a/python/pyfory/format/tests/test_infer.py
+++ b/python/pyfory/format/tests/test_infer.py
@@ -64,22 +64,6 @@ def test_infer_field():
     assert result.type.id == TypeId.STRUCT
 
 
-def test_infer_field_unannotated_class():
-    """Classes without annotations should be treated as structs, not raise TypeError."""
-
-    class Empty:
-        pass
-
-    class WithInit:
-        def __init__(self):
-            self.x = 1
-            self.y = "hello"
-
-    # Should not raise TypeError and must be treated as STRUCT
-    assert _infer_field("", Empty).type.id == TypeId.STRUCT
-    assert _infer_field("", WithInit).type.id == TypeId.STRUCT
-
-
 def test_infer_field_builtin_types_not_treated_as_struct():
     """Built-in types must NOT be routed to visit_customized (regression guard)."""
     assert _infer_field("", int).type.id == TypeId.INT64


chaokunyang:
@SURYAS1306 Could you add tests for cases which type inference goes wrong?

python/pyfory/format/tests/test_infer.py
```diff
@@ -63,6 +63,39 @@ class X:
...
+    class Empty:
+        pass
+
+    class WithInit:
+        def __init__(self):
+            self.x = 1
+            self.y = "hello"
+
+    # Should not raise TypeError and must be treated as STRUCT
+    assert _infer_field("", Empty).type.id == TypeId.STRUCT
```
chaokunyang: This may be wrong, only dataclass should be taken as struct. The current struct seiralizer is DataclassSerializer and it only support dataclass too

SURYAS1306:
Hi @chaokunyang,
Thanks for the clarification. I’ve updated the tests to align with the dataclass-only STRUCT behavior and removed the incorrect unannotated class case.

Please let me know if you’d like additional edge cases to be covered.

python/pyfory/type_util.py
```diff
@@ -249,11 +249,17 @@ def infer_field(field_name, type_, visitor: TypeVisitor, types_path=None):
         else:
             raise TypeError(f"Collection types should be {list, dict} instead of {type_}")
     else:
-        if is_function(origin) or not hasattr(origin, "__annotations__"):
+        if is_function(origin):
```
chaokunyang, line 252: So I do not understand, what does this fix exactly?

SURYAS1306: Previously, types without `__annotations__` (including non-dataclass classes) were being routed through the same branch as functions due to the `or not hasattr(origin, "__annotations__")` condition.

This caused incorrect type dispatch for custom classes.

By removing the `not hasattr(origin, "__annotations__")` check, only actual functions are handled in that branch, and class types now follow the intended dispatch path.

Please let me know if I should clarify this further in the PR description.

PR #3455, opened 2026-03-04
feat(c++): add configurable deserialization size guardrails

### LINKED ISSUE #3408: [C++] configurable size guardrails for untrusted payloads
### Feature Request

Add configurable deserialization size guardrails in Fory C++ for untrusted payloads.

Add two configurable size limits to C++ deserialization and enforce them in relevant preallocation-sensitive read paths.

Resolve task:

### Additional context

Related locations:
- `cpp/fory/serialization/config.h:31`
- `cpp/fory/serialization/collection_serializer.h:396`
- `cpp/fory/serialization/collection_serializer.h:719`
- `cpp/fory/serialization/map_serializer.h:541`

shivendra-dev54:
## Why?

We currently don't have any size limits for incoming payloads in the C++ implementation. This is a security risk because a malicious or malformed payload can claim to have a massive collection or binary length, forcing the system to pre-allocate gigabytes of memory (via `.reserve()` or constructors) before actually reading the data. This makes the system vulnerable to simple Out-of-Memory (OOM) Denial-of-Service attacks.

## What does this PR do?

This PR adds two essential security guardrails to the deserialization path: `max_binary_size` and `max_collection_size`.

**Changes included:**

* **Config & API**: Added the two new limits to `serialization::Config` and updated `ForyBuilder` so users can easily set these at runtime. Defaults are 64MB for binary and 1M entries for collections.
* **Security Enforcement**:
* Integrated checks into all sensitive pre-allocation paths, including `std::vector`, `std::list`, `std::deque`, `std::set`, and `std::unordered_set`.
* Added entry-count validation for Maps (both fast and slow paths).
* Specifically handled arithmetic vectors by converting byte-lengths to element counts to ensure `max_collection_size` is respected.

* **Context Access**: Exposed a public `config()` accessor in `ReadContext` and `WriteContext` so internal serializers can reach these settings.
* **Tests**: Added new test cases in `collection_serializer_test.cc` and `map_serializer_test.cc` to verify that deserialization fails immediately with a descriptive error when limits are exceeded.

## Related issues

Fixes #3408

Yes, it adds two new methods (`max_binary_size` and `max_collection_size`) to the `ForyBuilder`.

* [x] Does this PR introduce any public API change?
* [ ] Does this PR introduce any binary protocol compatibility change?

## Benchmark

The performance impact is negligible. The checks are simple integer comparisons performed once per collection/binary read, occurring right before the expensive allocation phase. All 30 existing C++ test targets pass with no measurable change in execution time.

### DIFF
--- a/cpp/fory/serialization/collection_serializer.h
+++ b/cpp/fory/serialization/collection_serializer.h
@@ -925,6 +925,12 @@ template <typename Alloc> struct Serializer<std::vector<bool, Alloc>> {
     if (FORY_PREDICT_FALSE(ctx.has_error())) {
       return std::vector<bool, Alloc>();
     }
+
+    if (FORY_PREDICT_FALSE(size > ctx.config().max_binary_size)) {
+      ctx.set_error(Error::invalid_data("Binary size exceeds max_binary_size"));
+      return std::vector<bool, Alloc>();
+    }
+
     std::vector<bool, Alloc> result(size);
     // Fast path: bulk read all bytes at once if we have enough buffer
     Buffer &buffer = ctx.buffer();
--- a/cpp/fory/serialization/unsigned_serializer.h
+++ b/cpp/fory/serialization/unsigned_serializer.h
@@ -703,6 +703,12 @@ template <> struct Serializer<std::vector<uint8_t>> {
 
   static inline std::vector<uint8_t> read_data(ReadContext &ctx) {
     uint32_t length = ctx.read_var_uint32(ctx.error());
+
+    if (FORY_PREDICT_FALSE(length > ctx.config().max_binary_size)) {
+      ctx.set_error(Error::invalid_data("Binary size exceeds max_binary_size"));
+      return std::vector<uint8_t>();
+    }
+
     if (FORY_PREDICT_FALSE(length > ctx.buffer().remaining_size())) {
       ctx.set_error(
           Error::invalid_data("Invalid length: " + std::to_string(length)));
@@ -798,6 +804,12 @@ template <> struct Serializer<std::vector<uint16_t>> {
     if (FORY_PREDICT_FALSE(ctx.has_error())) {
       return std::vector<uint16_t>();
     }
+
+    if (FORY_PREDICT_FALSE(total_bytes > ctx.config().max_binary_size)) {
+      ctx.set_error(Error::invalid_data("Binary size exceeds max_binary_size"));
+      return std::vector<uint16_t>();
+    }
+
     if (total_bytes % sizeof(uint16_t) != 0) {
       ctx.set_error(Error::invalid_data("Invalid length: " +
                                         std::to_string(total_bytes)));
@@ -900,6 +912,12 @@ template <> struct Serializer<std::vector<uint32_t>> {
     if (FORY_PREDICT_FALSE(ctx.has_error())) {
       return std::vector<uint32_t>();
     }
+
+    if (FORY_PREDICT_FALSE(total_bytes > ctx.config().max_binary_size)) {
+      ctx.set_error(Error::invalid_data("Binary size exceeds max_binary_size"));
+      return std::vector<uint32_t>();
+    }
+
     if (total_bytes % sizeof(uint32_t) != 0) {
       ctx.set_error(Error::invalid_data("Invalid length: " +
                                         std::to_string(total_bytes)));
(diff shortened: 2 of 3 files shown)


cpp/fory/serialization/collection_serializer.h
```diff
@@ -611,15 +618,28 @@ struct Serializer<
...
+      ctx.set_error(Error::invalid_data("Binary size exceeds max_binary_size"));
+      return std::vector<T, Alloc>();
+    }
     if (sizeof(T) == 0) {
       return std::vector<T, Alloc>();
     }
+
+    // 2. Convert bytes to element count and check max_collection_size
+    size_t elem_count = total_bytes_u32 / sizeof(T);
+    if (FORY_PREDICT_FALSE(elem_count > ctx.config().max_collection_size)) {
```
chaokunyang: Could we use binary size for check instead? The collectionm with basic numeric types are taken as binary array?

shivendra-dev54: I have updated the arithmetic vector path to use `max_binary_size` instead of `max_collection_size`, as suggested. I've also updated the unit tests to verify this behavior.

cpp/fory/serialization/collection_serializer.h
```diff
@@ -677,6 +691,13 @@ struct Serializer<
     if (FORY_PREDICT_FALSE(ctx.has_error())) {
       return std::vector<T, Alloc>();
     }
+
+    if (FORY_PREDICT_FALSE(length > ctx.config().max_collection_size)) {
+      ctx.set_error(
+          Error::invalid_data("Collection length exceeds max_collection_size"));
+      return std::vector<T, Alloc>();
+    }
+
```
chaokunyang, line 700: Nice addition on the read() path. One gap remains: several read_data() paths still trust payload size without max_collection_size checks (for example vector<T> non-arithmetic, vector<bool>, list, deque, and forward_list). Those paths are used in native/nested deserialization, so guardrails can still be bypassed there.

cpp/fory/serialization/collection_serializer.h
```diff
@@ -1161,6 +1189,13 @@ template <typename T, typename Alloc> struct Serializer<std::deque<T, Alloc>> {
     if (FORY_PREDICT_FALSE(ctx.has_error())) {
       return std::deque<T, Alloc>();
     }
+
+    if (FORY_PREDICT_FALSE(length > ctx.config().max_collection_size)) {
+      ctx.set_error(
+          Error::invalid_data("Collection length exceeds max_collection_size"));
+      return std::deque<T, Alloc>();
+    }
+
```
chaokunyang, line 1218: There is still an unguarded preallocation in forward_list::read: temp.reserve(length) happens before any max_collection_size validation. A malicious length can force a large allocation before we fail. Please validate length immediately after reading it, before reserve.

cpp/fory/serialization/collection_serializer_test.cc
```diff
@@ -620,6 +620,47 @@ TEST(CollectionSerializerTest, ForwardListEmptyRoundTrip) {
   EXPECT_TRUE(deserialized.strings.empty());
 }
 
+// Test max_collection_size using objects (e.g., strings)
+TEST(CollectionSerializerTest, MaxCollectionSizeGuardrail) {
+  auto fory = Fory::builder()
```
chaokunyang: These tests are helpful, but they only cover xlang(true). The new limits should also be validated in xlang(false)/read_data paths, otherwise we can miss regressions in native mode (especially for vector/list/deque/forward_list).

cpp/fory/serialization/context.h
```diff
@@ -361,6 +361,9 @@ class WriteContext {
   /// reset context for reuse.
   void reset();
 
+  /// get associated configuration.
+  inline const Config &config() const { return *config_; }
```
chaokunyang: Read-side code now uses config(), but I do not see write-side usage of WriteContext::config(). If no write-path consumer is planned, we can drop this accessor to keep the API surface smaller.

shivendra-dev54: If it do not have any use then I think dropping it is a better idea

shivendra-dev54:
@chaokunyang 
I have gone through all the points you mentioned.
please check if I am wrong somewhere or there are any additional things i should include

chaokunyang:
Please also update unsigned_serializer.h and also handle `std::vector<bool>::read_data`

shivendra-dev54:
@chaokunyang 
Just as you said I have updated the `cpp/fory/serialization/unsigned_serializer.h` for the `uint8_t`, `uint16_t`, `uint32_t`, `uint64_t`
and also for `bool` type in `cpp/fory/serialization/collection_serializer.h`
and also I have added the tests for unsigned serializer in `cpp/fory/serialization/unsigned_serializer_test.cc`

PR #3456, opened 2026-03-07
perf(go): remove //go:inline directives and mark cold paths as //go:noinline

### LINKED ISSUE #3446: [BUG][GO]: //go:inline pragmas are no-ops — hot-path functions not being inlined
### Search before asking

### Version

Current

### Component(s)

Go

It was introduced in the PR #3071 

### Minimal reproduce step

Run `go build -gcflags='-m -m'`
Output - 
```bash
buffer.go:71:  cannot inline (*ByteBuffer).WriteByte:   cost 86 > budget 80
buffer.go:86:  cannot inline (*ByteBuffer).WriteInt8:   cost 85 > budget 80
buffer.go:114: cannot inline (*ByteBuffer).WriteUint32: cost 85 > budget 80
buffer.go:121: cannot inline (*ByteBuffer).WriteInt32:  cost 86 > budget 80
buffer.go:199: cannot inline (*ByteBuffer).ReadByte:    cost 94 > budget 80
buffer.go:251: cannot inline (*ByteBuffer).ReadUint32:  cost 96 > budget 80
buffer.go:264: cannot inline (*ByteBuffer).ReadUint64:  cost 96 > budget 80
```

## Summary

`buffer.go` uses `//go:inline` on functions. **This is not a valid Go compiler directive** — the standard `gc` compiler silently ignores it. As a result, critical hot-path functions like `WriteInt32`, `ReadByte`, `ReadUint64` etc. are **not being inlined**, paying full function call overhead on every invocation.

## Root Cause

- **Write functions**: the `grow()` call adds enough cost (~40) to push the total over 80.
- **Read functions**: the inline `BufferOutOfBoundError()` construction adds ~15–20 cost, pushing totals to 94–97.

## Proposed Fix

1. **Remove all `//go:inline` pragmas** — they are no-ops and misleading.
2. **Write path**: inline the fast-path bounds check directly and keep only `growSlow()` as `//go:noinline`:
   ```go
   func (b *ByteBuffer) WriteInt8(value int8) {
       if b.writerIndex+1 > len(b.data) {
           b.growSlow(1)
       }
       b.data[b.writerIndex] = byte(value)
       b.writerIndex++
   }
   ```
3. **Read path**: move error construction to a `//go:noinline` cold-path helper:
   ```go
   func (b *ByteBuffer) ReadInt8(err *Error) int8 {
       if b.readerIndex+1 > len(b.data) {
           b.readBoundsError(err, 1)
           return 0
       }
       v := int8(b.data[b.readerIndex])
       b.readerIndex++
       return v
   }

   //go:noinline
   func (b *ByteBuffer) readBoundsError(err *Error, n int) {
       *err = BufferOutOfBoundError(b.readerIndex, n, len(b.data))
   }
   ```

Both changes bring function costs well under 80, enabling **automatic** compiler inlining with no pragmas needed.

## Impact

These are the most frequently called functions in the entire serialization pipeline. Enabling inlining eliminates function call overhead (stack frame, argument spilling) on every primitive read/write operation.

ayush00git:
## Why?

//go:inline directives were introduced in the go runtime, expecting the gc compiler would inline that specific function. But //go:inline isn't a directive that is recognized by the gc compiler. gc compiler automatically inlines a function which costs less than 80.

## What does this PR do?

1. Removes all occurrences of //go:inline. 
2. Mark cold paths as //go:noinline.

## Related issues
Closes #3446 

## Benchmark
main branch -
image

this branch - 
image

### DIFF
--- a/go/fory/fory.go
+++ b/go/fory/fory.go
@@ -193,6 +193,7 @@ func validateUserTypeID(typeID uint32) error {
 // type_ can be either a reflect.Type or an instance of the type
 // typeID should be the user type ID in the range 0-0xfffffffe (0xffffffff is reserved for "unset").
 // Note: For enum types, use RegisterEnum instead.
+//
 //go:noinline
 func (f *Fory) RegisterStruct(type_ any, typeID uint32) error {
 	if err := validateUserTypeID(typeID); err != nil {
@@ -225,6 +226,7 @@ func (f *Fory) RegisterStruct(type_ any, typeID uint32) error {
 // type_ can be either a reflect.Type or an instance of the union type.
 // typeID should be the user type ID in the range 0-0xfffffffe (0xffffffff is reserved for "unset").
 // serializer must implement union payload encoding/decoding.
+//
 //go:noinline
 func (f *Fory) RegisterUnion(type_ any, typeID uint32, serializer Serializer) error {
 	if serializer == nil {
@@ -251,6 +253,7 @@ func (f *Fory) RegisterUnion(type_ any, typeID uint32, serializer Serializer) er
 // RegisterNamedUnion registers a union type with a namespace + type name for cross-language serialization.
 // type_ can be either a reflect.Type or an instance of the union type.
 // serializer must implement union payload encoding/decoding.
+//
 //go:noinline
 func (f *Fory) RegisterNamedUnion(type_ any, typeName string, serializer Serializer) error {
 	if serializer == nil {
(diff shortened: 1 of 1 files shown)


ayush00git:
Hii @chaokunyang 
I have a suggestion, in write paths if we just place the `grow()'s` logic directly, it would automatically reduce the cost and may inline many of them. 
Let me know should I implement this? i'll paste an proof as well

ayush00git:
Before  -
image


After adding the grow()'s logic to the write paths -
image

ayush00git:
Some of the functions cost is being reduced, let me know to do this or not?

chaokunyang:
Please share the benchmark results in apache/main and this branch

ayush00git:
main branch -
image


this branch - 
image

chaokunyang:
I see some cases has performance regression, such as fory structlist serialize. Could address those regression and share a compare report for fory?

ayush00git:
> I see some cases has performance regression, such as fory structlist serialize. Could address those regression and share a compare report for fory?

Hii @chaokunyang 
i think there's a confusion.
In main branch the parameter you mentioned has the value 1197 ns/op
And the same parameter in the fix/gc-flags has a value 983 ns/op.
And even every parameter's performance has significantly increased.

chaokunyang (approved):
LGTM

Issue #1203, opened 2023-12-03
[C++] Add sanitizer configuration in bazel build and CI


PragmaTwice:
We need to run unit test with address sanitizer and/or ub sanitizer enabled to discover memory bugs.

BaldDemian:
Hi @chaokunyang , I want to try this issue, could you please assign this to me? As for the sanitizer, would [AddressSanitizer](https://github.com/google/sanitizers/wiki/addresssanitizer) be a proper choice?

chaokunyang:
@BaldDemian Assigned to you

BaldDemian:
Hi @chaokunyang, I successfully integrate AddressSanitizer(ASan) and UndefinedBehaviorSanitizer(UBSan) in bazel and CI. But I encounter some issues and detect UB in the existing Fory code when integrating UBSan. You can see this [PR](https://github.com/apache/fory/pull/3458) for more details.
Thanks for your time and any feedback!

PR #3458, opened 2026-03-08
feat(c++): add sanitizer configuration in bazel build and CI

BaldDemian:
## Why?

Close https://github.com/apache/fory/issues/1203

## What does this PR do?

- Add ASan and UBSan integration in `.bazelrc`.
- I noticed that directly running `bazel test $(bazel query //cpp/...)` triggers compilation errors because the `--config=x86_64` flag is not provided. In the new CI Python script `run_ci.py`, the `cpp.py` task invoked by it indeed contains a check for this flag:
  https://github.com/apache/fory/blob/af6e8b25e8c2d188cba2d99ccb276c47ec6c9c31/ci/tasks/cpp.py#L40-L41
  The `run_ci.sh` script does not contain this check. Although this script appears to no longer be used, I still added the corresponding flag check there.
Similarly, I also added documentation about this flag in `AGENTS.md`.

- In `ci.yml`, I added a new job to enable tests with sanitizers in CI. Currently, this job only runs on ubuntu-latest.

## Related issues

https://github.com/apache/fory/issues/1203

## Benchmark

I believe this feature only introduces extra burden during testing.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -56,3 +56,16 @@ build:x86_64 --copt=-mbmi2
 # ARM64-specific optimizations (if any needed in the future)
 build:arm64 --copt=-march=armv8-a
 
+# AddressSanitizer
+build:asan --strip=never
+build:asan --copt=-fsanitize=address
+build:asan --copt=-g
+build:asan --copt=-fno-omit-frame-pointer
+build:asan --linkopt=-fsanitize=address
+
+# UndefinedBehaviorSanitizer
+build:ubsan --strip=never
+build:ubsan --copt=-fsanitize=alignment,bool,bounds,bounds-strict,builtin,enum,integer-divide-by-zero,object-size,pointer-overflow,return,shift,signed-integer-overflow,unreachable,vla-bound,vptr
+build:ubsan --copt=-g
+build:ubsan --copt=-fno-omit-frame-pointer
+build:ubsan --linkopt=-fsanitize=alignment,bool,bounds,bounds-strict,builtin,enum,integer-divide-by-zero,object-size,pointer-overflow,return,shift,signed-integer-overflow,unreachable,vla-bound,vptr
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -609,6 +609,69 @@ jobs:
             bazel-out/*/testlogs/**/*.xml
           if-no-files-found: ignore
 
+  cpp_sanitizers:
+    name: C++ Sanitizer (${{ matrix.sanitizer }})
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        sanitizer: [asan, ubsan]
+    steps:
+      - uses: actions/checkout@v5
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
+          cache: 'pip'
+      - name: Cache Bazel binary
+        uses: actions/cache@v4
+        with:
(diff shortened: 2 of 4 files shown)


BaldDemian:
Results

Local testing:
- ASan:
  Run `bazel test --cache_test_results=no --config=x86_64 --config=asan $(bazel query //...)`
  All tests passed, and no memory access errors were detected. Logs are available in this file: [asan_report.txt](https://github.com/user-attachments/files/25822903/asan_report.txt)
- UBSan:
  Run `bazel test --cache_test_results=no --config=x86_64 --config=ubsan $(bazel query //...)`
  Normally, when integrating UBSan, it is necessary to add the following line in `.bazelrc`: `build:ubsan --copt=-fsanitize=undefined`.
  However, with gcc, running the above `bazel test` command leads to a compilation failure.
  See this file for details: [ubsan_compiling_error.txt](https://github.com/user-attachments/files/25822911/ubsan_compiling_error.txt).
  This compilation failure appears to be caused by a known and unresolved bug in gcc. See   this discussion for reference: https://stackoverflow.com/questions/79775145/adding-fsanitize-undefined-in-gcc-makes-reflect-cpp-header-not-compile-anymore
  `undefined` is an aggregate sanitizer flag, and the list of its sub-options can be found here: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dundefined
  After some investigation, it seems that enabling the sub-options `null`, `nonnull-attribute`, and `returns-nonnull-attribute` will all trigger the compilation failure. The current workaround is to manually expand the sub-options included in `undefined` and exclude the three options above. With this configuration, the compilation error no longer occurs.
  Test results: several **misaligned address runtime errors** were indeed detected.
However, due to the default behavior of UBSan(continuing execution even when errors are reported), the final result still shows that all tests passed. From the log file 
[ubsan_report.txt](https://github.com/user-attachments/files/25822916/ubsan_report.txt)
(search for `runtime error`), the error appears to originate mainly from `buffer.h`. I will further investigate the code to check whether this is indeed a bug.

CI testing:
Newly added jobs and the existing ones all passed in my local fork: https://github.com/BaldDemian/fory/commit/95334b2bcd7ef8361b19398bd74869438a7dc7e8

BaldDemian:
[This sole failed CI job](https://github.com/apache/fory/actions/runs/22818663268/job/66187819482?pr=3458) seems to be a network issue 🤔. It passed in my local fork, so simply rerunning it may solve the problem.

ayush00git:
> [This sole failed CI job](https://github.com/apache/fory/actions/runs/22818663268/job/66187819482?pr=3458) seems to be a network issue 🤔. It passed in my local fork, so simply rerunning it may solve the problem.

pushing an empty commit would re-run the ci, try `git commit --allow-empty -m "ci"` and then push it

BaldDemian:
After some investigation, I can confirm that the misaligned address access UB mentioned above is caused by the way `buffer.h` creates types that require alignment using `reinterpret_cast` directly, e.g.
https://github.com/apache/fory/blob/ed586af222327bb2c3e7220832c991f211920005/cpp/fory/util/buffer.h#L885-L892
In this method, `int16` requires 2 byte alignment, but `reinterpret_cast` cannot guarantee or check this requirement.
I am preparing a PR to fix this issue.

BaldDemian:
I manually closed this PR since [an upstream PR](https://github.com/apache/fory/pull/3479) has been merged.

PR #3476, opened 2026-03-13
chore(java): Update Spotless and Checkstyle to allow building with JDK25

stevenschlansker:
## Why?

Spotless fails when run in JDK25:
```
Execution default-cli of goal com.diffplug.spotless:spotless-maven-plugin:2.41.1:apply failed: An API incompatibility was encountered while executing com.diffplug.spotless:spotless-maven-plugin:2.41.1:apply: java.lang.NoSuchMethodError: 'java.util.Queue com.sun.tools.javac.util.Log$DeferredDiagnosticHandler.getDiagnostics()'
```

## What does this PR do?

Updates build plugins to newer versions with JDK25 support

### DIFF
--- a/java/checkstyle.xml
+++ b/java/checkstyle.xml
@@ -220,7 +220,6 @@
                 value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF, METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>
     </module>
     <module name="JavadocMethod">
-      <property name="scope" value="public"/>
       <property name="allowMissingParamTags" value="true"/>
       <property name="allowMissingReturnTag" value="true"/>
       <property name="allowedAnnotations" value="Override, Test"/>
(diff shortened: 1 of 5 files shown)


java/pom.xml
```diff
@@ -245,7 +238,6 @@
         <configuration>
           <configLocation>${fory.java.rootdir}/checkstyle.xml</configLocation>
           <suppressionsLocation>${fory.java.rootdir}/checkstyle-suppressions.xml</suppressionsLocation>
-          <encoding>UTF-8</encoding>
```
stevenschlansker, line 248: > Warning:  Parameter 'encoding' is unknown for plugin 'maven-checkstyle-plugin:3.6.0:check (default-cli)'

java/checkstyle.xml
```diff
@@ -220,7 +220,6 @@
                 value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF, METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>
     </module>
     <module name="JavadocMethod">
-      <property name="scope" value="public"/>
```
stevenschlansker, line 223: >  cannot initialize module TreeWalker - cannot initialize module JavadocMethod - Property 'scope' does not exist, please check the documentation

java/pom.xml
stevenschlansker: (the default is source encoding, so this should be okay I think?)

Issue #3459, opened 2026-03-09
UBSan runtime error: load/store of misaligned address


BaldDemian:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fory/issues) and found no similar issues.

### Version

Fory: latest
OS: Ubuntu 24.04.3 LTS x86_64
GCC: 13.3.0

### Component(s)

C++

### Minimal reproduce step

Apply changes in [this PR](https://github.com/apache/fory/pull/3458) and run `bazel test --cache_test_results=no --config=x86_64 --config=ubsan $(bazel query //...)`

### What did you expect to see?

No UB should be detected by UBSan

### What did you see instead?

Several misaligned address load/store errors. See this log file for more details: https://github.com/user-attachments/files/25822916/ubsan_report.txt

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #3479, opened 2026-03-14
fix(c++): fix misaligned address access errors detected by UBSan in buffer.h

BaldDemian:
## Why?
`buffer.h` reads and writes multi-byte integers (16/32/64-bit) directly into a raw `uint8_t*` buffer. The original code cast the byte pointer to a **typed** pointer and dereferenced it: 

- Read:                                                                                                                                                                
  `return reinterpret_cast<const T*>(data_ + offset)[0];`                                                                                                                                                                                  
                                                                                                                                            
- Write:                                                                                                                               
  `*reinterpret_cast<T*>(data_ + offset) = value;`                                                                                          
                                                                                                                                            
Dereferencing a pointer that is not aligned to `alignof(T)` is considered UB in C++. UBSan correctly flagged these as misaligned address runtime errors, because `data_ + offset` can be at any byte boundary.


## What does this PR do?


Two helper templates were added to the `buffer.h`:

```c++
template <typename T>
FORY_ALWAYS_INLINE static T load_unaligned(const uint8_t *ptr) {
    T value;
    std::memcpy(&value, ptr, sizeof(T));
    return value;
}

template <typename T>
FORY_ALWAYS_INLINE static void store_unaligned(uint8_t *ptr, T value) {
    std::memcpy(ptr, &value, sizeof(T));
}
```

All `reinterpret_cast` calls that may lead to UB in the file were replaced with calls to these helpers.

No UB were detected when running `bazel test --cache_test_results=no --config=x86_64 --config=ubsan $(bazel query //...)` after applying this patch. 
Details can be found in [ubsan_report.txt](https://github.com/user-attachments/files/25990662/ubsan_report.txt).
Only a few `unused-but-set-parameter` warnings were detected by UBSan.

## Related issues



Fix https://github.com/apache/fory/issues/3459

## AI Contribution Checklist



## Benchmark


I believe replacing `reinterpret_cast` with `memcpy` won't incur much runtime burden, since `memcpy` can be optimized by both GCC and Clang effectively.


chaokunyang:
Does thos fix affect the performance? Could you run benchmarks/cpp And compare the performance

BaldDemian:
[after_README.md](https://github.com/user-attachments/files/25996210/after_README.md)
[before_README.md](https://github.com/user-attachments/files/25996211/before_README.md)
Hey @chaokunyang, here are the cpp benchmark results after and before this fix. From my understanding, there are no major fluctuations in the performance 🤔 (mostly within ±3%, with a maximum around 8%)

(in `benchmarks/cpp`, run `./run.sh --serializer fory`)

BaldDemian:
run `./fory_benchmark --benchmark_filter="Fory"` in `benchmarks/cpp/build`.
Before the commit: 
before
After the commit:
after

Still I don't think there is regression in the performance.

PR #3481, opened 2026-03-14
fix(rust): handle panics from fuzz-found edge cases

### LINKED ISSUE #3480: [Rust] several panics found by cargo-fuzz
### Search before asking

### Version

Fory: latest
OS: Ubuntu 24.04.3
GCC: 11.5.0

### Component(s)

Rust

### Minimal reproduce step

Clone https://github.com/BaldDemian/fory and change to the `cargo-fuzz` branch.
Change to `rust` dir and run
```shell
cargo +nightly fuzz run meta_string -- \                                                                     
    -max_total_time=30 -max_len=4096 \
    -fork=1 -ignore_crashes=1 -ignore_ooms=1 -ignore_timeouts=1
```
and 
```shell
cargo +nightly fuzz run field_info -- \                                                                     
    -max_total_time=30 -max_len=4096 \
    -fork=1 -ignore_crashes=1 -ignore_ooms=1 -ignore_timeouts=1
```

All following panics should be observed within 30 seconds fuzzing.

To my understanding, all input to a deserialization tool should be considered untrusted in case of [DoS](https://blog.trailofbits.com/2025/02/21/dont-recurse-on-untrusted-input/). Therefore, related code should return Err when encountering invalid or even **malicious** inputs, rather than directly panicking. However, by running cargo fuzz, I indeed found that some code paths panic under certain inputs.

### What did you see instead?

1. https://github.com/apache/fory/blob/8df6bc3db7c5b2e9584b8a4be743727c7025bd6d/rust/fory-core/src/meta/type_meta.rs#L279-L281
  Line 280 will panic if i16 add overflows.

2. https://github.com/apache/fory/blob/8df6bc3db7c5b2e9584b8a4be743727c7025bd6d/rust/fory-core/src/meta/type_meta.rs#L305-L307
  `decode` may return `Error` but line 307 doesn't propagate the error. It will panic if `unwrap` on `Error`.

3. https://github.com/apache/fory/blob/8df6bc3db7c5b2e9584b8a4be743727c7025bd6d/rust/fory-core/src/resolver/meta_string_resolver.rs#L41-L50
   This method will panic when matching the default case. Maybe we should change the return value type of this method from `Encoding` to `Result<Encoding, Error>`?

4. https://github.com/apache/fory/blob/8df6bc3db7c5b2e9584b8a4be743727c7025bd6d/rust/fory-core/src/resolver/meta_string_resolver.rs#L246
   If `len` is zero then this method will panic.

5. https://github.com/apache/fory/blob/8df6bc3db7c5b2e9584b8a4be743727c7025bd6d/rust/fory-core/src/resolver/meta_string_resolver.rs#L366-L368
   Similar to 2, the `Error` is not properly handled in line 368

### Anything Else?

Currently, testing of the Rust module in Fory mainly relies on manually constructed test cases. However, in serialization/deserialization-related projects, fuzzing is commonly used to test code robustness. I also noticed this:
https://github.com/apache/fory/blob/8df6bc3db7c5b2e9584b8a4be743727c7025bd6d/rust/Cargo.toml#L26-L28
Perhaps Fory previously had plans to integrate cargo-fuzz?

At the moment, I have only made a simple attempt to integrate cargo-fuzz and tested two modules. If fuzzing integration is indeed desired, we may also need to consider more advanced fuzzing features, such as more structured fuzzing supported by [arbitrary](https://github.com/rust-fuzz/arbitrary), or integrating [OSS-fuzz](https://github.com/google/oss-fuzz).

In addition, other modules such as the C++ implementation could also consider integrating fuzzing.

utafrali:
Closes #3480

Cargo-fuzz was hitting several panics in the serialization code. Switched from unwrap/unreachable to proper error handling:

- field_id calculation now uses checked_add to catch overflows instead of panicking
- field name decoding switched from unwrap() to Result propagation
- byte_to_encoding now returns Result for unknown encoding bytes instead of using unreachable!()
- MetaStringBytes::new() updated to return Result so callers can handle invalid inputs

All of these return the same error type so the call sites already handle it correctly.

## Why?
Fuzzing found inputs that triggered panics. Better to return an error than crash.

## What does this PR do?
Replaces panics with proper error returns in the metadata serialization paths.

## Related issues
#3480

## AI Contribution Checklist
N/A

No. Valid inputs behave identically. Invalid/malformed data now returns an error instead of panicking.

## Benchmark
N/A

### DIFF
--- a/rust/fory-core/src/meta/type_meta.rs
+++ b/rust/fory-core/src/meta/type_meta.rs
@@ -277,7 +277,9 @@ impl FieldInfo {
             // Field ID mode: | 0b11:2bits | field_id_low:4bits | nullable:1bit | track_ref:1bit |
             let mut field_id = ((header >> 2) & FIELD_NAME_SIZE_THRESHOLD as u8) as i16;
             if field_id == SMALL_FIELD_ID_THRESHOLD {
-                field_id += reader.read_varuint32()? as i16;
+                field_id = field_id
+                    .checked_add(reader.read_varuint32()? as i16)
+                    .ok_or_else(|| Error::invalid_data("field_id overflow"))?;
             }
 
             let mut field_type = FieldType::from_bytes(reader, false, Option::from(nullable))?;
@@ -302,9 +304,7 @@ impl FieldInfo {
 
             let field_name_bytes = reader.read_bytes(name_size)?;
 
-            let field_name = FIELD_NAME_DECODER
-                .decode(field_name_bytes, encoding)
-                .unwrap();
+            let field_name = FIELD_NAME_DECODER.decode(field_name_bytes, encoding)?;
             Ok(FieldInfo {
                 field_id: -1i16,
                 field_name: field_name.original,
--- a/rust/fory-core/src/resolver/meta_string_resolver.rs
+++ b/rust/fory-core/src/resolver/meta_string_resolver.rs
@@ -38,14 +38,16 @@ pub struct MetaStringBytes {
 
 const HEADER_MASK: i64 = 0xff;
 
-fn byte_to_encoding(byte: u8) -> Encoding {
+fn byte_to_encoding(byte: u8) -> Result<Encoding, Error> {
     match byte {
-        0 => Encoding::Utf8,
-        1 => Encoding::LowerSpecial,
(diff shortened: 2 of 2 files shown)


chaokunyang (approved):
LGTM

PR #3482, opened 2026-03-15
feat(dart): Introduce id based enum serialization

yash-agarwa-l:
## Why?
The Dart module, only supports ordinal enum serialization, where each enum
value is serialized as its position index (0, 1, 2...), which breaks in different
cases like adding or reordering values. This adds optional id based serialization
with so that we can provide unique id  to every while preserving ordinal serialization
as the default fallback.

## What does this PR do?

- Adds @ForyEnumId(int id) for enum values
- Registers ForyEnumId in analysis type detection
- Updates enum analysis to collect IDs and emit build-time warnings for:
         - partial @ForyEnumId usage
         - duplicate IDs
- Falls back to ordinal serialization when annotations are incomplete or duplicated
- Extends EnumSpec with an optional idToValue map
- Updates generated enum specs to include the ID map when valid
- Updates EnumSerializer to use ID-based read/write when the map is present, otherwise keep ordinal behavior
- Adds codegen and serializer tests

## AI Contribution Checklist

AI Usage Disclosure
AI assistance was used to suggest tests and for better warnings in enum_analyzer.
    dart/packages/fory/lib/src/codegen/analyze/impl/struct/enum_analyzer_impl.dart
    dart/packages/fory-test/test/

- [x] I included a completed [AI Contribution Checklist](https://github.com/apache/fory/blob/main/AI_POLICY.md#9-contributor-checklist-for-ai-assisted-prs) in this PR description and the required AI Usage Disclosure.
- [x] I can explain and defend all important changes without AI help.
- [x] I reviewed AI-assisted code changes line by line before submission.
- [x] I ran adequate human verification and recorded evidence (checks run locally or in CI, pass/fail summary, and confirmation I reviewed results).
- [x] I added/updated tests and specs where required.
- [x] I validated protocol/performance impacts with evidence when applicable.
- [x] I verified licensing and provenance compliance.

          `@ForyEnumId` is a new annotation users will write in their code
          `ForyEnumId` class is publicly accessible

## Benchmark
General struct performance remained similar because those benchmarks do not exercise enum serialization paths.

### DIFF
--- a/dart/packages/fory-test/test/datatype_test/enum_serializer_test.dart
+++ b/dart/packages/fory-test/test/datatype_test/enum_serializer_test.dart
@@ -106,7 +106,7 @@ void main() {
       check(value).equals(EnumWithIds.C);
     });
 
-    test('falls back to ordinal serialization when id mapping is absent', () {
+    test('uses ordinal serialization when no @ForyEnumId annotations are present', () {
       final EnumSerializer serializer = EnumSerializer(false, [
         EnumFoo.A,
         EnumFoo.B,
--- a/dart/packages/fory/lib/src/codegen/analyze/impl/struct/enum_analyzer_impl.dart
+++ b/dart/packages/fory/lib/src/codegen/analyze/impl/struct/enum_analyzer_impl.dart
@@ -70,7 +70,7 @@ class EnumAnalyzerImpl implements EnumAnalyzer {
 
     final String? idFieldName = _findIdField(enumElement);
     final Map<String, int> enumIds = <String, int>{};
-    final Map<int, String> usedIds = <int, String>{};
+    final Map<int, String> seenIds = <int, String>{};
     final List<String> duplicateIds = <String>[];
     final List<String> missingIdValues = <String>[];
     final List<String> outOfRangeIds = <String>[];
@@ -90,12 +90,12 @@ class EnumAnalyzerImpl implements EnumAnalyzer {
         continue;
       }
 
-      final String? firstValueWithId = usedIds[id];
+      final String? firstValueWithId = seenIds[id];
       if (firstValueWithId != null) {
         duplicateIds.add('$id for $firstValueWithId and ${enumField.name}');
         continue;
       }
-      usedIds[id] = enumField.name;
+      seenIds[id] = enumField.name;
       enumIds[enumField.name] = id;
     }
 


yash-agarwa-l:
Firstly, Thank you sm for the clarification about this in the discussions, it took some time to complete it.
Few Questions:

-    Shall I create an issue for this and tag it there? 
-    And I did a performance test by creating a different file specific to enums, shall I add that?

chaokunyang:
Could we also support this pattern?
```dart
@ ForyEnum()
enum UserRole {
  guest(0),
  member(1),
  admin(2);

  @ForyEnumId()
  final int code;

  const UserRole(this.code);
}
```

dart/packages/fory-test/lib/entity/enum_id_foo.dart
```diff
@@ -0,0 +1,51 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import 'package:fory/fory.dart';
+
+part '../generated/enum_id_foo.g.dart';
+
+@foryEnum
```
chaokunyang, line 24: use uppercase?

yash-agarwa-l:
> Could we also support this pattern?
> 
> ```dart
> @ ForyEnum()
> enum UserRole {
>   guest(0),
>   member(1),
>   admin(2);
> 
>   @ForyEnumId()
>   final int code;
> 
>   const UserRole(this.code);
> }
> ```

Yes, I had this in mind initially too. I went with per-value annotation because I feel the id is a metadata for values and since Dart enums don't support private fields, so the `foryId` ends up publicly accessible in user's code which felt a bit annoying. But I agree the field-based approach looks much cleaner for long enums. I'd be happy to change it if you prefer.

dart/packages/fory-test/lib/entity/enum_id_foo.dart
yash-agarwa-l: Yes, we should do it as per conventions, but the existing files also follow the same pattern. If you prefer, shall I make a separate PR for that first before this one?

chaokunyang:
Sometimes users also want to use that id too. So I think we could support both. Could you support both in this pr?

yash-agarwa-l:
Yes, we can do that! I have updated the PR to support both methods and added the relevant tests.

dart/packages/fory/lib/src/codegen/analyze/impl/struct/enum_analyzer_impl.dart
```diff
@@ -17,26 +17,103 @@
...
+        enumFields.map((FieldElement e) => e.name).toList();
+
+    final String? idFieldName = _findIdField(enumElement);
+    final Map<String, int> enumIds = <String, int>{};
+    final Map<int, String> usedIds = <int, String>{};
+    final List<String> duplicateIds = <String>[];
+    final List<String> missingIdValues = <String>[];
+
+    for (final FieldElement enumField in enumFields) {
+      final int? id = idFieldName != null
```
chaokunyang, line 79: Validate @ForyEnumId range before generating enum specs

This accepts any `int` from `@ForyEnumId` and later feeds it to `writeVarUint32Small7(...)`. Negative values do not fail cleanly here: I reproduced `writeVarUint32Small7(-1)` decoding back as `34359738367`, so `@ForyEnumId(-1)` silently writes a different wire id than the one the user declared. Please reject ids outside the unsigned 32-bit range during analysis, and ideally keep a serializer-side assertion as defense in depth.

dart/packages/fory-test/test/datatype_test/enum_serializer_test.dart
```diff
@@ -0,0 +1,191 @@
...
+    DeserializationDispatcher.I,
+    DeserializationRefResolver.getOne(false),
+    _typeResolver,
+    Stack<TypeSpecWrap>(),
+  );
+}
+
+void main() {
+  group('Enum serializer', () {
+    test('writes and reads annotated enum ids when all values are annotated',
```
chaokunyang, line 77: Add a public Fory round-trip for annotated enums

These tests prove the generated `EnumSpec` shape and the direct `EnumSerializer` behavior, but they still do not exercise the public `Fory.serialize` / `Fory.deserialize` path with an annotated enum. That leaves a gap around the real registration and resolver wiring for this new feature. Please add at least one end-to-end round-trip test for an enum that uses `@ForyEnumId`.

yash-agarwa-l:
Thanks for the review! Now, For the codegen path, if Ids are out of range it fall back to ordinal serialization with a warning while `EnumSerializer.write()` throw a `RangeError` for manually constructed specs. 
Is this fine? Or shall we also give a hard error through the codegen?

chaokunyang:
> Thanks for the review! Now, For the codegen path, if Ids are out of range it fall back to ordinal serialization with a warning while `EnumSerializer.write()` throw a `RangeError` for manually constructed specs. Is this fine? Or shall we also give a hard error through the codegen?

Need strict errors instead

yash-agarwa-l:
I've added the round-trip tests and updated for out-of-range. Should `missingIdValues` and `duplicateIds` also be strict? I think they cause a similar problem.

chaokunyang:
> I've added the round-trip tests and updated for out-of-range. Should `missingIdValues` and `duplicateIds` also be strict? I think they cause a similar problem.

Yes. missingIdValues and duplicateIds should also be strict.

The rule should be:

- No @ForyEnumId anywhere: use ordinal mode.
- Any @ForyEnumId present: the whole enum ID configuration must be valid, or codegen should fail.

yash-agarwa-l:
Thanks! I have made the revision.

PR #3483, opened 2026-03-15
fix(rust): fix several panics detected by cargo-fuzz

BaldDemian:
## Why?

Fix several new panics when feeding corner-case input found by cargo-fuzz

## What does this PR do?
- In `rust/README.md`, the right command to run all tests seems to be `cargo test --workspace`. Run `cargo test --features tests` will get: 
  Screenshot 2026-03-15 at 6 16 51 AM

- In `rust/fory-core/src/meta/type_meta.rs`, 
  - https://github.com/apache/fory/blob/5fc06f1db45337346db4ed380906c013f1e2f3f7/rust/fory-core/src/meta/type_meta.rs#L645
    will panic if `encoding_idx` exceeds the size of `encodings`.
  - https://github.com/apache/fory/blob/5fc06f1db45337346db4ed380906c013f1e2f3f7/rust/fory-core/src/meta/type_meta.rs#L836
    will cause OOM if `num_fields` is too large. I limit the max value of `num_fields` to `i16::MAX` since `field_id` is `i16`

- In `rust/fory-core/src/row/bit_util.rs`, use saturating_add/mul to prevent potential overflow panic. But would it be better to return error instead of saturating_add/mul ?🤔

- In `rust/fory-core/src/row/reader.rs`, direct access into slice using `[]` may cause out-of-bounds panic.
  
- In `rust/fory-core/src/serializer/collection.rs`, `rust/fory-core/src/serializer/map.rs` and `rust/fory-core/src/serializer/primitive_list.rs`, we should check the remaining bytes in the buffer **before** allocating `Vec`. This can also prevent OOM.

- In `rust/fory-core/src/serializer/skip.rs`, `generics.first().unwrap()` and `generics.get(1).unwrap()` will panic if the size of `generics` is not long enough.

## Related issues
N/A

## AI Contribution Checklist

N/A

N/A

## Benchmark

This PR only adds additional check in case of corner-case input and thus won't has major influence on the performance.

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -670,7 +670,7 @@ cargo build
 
 ```bash
 # Run all tests
-cargo test --features tests
+cargo test --workspace
 
 # Run specific test
 cargo test -p tests --test test_complex_struct
--- a/rust/fory-core/src/meta/type_meta.rs
+++ b/rust/fory-core/src/meta/type_meta.rs
@@ -60,6 +60,7 @@ use std::collections::HashMap;
 use std::rc::Rc;
 
 const SMALL_NUM_FIELDS_THRESHOLD: usize = 0b11111;
+const MAX_TYPE_META_FIELDS: usize = i16::MAX as usize;
 const REGISTER_BY_NAME_FLAG: u8 = 0b100000;
 const FIELD_NAME_SIZE_THRESHOLD: usize = 0b1111;
 /// Marker value in encoding bits to indicate field ID mode (instead of field name)
@@ -642,6 +643,9 @@ impl TypeMeta {
             length as usize
         };
         let bytes = reader.read_bytes(length)?;
+        if encoding_idx as usize >= encodings.len() {
+            return Err(Error::invalid_data("encoding_index out of bounds"));
+        }
         let encoding = encodings[encoding_idx as usize];
         decoder.decode(bytes, encoding)
     }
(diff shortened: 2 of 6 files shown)


rust/fory-core/src/row/bit_util.rs
```diff
@@ -18,5 +18,5 @@
 const WORD_SIZE: usize = 8;
 
 pub fn calculate_bitmap_width_in_bytes(num_fields: usize) -> usize {
-    ((num_fields + 63) / 64) * WORD_SIZE
+    (num_fields.saturating_add(63) / 64).saturating_mul(WORD_SIZE)
```
chaokunyang: This change behaviour, I think you should let it crash. Such crash is basically programming error

rust/fory-core/src/row/reader.rs
```diff
@@ -103,10 +121,17 @@ pub struct MapViewer<'r> {
...
     pub fn new(row: &'r [u8]) -> MapViewer<'r> {
-        let key_byte_size = LittleEndian::read_u64(&row[0..8]) as usize;
+        let Some(header) = row.get(0..8) else {
+            return MapViewer {
+                key_row: &[],
+                value_row: &[],
+            };
+        };
+        let key_byte_size = LittleEndian::read_u64(header) as usize;
+        let key_end = (8usize).saturating_add(key_byte_size).min(row.len());
```
chaokunyang: ditto here.

chaokunyang: For row format, it's not used for rpc. it's mostly used in data processing cases. So we don't need to handle malicous/invalid data.

BaldDemian:
Thanks for your feedback, the mentioned changes are removed.

chaokunyang (approved):
LGTM

PR #3487, opened 2026-03-16
feat(cpp): add float16 to c++

### LINKED ISSUE #3208: [C++] add float16 to c++
### Feature Request

Add full IEEE 754 half-precision floating point (`float16`, binary16) support to Fory/FDL, including a complete **C++** runtime implementation that uses a strong type **`float16_t`** (internally storing **`uint16_t` bits**), plus exhaustive tests for conversion/rounding/NaN/subnormal semantics.

C++ does not have a universally-available standard `float16` primitive type (support varies across compilers/architectures). To guarantee portability and consistent cross-language semantics, we need a first-party implementation that:
- stores IEEE 754 binary16 exactly (2 bytes),
- has well-defined IEEE conversion/rounding behavior,
- exposes a **single strong type** `float16_t` in public APIs (no passing raw `uint16_t` bits around).

#### 1) FDL / Type System
  - message fields
  - repeated fields

- Encode `float16` as **2 bytes** representing the raw IEEE 754 binary16 bit pattern.

#### 3) C++ Runtime (core requirement): `float16_t` strong type only
Provide a public strong type `float16_t`. **All runtime APIs must accept/return `float16_t` only** (no `uint16_t` parameters in public APIs).

##### 3.1 Type definition
- Provide a trivial, standard-layout type:
  - `struct float16_t { uint16_t bits; };`
- Storage remains exactly 16 bits (2 bytes), with no heap allocation.
- Provide explicit bit access only via named methods (still returning `uint16_t` is OK, but APIs should pass `float16_t`):
  - `uint16_t to_bits() const;`
  - `static float16_t from_bits(uint16_t bits);`

- `static float16_t from_float(float f);`
  - Convert float32 -> binary16 using IEEE 754 rules and **round-to-nearest, ties-to-even**.
  - Must correctly handle:
    - ±Inf
    - ±0 (preserve sign)
    - normalized values
    - overflow -> ±Inf
    - underflow -> subnormal/±0
- `float to_float() const;`
  - Convert binary16 -> float32 (exact for all half values).

All take `float16_t`:
- `static bool is_nan(float16_t h);`
- `static bool is_inf(float16_t h);` and/or `static bool is_inf(float16_t h, int sign);` (sign: +1/-1/0)
- `static bool is_zero(float16_t h);` (treat +0/-0 as zero)
- `static bool signbit(float16_t h);`
- `static bool is_subnormal(float16_t h);`
- `static bool is_normal(float16_t h);`
- `static bool is_finite(float16_t h);`

C++ supports operator overloading, we should add operator overloading, and add explicit functions/methods so semantics are obvious and consistent with other languages:

- `static float16_t add(float16_t a, float16_t b);`
- `static float16_t sub(float16_t a, float16_t b);`
- `static float16_t mul(float16_t a, float16_t b);`
- `static float16_t div(float16_t a, float16_t b);`
- `static float16_t neg(float16_t a);`
- `static float16_t abs(float16_t a);`

Optional math parity (if needed by users):
- `sqrt`, `min`, `max`, `copysign`, `floor`, `ceil`, `trunc`, `round`, `round_to_even`

Implementation rule for arithmetic (unless full half-FPU emulation is desired):
- compute in float32 and round back to half each operation:
  - `return from_float(a.to_float() op b.to_float());`

##### 3.5 Comparisons
Provide IEEE-consistent comparisons using `float16_t` only:
- `static bool equal(float16_t a, float16_t b);`
  - NaN != NaN
  - +0 == -0
- `static bool less(float16_t a, float16_t b);`
  - if NaN involved => false
- `static bool less_eq(float16_t a, float16_t b);`
- `static bool greater(float16_t a, float16_t b);`
- `static bool greater_eq(float16_t a, float16_t b);`
- `static int compare(float16_t a, float16_t b);`
  - Define NaN behavior explicitly (unordered). If a total ordering is needed, add:
    - `static int total_order_compare(float16_t a, float16_t b);` (optional)

##### 3.6 Formatting
- `static std::string to_string(float16_t h);` (format via `h.to_float()`)

##### 3.7 Codegen requirement (C++)
- Generated C++ fields for `float16` must use **`float16_t`**, not `uint16_t`.
- Repeated `float16` should use `std::vector<float16_t>` (or existing Fory container type).

##### 3.8 Serialization boundary (still uses bits, but not in public API)
- Serializer/deserializer should read/write the underlying 16-bit pattern.
- It is acceptable for the codec implementation (internal) to access `h.to_bits()` / `from_bits(...)`, but **public helper APIs must not take raw `uint16_t`**.

#### 4) Compiler / Reflection Integration
- Ensure reflection/dynamic serialization recognizes `float16_t` as the `float16` primitive (distinct from `uint16_t` integer).
- Clarify schema evolution:

#### 5) Codegen & Type Mappings Across Languages
- C++: `float16_t` strong type with `uint16_t` internal bits storage and IEEE conversions.
- Other languages: implement mappings in future PRs, but wire format must remain IEEE binary16 bits.

1. **Conversion tests (C++)**
   - ±0, ±Inf, NaN
   - max finite 65504
   - min normal 2^-14
   - min subnormal 2^-24
   - overflow -> Inf, underflow -> subnormal/0
     - `h = float16_t::from_bits(bits);`
     - `h2 = float16_t::from_float(h.to_float());`

   - Ensure wire output matches expected 16-bit patterns for known values.

1) Store `float16` as `float` in C++ and convert to `float16` only during serialization.

2) Rely on compiler-specific half types (`__fp16`, `_Float16`) directly in public APIs.
- Rejected: portability concerns and inconsistent behavior across platforms; we still want a stable `float16_t` strong type with `uint16_t` storage.

3) Use third-party half libraries.

### Additional context

#3099

UninspiredCarrot:
## Why?

Implement float16_t (IEEE 754 binary16 / half-precision) as a primitive type in the C++ runtime, as required by issue https://github.com/apache/fory/issues/3208. No C++ standard type represents float16, so the framework needs its own strong type with correct IEEE 754 semantics and serialiser integration.

## What does this PR do?

cpp/fory/util/float16.h / float16.cc — new fory::float16_t strong type:

- Trivial, standard-layout, exactly 2 bytes; internal storage is uint16_t bits accessed only via to_bits()/from_bits()
- from_float / to_float — IEEE 754 compliant conversion with round-to-nearest ties-to-even, correct handling of ±0, ±Inf, NaN (payload preserved, signaling→quiet), subnormals, overflow→Inf, underflow→subnormal/±0
- Classification: is_nan, is_inf (two overloads), is_zero, signbit, is_subnormal, is_normal, is_finite
- Arithmetic: add, sub, mul, div, neg, abs; optional math: sqrt, min, max, copysign, floor, ceil, trunc, round, round_to_even; compound assignment and binary operator overloads
- Comparisons: equal, less, less_eq, greater, greater_eq, compare (NaN unordered, +0 == −0); comparison operator overloads (==, !=, <, <=, >, >=)

cpp/fory/serialization/struct_serializer.h — serializer integration:
- Serializer<float16_t> specialization wired to TypeId::FLOAT16 (type ID 17)

cpp/fory/util/float16_test.cc — exhaustive tests (1300+ lines, 61 test cases):
- Stress-tests all 65 536 bit patterns for round-trip correctness
- Ties-to-even rounding, subnormal gradual underflow, overflow→Inf, NaN payload preservation
- Buffer wire-format goldens (little-endian), serializer round-trips (scalar, vector, map, optional), type ID check
- Full comparison test suite including NaN unordered and ±0 equality edge cases

## Related issues

- Closes #3208 

## AI Contribution Checklist
- [X] Substantial AI assistance was used in this PR:  `yes`
- [X] If `yes`, I included a completed [AI Contribution Checklist](https://github.com/apache/fory/blob/main/AI_POLICY.md#9-contributor-checklist-for-ai-assisted-prs) in this PR description and the required `AI Usage Disclosure`.
  - [X] Substantial AI assistance was used in this PR: yes
  - [X] If yes, I included a completed AI Contribution Checklist in this PR description and the required AI Usage Disclosure.
  - [X] If yes, I can explain and defend all important changes without AI help.
  - [X] If yes, I reviewed AI-assisted code changes line by line before submission.
  - [X] If yes, I ran adequate human verification and recorded evidence (checks run locally or in CI, pass/fail summary, and confirmation I reviewed results).
  - [X] If yes, I added/updated tests and specs where required.
  - [X] If yes, I validated protocol/performance impacts with evidence when applicable.
  - [X] If yes, I verified licensing and provenance compliance. 
  
AI Usage Disclosure
  - scope: tests
  - affected_files_or_subsystems: cpp/fory/util/float16_test.cc
  - human_verification: bazel test //cpp/fory/util:float16_test — 61/61 passed locally; contributor reviewed all diffs line by line
  - performance_verification: N/A
  - provenance_license_confirmation: Apache-2.0-compatible provenance confirmed; no incompatible third-party code introduced

- [X] Does this PR introduce any public API change?
- [X] Does this PR introduce any binary protocol compatibility change?

### DIFF
--- a/compiler/examples/service.fdl
+++ b/compiler/examples/service.fdl
@@ -1,0 +1,31 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package demo.service;
+
+message HelloRequest {
+    string name = 1;
+}
+
+message HelloReply {
+    string message = 1;
+}
+
+service Greeter {
+    rpc SayHello (HelloRequest) returns (HelloReply);
+}
+
--- a/compiler/fory_compiler/ir/validator.py
+++ b/compiler/fory_compiler/ir/validator.py
@@ -386,6 +386,17 @@ class SchemaValidator:
             for f in union.fields:
                 check_type_ref(f.field_type, f, None)
 
+        # Also validate service RPC request/response type references.
+        for service in self.schema.services:
+            for method in service.methods:
+                for named_type in (method.request_type, method.response_type):
+                    type_name = named_type.name
+                    if self.schema.get_type(type_name) is None:
+                        self._error(
+                            f"Unknown type '{type_name}'",
+                            named_type.location,
+                        )
+
     def _check_ref_rules(self) -> None:
         def is_any_type(field_type: FieldType) -> bool:
             return (
--- a/compiler/fory_compiler/tests/test_fbs_service.py
+++ b/compiler/fory_compiler/tests/test_fbs_service.py
@@ -20,6 +20,7 @@
 from fory_compiler.frontend.fbs.lexer import Lexer
 from fory_compiler.frontend.fbs.parser import Parser
 from fory_compiler.frontend.fbs.translator import FbsTranslator
+from fory_compiler.ir.validator import SchemaValidator
 
 
 def parse_and_translate(source):
@@ -107,3 +108,43 @@ def test_streaming_attributes():
     assert m3.name == "BidiStream"
     assert m3.client_streaming is True
     assert m3.server_streaming is True
+
+
+def test_service_unknown_request_type_fails_validation():
+    source = """
+    namespace demo;
+
+    table Response {
+        result: string;
+    }
+
+    rpc_service Greeter {
+        SayHello(UnknownRequest):Response;
+    }
+    """
+    schema = parse_and_translate(source)
+    validator = SchemaValidator(schema)
+    assert not validator.validate()
+    assert any(
+        "Unknown type 'UnknownRequest'" in err.message for err in validator.errors
+    )
+
+
+def test_service_unknown_response_type_fails_validation():
+    source = """
+    namespace demo;
+
+    table Request {
+        id: int;
+    }
+
+    rpc_service Greeter {
+        SayHello(Request):UnknownResponse;
+    }
+    """
+    schema = parse_and_translate(source)
+    validator = SchemaValidator(schema)
+    assert not validator.validate()
+    assert any(
+        "Unknown type 'UnknownResponse'" in err.message for err in validator.errors
+    )
--- a/compiler/fory_compiler/tests/test_fdl_service.py
+++ b/compiler/fory_compiler/tests/test_fdl_service.py
@@ -17,12 +17,20 @@
 
 import pytest
 from fory_compiler.frontend.fdl.parser import Parser, ParseError
+from fory_compiler.ir.validator import SchemaValidator
 
 
 def parse(source: str):
     parser = Parser.from_source(source)
     schema = parser.parse()
     return schema
+
+
+def validate(source: str) -> SchemaValidator:
+    schema = parse(source)
+    validator = SchemaValidator(schema)
+    validator.validate()
+    return validator
 
 
 def test_empty_service():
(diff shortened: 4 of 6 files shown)


cpp/fory/util/float16.h
```diff
@@ -0,0 +1,265 @@
...
+
+#include <cmath>
+#include <cstdint>
+#include <string>
+#include <type_traits>
+namespace fory {
+
+// A 16-bit floating point representation with 1 sign bit, 5 exponent bits, and
+// 10 mantissa bits.
+struct float16_t {
```
chaokunyang, line 30: Please add a `std::hash<fory::float16_t>` specialization alongside the comparison operators. Right now `std::unordered_map<float16_t, ...>` / `std::unordered_set<float16_t>` do not compile at all; a minimal local compile probe fails immediately because the default `std::hash` is deleted.

cpp/fory/util/float16_test.cc
```diff
@@ -0,0 +1,1349 @@
...
+    auto deser =
+        fory.deserialize<Float16Scalar>(ser.value().data(), ser.value().size());
+    ASSERT_TRUE(deser.ok()) << "deserialize failed bits=0x" << std::hex << bits
+                            << ": " << deser.error().to_string();
+    EXPECT_EQ(deser.value(), original)
+        << "round-trip mismatch bits=0x" << std::hex << bits;
+  }
+}
+
+TEST(Float16SerializerTest, VectorRoundTrip) {
```
chaokunyang, line 1157: This round-trip only proves C++ can serialize and deserialize the same struct locally. It does not catch the wire-format regression here: `std::vector<float16_t>` still falls through the non-arithmetic vector serializer and emits `LIST`, because `std::is_arithmetic_v<float16_t>` is false. Please either wire `float16_t` into the typed-array path or assert here that the emitted type id is `FLOAT16_ARRAY`.

cpp/fory/util/float16_test.cc
```diff
@@ -0,0 +1,1349 @@
...
+  };
+  auto ser = fory.serialize(original);
+  ASSERT_TRUE(ser.ok()) << ser.error().to_string();
+  auto deser =
+      fory.deserialize<Float16Vector>(ser.value().data(), ser.value().size());
+  ASSERT_TRUE(deser.ok()) << deser.error().to_string();
+  EXPECT_EQ(deser.value(), original);
+}
+
+TEST(Float16SerializerTest, EmptyVectorRoundTrip) {
```
chaokunyang, line 1183: Please add a fixed-size array case in this serializer test block as well. `std::array<float16_t, N>` is still unsupported today, because the generic primitive-array serializer is gated on `std::is_arithmetic_v<T>` and there is no explicit `float16_t` specialization.

UninspiredCarrot:
> LGTM

thanks for your patience and reviewing, i was going to ping at some point but thank you for already taking a look at it. apologies for the earlier mistakes, im still getting used to open source and was struggling with the version control as im sure you can tell with the messy commit history

Issue #3484, opened 2026-03-15
[Question] Rust std::time::Duration is unsignd but the spec allows negative duration


BaldDemian:
### Question

According to Fory's [xlang spec](https://github.com/apache/fory/blob/main/docs/specification/xlang_serialization_spec.md#duration), the `seconds` field of a Duration is:                    
> encoded as a signed varint64. Can be positive or negative.

However, Fory's Rust implementation uses `std::time::Duration`, which is **unsigned** and therefore cannot represent negative durations.

https://doc.rust-lang.org/src/core/time.rs.html#82

The deserialization code in `rust/fory-core/src/serializer/datetime.rs` currently does:

`let secs = context.reader.read_varint64()? as u64;`

`read_varint64()` correctly decodes the ZigZag-encoded wire bytes back to an i64. However, the u64 cast is only a bitwise reinterpretation which doesn't check the sign. A value of `-1i64` becomes `u64::MAX`, resulting in a Duration of approximately 584 billion years.

Also, the `nanos` field is allowed to be negative in the spec, but Rust's Nanosecond type is also unsigned.

Maybe we should replace Rust's `std::time::Duration` to `chrono::Duration`, whose `seconds` and `nanos` are all signed: https://docs.rs/chrono/latest/src/chrono/time_delta.rs.html#64

BaldDemian:
I created two files in my fork to illustrate the above scenario with an ene-to-end Fory usage. 
To reproduce:
1. `git clone git@github.com:BaldDemian/fory.git`
2. `git checkout negative-durarion`
3. `cd examples/java`
4. `mvn -q compile exec:java -Dexec.mainClass=NegativeDurationDemo`
5. `cd ../../rust`
6. `cargo run -p fory --example deserialize_duration`

The Java file simply encodes a `Duration.ofSeconds(-1)` using Fory's xlang protocol, flush the bytes to a temp file; then the Rust file reads the bytes from the same file, decodes them, and prints the obtained `std::time::Duration`.

BaldDemian:
[...]

PR #3490, opened 2026-03-17
fix(rust): add error handling logic for std duration; add support for chrono duration

BaldDemian:
## Why?

The default duration type in Rust is `std::time::Duration`, which is unsigned, causing potential problems mentioned in https://github.com/apache/fory/issues/3484.

## What does this PR do?

For compatibility purposes, code related to `std::time::Duration` is not removed, instead I add some error handling logic around it.

I add write/read logic for `chrono::Duration` similar to `std::time::Duration`. It is worth noting that the maximum value of `seconds` in `chrono::Duration` is actually `i64::MAX/100`, smaller than what the spec allows, but I think this is fine due to the following reasons:
1. `i64::MAX/100` is already large enough for real use cases
2. duration types in C++, C# module in Fory also have maximum limits less than `i64::MAX`

## Related issues
Close https://github.com/apache/fory/issues/3484

## AI Contribution Checklist

No

No. Code related to `std::time::Duration` is not removed

## Benchmark

Existing benchmark code doesn't cover `std::time::Duration`.

### DIFF
--- a/rust/fory-core/src/serializer/datetime.rs
+++ b/rust/fory-core/src/serializer/datetime.rs
@@ -24,7 +24,7 @@ use crate::serializer::ForyDefault;
 use crate::serializer::Serializer;
 use crate::types::TypeId;
 use crate::util::EPOCH;
-use chrono::{NaiveDate, NaiveDateTime};
+use chrono::{Duration as ChronoDuration, NaiveDate, NaiveDateTime};
 use std::mem;
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


rust/fory-core/src/serializer/datetime.rs
```diff
@@ -243,4 +340,52 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_chrono_duration_serialization() {
+        let fory = Fory::default();
+
+        // Test various durations
+        let test_cases = vec![
```
chaokunyang, line 349: could you add some negative duration here? and also add some edge cases such as max/min allowed values

BaldDemian: done

BaldDemian:
BTW, at first I didn't notice that there is no benchmark case covering Duration so I ran the whole `cargo bench`. When hitting the `8_threads_shared_fory` case, error occurred. The error can be stably reproduced in `benchmarks/rust` and run `cargo -q bench --bench threaded_bench -- "ThreadedSerialization/8_threads_shared_fory" --exact --noplot`

Screenshot 2026-03-18 at 12 11 19 AM

I may spend some time finding out why.

chaokunyang (approved):
LGTM

Issue #3572, opened 2026-04-15
Apache Fory regression bug - 0.11.1 version to 0.16.0 : cannot serialise cyclic object graph anymore


PiotrDuz:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/fory/issues) and found no similar issues.

### Version

Hello, 
We were using apache fory 0.11.1 successfully. After updating to 0.16.0 , we no longer can serialise our complex cyclic object graph. Apache fory was chosen specifically because it had that capability. 
I am reporting is as a bug because I hope the functionality can be bring back. 

### Component(s)

Java

### Minimal reproduce step

Reproduction example is here: 
https://github.com/PiotrDuz/fory-compiler-error-repro
Just please run a test. 

### What did you expect to see?

I want the object to be serialised and deserialised properly. 
In reproduction example, test should be green.

### What did you see instead?

The exception in auto-generated code: 

`Caused by: org.apache.fory.shaded.org.codehaus.commons.compiler.CompileException: File 'io/cashforce/ng/platformapi/data/masterdata/mdsnapshot/ReproNodeForyRefCodec_0.java', Line 387, Column 9: Assignment conversion not possible from type "java.lang.Enum" to type "io.cashforce.ng.platformapi.data.masterdata.mdsnapshot.ReproType"
	at org.apache.fory.shaded.org.codehaus.janino.UnitCompiler.compileError(UnitCompiler.java:13228)
	at org.apache.fory.shaded.org.codehaus.janino.UnitCompiler.assignmentConversion(UnitCompiler.java:11445)
	at org.apache.fory.shaded.org.codehaus.janino.UnitCompiler.compile2(UnitCompiler.java:4186)
	at org.apache.fory.shaded.org.codehaus.janino.UnitCompiler.access$6300(UnitCompiler.java:240)
	at org.apache.fory.shaded.org.codehaus.janino.UnitCompiler$13.visitAssignment(UnitCompiler.java:4143)
	at org.apache.fory.shaded.org.codehaus.janino.UnitCompiler$13.visitAssignment(UnitCompiler.java:4126)
	at org.apache.fory.shaded.org.codehaus.janino.Java$Assignment.accept(Java.java:4864)
	at org.apache.fory.shaded.org.codehaus.janino.UnitCompiler.compile(UnitCompiler.java:4126)`

### Anything Else?

I have used AI to generate simple reproducible example, as I could not attach our production code. 
Please do not be repelled by some AI artefacts.
[...]

PR #3504, opened 2026-03-24
fix(java): handle private final map codegen

chaokunyang:
## What does this PR do?
- use a source-accessible local type when generated map chunk code handles monomorphic private final keys or values
- keep serializer selection and wire behavior based on the declared private final type
- add regression coverage for private final map keys and values under codegen with and without ref tracking

## Related issues

Closes #3503

## AI Contribution Checklist

- [ ] Substantial AI assistance was used in this PR: `yes` / `no`
- [ ] If `yes`, I included a completed [AI Contribution Checklist](https://github.com/apache/fory/blob/main/AI_POLICY.md#9-contributor-checklist-for-ai-assisted-prs) in this PR description and the required `AI Usage Disclosure`.

## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

PR #3573, opened 2026-04-15
fix(java): add test for previous uncaught regression. 

PiotrDuz:
Generated code should cast non-public class to the parent.
Non-public classes should be serialised correctly when used in cyclic graphs. 
Originated from: #3572
Related to #3504


chaokunyang (approved):
LGTM

PR #3592, opened 2026-04-20
fix(dart): use getUint32 for correctly encoding u64 value

ayush00git:
## Why?
`getInt32` treats the top bit as a sign bit. For large values where that bit is set (0xFFFFFFFE), the number comes back negative. Even though >>> is unsigned shift, it is shifting the already-corrupted signed value, garbage in, garbage out.
## What does this PR do?
Use `getUint32` instead.
`getUint32` has no sign bit, so all 32 bits are treated as magnitude.
## Related issues
N/A

## AI Contribution Checklist

### DIFF
--- a/dart/packages/fory/lib/src/buffer.dart
+++ b/dart/packages/fory/lib/src/buffer.dart
@@ -402,7 +402,7 @@ final class Buffer {
   /// Reads an unsigned 64-bit integer written by [writeTaggedUint64].
   int readTaggedUint64() {
     final readIndex = _readerIndex;
-    final first = _view.getInt32(readIndex, Endian.little);
+    final first = _view.getUint32(readIndex, Endian.little);
     if ((first & 1) == 0) {
       _readerIndex = readIndex + 4;
       return first >>> 1;
--- a/dart/packages/fory/lib/src/codegen/generated_support.dart
+++ b/dart/packages/fory/lib/src/codegen/generated_support.dart
@@ -345,7 +345,7 @@ final class GeneratedReadCursor {
 
   int readTaggedUint64() {
     final readIndex = _offset;
-    final first = _view.getInt32(readIndex, Endian.little);
+    final first = _view.getUint32(readIndex, Endian.little);
     if ((first & 1) == 0) {
       _offset = readIndex + 4;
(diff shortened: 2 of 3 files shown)


ayush00git:
@chaokunyang 
why do the dart files misses the apache LICENSE ?

dart/packages/fory/lib/src/buffer.dart
```diff
@@ -402,7 +402,7 @@ final class Buffer {
   /// Reads an unsigned 64-bit integer written by [writeTaggedUint64].
   int readTaggedUint64() {
     final readIndex = _readerIndex;
-    final first = _view.getInt32(readIndex, Endian.little);
+    final first = _view.getUint32(readIndex, Endian.little);
```
chaokunyang, line 424: This fixes `Buffer.readTaggedUint64()`, but `GeneratedReadCursor.readTaggedUint64()` in `dart/packages/fory/lib/src/codegen/generated_support.dart` still reads the first word with `getInt32`. Generated serializers route tagged `uint64` fields through that cursor path, so `0x7fffffff` still round-trips there as `9223372036854775807` instead of `2147483647`. The generated path needs the same `getUint32` change and a regression test.

chaokunyang:
> @chaokunyang why do the dart files misses the apache LICENSE ?

I missed that, let me add it

ayush00git:
> I missed that, let me add it

@chaokunyang 
also aren't we following any linting format for dart files?

chaokunyang:
> > I missed that, let me add it
> 
> @chaokunyang also aren't we following any linting format for dart files?

it seems licenserc doesn't support check dart, could you check it?

chaokunyang (approved):
LGTM

ayush00git:
> it seems licenserc doesn't support check dart, could you check it?

@chaokunyang 
I was actually asking about code formatting (like `dart format`) and linting (`dart analyze`), not the license headers. we already run `dart analyze` in ci, but we don't enforce `dart format` yet. should we add a formatting check?

and for the licenserc issue, we just need to add the `.dart` extension to licenserc.toml so the ci explicitly enforces it. should i add them ?

PR #3618, opened 2026-04-25
fix(go): added pre-allocation bounds checks for slices and strings

### LINKED ISSUE #3617: [Go] unbounded allocations prior to buffer validation
### Search before asking

### Version

v0.17.0

### Component(s)

Go

### Minimal reproduce step

Run these tests - 
```go
func TestReadLatin1OOMBug(t *testing.T) {
	// Missing Error Check Causes Unbounded Rune Allocation
	// We claim a massive size of 10,000 bytes, but provide an empty buffer.
	buf := NewByteBuffer(nil)
	
	err := &Error{}
	// readLatin1 doesn't read the length itself, it takes it as an argument
	result := readLatin1(buf, 10000, err)

	// Before the fix, this allocates a slice of 10000 runes (zeros) and returns it as a string of null bytes.
	// After the fix, it should return an empty string immediately when bounds check fails.
	require.True(t, err.HasError(), "Expected an error due to out of bounds buffer")
	require.Equal(t, "", result, "Expected an empty string due to missing data")
}

func TestReadInt32SliceOOMBug(t *testing.T) {
	// Unbounded Allocation in Primitive Slice Deserializers
	// We claim a size of 40,000 bytes, but provide no actual data.
	buf := NewByteBuffer(nil)
	buf.WriteLength(40000) 
	
	// Reset reader index so we can read what we just wrote
	buf.SetReaderIndex(0)

	err := &Error{}
	result := ReadInt32Slice(buf, err)

	// Before the fix, this allocates a slice of 10000 int32s (zeros) and returns it.
	// After the fix, it should return an empty/nil slice immediately when bounds check fails.
	assert.True(t, err.HasError(), "Expected an error due to out of bounds buffer")
	assert.Equal(t, 0, len(result), "Expected an empty slice due to missing data")
}

```

ayush00git:
## Why?

The code immediately allocated bytes based on the untrusted size. If any user sent a size of 2 billion but only actually sent 5 bytes of data, the server would instantly try to allocate 2GB of RAM. If many such requests were sent, the server would run out of memory (OOM) and crash.

## What does this PR do?
Add check which ensures we never allocate memory for data that hasn't actually arrived yet. It forces the allocation to be bounded by the physical size of the data we have already received and verified.

## Related issues
Closes #3617 

## AI Contribution Checklist

## Benchmark

### DIFF
--- a/go/fory/slice_primitive.go
+++ b/go/fory/slice_primitive.go
@@ -695,7 +695,7 @@ func ReadByteSlice(buf *ByteBuffer, err *Error) []byte {
 	if size == 0 {
 		return make([]byte, 0)
 	}
-	if size > buf.writerIndex-buf.readerIndex {
+	if buf.reader == nil && size > buf.writerIndex-buf.readerIndex {
 		*err = BufferOutOfBoundError(buf.readerIndex, size, buf.writerIndex)
 		return nil
 	}
@@ -720,7 +720,7 @@ func ReadBoolSlice(buf *ByteBuffer, err *Error) []bool {
 	if size == 0 {
 		return make([]bool, 0)
 	}
-	if size > buf.writerIndex-buf.readerIndex {
+	if buf.reader == nil && size > buf.writerIndex-buf.readerIndex {
 		*err = BufferOutOfBoundError(buf.readerIndex, size, buf.writerIndex)
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


ayush00git:
@chaokunyang have a look

chaokunyang (approved):
LGTM

PR #3629, opened 2026-04-28
fix(dart): guard ref and meta-string index reads

### LINKED ISSUE #3627: [Dart] runtime crashes on malformed reference metadata instead of rejecting it cleanly
### Search before asking

### Version

0.17.0

### Component(s)

Other

### Minimal reproduce step

Just run this test script for reproducing the issue:

```dart
import 'package:fory/src/config.dart';
import 'package:fory/src/context/meta_string_reader.dart';
import 'package:fory/src/context/ref_reader.dart';
import 'package:fory/src/context/ref_writer.dart';
import 'package:fory/src/memory/buffer.dart';
import 'package:fory/src/resolver/type_resolver.dart';
import 'package:test/test.dart';

void main() {
  test('RefReader.readRefOrNull does not throw for an out-of-range ref id',
      () {
    final reader = RefReader();
    final buffer = Buffer();

    buffer.writeByte(RefWriter.refFlag);
    buffer.writeVarUint32(9999);
    bufferSetReaderIndex(buffer, 0);

    expect(
      () => reader.readRefOrNull(buffer),
      returnsNormally,
    );
  });

  test(
      'MetaStringReader.readMetaString does not throw for a negative ref index',
      () {
    final reader = MetaStringReader(TypeResolver(const Config()));
    final buffer = Buffer();

    buffer.writeVarUint32Small7(1);
    bufferSetReaderIndex(buffer, 0);

    expect(
      () => reader.readMetaString(buffer),
      returnsNormally,
    );
  });
}

```

I expected the Dart runtime to reject the malformed reference data cleanly, without throwing a RangeError or crashing.
For the ref test, I expected readRefOrNull to handle an out-of-range ref id safely.
For the meta-string test, I expected readMetaString to handle the crafted reference header safely instead of indexing -1.

### What did you see instead?

The current Dart runtime throws RangeError for both cases.
RefReader.readRefOrNull indexes _refs[id] directly when the ref flag is set, so a large ref id like 9999 fails with an out-of-range error.
MetaStringReader.readMetaString computes length - 1 for a reference header, so header 1 becomes index -1 and fails with a RangeError.

### Anything Else?

_No response_

miantalha45:
## Why

This PR fixes two out-of-bounds bugs in the Dart runtime that were caused by trusting integer values read directly from the wire.

A malformed payload can provide an invalid reference id or meta-string index, which currently leads to a `RangeError` instead of a controlled failure. In a server environment, this can be used as a denial-of-service vector by crashing the Dart process with a single crafted packet.

## What does this PR do?

Bounds checking in `RefReader`:
- Added range validation before reading or resolving a ref id.
- This prevents out-of-bounds access when a malicious payload supplies a ref id larger than the current reference table.

Bounds checking in `MetaStringReader`:
- Added range validation before resolving a dynamic meta-string reference.
- This prevents negative or out-of-range indexing when the header encodes an invalid reference.

Security and regression tests:
- Added focused tests for the malformed ref-id and meta-string-reference cases.
- Added coverage for valid boundary cases to make sure normal edge values still work correctly.

## Related issues
Closes #3627

## AI Contribution Checklist


miantalha45:
@chaokunyang Have a look on this PR when you have time

PR #3638, opened 2026-04-29
fix(java): use REPLACE_STUB_ID for unregistered writeReplace classes to prevent cross-JVM ClassNotFoundException

### LINKED ISSUE #3635: [Java] Unregistered writeReplace classes cause ClassNotFoundException on cross-JVM deserialization
### Search before asking

### Version

0.18.0-SNAPSHOT (also affects 0.17.x and earlier)
Java 8 / Java 21, macOS

### Component(s)

Java

### Minimal reproduce step

When serializing an unregistered class that has `writeReplace()` returning a different type (e.g., a Hibernate/ByteBuddy proxy), Fory writes the **proxy class name** in the outer type info using `NAMED_EXT`. On a different JVM where that proxy class doesn't exist, deserialization fails with `ClassNotFoundException`.

```java
// "Proxy" class — only exists on JVM A
public class HibernateProxy extends RealEntity {
    protected Object writeReplace() {
        return new RealEntity(name, value);
    }
}

// JVM A: serialize
Fory fory1 = Fory.builder().requireClassRegistration(false).build();
byte[] bytes = fory1.serialize(new HibernateProxy("test", 42));

// JVM B: deserialize (HibernateProxy class not available)
Fory fory2 = Fory.builder().requireClassRegistration(false)
    .withClassLoader(classLoaderWithoutProxy).build();
Object result = fory2.deserialize(bytes); // ClassNotFoundException!
```

The root cause is in `ClassResolver.buildUnregisteredTypeId()` which returns `Types.NAMED_EXT` for writeReplace classes. This causes `writeTypeInfo()` to write the proxy class name bytes to the stream. The `ReplaceResolveSerializer` correctly handles the substitution *inside* `writeData()`, but the outer type info already contains the proxy class name — and the deserializer tries to load it before the serializer can handle the replacement.

Stack trace:
```
java.lang.IllegalStateException: Class HibernateProxy not found from classloaders [...]
  at org.apache.fory.resolver.TypeResolver.loadClass(TypeResolver.java:1095)
  at org.apache.fory.resolver.ClassResolver.populateBytesToTypeInfo(ClassResolver.java:1959)
  at org.apache.fory.resolver.TypeResolver.readTypeInfoFromBytes(TypeResolver.java:754)
  at org.apache.fory.resolver.TypeResolver.readTypeInfo(TypeResolver.java:546)
  at org.apache.fory.context.ReadContext.readRef(ReadContext.java:528)
Caused by: java.lang.ClassNotFoundException: HibernateProxy
```

Deserialization should succeed and return a `RealEntity` object (the result of `writeReplace()`), regardless of whether the proxy class exists on the deserializing JVM. The proxy class name should never appear in the serialized bytes since `writeReplace()` substitutes it before the data is written.

This is the same pattern Fory already handles correctly for lambdas (`LAMBDA_STUB_ID`) and JDK proxies (`JDK_PROXY_STUB_ID`) — a compact internal type ID is written instead of the class name.

### What did you see instead?

`ClassNotFoundException` during deserialization because `readTypeInfo()` tries to load the proxy class name from the byte stream. The proxy class name is baked into the outer type info by `writeTypeInfo()` before `ReplaceResolveSerializer.write()` gets a chance to handle the substitution.

This affects any framework that generates proxy classes with `writeReplace()`, including:
- Hibernate (ByteBuddy proxies for lazy-loaded entities)
- Spring AOP proxies
- Any custom proxy framework using `writeReplace()` for serialization transparency

wakilurislam:
# PR Description

## Why?

When Fory serializes an unregistered class with `writeReplace()` returning a different type (e.g., a Hibernate/ByteBuddy proxy), the outer type info writes the **proxy class name** to the byte stream. On a different JVM where that proxy class doesn't exist, deserialization fails with `ClassNotFoundException`.

## What does this PR do?

Adds an `instanceof ReplaceResolveSerializer` check in `ClassResolver.buildUnregisteredTypeId()` to return `REPLACE_STUB_ID` instead of `NAMED_EXT`. This writes a compact 1-byte internal type ID without the class name, matching the existing pattern for lambdas (`LAMBDA_STUB_ID`) and JDK proxies (`JDK_PROXY_STUB_ID`).

The original `NAMED_EXT` fallback for `serializer == null` (before serializer creation) is preserved unchanged.

### Changes

**`ClassResolver.java`**:

1. `buildUnregisteredTypeId()` — return `REPLACE_STUB_ID` when serializer is `ReplaceResolveSerializer`:

```diff
   protected int buildUnregisteredTypeId(Class<?> cls, Serializer<?> serializer) {
-    if (serializer == null && !cls.isEnum() && useReplaceResolveSerializer(cls)) {
-      return Types.NAMED_EXT;
+    if (!cls.isEnum()) {
+      if (serializer instanceof ReplaceResolveSerializer) {
+        return REPLACE_STUB_ID;
+      }
+      if (serializer == null && useReplaceResolveSerializer(cls)) {
+        return Types.NAMED_EXT;
+      }
     }
```

2. `addSerializer()` — guard against overwriting `typeIdToTypeInfo[REPLACE_STUB_ID]`:

```diff
+      if (typeId == REPLACE_STUB_ID) {
+        classInfoMap.put(type, typeInfo);
+      } else {
+        updateTypeInfo(type, typeInfo);
+      }
```

**`WriteReplaceCrossJvmTest.java`** (new): 6 test methods × 4 configurations = 24 tests covering cross-JVM proxy, same-JVM regression, same-type replace, round-trip, and nested proxy in DTO.

## Related issues

- Fixes #3635

## AI Contribution Checklist

```text
AI Usage Disclosure
- scope: code drafting, tests, design analysis
- affected_files_or_subsystems: java/fory-core ClassResolver.java, WriteReplaceCrossJvmTest.java (new)
- ai_review: two fresh AI reviewers on current HEAD — both found 0 actionable issues.
- human_verification: full fory-core test suite (1820 tests, 0 failures). ObjectStreamSerializerTest (82), URLSerializerTest (9), ReplaceResolveSerializerTest (58) all pass. Cross-JVM test fails without fix (4 failures), passes with fix (24/24).
- performance_verification: N/A — one-time cold-path change only
- provenance_license_confirmation: Apache-2.0 compatible, all original work.
```

No public API or binary protocol changes. Wire format changes only for unregistered `ReplaceResolveSerializer` classes (bug fix — previous format was broken for cross-JVM).

## Benchmark

N/A — one-time `buildUnregisteredTypeId()` call, not per-element.

[review_1_with_skill.md](https://github.com/user-attachments/files/27203956/review_1_with_skill.md)
[review_2_without_skill.md](https://github.com/user-attachments/files/27203958/review_2_without_skill.md)

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/resolver/ClassResolver.java
+++ b/java/fory-core/src/main/java/org/apache/fory/resolver/ClassResolver.java
@@ -686,8 +686,13 @@ public class ClassResolver extends TypeResolver {
 
   @Override
   protected int buildUnregisteredTypeId(Class<?> cls, Serializer<?> serializer) {
-    if (serializer == null && !cls.isEnum() && useReplaceResolveSerializer(cls)) {
-      return Types.NAMED_EXT;
+    if (!cls.isEnum()) {
+      if (serializer instanceof ReplaceResolveSerializer) {
+        return REPLACE_STUB_ID;
+      }
+      if (serializer == null && useReplaceResolveSerializer(cls)) {
+        return Types.NAMED_EXT;
+      }
     }
     return super.buildUnregisteredTypeId(cls, serializer);
   }
@@ -1215,7 +1220,11 @@ public class ClassResolver extends TypeResolver {
       } else {
         typeInfo = typeInfo.copy(typeId);
       }
-      updateTypeInfo(type, typeInfo);
+      if (typeId == REPLACE_STUB_ID) {
+        classInfoMap.put(type, typeInfo);
+      } else {
+        updateTypeInfo(type, typeInfo);
+      }
       // Add to compositeNameBytes2TypeInfo for unregistered classes so that
       // readTypeInfo can find the TypeInfo by name bytes during deserialization.
       // This is important for dynamically created classes that can't be loaded by name.
--- a/java/fory-core/src/test/java/org/apache/fory/serializer/WriteReplaceCrossJvmTest.java
+++ b/java/fory-core/src/test/java/org/apache/fory/serializer/WriteReplaceCrossJvmTest.java
@@ -1,0 +1,408 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fory.serializer;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+
+import java.io.Serializable;
+import java.util.Objects;
+import org.apache.fory.Fory;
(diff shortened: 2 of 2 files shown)


chaokunyang (approved):
LGTM

PR #3658, opened 2026-05-08
fix(compiler): handle Rust identifier escaping and name collisions

BaldDemian:
## Why?

Identifiers colliding with Rust keywords should be escaped to prevent compiling errors.

### Workflow

The current implementation broadly follows the approach used by prost-build. It first normalizes identifiers from the IDL according to Rust naming conventions, e.g. converting struct names to UpperCamelCase and field names to snake_case. It then checks whether the normalized identifiers are Rust keywords and escapes them when necessary.

This process is not trivial as I first expected. For example, we need to cache the escaped name of each identifier to make sure the definition site and all use sites refer to the same name consistently.

### Known issues

Although I have tested the implementation against many corner cases, this handling is still **incomplete**, much like prost itself. It may still generate Rust code that fails to compile for valid (but rare, I suppose 🤔) IDL. In most cases, this happens because an identifier from the IDL conflicts with an identifier already used by the Rust code generated by Fory. 

Example 1:

IDL:

```protobuf
package demo;
message Fory {
  string value = 1;
}
```

Generated Rust code:

```rust
use fory::Fory;
pub struct Fory {
    pub value: String,
}
pub fn register_types(fory: &mut Fory) -> Result<(), fory::Error> {
    fory.register::<Fory>(...)?;
    Ok(())
}
```

Compiling errors:

```
error[E0255]: the name `Fory` is defined multiple times
error[E0117]: derive/impl resolves to the external `fory::Fory` type
```

----

Example 2:

IDL:

```protobuf
package demo;
message HashMap {
  string value = 1;
}
message Holder {
  map<string, string> values = 1;
  HashMap item = 2;
}
```

Generated Rust code:

```rust
use std::collections::HashMap;
pub struct HashMap {
    pub value: String,
}
pub struct Holder {
    pub values: HashMap<String, String>,
    pub item: HashMap,
}
```

Compiling errors:

```
error[E0255]: the name `HashMap` is defined multiple times
error[E0107]: missing generics for struct `std::collections::HashMap`
```

However, I think IDL like the above are rare in real practice. The resulting compiling errors are clear and straightforward and the user can handle the conflicting names easily in the IDL.

### Changes made to the FBS compiler

In the current implementation, I use the tuple `<file_path, line, column>` as the stable key for an identifier. This lets the generator connect an identifier's definition site with all of its use sites, so they consistently resolve to the same name after normalization and Rust keyword escaping.

However, when translating a FlatBuffers union into the Fory IR, the FBS translator currently assigns every generated union case the same source location: the line and column of the union declaration itself.

https://github.com/apache/fory/blob/389b66c81d197bc7a03129e4dfefde34dd2fcf1d/compiler/fory_compiler/frontend/fbs/translator.py#L243-L244

As a result, different cases in the same union can end up sharing the same identifier key. This causes the later case to overwrite the earlier one in the Rust generator's name cache, and can incorrectly make distinct union variants resolve to the same Rust name.

Since both the FDL parser and the Protobuf translator preserve finer-grained source locations for union/oneof cases:

https://github.com/apache/fory/blob/389b66c81d197bc7a03129e4dfefde34dd2fcf1d/compiler/fory_compiler/frontend/proto/translator.py#L132-L133

https://github.com/apache/fory/blob/389b66c81d197bc7a03129e4dfefde34dd2fcf1d/compiler/fory_compiler/frontend/fdl/parser.py#L598-L599

the FBS translator should do the same by assigning each union case the location of its own token.

## Related issues

Close https://github.com/apache/fory/issues/3544.

## AI Contribution Checklist

### DIFF
--- a/compiler/fory_compiler/cli.py
+++ b/compiler/fory_compiler/cli.py
@@ -123,6 +123,7 @@ def resolve_imports(
     imported_enums = []
     imported_messages = []
     imported_unions = []
+    source_packages: Dict[str, Optional[str]] = {str(file_path): schema.package}
 
     for imp in schema.imports:
         # Resolve import path using search paths
@@ -149,6 +150,7 @@ def resolve_imports(
         imported_enums.extend(imported_schema.enums)
         imported_messages.extend(imported_schema.messages)
         imported_unions.extend(imported_schema.unions)
+        source_packages.update(imported_schema.source_packages)
 
     # Create merged schema with imported types first (so they can be referenced)
     merged_schema = Schema(
@@ -161,6 +163,7 @@ def resolve_imports(
         options=schema.options,
         source_file=schema.source_file,
         source_format=schema.source_format,
+        source_packages=source_packages,
     )
 
     cache[file_path] = copy.deepcopy(merged_schema)
@@ -482,6 +485,7 @@ def compile_file(
         package_override: Optional package name override
         import_paths: List of import search paths
     """
+    file_path = file_path.resolve()
     print(f"Compiling {file_path}...")
 
     # Parse and resolve imports
@@ -543,11 +547,25 @@ def compile_file(
 
         generator_class = GENERATORS[lang]
         generator = generator_class(schema, options)
-        files = generator.generate()
+        try:
+            files = generator.generate()
 
-        if grpc:
-            service_files = generator.generate_services()
-            files.extend(service_files)
+            if grpc:
+                service_files = generator.generate_services()
+                files.extend(service_files)
+        except ValueError as e:
+            print(f"Error: {e}", file=sys.stderr)
+            return False
+
+        if lang == "rust":
+            for f in files:
+                target = (lang_output / f.path).resolve()
+                if target.exists() and not is_generated_file(target):
+                    print(
+                        f"Error: refusing to overwrite non-generated Rust file: {target}",
+                        file=sys.stderr,
+                    )
+                    return False
 
         generator.write_files(files)
 
--- a/compiler/fory_compiler/generators/rust.py
+++ b/compiler/fory_compiler/generators/rust.py
@@ -38,6 +38,14 @@ from fory_compiler.ir.ast import (
 from fory_compiler.ir.types import PrimitiveKind
 
 
+class RustNameError(ValueError):
+    """Base error for invalid Rust names generated from IDL names."""
+
+
+class RustNameCollisionError(RustNameError):
+    """Raised when two IDL names map to the same Rust identifier."""
+
+
 class RustGenerator(BaseGenerator):
     """Generates Rust types with Fory derive macros."""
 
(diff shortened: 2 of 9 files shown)


BaldDemian:
~I am aware of the root cause of the failing xlang test. Working on the fix~ Now CI passes

compiler/fory_compiler/generators/rust.py
```diff
@@ -68,6 +76,427 @@ class RustGenerator(BaseGenerator):
...
+        scope = self._local_top_level_messages(schema)
+        for part in parts:
+            match = next((message for message in scope if message.name == part), None)
+            if match is None:
+                return []
+            lineage.append(match)
+            scope = match.nested_messages
+        return lineage
+
+    def _allocate_type_scope_names(self, type_defs: List[object], scope: str) -> None:
```
chaokunyang, line 388: This allocator only checks IDL declarations against each other, but the generated Rust module also uses helper and prelude names unqualified. For example, `message String { string value = 1; }` generates `pub struct String { pub value: String, }`, so the field type resolves to the generated struct instead of `std::string::String` and the Rust code is recursive/invalid. `Vec` and `Fory` have similar failures with `Vec<T>` and `use fory::Fory`. Please reserve the generated/helper/prelude names in the relevant scopes or fully qualify those references, and add a generated Rust compile check for `String`, `Vec`, and `Fory`.

compiler/fory_compiler/cli.py
```diff
@@ -543,11 +547,25 @@ def compile_file(
...
+                service_files = generator.generate_services()
+                files.extend(service_files)
+        except ValueError as e:
+            print(f"Error: {e}", file=sys.stderr)
+            return False
 
-        if grpc:
-            service_files = generator.generate_services()
-            files.extend(service_files)
+        if lang == "rust":
```
chaokunyang, line 560: This guard only rejects overwriting existing non-generated files. It does not catch two inputs in the same invocation that normalize to the same Rust output path. For example, packages `foo.bar` and `foo_bar` both produce `foo_bar.rs`; the second file sees the generated marker, overwrites the first file, and the command exits successfully with only the second schema's types. Please track generated target paths during the compile invocation and fail when distinct source files map to the same Rust output file, with a two-file CLI regression test.

compiler/fory_compiler/generators/rust.py
BaldDemian: I'm considering changing the generated Rust code to use crate-root absolute paths for generator-owned dependencies, for example emitting `::std::string::String` instead of `std::string::String`, and removing generated `use` statements.

Prost follows this absolute path style.

Using absolute paths keeps Fory/runtime/std/core references out of the generated code's local namespace, so IDL-defined names such as `String`, `Vec` cannot collide with the generator-owned types.

I'd like to hear your opinion @chaokunyang

chaokunyang: good idea, absolute paths is better

BaldDemian: I opened [a separate PR](https://github.com/apache/fory/pull/3666) since using absolute paths is orthogonal to this issue

PR #3666, opened 2026-05-09
refactor(rust): use absolute path in generated Rust code

BaldDemian:
## Why?

For generated Rust code, it would be better to use absolute paths like `::std::option::Option` instead of `Option`. 
On the one hand, this follows common practice in [procedural macros](https://doc.rust-lang.org/reference/procedural-macros.html#r-macro.proc.hygiene) and prost; on the other hand, this will keep Fory/runtime/std/core references out of the generated code's local namespace, so IDL-defined names such as `String`, `Vec` won't collide with the generator-owned types.

## What does this PR do?

- Use absolute paths across the Rust code generator and derived macros.
- Remove collecting uses logic in the Rust code generator like
  https://github.com/apache/fory/blob/b4441d0ac55ba0dc89eb4a311966daa58074b9fc/compiler/fory_compiler/generators/rust.py#L199-L215

## Related issues

N/A.

## AI Contribution Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ bazel-*
 **/.bazel_user_root/
 .whl
 python/.cache
-python/pyfory/__pycache__/
+__pycache__/
 python/dist
 python/build
 python/pyfory.egg-info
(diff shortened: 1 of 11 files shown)


chaokunyang (approved):
Looks great

PR #3692, opened 2026-05-18
feat(compiler): add java/python gRPC support

chaokunyang:
## Why?

Add compiler-generated gRPC support for Fory IDL services so Java and Python users can build gRPC clients and servers directly from FDL, FlatBuffers IDL, and Protobuf service definitions.

This also covers the service-related gaps tracked by #3266, #3272, and #3273.

## What does this PR do?

- Generates Java grpc-java service classes, method descriptors, async/blocking/future stubs, server bindings, and Fory-backed request/response marshalling.
- Generates Python grpcio companion modules with servicer base classes, registration helpers, and unary, server-streaming, client-streaming, and bidirectional-streaming handlers.
- Preserves source package metadata during import resolution so service request and response types can be resolved across imported schemas.
- Adds validation and codegen tests for service generation, naming collisions, streaming RPCs, imports, Java package options, and Protobuf service translation.
- Adds Java/Python gRPC IDL interop fixtures and CI coverage through `integration_tests/idl_tests/run_grpc_tests.sh`.
- Updates compiler documentation for service generation and the new `--grpc` integration test generation path.

## Related issues

#3266 
Closes #3272 
Closes #3273  

## AI Contribution Checklist

- [ ] Substantial AI assistance was used in this PR: `yes` / `no`
- [ ] If `yes`, I included a completed [AI Contribution Checklist](https://github.com/apache/fory/blob/main/AI_POLICY.md#9-contributor-checklist-for-ai-assisted-prs) in this PR description and the required `AI Usage Disclosure`.
- [ ] If `yes`, my PR description includes the required `ai_review` summary and screenshot evidence of the final clean AI review results from both fresh reviewers on the current PR diff or current HEAD after the latest code changes.

## Does this PR introduce any user-facing change?

- [ ] Does this PR introduce any public API change?
- [ ] Does this PR introduce any binary protocol compatibility change?

PR #3698, opened 2026-05-22
feat(go): implement grpc stub generation

### LINKED ISSUE #3274: [Compiler][gRPC][Go] Generate gRPC stubs with Fory codec
Parent issue: #3266

## Scope
Implement Go gRPC stub generation using Fory serialization only.

## Detailed tasks
- Emit `*_service.go` interface and `*_grpc.go` binding.
- Generate `ServiceDesc`, server interface, and client wrapper.
- Register a custom Fory codec and use it in `grpc.CallOption`.

ayush00git:
## Why?
Go implementation only provides type generation and misses the gRPC service stub generations for now. this PR adds it.

## Related issues
Closes #3274 

#### `ImportTracker` class in `base.py`
1. which returns the foreign imports defined in a RpcMethod inside a schema file's service interface.

#### `generate_services()`
1. which returns a service generated file
2. License header, package header using `get_package_name()`
3. the imports block of go using `_build_import_block(tracker)`
`_build_import_block(tracker)` function returns the import block with fixed grpc, context imports which are required by every single generated file for each service. and a foreign type imports statement returned by the `tracker.go_imports()`. at end after all the service code generation is done, this function inserts the import block to the saved placeholder index.


ayush00git:
@chaokunyang have a look

chaokunyang:
Do you plan to add grpc intgeration tests like we did in #3692 ?

ayush00git:
> Do you plan to add grpc intgeration tests like we did in #3692 ?

actually yes, but I had planned integrations test with => go server <-> js client. 
or should we design the interop tests using each of the languages implemented ?

chaokunyang:
You should use Java driven integration tests for consistency. And Java grpc has been supported. You can do it directly now

ayush00git:
@chaokunyang  have a look

go/fory/grpc/codec.go
```diff
@@ -0,0 +1,60 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package forygrpc
+
+import (
+	"github.com/apache/fory/go/fory"
+	"google.golang.org/grpc/mem"
```
chaokunyang: This is not allowed. We must not introduce any grpc dependency into fory go serialization library.

You could generate code into ForyXXXModule.go in generated code when grpc service is enabled.

ayush00git:
@chaokunyang have a look now.
i've moved the codec to the generated grpc files.

chaokunyang (approved):
LGTM. Could  you open a new PR to reactor integration_tests/grpc_tests/java/src/test/java/org/apache/fory/grpc_tests/GrpcInteropTest.java into base class and subclass for different languages. You could take XlangTestBase.java as reference.

And also create separate ci job for Java/Go grpc job

ayush00git:
> LGTM. Could you open a new PR to reactor integration_tests/grpc_tests/java/src/test/java/org/apache/fory/grpc_tests/GrpcInteropTest.java into base class and subclass for different languages. You could take XlangTestBase.java as reference.
> 
> And also create separate ci job for Java/Go grpc job

i'll look into it and raise a PR soon

integration_tests/grpc_tests/go/main.go
```diff
@@ -0,0 +1,299 @@
...
+	serverPortFile := serverCmd.String("port-file", "", "path to write bound port")
+
+	clientCmd := flag.NewFlagSet("client", flag.ExitOnError)
+	clientTarget := clientCmd.String("target", "", "host:port to connect to")
+
+	if len(os.Args) < 2 {
+		log.Fatal("usage: grpc-interop <server|client> [flags]")
+	}
+
+	f := newFory()
```
chaokunyang, line 277: This is wrong design, the fory Go IDL already generated thread safe Fory, you should not create Fory again. And the Fory you created is not thread-safe, which won't work in multi-thread cases

compiler/fory_compiler/generators/services/go.py
```diff
@@ -0,0 +1,720 @@
...
+        lines.append("\tcc grpc.ClientConnInterface")
+        lines.append("\tfory *fory.Fory")
+        lines.append("}")
+        lines.append("")
+        return lines
+
+    def _generate_new_client(self, service: Service) -> List[str]:
+        lines: List[str] = []
+        lines.append(
+            f"func New{service.name}Client(cc grpc.ClientConnInterface, f *fory.Fory) {service.name}Client {{"
```
chaokunyang, line 153: Generated model files already own a package-level `getFory() *threadsafe.Fory` with local and imported schema registrations installed. Requiring every generated client and `CodecV2` to receive a separate `*fory.Fory` splits that owner model, makes users duplicate generated registration, and is unsafe for concurrent RPCs because plain Go `Fory` reuses internal buffers and state. The gRPC companion should reuse the generated package runtime and make the codec/client constructor stateless from the user's perspective.

compiler/fory_compiler/generators/services/go.py
```diff
@@ -0,0 +1,720 @@
...
+            f"// {service.name}Client is the client API for {service.name} service."
+        )
+        lines.append(f"type {service.name}Client interface {{")
+        for method in service.methods:
+            req_type = self._resolve_go_type(method.request_type, tracker)
+            res_type = self._resolve_go_type(method.response_type, tracker)
+            mode = streaming_mode(method)
+            if mode is StreamingMode.UNARY:
+                signature = f"(ctx context.Context, in {req_type}, opts ...grpc.CallOption) ({res_type}, error)"
+                lines.append(f"\t{self.to_pascal_case(method.name)}{signature}")
```
chaokunyang, line 129: This emits Go identifiers from `to_pascal_case(method.name)` without checking for collisions. A service with `rpc Foo` and `rpc foo` passes validation but generates duplicate `Foo` methods, duplicate handlers, and an uncompilable `_grpc.go` file. Add a Go-specific service/method collision check before emitting code, similar to the collision checks already present for the other generated gRPC targets.

compiler/fory_compiler/generators/go.py
```diff
@@ -206,6 +207,10 @@ def generate(self) -> List[GeneratedFile]:
         # Generate a single Go file with all types
         files.append(self.generate_file())
 
+        # Generate gRPC service stubs if requested
+        if self.options.grpc:
```
chaokunyang, line 211: The CLI already appends `generator.generate_services()` when `--grpc` is set, so adding service files again inside `GoGenerator.generate()` makes Go emit the same `_grpc.go` file twice. The duplicate currently overwrites identical content and prints duplicate `Generated:` lines, but it leaves two owners for the same output path and can hide future generation divergence. Keep service companion emission in the common CLI path only.

integration_tests/grpc_tests/go/main.go
```diff
@@ -0,0 +1,299 @@
...
+	if err != nil {
+		return fmt.Errorf("listen: %w", err)
+	}
+	codec := grpc_fdl.CodecV2{Fory: f}
+	// "fory" (v2): used by Go clients via grpc.ForceCodecV2.
+	encoding.RegisterCodecV2(codec)
+	// "proto" (v1): overrides the built-in proto codec so Java clients are
+	// also handled by Fory. RegisterCodecV2 writes to a separate registry and
+	// does not replace the v1 "proto" entry; RegisterCodec must be used.
+	encoding.RegisterCodec(protoFallbackCodec{codec})
```
chaokunyang, line 158: Registering a Fory codec under the global `proto` name is a test-only workaround for the server not forcing the generated codec. It mutates grpc-go process-global codec state and can affect unrelated gRPC clients or services in the same process. The Go server should be created with `grpc.ForceServerCodecV2(...)` using the generated codec, then this fallback codec and global registry mutation can be removed.

compiler/fory_compiler/generators/services/go.py
```diff
@@ -0,0 +1,720 @@
...
+        return lines
+
+    def _generate_unary_type_desc(self, service: Service) -> List[str]:
+        lines: List[str] = []
+        for method in service.methods:
+            mode = streaming_mode(method)
+            if mode is StreamingMode.UNARY:
+                lines.append("\t\t{")
+                lines.append(
+                    f'\t\t\tMethodName:\t"{self.to_pascal_case(method.name)}",'
```
chaokunyang, line 687: The generated service descriptor should keep the exact IDL RPC name here. Client calls use get_grpc_method_path(...), which preserves method.name, but the server registers MethodName/StreamName with to_pascal_case(method.name). For valid names such as sayHello or say_hello, grpc-go registers SayHello while the client invokes /Service/sayHello, so routing returns UNIMPLEMENTED. Please use method.name for MethodName and StreamName, and keep PascalCase only for exported Go identifiers.

PR #3702, opened 2026-05-23
feat(java): remove sun.misc.Unsafe for jdk25

### LINKED ISSUE #2090: java memory.Platform uses terminally deprecated Unsafe api
### Search before asking

### Version

Since JDK 24, the `sun.misc.Unsafe.arrayBaseOffset` api is terminally deprecated for removal:

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by org.apache.fury.memory.Platform (file:/home/steven/.m2/repository/org/apache/fury/fury-core/0.10.0/fury-core-0.10.0.jar)
WARNING: Please consider reporting this to the maintainers of class org.apache.fury.memory.Platform
WARNING: sun.misc.Unsafe::arrayBaseOffset will be removed in a future release
```

For now, this is only a warning. In the future this will fail entirely.

### Component(s)

Java

### Minimal reproduce step

Any usage of Fury e.g. `ThreadSafeFury.deserialize` under JDK 24, as far as I understand

Using Fury should not print warnings about using terminally deprecated methods.

### What did you see instead?

A warning that the `Unsafe.arrayBaseOffset` api is deprecated for removal :)

### Anything Else?

_No response_

### LINKED ISSUE #2606: [Java] Displays warnings when starting my process with JDK25
### Feature Request

WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by org.apache.fory.memory.Platform (file:/G:/.m2/repository/org/apache/fory/fory-core/0.12.2/fory-core-0.12.2.jar)
WARNING: Please consider reporting this to the maintainers of class org.apache.fory.memory.Platform
WARNING: sun.misc.Unsafe::arrayBaseOffset will be removed in a future release

_No response_

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #3729: [Java26] set final fields when --illegal-final-field-mutation=deny
### Feature Request

JDK25+ zero-Unsafe Java runtime support should use a true `MethodHandles.Lookup.IMPL_LOOKUP` trusted lookup and trusted-lookup `VarHandle`s to populate fields, including final fields on JDK26 when ordinary reflection final-field mutation is denied.

For JDK26+, normal Fory operation should not require `--enable-final-field-mutation`. `@ForyConstructor` and `BaseFory.registerConstructor(...)` remain supported and preferred when a constructor is the clearest semantic object-creation model, but they should be explicit construction mechanisms rather than mandatory escape hatches for final fields.

Yes. The current Java25 lookup overlay uses:

```java
return implLookup().in(objectClass);
```

That is not equivalent to a true trusted target-class lookup. A local JDK26.0.1 experiment showed the difference:

```text
IMPL_LOOKUP:              /trusted, lookupModes=127
Lookup private ctor path: /trusted, lookupModes=127
IMPL_LOOKUP.in(type):     type lookup, lookupModes=95
```

Under `--illegal-final-field-mutation=deny`:

```text
lookup-created-by-private-Lookup-constructor.findVarHandle(...).set(...) succeeds
lookup-created-by-private-Lookup-constructor.findSetter(...)             succeeds
IMPL_LOOKUP.in(type).findVarHandle(...).set(...)                         fails
IMPL_LOOKUP.in(type).findSetter(...)                                     fails
Field.set*(finalField)                                                   fails
```

Without `--illegal-final-field-mutation=deny`, ordinary `Field.set*` currently succeeds on JDK26 because the default policy is not deny. Fory should not depend on that default policy. The runtime should use the trusted lookup owner path instead.

Update the JDK25+ lookup implementation so `_trustedLookup(type)` preserves true trusted lookup semantics.

Required code changes:

- Load original `MethodHandles.Lookup.IMPL_LOOKUP` through reflection, guarded by the required `java.base/java.lang.invoke` open to `org.apache.fory.core`.
- Do not implement `_trustedLookup(type)` with `IMPL_LOOKUP.in(type)`.
- Use original `IMPL_LOOKUP` to resolve the private `MethodHandles.Lookup` constructor:

  ```java
  Lookup(Class<?> lookupClass, Class<?> previousLookupClass, int allowedModes)
  ```

- Invoke that constructor with the target type, `null` previous lookup class, and trusted mode `-1`.
- Cache the constructor method handle as a cold-path lookup primitive.
- Use the resulting target-class `/trusted` lookup to create field `VarHandle`s.
- Make JDK25+ `FieldAccessorStrategy` use `VarHandle` as the owner for both normal and final field reads/writes.
- Remove any final-field-only lazy `MethodHandle` setter path from Java25 field access if it exists only because the current lookup implementation cannot create writable final-field `VarHandle`s.
- Keep `MethodHandle` for executable behavior such as constructors, methods, serialization hooks, and lambda/metafactory paths.
- Keep `@ForyConstructor` and `BaseFory.registerConstructor(...)` as explicit constructor-owned object-creation paths. They are semantically clearer for types where a constructor is the right reconstruction model, but the generic final-field population path should still work without requiring `--enable-final-field-mutation`.

Required tests:

- Add JDK25+/JDK26 coverage proving `_JDKAccess._trustedLookup(type).findVarHandle(...)` can read and write a final field on an object created by Fory's constructor-bypass allocator.
- Run the JDK26 test with `--illegal-final-field-mutation=deny` and verify it still passes.
- Include a regression check that the Java25+ implementation is not using `IMPL_LOOKUP.in(type)` as the final-field trusted lookup path.
- Verify ordinary reflection `Field.set*` is not the owner path by confirming it fails under denial mode.
- `findSetter(...)` may remain useful as an experiment or non-field-access primitive, but field access should not require a separate final-field `MethodHandle` setter path.

Required docs and comments:

- Update the JDK25 zero-Unsafe design document to state that Fory requires opening `java.base/java.lang.invoke` to `org.apache.fory.core` so it can obtain original `IMPL_LOOKUP`.
- State that Java25+ must preserve true trusted lookup mode and must not use `Lookup.in(type)` where writable final-field handles are required.
- State that JDK25+ field access is owned by trusted-lookup `VarHandle`s.
- State that JDK26+ normal operation should not require `--enable-final-field-mutation`.
- Document `@ForyConstructor` and `BaseFory.registerConstructor(...)` as explicit semantic construction paths, not required workarounds for final-field mutation.
- Add a concise source comment near Java25 `_Lookup._trustedLookup(type)` explaining why `IMPL_LOOKUP.in(type)` is invalid for writable final-field handles.

- Requiring `--enable-final-field-mutation` for normal user classes: rejected. It asks users to opt back into the behavior this task is avoiding and is not needed when true trusted lookup `VarHandle`s are available.
- Using ordinary reflection `Field.set*`: rejected as the primary path. It fails under `--illegal-final-field-mutation=deny`.
- Using `IMPL_LOOKUP.in(type)`: rejected. It drops the trusted mode needed to create writable final-field handles.
- Keeping a final-field-only `MethodHandle` setter path in field access: rejected unless a measured, documented reason proves it is necessary. Field get/set access should have one `VarHandle` owner path.
- Requiring constructor binding for all final-field classes: rejected. `@ForyConstructor` and `registerConstructor(...)` should be supported as explicit semantic construction paths, but they should not be mandatory for every normal final-field class.

### Additional context

This issue is about preserving Fory's serializer-family and object-population semantics across JDK versions. JDK25+ must not switch serializer families or require a different wire-compatible fallback only because Unsafe is unavailable.

The expected verification result on JDK26 is:

```text
_JDKAccess._trustedLookup(type).findVarHandle(finalField).set(instance, value) succeeds
Field.set*(finalField) fails under --illegal-final-field-mutation=deny
```

That proves Fory is using the intended trusted-lookup `VarHandle` owner path rather than depending on the current JDK26 default final-field mutation policy.

chaokunyang:
## Why?

JDK25+ removes the remaining viable `sun.misc.Unsafe` paths Fory used for Java object creation, field access, string internals, and memory operations. This PR makes the Java runtime continue to support modern JDKs while preserving the existing JDK8-24, Android, GraalVM, native-mode, and xlang behavior where applicable.

## What does this PR do?

- Builds `fory-core` as a multi-release jar with JDK25-specific classes and module metadata, while keeping the root implementation for JDK8-24.
- Replaces the JDK25+ runtime paths for field access, object instantiation, string internals, memory access, and generated serializers with trusted lookup, VarHandle, ReflectionFactory, and other non-Unsafe owners.
- Updates Java serializers, codegen, copy paths, collection/map serializers, object stream handling, and default-value support to use the new accessor and instantiator ownership model.
- Extends CI and integration coverage for JDK25/JDK26, JPMS zero-Unsafe runs, multi-release jar verification, GraalVM packaging, and cross-JDK compatibility expectations.
- Updates Java, Kotlin, Scala, and top-level docs for JDK25+ setup, including the required `java.base/java.lang.invoke` open for classpath and module-path users.
- Adds Java25 benchmark modules and MR-JAR benchmark verification for direct memory and direct-to-heap copy paths.

## Related issues

Closes #3729
Closes #2090
Closes #2606

## AI Contribution Checklist

Yes. Java users on JDK25+ must open `java.base/java.lang.invoke` to Fory:

- Classpath: `--add-opens=java.base/java.lang.invoke=ALL-UNNAMED`
- Module path: `--add-opens=java.base/java.lang.invoke=org.apache.fory.core`

No binary protocol compatibility change is intended.

## Benchmark

This PR adds Java25 benchmark coverage under `benchmarks/java25` and a benchmark MR-JAR check. No benchmark result table is included here.


pandalee99 (approved):

hrgdavor:
This is very welcome addition. I am still keeping my eye on this project to be able to use it when a fitting project comes along that will benefit form fory. Also, thanks for making it milti-release for easier use.

PR #3704, opened 2026-05-25
ci: declare workflow-level `contents: read` on ci and lint

arpitjain099:
Both workflows run pure CI / lint checks; no GitHub API writes from the workflows. Workflow-level `contents: read` is sufficient as the cap for the default `GITHUB_TOKEN`.

Same post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.


chaokunyang (approved):
LGTM, thanks

PR #3714, opened 2026-05-29
feat(format): schema evolution for the Java row codec

stevenschlansker:
Opt in with `.withSchemaEvolution()` on any row, array, or map codec builder. Fields carry `@ForyVersion(since, until)`; removed fields are listed on a nested interface referenced from
`@ForySchema(removedFields = ...)`. Older payloads are dispatched at read time; nothing changes when the flag is off. Standard and compact formats supported.

## Why?

Currently changing row format schema definition in any way invalidates all records

## What does this PR do?

Propose a new concept of format versions, each succeeding version may add or remove fields from types, and deserialization machinery picks version based on schema hash

```
  Schema evolution for the Java row codec

  Lets a consumer built from the current bean decode rows written by older versions of that bean.

  Adds opt-in schema evolution to the Java row codec, enabled with
  `withSchemaEvolution()` on the bean/array/map codec builders. A reader can decode
  payloads written against older versions of a bean by dispatching on a per-payload
  strict schema hash to a projection codec that reads the historical layout and
  maps it onto the current type, applying defaults for fields that did not yet
  exist and discarding fields that have been removed.

  Field history is declared with `@ForyVersion(since/until)` on live fields and on
  a `@ForySchema(removedFields = ...)` history interface for removed fields.
  `SchemaHistory` enumerates the version history, including the cross-product over
  nested versioned beans; each historical layout gets a projection codec whose row
  layout is precomputed once so projection decode costs the same per call as
  current-schema decode. The wire format uses an 8-byte strict-hash slot
  (reusing the row's existing hash slot, and new for evolution-enabled array/map elements),
  and producer/consumer must agree on the flag.

  Covers the standard and compact row formats, records (including
  `@ForyVersion` on record components) and interface beans, and nested versioned
  beans to arbitrary depth.


  API
  - `@ForyVersion` — since/until version window on a field, record component, or accessor method, defining when that field is present in the schema history.
  - `@ForySchema` — declares a bean's evolution intent and points at a nested interface describing removed fields.

  How it works

  Row payloads already carry an 8-byte schema hash. Evolution-enabled encoders keep
  a map of historical hashes; on decode, a hash mismatch is resolved against that
  map instead of failing.

  - `SchemaHistory` derives the ordered historical schemas from the version
    annotations and a strict hash per version (hash includes field name and
    nullability).
  - `RowCodecBuilder` generates one projection codec per historical version
    (`_V<n>` classes), each reading its historical schema and producing
    current-bean instances.
  - `BinaryRowEncoder.decode` dispatches on the peer hash: exact match → current
    codec; known historical hash → projection codec; else
    `ClassNotCompatibleException`.
  - Nested versioned beans dispatch recursively by strict hash, routed through
    array and map projection codecs so versioning composes through `List<Bean>` and
    `Map<_, Bean>`.

  Performance

  Both decode paths reuse the `CompactRowLayout` cache from #3717. The
  current-schema path allocates rows via `writer.newRow()`; the historical path via
  a per-projection `RowFactory` that holds its historical schema's layout, built
  once. Per-decode cost is the same on both — no layout recomputation. The
  evolution-enabled array/map codecs keep a single allocation per encode.
```

## AI Contribution Checklist

AI Usage Disclosure
- scope: all
- affected_files_or_subsystems: row format Java
- ai_review: <line-by-line self-review completed; summarize the two-reviewer loop and final no-further-comments result>
- ai_review_artifacts: <embedded screenshots or links showing the final clean review results from both fresh reviewers on the current PR diff or current HEAD after the latest code changes>
- human_verification: <checks run locally or in CI + pass/fail summary + contributor reviewed results>
- [x] If `yes`, I included a completed [AI Contribution Checklist](https://github.com/apache/fory/blob/main/AI_POLICY.md#9-contributor-checklist-for-ai-assisted-prs) in this PR description and the required `AI Usage Disclosure`.

New codec option: schema evolution. Some small annotations and a builder method.
Existing row format compatibility unchanged

Cross-language compatibility is documented and ready to share with any other interested implementations.

## Benchmark

  `withSchemaEvolution()` is an opt-in feature that adds a new row-codec path; it does not modify any existing serialization hot path.
  There is no `apache/main` baseline to compare against — `SchemaEvolutionSuite` exercises `withSchemaEvolution()`, which does not exist
  on `main` — so the benchmark measures two things directly: the steady-state cost of enabling the flag, and projection-vs-current
  parity.

  Bounded JMH run (JDK 26, 2 forks × 4 iterations, 1s each, `-prof gc`). `B/op` is `gc.alloc.rate.norm` (bytes allocated per operation).

  | Benchmark | Throughput (ops/s) | B/op |
  |---|---|---|
  | `currentDecode` | 17.6M | 312 |
  | `currentDecodeNoEvolution` | 16.6M | 312 |
  | `encode` | 15.9M | 152 |
  | `encodeNoEvolution` | 15.8M | 152 |
  | `compactCurrentDecode` | 16.4M | 280 |
  | `compactCurrentDecodeNoEvolution` | 16.3M | 280 |
  | `compactEncode` | 16.4M | 144 |
  | `compactEncodeNoEvolution` | 15.5M | 144 |
  | `olderDecode` | 24.5M | 216 |
  | `compactOlderDecode` | 24.9M | 192 |

  Findings:

  - **Enabling evolution adds zero allocation on the current path.** `B/op` is byte-identical between the evolution-on and `*NoEvolution`
  variants on every path (decode 312/312, encode 152/152, compact decode 280/280, compact encode 144/144).
  - **Throughput overhead of the flag is within the run's noise band.** Every on-vs-off pair overlaps within error. This bounded run has
  ~10% confidence intervals on the no-evolution variants, so the throughput claim is "no measurable difference," not a tight bound;
  allocation is exact.
  - **Projection (older-version) decode is not penalized versus current decode.** It allocates less here (216 vs 312 B/op standard; 192
  vs 280 compact) because it reads the narrower V1 schema, not because projection is inherently cheaper. Each projection codec holds its
  historical schema's precomputed row layout, so there is no per-decode rebuild.

---

### Limitations

  - Producer and consumer must agree on the `withSchemaEvolution()` flag; the two
    framings are not wire-compatible. A flag-mismatched peer fails loudly with
    `ClassNotCompatibleException` (except evolution-off reading evolution-on bytes,
    which is undefined). Adopt by enabling the flag on both sides in a release that
    changes no schema, then evolve schemas once every peer is on the new build.
    Data cannot be upgraded since the original format has no hash slot for array or map.
  - Evolution-enabled payloads are Java-only; cross-language consumers (Python,
    C++) cannot read them.
  - The number of generated projection codec classes grows as the product of the
    version counts of the *distinct* nested versioned bean classes. Retire entries
    from a bean's `History` interface once you no longer need to read payloads from
    that range to bound the growth.


docs/row-format/java.md
```diff
@@ -113,6 +113,78 @@ Row format is ideal for:
 - **Data pipelines**: Processing data without full object reconstruction
 - **Cross-language data sharing**: When data needs to be accessed from multiple languages
 
+## Schema evolution
```
stevenschlansker, line 280: This section should probably move down

docs/guide/java/row-format.md
```diff
@@ -113,6 +113,78 @@ Row format is ideal for:
...
+
+Remove a field by deleting the Java member and declaring it on a nested history interface as a
+method with a `@ForyVersion(until = N)`. The method's return type carries any parameterized
+type information from the original field.
+
+```java
+@Data
+@ForySchema(removedFields = Person.History.class)
+public class Person {
+  private String name;
```
stevenschlansker: remove private

docs/guide/java/row-format.md
```diff
@@ -113,6 +113,78 @@ Row format is ideal for:
...
+  private int age;
+
+  @ForyVersion(since = 2)
+  private String email;
+}
+```
+
+A v1 payload (with `name` and `age` only) decodes to a `Person` whose `email` is `null`.
+Primitive fields added later default to `0`, `0.0`, or `false`. If a class adopts versioning
+after its v1 is already in the wild, set `@ForySchema(baseVersion = N)` so unannotated fields
```
stevenschlansker: this is nonsense?

docs/row-format/java.md
```diff
@@ -113,6 +113,78 @@ Row format is ideal for:
...
+```
+
+The history method name matches the original live descriptor name: the field name for Lombok
+`@Data` or records (`age`, `tags`), or the full accessor name for JavaBeans-style classes and
+interfaces (`getAge`).
+
+### Wire format and limitations
+
+Producers and consumers must agree on the `withSchemaEvolution()` flag — they are not
+wire-compatible otherwise. Row payloads always carry an 8-byte hash slot; under evolution its
```
stevenschlansker, line 356: what about the upgrade path?

stevenschlansker:
@chaokunyang , this is not 100% done yet, but I think it is getting very close. Would you mind doing a design review to make sure that this feature will be acceptable to commit into the project? It is not a small feature, but I think the complexity is justified by giving the row format a low-cost method of achieving schema evolution.

If there is some way to streamline the eventual code review of a large volume of LLM-typed code, please let me know.
I have tried to keep the bar for quality high (many, many review iterations per ai policy) and testing thorough.
I considered breaking it up into multiple PRs but there is not a "slice" of the feature that stands on its own.

I intend to test this over the coming month or so internally, and mark the PR as ready once we have some real world experience running it.

java/fory-format/src/main/java/org/apache/fory/format/type/TypeInference.java
```diff
@@ -258,6 +263,16 @@ private static Field inferField(String name, TypeRef<?> typeRef, TypeResolutionC
...
   }
 
+  /**
+   * Infer a single named field from its Java type, used by schema-evolution code paths that need to
+   * reconstruct historical fields by name and type without going through a Java member.
+   */
+  static Field inferNamedField(String name, TypeRef<?> typeRef) {
+    TypeResolutionContext ctx =
+        new TypeResolutionContext(CustomTypeEncoderRegistry.customTypeHandler(), true);
+    return inferField(name, typeRef, ctx);
```
wenshao: **[Critical]** `inferNamedField` creates a `TypeResolutionContext` with an empty walked-type-path, so `getEnclosingType()` returns `Object.class`. This misses bean-scoped custom codecs that the normal `inferSchema` path and `SchemaHistory.Wrapper.classify` both find via the actual bean class. — Failure scenario: register a custom codec scoped to `(Foo.class, Bar.class)` whose `encodedType()` returns `Baz`, and enable schema evolution on `Foo`. `SchemaHistory.enumerate` reconstructs fields via `inferNamedField` which calls `findCodec(Object.class, Bar)` and misses the `(Foo, Bar)` registration. The history-derived schema has `Bar`'s struct, but the generated codec expects `Baz`'s layout — if field counts differ, `fieldOfSchema` throws `IndexOutOfBoundsException`; if types differ, the writer silently corrupts the row.

```suggestion
 static Field inferNamedField(String name, TypeRef<?> typeRef, Class<?> enclosingType) {
 TypeResolutionContext ctx =
 new TypeResolutionContext(CustomTypeEncoderRegistry.customTypeHandler(), true)
 .appendTypePath(enclosingType);
 return inferField(name, typeRef, ctx);
```

_— qwen3-coder-plus via Qwen Code /review (v0.21.2)_

java/fory-format/src/main/java/org/apache/fory/format/type/SchemaHistory.java
```diff
@@ -0,0 +1,936 @@
...
+ * from {@link DataTypes#computeSchemaHash} and is used only by versioning code paths.
+ */
+@Internal
+public final class SchemaHistory {
+
+  /** Implicit version of a live field that carries no {@link ForyVersion}. */
+  private static final int FIRST_VERSION = 1;
+
+  /** FNV-1a 64-bit offset basis, the seed for every strict-hash mix in this class. */
+  private static final long FNV_OFFSET_BASIS = 1469598103934665603L;
```
wenshao: **[Suggestion]** The FNV-1a 64-bit offset basis is `14695981039346656037` (`0xcbf29ce484222325`); the literal here is missing the trailing `7`, producing a different seed. Every other Fory runtime uses the correct value — C# `TypeResolver.cs:373`, C++ `type_index.h:53`, Rust `type_meta.rs:569`. — Concrete cost: the hash is self-consistent within Java so dispatch works today, but if schema-evolution hash comparison is ever extended cross-runtime or a user compares hash values for debugging, Java hashes will silently disagree. The comment explicitly claims this IS the FNV-1a 64-bit offset basis, making the code misleading.

```suggestion
 private static final long FNV_OFFSET_BASIS = 0xcbf29ce484222325L;
```

_— qwen3-coder-plus via Qwen Code /review (v0.21.2)_

java/fory-format/src/main/java/org/apache/fory/format/encoder/MapCodecBuilder.java
```diff
@@ -55,23 +66,246 @@ public class MapCodecBuilder<M extends Map<?, ?>> extends BaseCodecBuilder<MapCo
...
       }
     };
   }
 
+  /**
+   * Generate the projection row codec(s) and projection map codec for one (key-version,
+   * value-version) combination, and rebuild the map field with each position projected onto its
+   * historical schema. A position at its current schema gets an empty suffix and keeps its current
+   * field, so a map where only one side evolves pays no codegen for the unchanged side.
+   */
```
wenshao: **[Suggestion]** Orphaned Javadoc block: this comment is immediately followed by a second Javadoc block (the real documentation for `projectedPositionField`), so it attaches to no declaration and is invisible to javadoc and IDEs. — Concrete cost: the text describes `ProjectionSource.compile`'s contract but sits before `projectedPositionField`. A maintainer reading the source sees what looks like the method's contract, but it documents the wrong method; the real `compile` method has no equivalent top-level summary. The stale block will drift out of sync because nothing tooling-wise references it.

Suggested fix: move this Javadoc block to precede the `ProjectionSource.compile` method, or delete it if `compile`'s existing inline comments are sufficient.

_— qwen3-coder-plus via Qwen Code /review (v0.21.2)_

java/fory-format/src/test/java/org/apache/fory/format/encoder/ImplementInterfaceTest.java
```diff
@@ -329,6 +331,41 @@ public void testListTooLazy() {
...
+  }
+
+  /**
+   * Interface bean as a map value. Type inference reaches the map value type through {@code
+   * isSupported}, which must recognize the interface as a synthesizable bean the same way it does
+   * for a direct field or list element.
+   */
+  @Test
+  public void testMapValueInterface() {
+    final Map<String, ListInner> map = new HashMap<>();
```
wenshao, line 358: **[Suggestion]** The `TypeUtils.isSupported` ctx-threading fix is symmetric across map key and value, but only the value side is exercised by a test; no test uses an interface bean as a map key through this path. — Concrete cost: a future refactoring that accidentally reverts only the key-side call to single-arg `isSupported` would cause interface beans used as map keys to be rejected. No existing test would turn red: `testMapValueInterface` uses `String` as the key (always supported), and the map-key evolution tests in `SchemaEvolutionStressTest` use concrete `@Data` classes.

Suggested fix: add a counterpart `testMapKeyInterface` that uses an interface bean as the map key.

_— qwen3-coder-plus via Qwen Code /review (v0.21.2)_

wenshao (changes requested):

java/fory-format/src/main/java/org/apache/fory/format/type/TypeInference.java
stevenschlansker: Fixed in c36416a6f. inferNamedField now takes the enclosing bean class and appends it to the resolution path; SchemaHistory.enumerate threads the declaring bean through (void.class for top-level elements, matching Wrapper.classify).

One correction to the failure scenario: the row-level corruption does not reproduce, because the writer's wire hash and the reader's dispatch table both derive from the same reconstruction, and the generated codecs resolve codecs from the bean walk, so the two sides stay byte-consistent even when reconstruction is wrong. The observable break is the encoder's public schema(), which carried the raw domain type instead of the codec's declared column type. The regression test (evolvingBeanWithScalarCodec) pins that: without the fix, schema().int64Field("balance") fails with "Field balance is struct<cents: int64 not null>, expected int64".

java/fory-format/src/main/java/org/apache/fory/format/encoder/MapCodecBuilder.java
stevenschlansker: Fixed in c36416a6f: moved onto ProjectionSource.compile, the method it describes.

java/fory-format/src/test/java/org/apache/fory/format/encoder/ImplementInterfaceTest.java
stevenschlansker: Added testMapKeyInterface in c36416a6f: an interface bean as the map key with a String value, round-tripped through Encoders.bean. It fails if the key-side isSupported call reverts to the single-arg form.

java/fory-format/src/main/java/org/apache/fory/format/type/SchemaHistory.java
stevenschlansker: Fixed in c36416a6f, using the hex form 0xcbf29ce484222325L

stevenschlansker:
Thank you for the review @wenshao , I have incorporated your feedback and refreshed the PR with current main branch.

@chaokunyang , I am sorry to ping you again, but I just want to confirm that this architecture is a good direction for the Fory project. I will test this internally before asking for full review, but do not want to spend that time and risk if the feature would be rejected anyway. Thanks!

chaokunyang:
@stevenschlansker Schema evolution for the row format is a good direction. Since we’ll eventually want consistent support across all languages, it may be worth taking a broader look at the API and protocol design from a cross-language perspective.

stevenschlansker:
Thank you very much for the feedback. There were a few small xlang issues, now the design should be compatible out of the box.

PR #3719, opened 2026-05-30
fix(go): return nil serializer on getTypeInfo err

### LINKED ISSUE #3718: [Go] lazy serializer init silently swallows createSerializer error in getTypeInfo
## Bug

In `go/fory/type_resolver.go` line 1018, the lazy-serializer-initialization path inside `getTypeInfo` discards the error returned by `createSerializer`:

```go
serializer, err := r.createSerializer(value.Type(), false)
if err != nil {
    fmt.Errorf("failed to create serializer: %w", err) // result unused — error lost
}
info.Serializer = serializer  // nil serializer stored in cache
return info, nil              // false success returned to caller
```

The `fmt.Errorf` result is never used (detected by \`go vet\`). On failure, a \`nil\` serializer is stored in both \`typesInfo\` and `typePointerCache`, and the function returns \`(info, nil)\` — hiding the failure from all callers. This leads to a nil-pointer panic or silent data corruption far from the actual failure site.

All sibling call sites in the same file propagate the error correctly (e.g. \`getSerializerByType\` at line 942).

## Fix

```go
if err != nil {
    return nil, fmt.Errorf("failed to create serializer: %w", err)
}
```

## Regression test

A white-box test seeds \`typesInfo\` with a nil-serializer entry for \`**int\` (pointer-to-pointer is explicitly rejected by `createSerializer`), calls \`getTypeInfo\`, and asserts:
- a non-nil error is returned
- no nil-serializer entry is left in \`typePointerCache\`

## Detected by

`go vet`: `type_resolver.go:1018:5: result of fmt.Errorf call not used`

ayush00git:
## Related issues
Closes #3718 

## AI Contribution Checklist
Issue was flagged by claude
```
Confirmed bug

1. type_resolver.go:1018 — serializer-creation error is silently swallowed

Confidence: high (confirmed by go vet and by reading the code). Severity: medium–high.

serializer, err := r.createSerializer(value.Type(), false)
if err != nil {
    fmt.Errorf("failed to create serializer: %w", err)   // <-- result discarded
}
info.Serializer = serializer        // serializer is nil here on error
// ...
r.typePointerCache[typePtr] = info  // nil serializer gets cached
return info, nil                    // returns nil error despite failure

The error is built with fmt.Errorf but never returned or assigned. On the lazy-init path, when createSerializer fails, serializer is nil, that nil is stored into info.Serializer, cached, and the function returns (info, nil) — pretending success. A later serialize/deserialize then uses a nil serializer → wrong behavior or a nil-deref panic, far from the real cause.

Fix: return nil, fmt.Errorf("failed to create serializer: %w", err)

go vet flags this directly:
type_resolver.go:1018:5: result of fmt.Errorf call not used
```

### DIFF
--- a/go/fory/type_resolver.go
+++ b/go/fory/type_resolver.go
@@ -1015,7 +1015,7 @@ func (r *TypeResolver) getTypeInfo(value reflect.Value, create bool) (*TypeInfo,
 			*/
 			serializer, err := r.createSerializer(value.Type(), false)
 			if err != nil {
-				fmt.Errorf("failed to create serializer: %w", err)
+				return nil, fmt.Errorf("failed to create serializer: %w", err)
 			}
 			info.Serializer = serializer
 		}
--- a/go/fory/type_resolver_test.go
+++ b/go/fory/type_resolver_test.go
@@ -1,0 +1,64 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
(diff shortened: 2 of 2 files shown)


ayush00git:
@chaokunyang have a look

chaokunyang (approved):
LGTM

PR #3731, opened 2026-06-02
fix(compiler): handle nested container ref pointer options in Rust compiler correctly

### LINKED ISSUE #3730: [Rust] Rust compiler ignores nested container ref pointer options
### Search before asking

### Version

Fory: latest
OS: Ubuntu 24.04.3
GCC: 11.5.0

### Component(s)

Rust

### Minimal reproduce step

Run this testcase:
```python
def test_rust_nested_list_ref_uses_container_local_pointer_type():
    schema = parse_fdl(
        dedent(
            """
                package gen;

                message Node {
                    string value = 1;
                }

                message Request {
                    list<list<ref(thread_safe=true) Node>> groups = 1;
                }
                """
        )
    )
    rust_output = render_files(generate_files(schema, RustGenerator))
    assert "pub groups: ::std::vec::Vec<::std::vec::Vec<::std::sync::Arc<Node>>>," in rust_output
    assert "::std::vec::Vec<::std::vec::Vec<::std::rc::Rc<Node>>>" not in rust_output
```

The testcase should pass, i.e. `groups` should be compiled into `::std::vec::Vec<::std::vec::Vec<::std::sync::Arc<Node>>>`, not `::std::vec::Vec<::std::vec::Vec<::std::rc::Rc<Node>>>`.

### What did you see instead?

The testcase failed.

### Anything Else?

The Rust compiler does not correctly apply ref pointer options for **nested** container element/value references. The same issue can affect nested map value references, such as `map<string, map<string, ref(thread_safe=true) Node>>`.

The root cause is that `RustGenerator.generate_type` receives a single `pointer_type` argument computed once from the outer `Field` by `get_field_pointer_type(field)`:

https://github.com/apache/fory/blob/810943c24db7866bf4e57960c2fd8a3c1e4329fe/compiler/fory_compiler/generators/rust.py#L909-L918

That pointer type is then recursively reused for nested container types.

When recursion reaches a nested `ListType.element_ref` or `MapType.value_ref`, the generator does not recompute the pointer type from the nested container's own `element_ref_options` or `value_ref_options`. As a result, nested `ref(thread_safe=true)` metadata is ignored and the default pointer type `Rc` is used instead of `Arc`.

BaldDemian:
## Why?

The Rust compiler should correctly apply ref pointer options for **nested** container element/value references.

## What does this PR do?

- Handle nested container ref pointer options correctly.
- Add corresponding testcase.

## Related issues

Fixes https://github.com/apache/fory/issues/3730.

## AI Contribution Checklist

N/A.

## Benchmark

N/A.

### DIFF
--- a/compiler/fory_compiler/generators/rust.py
+++ b/compiler/fory_compiler/generators/rust.py
@@ -1039,12 +1039,18 @@ class RustGenerator(BaseGenerator):
         elif isinstance(field_type, ListType):
             effective_element_optional = element_optional or field_type.element_optional
             effective_element_ref = element_ref or field_type.element_ref
+            element_pointer_type = pointer_type
+            if field_type.element_ref:
+                element_pointer_type = self.get_pointer_type(
+                    field_type.element_ref_options,
+                    field_type.element_ref_options.get("weak_ref") is True,
+                )
             element_type = self.generate_type(
                 field_type.element_type,
                 nullable=effective_element_optional,
                 ref=effective_element_ref,
                 parent_stack=parent_stack,
-                pointer_type=pointer_type,
+                pointer_type=element_pointer_type,
             )
             list_type = f"::std::vec::Vec<{element_type}>"
             if ref:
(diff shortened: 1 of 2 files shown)


chaokunyang (approved):
LGTM

Issue #3266, opened 2026-02-05
[Compiler] Grpc Support


chaokunyang:

chaokunyang:
Created sub-tasks (tracking in this issue):

- #3267 IR nodes + validation rules
- #3268 FDL service/rpc/stream parsing
- #3269 Protobuf service/rpc parsing + translation
- #3270 FlatBuffers rpc_service parsing (table + struct)
- #3271 CLI options + generator service hook
- #3272 Java gRPC stubs with Fory codec
- #3273 Python gRPC stubs with Fory codec
- #3274 Go gRPC stubs with Fory codec
- #3275 Rust gRPC stubs with tonic + Fory codec
- #3276 C++ gRPC stubs with Fory codec
- #3277 Test plan + smoke tests

chaokunyang:
Added JS/Dart sub-tasks:

- #3278 JavaScript/TypeScript gRPC stubs with Fory codec
- #3279 Dart gRPC stubs with Fory codec

eayushsingh:
Hi, I’m interested in this GSoC project.

I’ve started exploring the repository and trying to understand the compiler and Java modules, especially around code generation and serialization.

I plan to begin with smaller contributions and gradually work towards this feature. Any suggestions on where to start would be really helpful.

Thanks!

PR #3732, opened 2026-06-02
feat(compiler): reject non-thread-safe payload types in Rust compiler

BaldDemian:
## Why?

This is a prerequisite PR for Rust gRPC code generation. 

In the Rust gRPC code generation requirement, we chose tonic as the network transport layer for gRPC, so the generated Rust code needs to be compatible with tonic.

In tonic, a gRPC payload (a request or a response) type must be both `Send` and `Sync`, e.g. 

https://docs.rs/tonic/latest/tonic/client/struct.Grpc.html#method.unary

However, in Fory, users can define the following types, which cause the generated Rust types to be neither `Send` nor `Sync`:

- Non-thread-safe refs
- `any`

In addition, Fory currently generates an `UnknownCase` by default for a union type.

https://github.com/apache/fory/blob/33b0d6e8b73d1769d98824195b9155043266f3fe/compiler/fory_compiler/generators/rust.py#L379-L380

However, the type of its `value` field is `Arc<dyn Any>`, rather than something like `Arc<dyn Any + Send + Sync>`.

https://github.com/apache/fory/blob/33b0d6e8b73d1769d98824195b9155043266f3fe/rust/fory-core/src/types/unknown_case.rs#L24-L32

## What does this PR do?

In this PR, I made the following changes:

1. Before generating Rust gRPC stub code, all types used by gRPC payloads are checked via graph traversal. If their fields contain non-thread-safe refs or `any`, an error will be raised directly with a corresponding diagnostic message for the user.
2. `UnknownCase` is not generated for unions used by gRPC payloads. Currently, I use a special macro attribute `#[fory(no_unknown_case)]` to make the derive macro logic handle this case specially, rather than directly panicking when encountering a union without an `UnknownCase`.
3. Add testcases accordingly.

It is worth noting that, for simplicity, if a gRPC payload type contains imported types, the current implementation will directly raise an error and exit.

## Related issues

https://github.com/apache/fory/issues/3266

## AI Contribution Checklist

- [ ] Substantial AI assistance was used in this PR: `no`

All unions in gRPC payload types will not generate `UnknownCase`.

## Benchmark

N/A.

### DIFF
--- a/compiler/fory_compiler/generators/rust.py
+++ b/compiler/fory_compiler/generators/rust.py
@@ -93,11 +93,127 @@ class RustGenerator(BaseGenerator):
     def generate(self) -> List[GeneratedFile]:
         """Generate Rust files for the schema."""
         files = []
+        self._grpc_reachable_type_ids = (
+            self.validate_grpc_payload_safe() if self.options.grpc else set()
+        )
 
         # Generate a single module file with all types
         files.append(self.generate_module())
 
         return files
+
+    def is_grpc_reachable_type(self, type_def: object) -> bool:
+        if not hasattr(self, "_grpc_reachable_type_ids"):
+            raise AttributeError(
+                "RustGenerator gRPC reachability is initialized by generate()"
+            )
+        return id(type_def) in self._grpc_reachable_type_ids
+
+    def grpc_payload_root_names(self) -> Set[str]:
+        names: Set[str] = set()
+        for service in self.schema.services:
+            if self.is_imported_type(service):
+                continue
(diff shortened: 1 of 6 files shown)


BaldDemian:
To address the Send and Sync issue, I have tried several other approaches. However, the current implementation appears to be the simplest one, although it may slightly affect the original design intent of `UnknownCase`. Happy to hear your feedback or advice @chaokunyang

BaldDemian:
@chaokunyang Thanks for addressing the issue, this really saved my day 🚀

PR #3744, opened 2026-06-08
feat(compiler): handle Rust identifier escaping and name collisions

BaldDemian:
## Why?

- Identifiers colliding with Rust keywords should be escaped.
- After naming normalization, some identifiers may collide with each other.

The Fory compiler should handle above edge-cases carefully by sanitizing the identifiers or throwing collision errors to the user, instead of generating Rust code that cannot be compiled by the Rust compiler.

### Scope

This PR handles Rust identifier escaping and name collisions in both normal `message` definition and gRPC service definition.

**Note**: for now the handling logic for gRPC service definition is also defined in `compiler/fory_compiler/generators/rust.py`, I plan to move them into `compiler/fory_compiler/generators/services/rust.py` in the Rust gRPC code generation PR.

### Workflow

The current implementation broadly follows the approach used by prost-build.
It first normalizes identifiers from the IDL according to Rust naming conventions, e.g. converting struct names to UpperCamelCase and field names to snake_case. 
It then checks whether the normalized identifiers are Rust keywords and escapes them when necessary. 
Finally it checks the sanitized (i.e. normalized + escaped) identifier with already allocated identifiers in the same scope for any collision. The sanitized identifiers are cached to make sure the definition site and all use sites refer to the same name.

## Related issues

Close https://github.com/apache/fory/issues/3544.

## AI Contribution Checklist

N/A.

## Benchmark

N/A.

### DIFF
--- a/compiler/fory_compiler/cli.py
+++ b/compiler/fory_compiler/cli.py
@@ -129,6 +129,7 @@ def resolve_imports(
     imported_messages = []
     imported_unions = []
     resolved_import_files = []
+    source_packages: Dict[str, Optional[str]] = {str(file_path): schema.package}
 
     for imp in schema.imports:
         # Resolve import path using search paths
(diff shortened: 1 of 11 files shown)


chaokunyang (approved):
LGTM

Issue #3370, opened 2026-02-19
[Swift] Swift gRPC Integration for Fory


chaokunyang:
This issue is reserved for gsoc, do not take over it currently.

### Feature Request

Swift gRPC Integration in Apache Fory

## Problem Statement

Fory Swift currently lacks:

- Generated Swift transport bindings for services.
- A Fory-backed codec for grpc-swift.
- Official examples, integration tests, and comprehensive documentation for Swift gRPC.

## Expected Outcomes

- Generate Swift gRPC transport code from service definitions.
- Support unary, client-streaming, server-streaming, and bidirectional-streaming RPCs.
- Provide a grpc-swift codec backed by Fory serialization and deserialization.
- Provide deterministic, safe decode behavior, including zero-copy when valid and copy fallback when required.
- Add integration tests and runnable examples.
- Provide comprehensive Swift gRPC integration documentation.

### 1) Service gRPC Code Generation

- Generate `service.swift` for service protocols and method signatures.
- Generate `service_grpc.swift` for grpc-swift client/server bindings.
- Ensure generated signatures align with existing Swift model types.

### 2) grpc-swift Codec Integration

- Implement codec encode and decode using Fory Swift runtime.
- Map request and response types to generated service APIs.
- Provide clear error mapping for invalid payloads and type mismatches.

### 3) Decode Path Behavior

- Add zero-copy decode path when ownership and lifetime constraints are safe.
- Add copy fallback path when zero-copy cannot be guaranteed.
- Keep behavior deterministic and memory-safe.

### 4) Tests

Compiler and integration tests should cover:

- Generated file names and key service signatures.
- Unary, client-streaming, server-streaming, and bidi-streaming coverage.
- Codec round-trip and error-path behavior.
- Zero-copy path and fallback path validation.

### 5) Examples and Comprehensive Documentation

- Add runnable Swift gRPC server and client examples using generated code and codec.
- Update compiler and Swift docs for gRPC generation and integration usage.
- Document supported grpc-swift version(s), Swift toolchain version(s), and compatibility constraints.
- Add a clear end-to-end quickstart: IDL -> codegen -> build -> run server -> run client.
- Document generated files and API shape (`service.swift`, `service_grpc.swift`) with method mapping by RPC mode.
- Document codec behavior, error mapping, and zero-copy vs fallback rules and safety constraints.
- Add troubleshooting guidance for common setup/runtime errors and how to diagnose them.
[...]

PR #3776, opened 2026-06-21
feat(compiler): Add grpc support for Swift

yash-agarwa-l:
## Why?
Swift users can generate Fory model types today, but schemas that define services
do not produce Swift gRPC companions. This leaves Swift out of the existing `--grpc`
workflow used by the other supported service-generation targets.

## What does this PR do?
- Adds Swift gRPC companion generation for Fory compiler services, including service
  metadata descriptors, an `EventLoopFuture` provider, an async/await provider, an
  async client, and Fory-backed request/response stream adapters, targeting grpc-swift 1.x.
- Serializes request and response bodies with Fory instead of protobuf through an
  internal `GRPCPayload` wrapper. Because the Swift `Fory` runtime is single-threaded,
  the wrapper builds one `Fory` per thread from the schema module's own configuration
  and registrations, so concurrent RPCs are race-free (verified under ThreadSanitizer).
- Emits an async-only client; the `EventLoopFuture` client and interceptor hooks are
  omitted because their generated types would expose the internal wrapper.
- Adds Swift preflight validation for generated output-path and top-level symbol
  collisions, and reserves inherited provider/client member names (`handle`,
  `serviceName`, `channel`, `defaultCallOptions`) so a clashing rpc fails codegen
  with a clear message.
- Updates compiler and service-codegen tests to cover the four streaming shapes,
  identifier escaping, imported and nested service types, the default package, the
  protobuf and FlatBuffers frontends, collision handling, a SwiftPM build-and-run
  fixture, and a marshaller concurrency test under ThreadSanitizer (gated behind
  `FORY_SWIFT_TSAN`).
- Wires Swift output into the cross-language gRPC generation helper.
- Documents Swift gRPC support, dependencies, generated API shape, streaming, the
  shared-top-level-package limitation, troubleshooting, and compiler guide updates.

Draft: Java<->Swift cross-language interop tests (all four modes, all three IDL
frontends) are in progress and will be added before this PR is marked ready for review.

## Related issues
#3266
#3370

## AI Contribution Checklist
- [ ] Substantial AI assistance was used in this PR: `yes`
- [ ] If `yes`, I included a completed AI Contribution Checklist in this PR description and the required `AI Usage Disclosure`.

  - Adds generated Swift gRPC companion APIs when `foryc --swift_out=... --grpc` is used.
  - The generated services use Fory-encoded gRPC message bodies, but this PR does not
    change the Fory binary protocol.

## Benchmark
Not applicable.


chaokunyang:
@yash-agarwa-l Please git merge apache/main first to address the conflicts

yash-agarwa-l:
I already did that, so I'm not sure why it's still showing up. Let me mark it as ready for review and double-check.

yash-agarwa-l:
Hi @chaokunyang, small update and one thing I'd like your call on.

The Swift gRPC codegen is done and Swift↔Swift round-trips fine, but Java↔Swift interop fails. After digging in, it looks like a gap in the Fory Swift core serializer, not the gRPC code.

When refTracking is on, Swift writes `NOT_NULL_VALUE` (0xff) for a struct message, while Java/Python/Rust write `REF_VALUE` (0x00) and reserve a ref slot. (Go reaches `REF_VALUE` only because grpc-go marshals a pointer; its value-struct path opts out like Swift.) Without the reserved slot, the ref-id counters drift and Java's decode hits Index -1. I checked it's just that one flag byte by flipping it. The reason is that Swift structs are `isRefType` = `false`, so `foryWrite` skips tracking even when it's on. This might be on purpose, since structs are value types. Rust handles the same case by writing `REF_VALUE` + `reserve_ref_id()` for value-type structs under tracking.

If that sounds right to you, I'd like to do the same for Swift. One thing to flag: it changes tracking-mode output for all Swift structs, not just gRPC. Tracking is opt-in and off by default, so the impact should be small, but it's a core wire-format change, so I wanted to ask first. No gRPC module changes; the contract stays refTracking:true.

Two questions:
1. Would it be fine to make Swift value-type structs take part in ref tracking?
2. Is the current opt-out on purpose, or just not done yet? Want to be sure I'm not missing a reason it writes `NOT_NULL_VALUE`.

Since it's a wire-format change, I'm happy to do whatever works best for you here, a separate issue or PR, or just keep it on this PR. Let me know.

chaokunyang:
When ref tracking is enabled, java/python will still push an id -1 to ref reader, and when inner struct invoke `RefReader.reference()`, it will skip reference for -1 . Could you check this logic? `NOT_NULL_VALUE` should not be an issue. Use `REF_ID` is just working around the root cause.

```java
  /** Stores {@code object} under an already reserved read ref id. */
  @Override
  public void setReadRef(int id, Object object) {
    if (id >= 0) {
      readObjects.set(id, object);
    }
  }
```

There are some code invoking `preserveRefId(-1)`, it's exactly for such cases:

```java
      if (refMode != RefMode.NULL_ONLY || buffer.readByte() != Fory.NULL_FLAG) {
        refReader.preserveRefId(-1);
        return readContext.readNonRef(fieldInfo.typeInfo);
      }
```

yash-agarwa-l:
Thanks, I have traced the failure, you were right that this is a reader-side issue and that NOT_NULL_VALUE is legal. 

Under tracking, `tryPreserveRefId` pushes an id for `REF_VALUE` but not for `NOT_NULL_VALUE`. Swift sends `NOT_NULL_VALUE` for value-type structs, which the spec allows, so when the read path then calls `reference()` it pops an empty stack and we get `Index -1`.

The natural place to fix it is centrally, inside `tryPreserveRefId`, but that doesn't hold up. Not every caller pops. `readStringRef` goes through `tryPreserveRefId` and then into `readString`, which never calls `reference()`, so a push in the shared spot leaves a stray `-1` behind. Some later object pops that instead of its own, and the refs drift with nothing failing to show it. What works is a guarded push: `preserveRefId(-1)` on `NOT_NULL` only when the serializer actually tracks, gated on `needToWriteRef()`. 

Checked on current main, including the string path (round-trips with no stray id) and a tracked compatible struct with polymorphic leaves plus a back-reference. It's proven at the byte level, Swift bytes into Java; the live process-to-process harness is still parked.

PTAL if this is fine? Also shall I create a separate PR for the java fix or is it fine here?

chaokunyang:
@yash-agarwa-l You should only change swift code for this error, you should not change java code. Swift is similiar to rust/c++, both of them can work without changing java code. You could take how fory rust/c++ work as a reference.

yash-agarwa-l:
Thanks @chaokunyang, I have made the revision, One thing to note: this does change Swift's tracked wire output 0xff  to 0x00, so an older Swift peer won't match a new one under tracking.

compiler/fory_compiler/generators/services/swift.py
```diff
@@ -0,0 +1,628 @@
...
+    "serviceName",
+    "channel",
+    "defaultCallOptions",
+}
+
+
+class SwiftServiceMixin:
+    """Generates Swift gRPC service companions backed by Fory serialization."""
+
+    def generate_services(self) -> list[GeneratedFile]:
```
chaokunyang: Keep the compiler importable on Python 3.8 and 3.9

`fory-compiler` still declares Python 3.8+ support, but this module has no `from __future__ import annotations`; Python 3.8 evaluates `list[GeneratedFile]` during the eager `SwiftGenerator` import and raises `TypeError`. `cli.py` has the same problem, and its `list[Path] | None` annotations also break Python 3.9. Please retain compatible annotation syntax or postpone annotations in every affected module, and add Python 3.8/3.9 CLI-import coverage.

compiler/fory_compiler/generators/services/swift.py
```diff
@@ -0,0 +1,628 @@
...
+        )
+        file_name = f"{self.to_pascal_case(service.name)}Grpc.swift"
+        path = f"{package_path}/{file_name}" if package_path else file_name
+        return GeneratedFile(path=path, content=content)
+
+    def _marshaller(self, base: str, module: str) -> list[str]:
+        # The Swift Fory instance is single-threaded, so keep one per thread.
+        return [
+            "private enum ForyRuntime {",
+            f'    private static let key = "org.apache.fory.grpc.{module}"',
```
chaokunyang: Make the thread-local key unique across Swift modules

This key contains only the generated textual type path, so every default-package schema uses `org.apache.fory.grpc.ForyModule`. If two generated schemas live in separate Swift modules, the second wrapper retrieves the first module's `Fory`, skips its own registrations, and serializes with the wrong resolver. Please include the runtime Swift module/type identity in the key and cover two generated targets round-tripping sequentially on the same thread.

compiler/fory_compiler/generators/swift.py
```diff
@@ -201,6 +202,21 @@ def output_file_path(self) -> str:
             return f"{package_path}/{file_name}"
         return file_name
 
+    def swift_model_top_level_symbols(self) -> set[str]:
+        # In enum style with a package, types live nested under a shared namespace
+        # enum, so there are no collidable top-level names. Flatten and default
+        # packages put each type and the module helper at file scope.
+        components = self._package_components_for_schema(self.schema)
+        if self.get_namespace_style() == "enum" and components:
```
chaokunyang: Include the package namespace in collision preflight

Enum-style packages still emit a top-level namespace enum. For example, `demo.shared` and `demo.greeter` each emit `public enum Demo`, but this branch records no symbol, so preflight succeeds and Swift compilation later fails with `invalid redeclaration`. Please record the actual top-level namespace owner, or change generation so duplicate declarations are avoided.

compiler/fory_compiler/generators/swift.py
```diff
@@ -201,6 +202,21 @@ def output_file_path(self) -> str:
...
+            return set()
+        symbols: set[str] = set()
```
chaokunyang: Preserve duplicate symbol owners during preflight

Returning a `set` loses duplicate declarations produced within one schema. For a default-package schema declaring `message ForyModule {}`, the model and generated helper both become top-level `ForyModule`, but preflight sees only one symbol. Flattened helper collisions and normalized-name collisions have the same problem. Please preserve each declaration and its Swift scope instead of deduplicating before ownership validation.

compiler/fory_compiler/generators/services/swift.py
```diff
@@ -0,0 +1,628 @@
...
+        package = self.schema.package
+        package_path = package.replace(".", "/") if package else ""
+        file_name = f"{self.to_pascal_case(service.name)}Grpc.swift"
+        return f"{package_path}/{file_name}" if package_path else file_name
+
+    def swift_grpc_service_symbols(self, service: Service) -> list[str]:
+        base = self._service_symbol(service)
+        modes = {streaming_mode(m) for m in service.methods}
+        symbols = [
+            f"{base}Message",
```
chaokunyang: Do not reserve a marshaller symbol for empty services

`<Base>Message` is registered unconditionally, but the marshaller is emitted only when the service has methods. Consequently `message GreeterMessage {}; service Greeter {}` is rejected even though the service file contains no `GreeterMessage` wrapper. Please register this symbol only for non-empty services and add an empty-service preflight case.

compiler/fory_compiler/generators/services/swift.py
```diff
@@ -0,0 +1,628 @@
...
+                f"{base}ResponseStream",
+            ]
+        if StreamingMode.CLIENT_STREAMING in modes:
+            symbols.append(f"{base}UnaryResponseContext")
+        if modes & {StreamingMode.CLIENT_STREAMING, StreamingMode.BIDIRECTIONAL}:
+            symbols.append(f"{base}AsyncRequestStream")
+        return symbols
+
+    def _swift_grpc_method_name(self, method: RpcMethod) -> str:
+        return self.safe_member_name(method.name)
```
chaokunyang, line 87: Reject or rename an underscore-only RPC method

FDL accepts an RPC named `_`, but the generated member remains `_`, producing declarations and references such as `static let _`, `func _`, and `Methods._` that Swift cannot use as identifiers. Please generate a referencable name or reject this RPC name during preflight.

compiler/fory_compiler/generators/services/swift.py
```diff
@@ -0,0 +1,628 @@
...
+            "    }",
+            "}",
+            "",
+            "// Internal Fory wire wrapper for gRPC request and response messages.",
+            "// NIOCore.ByteBuffer is qualified because `import Fory` also exposes one.",
+            "// grpc-swift transfers this carrier between the calling task and the event",
+            "// loop, so the payload must be Sendable. The carrier itself only stores that",
+            "// payload, so @unchecked covers the wrapper while Value carries the guarantee.",
+            (
+                f"struct {base}Message<Value: Serializer & Sendable>: GRPCPayload,"
```
chaokunyang, line 180: Align the Sendable constraint with generated model types

This constraint applies to every request and response wrapper, but generated public structs, enums, unions, and classes do not conform to `Sendable`. Swift 6 therefore rejects even unary providers and clients when `Message<Request/Response>` is instantiated, not only client-streaming and bidi as the guide states. Please add a safe `Sendable` strategy plus Swift 6 build coverage, or state that the entire companion currently requires Swift 5 mode.

docs/grpc/swift.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Add Dependencies
+
+The `Fory` package does not depend on grpc-swift. Add grpc-swift in the package
+that compiles or runs the generated companions:
+
+```swift
+// Package.swift
+dependencies: [
+    .package(url: "https://github.com/apache/fory.git", from: "1.2.0"),
```
chaokunyang: Require a Fory version that contains this runtime support

`from: "1.2.0"` allows versions whose Swift `Serializer` lacks `Target`, and even 1.6.1 lacks this PR's tracked-value root ref-slot fix required for Java/Swift interoperability. Please use the first release containing both the compiler and runtime changes, or follow the existing `$version` convention.

integration_tests/grpc_tests/swift/interop/Tests/ForyGrpcTests/RoundTripTests.swift
```diff
@@ -0,0 +1,144 @@
...
+    for try await union in requestStream { try await responseStream.send(union) }
+  }
+}
+
+// Hosts the generated provider in-process and exercises the generated async client
+// across all four streaming modes (the relocated SwiftPM build-and-run fixture).
+final class RoundTripTests: XCTestCase {
+  func testInProcessAllStreamingModes() async throws {
+    let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+    defer { try? group.syncShutdownGracefully() }
```
chaokunyang: Use async cleanup in the async test

The current Java/Swift CI job reports async-context warnings for this blocking shutdown, lines 118/124, and `main.swift:456`; Swift 6 upgrades at least one to an error. Please use awaited `get()` or structured async lifecycle cleanup while preserving failure-path cleanup. Swift changes are required to compile without warnings.

.github/workflows/ci.yml
```diff
@@ -1016,6 +1016,49 @@ jobs:
...
+      - name: Generate gRPC test sources
+        run: python integration_tests/grpc_tests/generate_grpc.py
+      - name: Build Swift gRPC peer
+        run: |
+          cd integration_tests/grpc_tests/swift/interop
+          swift build -c release --product interop
+      - name: Run Swift gRPC package tests
+        run: |
+          cd integration_tests/grpc_tests/swift/interop
+          swift test
```
chaokunyang, line 1110: Run the concurrency regression under ThreadSanitizer

This runs only plain `swift test`. The sole `--sanitize=thread` command is behind `FORY_SWIFT_TSAN=1` in another script that this workflow never invokes, so the test and PR description's TSan safety claim is not exercised by CI. Please add a sanitized invocation, preferably filtered to the marshaller concurrency test.

integration_tests/grpc_tests/run_tests.sh
```diff
@@ -56,6 +56,19 @@ if has_test_class "DartGrpcTest"; then
   dart analyze bin lib/generated/*/*_grpc.dart
   dart format --output=none --set-exit-if-changed bin lib/generated/*/*_grpc.dart
 fi
+# Swift toolchain tests (generated marshaller round-trip and concurrency). These
+# need the Swift toolchain rather than the JVM, so they run in their own package.
+if command -v swift >/dev/null 2>&1 && [ -d "${SCRIPT_DIR}/swift/interop" ]; then
```
chaokunyang: Honor the selected test classes for Swift work

This block ignores `TEST_CLASSES`, so `./run_tests.sh GoGrpcTest` still resolves and runs the Swift package whenever Swift is installed. Please put both the Swift package tests and the `SwiftGrpcTest` build behind `has_test_class "SwiftGrpcTest"`, while retaining the tool-availability check before invoking Swift.

compiler/fory_compiler/generators/swift.py
yash-agarwa-l: I went with recording the namespace. Now, `demo.shared` + `demo.greeter` fails early with a Demo collision instead of reaching swiftc.

integration_tests/grpc_tests/swift/interop/Tests/ForyGrpcTests/MarshallerThreadSafetyTests.swift
```diff
@@ -0,0 +1,85 @@
...
+    XCTAssertEqual(fromShared.value, probe)
+
+    var outbound = allocator.buffer(capacity: 64)
+    try GrpcFdl_FdlGrpcServiceMessage(probe).serialize(into: &outbound)
+    let fromMarshaller: GrpcFdl.GrpcFdlRequest =
+      try GrpcFdl.ForyModule.getFory().deserialize(Data(outbound.readableBytesView))
+    XCTAssertEqual(fromMarshaller, probe)
+  }
+
+  func testSeparateModulesKeepOwnRuntimeOnOneThread() throws {
```
chaokunyang: Make the runtime-key regression test reproduce the old collision

`GrpcFdl` and `GrpcFbs` are both compiled into the single `ForyGrpcGenerated` Swift target, and the old textual keys were already different (`GrpcFdl.ForyModule` versus `GrpcFbs.ForyModule`). This test therefore passes even if the `String(reflecting:)` fix is removed. Please exercise two Swift targets/modules that expose the same textual helper path, ideally default-package `ForyModule`, so the test fails on the old implementation.

docs/grpc/swift.md
```diff
@@ -0,0 +1,235 @@
...
+returns call objects parameterized by the on-the-wire message type, which would
+expose the internal Fory wrapper, so it is not emitted. Both providers (async and
+`EventLoopFuture`) are generated.
+
+Interceptors are not generated. grpc-swift interceptors are typed on the
+on-the-wire message, which is the internal Fory wrapper; emitting interceptor
+hooks would expose that wrapper. Use a custom channel or server configuration for
+cross-cutting concerns instead.
+
+Swift models put each package under a nested `enum` namespace, so two schemas that
```
chaokunyang, line 219: Update the namespace limitation for the new preflight

The compiler now rejects schemas with the same root namespace before writing either file, so this section no longer describes the actual failure point. The separate-module workaround also requires separate generation invocations because one invocation runs the shared collision preflight before module assignment is known. Please document the preflight failure and that requirement.

compiler/fory_compiler/generators/services/swift.py
yash-agarwa-l: thanks,  I had ruff 0.16.0 locally while CI pins 0.15.22, and when I ran `ruff check --fix,` it perhaps rewrote the annotations. I have validated the ruff with the exact version mentioned in the ci file now. 

Also after adding the 3.8/3.9 import check, it surfaced some same issues with the frontend files. I have fixed them.

compiler/fory_compiler/generators/swift.py
yash-agarwa-l: Made the revision, I have also fixed duplicates nested inside a namespace enum, those were invisible to preflight before, PTAL.

chaokunyang: The top-level and helper cases are fixed, but nested IDL declarations still bypass the new walker. `swift_declared_symbols()` only iterates `schema.enums + schema.unions + schema.messages`; it never recurses through each message’s `nested_messages`, `nested_enums`, and `nested_unions`. For example, `message Parent { message my_type {} message MyType {} }` passes raw-name validation but emits two `Parent.MyType` declarations. The new test named `nested` covers package namespace nesting rather than an IDL nested type. Please collect declaration owners recursively by parent Swift scope and add a real nested-type case.

docs/grpc/swift.md
chaokunyang: The preflight wording is closer, but the separate-invocation workaround does not work for the service-import example above. `validate_swift_files()` recursively collects imports, so compiling `demo.greeter` still includes `demo.shared` in the graph and rejects the duplicate `Demo` even if the shared schema was generated in another invocation. Separate invocations only help unrelated schemas; an import graph currently needs disjoint top-level packages. Please narrow the workaround accordingly.

chaokunyang (approved):

PR #3789, opened 2026-06-25
fix(c++): make temporal types hashable

### LINKED ISSUE #3787: [C++] Generated C++ code fails to compile maps with `duration` or `timestamp` keys
### Search before asking

### Version

Latest Fory

### Component(s)

C++

### Minimal reproduce step

Consider this FDL:

```
package repro.temporal_key;

message Holder {
    map<duration, string> values = 1;
}
```

The generated code should compile.

### What did you see instead?

The generated code fails to compile.

## Root Cause

This bug is introduced by PR https://github.com/apache/fory/pull/3745. In this PR, we generate `unordered_map` for IDL's `map` type to match the behaviors of other languages. 

But in C++, the key type of an `unordered_map` must be hashable.

Fory documents `duration` and `timestamp` as valid IDL `map` key types. Currently, Fory maps `duration` to `fory::serialization::Duration` and timestamp to `fory::serialization::Timestamp`. But these two types are just aliases for standard library types:

https://github.com/apache/fory/blob/6c6ba3fb4172a71811c57664ae2a892f04208b5b/cpp/fory/serialization/temporal_serializers.h#L29-L38

In C++17, these standard library types do not provide usable `std::hash` specializations, so `std::unordered_map<fory::serialization::Duration, ...>` and `std::unordered_map<fory::serialization::Timestamp, ...>` fail to compile.

~~`date` does not hit the same problem because Fory maps it to a Fory-owned `Date` type instead of a standard library alias.~~ `date` can also hit the same problem.

## Proposed Fix

Replace `fory::serialization::Duration` and `fory::serialization::Timestamp` from standard-library aliases with Fory-owned wrapper types. 

Duration should store a normalized seconds/nanoseconds representation or total nanoseconds, and Timestamp should store seconds/nanoseconds since the Unix epoch. 
These wrapper types should provide equality, ordering, and hash methods, so they can be used directly as `unordered_map` keys.

The serializers can keep the existing wire format unchanged by reading from and writing to the wrapper fields using the same seconds/nanoseconds encoding.

BaldDemian:
## Why?

C++ temporal types (date, duration, timestamp) should be hashable so they can be used as key in `unordered_map`.

## What does this PR do?

1. Define `fory::Timestamp` and `fory::Duration` in `cpp/fory/type/temporal.h ` as wrapper types for `std::chrono::time_point` and `std::chrono::nanoseconds`, instead of their aliases. Inside these wrapper types, conversion APIs to and from the chrono representations are provided. I also move `fory::Serialization::Date` into `cpp/fory/type/temporal.h` as `fory::Date`.

2. Specify `std::hash` for temporal types (date, duration, timestamp).

3. Add serializers for raw `std::chrono::time_point` and `std::chrono::nanoseconds`. 

4. In `type_resolver.h`, reject registering `chrono` temporal types for `std::any`. chrono nanoseconds and chrono timestamp still work as explicit target types, but dynamic `std::any` always uses the Fory carriers.

5. Modify/Add testcases accordingly.

6. Modify the doc accordingly.

## Related issues

Fixes https://github.com/apache/fory/issues/3787.

## AI Contribution Checklist

Yes. The changes are documented accordingly.


BaldDemian:
Hi @chaokunyang, I also noticed that, the doc doesn't specify `any` `message` `union` as `map` key types:

https://github.com/apache/fory/blob/6c6ba3fb4172a71811c57664ae2a892f04208b5b/docs/compiler/schema-idl.md?plain=1#L1492-L1498

But the code 

https://github.com/apache/fory/blob/6c6ba3fb4172a71811c57664ae2a892f04208b5b/compiler/fory_compiler/ir/validator.py#L40-L46

and 

https://github.com/apache/fory/blob/6c6ba3fb4172a71811c57664ae2a892f04208b5b/compiler/fory_compiler/ir/validator.py#L550-L553

actually allows `any` `message` `union` as key types.

So which is the intention of Fory's design?

chaokunyang:
IDL should not allow any/union/message/list/map/float as map keys. Some types are mutable and don't have stable hash. Some types just don't have hash func

cpp/fory/serialization/temporal_serializers.h
```diff
@@ -21,21 +21,63 @@
...
+  bool operator<(const Duration &other) const { return ns_ < other.ns_; }
+
+private:
+  std::chrono::nanoseconds ns_;
+};
 
 /// Timestamp: point in time as nanoseconds since Unix epoch (Jan 1, 1970 UTC)
-using Timestamp = std::chrono::time_point<std::chrono::system_clock,
-                                          std::chrono::nanoseconds>;
+class Timestamp {
```
chaokunyang: put timestamp and duration into cpp/fory/type/temporal.h and cpp/fory/type/duration.h

BaldDemian: Do you mean we should put timestamp and duration in a separate header file? Should we include date in that file too?

cpp/fory/serialization/temporal_serializers.h
```diff
@@ -315,5 +358,142 @@ template <> struct Serializer<Date> {
...
+// These allow users to explicitly request chrono types as the deserialization
+// target (e.g., fory.deserialize<std::chrono::nanoseconds>(...)).  They share
+// the same wire encoding as the Fory-owned carrier serializers above and
+// delegate to them via conversion
+// ============================================================================
+
+/// Serializer for std::chrono::nanoseconds
+/// Per xlang spec: serialized as signed varint64 seconds + signed int32
+/// nanoseconds
+template <> struct Serializer<std::chrono::nanoseconds> {
```
chaokunyang, line 299: This makes chrono temporal types valid Any registrations for the same internal TypeId as the Fory carriers. Since `register_any_type` installs `any_read_fn` on the shared `TypeInfo` for `DURATION`/`TIMESTAMP`, the dynamic deserialization carrier becomes registration-order dependent. We should keep Any/polymorphic reads canonical: `DURATION`/`TIMESTAMP` should deserialize to the Fory carrier types, while chrono remains an explicit static target adaptation. Please add coverage for `std::any` duration/timestamp and prevent chrono registration from overriding the canonical carrier.

cpp/fory/serialization/temporal_serializers.h
```diff
@@ -21,21 +21,63 @@
...
 namespace serialization {
 
 // ============================================================================
-// Temporal Type Aliases
+// Fory-owned temporal carrier types
 // ============================================================================
 
 /// Duration: absolute length of time as nanoseconds
-using Duration = std::chrono::nanoseconds;
+class Duration {
```
chaokunyang: Now that `Duration`/`Timestamp` are Fory-owned carrier types, they should not be defined in the serializer header/namespace. Please move the value types to `cpp/fory/type`, expose the canonical API as `fory::Duration` / `fory::Timestamp` / `fory::Date`, and leave aliases in `fory::serialization` if needed for existing call sites. `temporal_serializers.h` should only own the `Serializer` specializations and chrono adapters.

docs/guide/cpp/supported-types.md
```diff
@@ -189,29 +189,44 @@ OptionalInt value = 42;
 
 ## Temporal Types
 
+`Duration`, `Timestamp`, and `Date` are Fory-owned carrier types defined in
```
chaokunyang: Please update the canonical type-mapping docs as part of this public carrier change. The C++ guide now describes Fory-owned carriers, but `docs/specification/xlang_type_mapping.md` still maps C++ timestamp to `std::chrono::nanoseconds`. The docs should consistently say that FDL/codegen and dynamic/Any use the Fory carrier by default, while chrono is an explicit C++ adaptation target.

cpp/fory/serialization/temporal_serializers.h
chaokunyang: Create a  cpp/fory/type/temporal.h, put duraton/timestamp into it. no cpp/fory/type/duration.h.

I think this is better

BaldDemian:
Hi @chaokunyang, I have addressed all above issues, PTAL. Thanks for your review!

chaokunyang (approved):

PR #3800, opened 2026-06-29
feat(format): support custom codecs keyed on Optional

stevenschlansker:
A row-format field typed Optional<X> is normally unwrapped to a nullable X with emptiness mapped to the column null bit, so a codec keyed on X only sees the present value and cannot control how absence is encoded. Allow a codec keyed on Optional itself to own the whole field: it receives both present and empty values and encodes optionality in-band (e.g. a sentinel) into a single non-nullable column, matching external wire formats an element codec cannot reproduce.

TypeInference.inferField and RowEncoderBuilder.addDecoderMethods now resolve a codec on the raw field type before the Optional unwrap, matching the serialize/deserialize paths. The reorder is behavior-preserving: without an Optional-keyed codec the lookup is null and the existing unwrap runs unchanged, so element codecs inside an Optional keep working.

Because an Optional-keyed codec declares a non-nullable column, serializeFor normalizes a null field reference to Optional.empty() rather than setNullAt (which the compact row writer rejects). The decode side returns a non-null Optional; this contract is documented on CustomCodec.decode and the Encoders javadoc.

Collection and map element codecs were keyed on BinaryArray.class, so a bean-scoped codec applied to a bean's fields but not its collection or map elements, diverging schema from data. Resolve elements against the enclosing type the schema uses: a bean field encoder keys on its own bean, a standalone array/map encoder keys on Object to match inferField. The new codecEnclosingType keeps codegen and schema in agreement.

Document the CustomCodec.getForyField width contract: a fixed-width declaration must equal the exact byte width encode() produces, since the compact layout sizes the slot from it. This cannot be statically validated, so the compact writers assert it on every fixed-slot write via a checkFixedWidth hook (a no-op for the uniform-slot default layout). assert keeps the cost at zero in production and catches a misconfigured codec instead of corrupting the row.

Tests cover present/empty/null-reference round-trips across the eager, compact, and lazy paths; Optional-keyed codecs in List, Set, nested List, and Map key/value; scope isolation between bean fields and top-level collections; and the width-mismatch assert on the row and array paths.

## Why?

Bug exploration on future feature request PR yielded an existing path that is not well supported, therefore
peel off feature / bugfix aspect as a separate PR

## What does this PR do?

Correct inconsistent handling of Optional type in row format - only mattered in the case of custom codecs.
Includes tests showing an actual bug was fixed.

## Related issues

N/A

## AI Contribution Checklist

- [x] Substantial AI assistance was used in this PR
- [x] AI Usage Disclosure
- scope: all
- affected_files_or_subsystems: row-format
- ai_review: included below
- human_verification: :heavy_check_mark: 

```
  Both reviewers independently reached approve with no blockers or majors, and that holds under scrutiny. The change
  correctly adds Optional-keyed custom codecs and fixes a real bug along the way: switching the codec-enclosing key from
  the internal BinaryArray.class sentinel to the enclosing bean, so bean-scoped codecs now reach a bean's collection and
  map elements instead of falling through to Object. The "raw-type codec wins before Optional unwrap" invariant is applied
  consistently across all four resolution sites with symmetric encode/decode, every @Internal signature change is
  non-breaking, and the test coverage is thorough — present/empty/null-ref cases, interface decode, collection-element
  scope, mixed-nullability asserted against BinaryRow.isNullAt wire state rather than the decoded value, and three
  width-mismatch traps gated for assertions-off runs. The only open question either review raised — whether the new
  checkFixedWidth hook on the primitive write paths costs throughput — is now settled by measurement: it is statistically
  zero-cost in production. The branch is ready to merge.

  Signed off by: Claude Opus 4.8 (model ID claude-opus-4-8, 1M-token context), Anthropic's Claude Code CLI.
```

Minor change to row-format custom codec handling around Optional types. The behavior before was broken, so this should be a strict improvement.

## Benchmark

```
  A focused JMH harness in the existing benchmarks/java module, built against this branch's fory-format, run on JDK 26
  with assertions off (the production and default config). I isolated the hook by pairing the real hooked
  CompactBinaryRowWriter.write(int, long) against an identical hook-free store on the same buffer and offset, with both
  the eager and compact array writers loaded so the checkFixedWidth call site is bimorphic — the exact dispatch the review
  worried about. 3 forks × 6 iterations, 18 measurements per benchmark:

  - hook-free store: 10.954 ± 0.112 ns/op (per 256-row batch)
  - real hooked write: 11.312 ± 0.468 ns/op

  The delta is ~0.36 ns across a 256-row batch and the confidence intervals overlap, so with assertions off the hook is
  statistically indistinguishable from zero cost. A negative control with -ea on showed a small but real gap (11.690 vs
  11.087), which confirms the benchmark genuinely exercises the hook rather than measuring a no-op — the cost simply
  vanishes once $assertionsDisabled folds the guard and dead-code elimination drops the now-unused array load, exactly as
  the bytecode predicted.
```

### DIFF
--- a/java/fory-format/src/main/java/org/apache/fory/format/encoder/ArrayDataForEach.java
+++ b/java/fory-format/src/main/java/org/apache/fory/format/encoder/ArrayDataForEach.java
@@ -61,9 +61,10 @@ public class ArrayDataForEach extends AbstractExpression {
    */
   public ArrayDataForEach(
       Expression inputArrayData,
+      Class<?> enclosingType,
       TypeRef<?> elemType,
       SerializableBiFunction<Expression, Expression, Expression> notNullAction) {
-    this(inputArrayData, elemType, notNullAction, null);
+    this(inputArrayData, enclosingType, elemType, notNullAction, null);
   }
 
   /**
@@ -72,6 +73,7 @@ public class ArrayDataForEach extends AbstractExpression {
    */
   public ArrayDataForEach(
       Expression inputArrayData,
+      Class<?> enclosingType,
       TypeRef<?> elemType,
       SerializableBiFunction<Expression, Expression, Expression> notNullAction,
       SerializableFunction<Expression, Expression> nullAction) {
@@ -79,8 +81,12 @@ public class ArrayDataForEach extends AbstractExpression {
     Preconditions.checkArgument(getRawType(inputArrayData.type()) == BinaryArray.class);
     this.inputArrayData = inputArrayData;
     CustomTypeHandler customTypeHandler = CustomTypeEncoderRegistry.customTypeHandler();
+    // Resolve the element codec scoped to the same enclosing type the schema used: the bean for a
+    // bean's collection field, Object for a top-level collection, falling back to an Object-scoped
+    // registration. A bean-scoped codec applies to that bean's collection elements, not only its
+    // direct fields, but does not bind to a top-level collection the schema never applied it to.
     CustomCodec<?, ?> customEncoder =
-        customTypeHandler.findCodec(BinaryArray.class, elemType.getRawType());
+        customTypeHandler.findCodec(enclosingType, elemType.getRawType());
     TypeRef<?> accessType;
     if (customEncoder == null) {
       accessType = elemType;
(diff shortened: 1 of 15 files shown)


chaokunyang (approved):
LGTM

PR #3804, opened 2026-06-30
fix(compiler): reject any, message and union as map key types

BaldDemian:
## Why?

As discussed in https://github.com/apache/fory/pull/3789#issuecomment-4798805385, we should reject `any` `message` and `union` as IDL `map` key types. But the validator misses them in the check.

## What does this PR do?

- Reject `any`, `message` and `union` as map key types
- Modify the test and doc accordingly

## Related issues

N/A.

## AI Contribution Checklist

N/A.

## Benchmark

N/A.


chaokunyang (approved):
LGTM

PR #3809, opened 2026-07-02
perf(java): avoid quadratic buffer growth in stream deserialization

temni:
## Why?

Since v1.2.0 (#3748), deserializing from a `ForyInputStream` (or a seekable `ForyReadableChannel`) is **O(n²)** in the payload size. For multi-MB payloads of many small objects this looks like an infinite hang: the deserializing thread spins at 100% CPU for hours inside buffer reallocation.

Thread dump of a real workload (3.7 MB payload, Fory 1.2.0 — identical stack for minutes, ~95% CPU):

```
"main" ... RUNNABLE
    at org.apache.fory.io.ForyInputStream.growBuffer(ForyInputStream.java:108)
    at org.apache.fory.io.ForyInputStream.fillBuffer(ForyInputStream.java:85)
    at org.apache.fory.memory.MemoryBuffer.readByte(MemoryBuffer.java:2086)
    ...
```

Root cause: `fillBuffer` grows the internal buffer to the **exact** target size:

- When `stream.available() >= remainingNeeded`, it sets `newSize = targetSize` — for a `readByte()`-triggered `fillBuffer(1)` on a full buffer that is `currentCapacity + 1`.
- The doubling fallback `nextBufferSize` is also capped with `Math.min(grown, targetSize)`, so it degenerates to the same exact fit for small fills.

Because the stream buffer accumulates the whole payload during one `deserialize()` call, the buffer is always exactly full after a fill, so **every small read reallocates and copies the entire buffer** (`readByte` → grow by 1 byte → copy N bytes). Total work is Σ ≈ n²/2 byte copies plus an n-byte allocation per few bytes read. Payloads ≤ 1.1.0 were unaffected (growth was 4x/1.5x with greedy reads).

## What does this PR do?

Makes stream buffer growth geometric again while keeping the allocation-bounding intent of #3748:

- `ForyInputStream.fillBuffer` / `ForyReadableChannel.fillBuffer`: on the verified path (stream/channel proved it has the bytes), grow to `max(targetSize, 2 * capacity)` instead of exactly `targetSize`.
- `nextBufferSize`: drop the `Math.min(grown, targetSize)` cap so the unverified fallback doubles instead of creeping by the requested amount. Allocation on this path is still bounded by ~2× the bytes actually received, so truncated/hostile streams still fail before large allocations — all assertions of `testStreamFillGrowsFromBufferedBytes` (added in #3748) are unchanged and still pass.
- The growth policy now lives in one place: `ForyStreamReader` gains a `MAX_BUFFER_SIZE` constant (`Integer.MAX_VALUE - 8`, the largest array size commonly supported by JVMs) and a static `nextBufferSize(int)` helper, used by both stream readers. This also aligns the two readers' previously inconsistent max-size clamps.
- Updates `docs/security/deserialization.md` (stream-backed fill buffer growth bullet), which previously prescribed the cap-to-target behavior; it now requires geometric growth and explains why cap-to-target is O(n²).

Adds `StreamTest#testStreamBufferGrowthIsGeometric`, which reads 64 KB byte-by-byte through both `ForyInputStream` and `ForyReadableChannel` and asserts the backing buffer is reallocated ≤ 20 times (geometric ⇒ ~10). Without the fix the test fails within milliseconds:

```
StreamTest.testStreamBufferGrowthIsGeometric:389 stream buffer must grow
geometrically, but already grew 21 times expected [true] but found [false]
```

With the fix, all 14 `StreamTest` cases pass, including `testStreamFillGrowsFromBufferedBytes` from #3748.

## Related issues

Regression introduced by #3748 (first released in v1.2.0).

## AI Contribution Checklist

- [x] I included a completed AI Contribution Checklist and the required `AI Usage Disclosure` below.
- [x] Two fresh AI review agents ran on the current HEAD: one Fory-guided reviewer using `AGENTS.md` and `.agents/ci-and-pr.md`, and one independent general reviewer in a separate clean-context session not pointed at Fory-specific checklists.
- [x] All AI review comments were addressed and the review loop repeated until both reviewers reported no further actionable comments.
- [x] Review results from both reviewers are persisted in the collapsible sections below.
- [x] Human verification was run and recorded (see `human_verification` below).
- [x] Tests added (regression test that fails on the unfixed code).
- [x] Performance impact validated with a benchmark (see Benchmark section).
- [x] Licensing and provenance compliance verified.
- [x] Line-by-line human self-review — the human contributor will confirm this on the PR after review.

```text
AI Usage Disclosure
- scope: design drafting | code drafting | tests | docs
- affected_files_or_subsystems: java/fory-core stream readers (org.apache.fory.io.ForyInputStream, org.apache.fory.io.ForyReadableChannel), StreamTest, docs/security/deserialization.md
- ai_review: line-by-line AI self-review completed. Round 1: Fory-guided reviewer reported 1 MAJOR (security doc still prescribed cap-to-target growth) + 2 MINOR (perf title/benchmark, direct-ByteBuffer test gap); independent reviewer approved with 4 MINOR (max-size clamp inconsistency, direct-ByteBuffer test gap, 2 informational). All actionable findings fixed. Round 2: both reviewers re-reviewed the updated diff and reported no further actionable comments.
- ai_review_artifacts: full final review reports from both reviewers embedded in the collapsible sections of this PR description.
- human_verification: StreamTest run locally on JDK 21 (14/14 pass, includes existing #3748 assertions); the new regression test verified to FAIL on unfixed code ("grew 21 times" assertion) and PASS with the fix; end-to-end benchmark on a 4.4 MB payload (see Benchmark).
- performance_verification: benchmark table in this PR description (unfixed: did not finish in 60 s; fixed: 30.7 ms).
- provenance_license_confirmation: Apache-2.0-compatible provenance confirmed; no incompatible third-party code introduced.
```

<details>
<summary>Reviewer 1 (Fory-guided, AGENTS.md + .agents/ci-and-pr.md) — final result</summary>

**Round 1 findings:** 1 MAJOR — `docs/security/deserialization.md` still prescribed the cap-to-target growth this PR removes (doc/code drift; risk that a future change "restores" the cap and reintroduces the O(n²) regression). 2 MINOR — use `perf(java):` Conventional-Commits type with benchmark data; direct-ByteBuffer channel growth path not exercised by the test. Verified as non-issues: security invariants preserved (fallback only doubles from a full buffer of real bytes, never from attacker-declared sizes; verified path allocates targetSize only after available()/channel size proves the bytes); #3748's pinned test assertions still hold; no integer overflow; test deterministic.

**Round 2 (after fixes):** "No further actionable comments." Verified all four fixes on disk: doc now requires geometric growth and explains the cap-to-target O(n²) degeneration, matching the code; perf title + benchmark present; direct-buffer detection confirmed sound (`growBuffer` → `initByteBuffer` → `initOffHeapBuffer` reassigns the `offHeapBuffer` field each grow, and `getHeapMemory()` is null for direct buffers so the helper falls back correctly); channel clamp aligned to `Integer.MAX_VALUE - 8`.

**Verdict: Approve** — "fix is correct, security-safe, and the whole surface (code, tests, docs, PR metadata) now agrees."

</details>

<details>
<summary>Reviewer 2 (independent, clean context) — final result</summary>

**Round 1 findings (all MINOR):** (1) `nextBufferSize` max-size clamp inconsistent between the two readers (`Integer.MAX_VALUE` vs `Integer.MAX_VALUE - 8`); (2) direct-ByteBuffer channel growth path not covered by the test; (3) informational: verified path may allocate up to ~2× target (standard geometric-growth trade-off, intended); (4) nit: temp-file creation outside try (consistent with surrounding tests). Verified as correct: no integer overflow, no false `growBuffer` guard trips, fix targets the real O(n²) path, test deterministic and genuinely pins the behavior.

**Round 2 (after fixes):** Findings 2–4 fully addressed; verified `getOffHeapBuffer()` returns the stored field (not a duplicate) and each grow swaps it, so the direct-buffer identity check genuinely detects reallocation. One residual optional one-liner: align the channel's `targetSize` guard to `Integer.MAX_VALUE - 8L` — **applied in the final HEAD**.

**Verdict: Approve** — "The fix is correct and well-tested across heap, heap-channel, and direct-channel paths."

</details>

- [x] Does this PR introduce any public API change? — Additive only: `ForyStreamReader` gains a `MAX_BUFFER_SIZE` constant and a static `nextBufferSize(int)` helper (the shared growth policy); no existing signature changed.
- [ ] Does this PR introduce any binary protocol compatibility change? — No.

## Benchmark

Deserializing a 4.4 MB payload (a `HashMap<String, Long>` with 300k entries, i.e. many small reads) built with `Fory.builder().requireClassRegistration(false).withIntCompressed(true).withLongCompressed(true).withRefTracking(true)`, serialized via `fory.serialize(OutputStream, obj)`:

| path | Fory 1.2.0 (unfixed) | this PR |
|---|---|---|
| `deserialize(new ForyInputStream(in))` | **did not finish within 60 s** (100% CPU, killed) | **30.7 ms** |
| `deserialize(byte[])` (reference) | 25.6 ms | 24.2 ms |

### DIFF
--- a/docs/security/deserialization.md
+++ b/docs/security/deserialization.md
@@ -164,13 +164,18 @@ For stream-backed input:
 - A stream-backed buffer may hold the full requested encoded body after that
   body has been read from the stream. It must not reserve the attacker-declared
   length before input bytes prove that length exists.
-- Stream-backed fill buffers should grow from the current proven buffer size,
-  such as by doubling current capacity, and cap only to the immediate target
-  when the next bounded growth step reaches it. A byte owner may use an
-  owner-local availability signal as a one-shot growth hint when the stream
-  implementation itself is caller-owned trusted code; if that hint is absent or
-  insufficient, the reader must fall back to bounded growth from already
-  buffered bytes. Serializers should not add their own availability branches.
+- Stream-backed fill buffers should grow geometrically from the current proven
+  buffer size, such as by doubling current capacity. Growth must not be capped
+  to the immediate fill target: for small fills the target is barely above the
+  current capacity, so cap-to-target degenerates into constant-size growth
+  steps that copy the whole buffer on every small read and make stream
+  deserialization O(n^2) overall. A byte owner may use an owner-local
+  availability signal as a one-shot growth hint when the stream implementation
+  itself is caller-owned trusted code, and may then reserve the full immediate
+  target at once while keeping at least the geometric growth step; if that hint
+  is absent or insufficient, the reader must fall back to bounded geometric
+  growth from already buffered bytes. Serializers should not add their own
+  availability branches.
 - A truncated stream should fail before allocating the final deserialized value
   and should allocate only for bytes actually read plus bounded spare capacity.
 
--- a/java/fory-core/src/main/java/org/apache/fory/io/ForyInputStream.java
+++ b/java/fory-core/src/main/java/org/apache/fory/io/ForyInputStream.java
@@ -64,7 +64,7 @@ public class ForyInputStream extends InputStream implements ForyStreamReader {
         int offset = buffer.size();
         int remainingNeeded = minFillSize - totalRead;
         long targetSize = (long) offset + remainingNeeded;
-        if (targetSize > Integer.MAX_VALUE - 8L) {
+        if (targetSize > MAX_BUFFER_SIZE) {
           throw new IndexOutOfBoundsException("Stream buffer size exceeds supported range");
         }
         if (targetSize > heapMemory.length) {
@@ -72,14 +72,17 @@ public class ForyInputStream extends InputStream implements ForyStreamReader {
           if (!checkedAvailable) {
             checkedAvailable = true;
             // Use available() only as a one-shot growth hint. It may be expensive or
-            // conservative, so failed hints fall back to bounded doubling. Final value
-            // allocation still waits for fillBuffer to complete successfully.
+            // conservative, so failed hints fall back to doubling. Grow by at least a
+            // doubling step so that repeated small fills stay amortized O(1); growing to
+            // the exact target size would copy the whole buffer on every small read,
+            // making stream deserialization O(n^2) overall.
             if (stream.available() >= remainingNeeded) {
-              newSize = (int) targetSize;
+              newSize =
+                  (int) Math.max(targetSize, ForyStreamReader.nextBufferSize(heapMemory.length));
             }
           }
           if (newSize == 0 && offset == heapMemory.length) {
-            newSize = nextBufferSize(heapMemory.length, (int) targetSize);
+            newSize = ForyStreamReader.nextBufferSize(heapMemory.length);
           }
           if (newSize != 0) {
             heapMemory = growBuffer(buffer, newSize);
@@ -111,15 +114,6 @@ public class ForyInputStream extends InputStream implements ForyStreamReader {
     buffer.initHeapBuffer(newBuffer, 0, size);
     heapMemory = newBuffer;
     return heapMemory;
-  }
-
-  private static int nextBufferSize(int size, int targetSize) {
-    long grown = size == 0 ? 1L : (long) size << 1;
-    int maxSize = Integer.MAX_VALUE - 8;
-    if (grown > maxSize) {
-      grown = maxSize;
-    }
-    return (int) Math.min(grown, targetSize);
   }
 
   @Override
--- a/java/fory-core/src/main/java/org/apache/fory/io/ForyReadableChannel.java
+++ b/java/fory-core/src/main/java/org/apache/fory/io/ForyReadableChannel.java
@@ -98,7 +98,7 @@ public class ForyReadableChannel implements ForyStreamReader, ReadableByteChanne
         int position = byteBuf.position();
         int remainingNeeded = minFillSize - totalRead;
         long targetSize = (long) position + remainingNeeded;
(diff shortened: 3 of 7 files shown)


temni:
The `JavaScript Xlang Test` failure is unrelated to this PR (the diff only touches `java/fory-core` and one docs page; all Java jobs are green).

The job fails during **Install JavaScript dependencies**: `@types/node@26.1.0` (published 2026-07-01) gets hoisted into the root `node_modules`, and its new `ffi.d.ts` uses syntax that TypeScript `^4.8.4` (the compiler version used in `javascript/`) cannot parse:

```
##[error]../../node_modules/@types/node/ffi.d.ts(94,21): error TS1139: Type parameter declaration expected.
##[error]../../node_modules/@types/node/ffi.d.ts(94,29): error TS1005: ',' expected.
```

Any run of this job started after 2026-07-01 should hit the same failure until `@types/node` is pinned (the direct dep is `^18.7.21`, but a transitive `@types/node: *` wins the hoist) or TypeScript is upgraded in the `javascript` workspace.

chaokunyang (approved):
LGTM, thanks for this fix

chaokunyang:
fory 1.4.0 will be released in next week.

chuckbenger:
> fory 1.4.0 will be released in next week.

Looking forward to the release! We just encountered this on our flink jobs which use Fory for some pretty massive state (TB's). We reverted to 1.1 for now, but will move to 1.4 next!

PR #3810, opened 2026-07-02
fix(compiler): never generate C++ equality methods for message and union containing any

BaldDemian:
## Why?

Fory IDL `any` maps to `std::any` in generated C++ code.
`std::any` does not define `operator==`, so generated equality code MUST NOT ask the C++ standard library to compare two `std::any` directly.

But the current C++ compiler still does this in some generated `operator==` implementations, especially when `any` appeared inside standard-comparable containers or union alternatives.

For example, generated code such as `values_ == other.values_` for `std::vector<std::any>`, 
`by_name_ == other.by_name_` for `std::unordered_map<std::string, std::any>`, or
`value_ == other.value_` for `std::variant<std::any, ...>` causes the standard
container or variant equality operator to instantiate `std::any == std::any`.

This causes compilation error like:

```text
error: no match for 'operator==' (operand types are 'const std::any' and 'const std::any')
```

The Rust compiler avoids the equivalent derived-equality problem by not deriving comparison traits for generated types that contain `any`:

https://github.com/apache/fory/blob/9b2bcec9618702d6aa5ba4b166f86619fce51baf/compiler/fory_compiler/generators/rust.py#L1149-L1158

## What does this PR do?

- Follow the Rust handling, i.e. don't generate C++ equality methods for message and union containing `any`.
- Add one testcase.

## Related issues

N/A.

## AI Contribution Checklist

N/A.

## Benchmark

N/A.

### DIFF
--- a/compiler/fory_compiler/generators/cpp.py
+++ b/compiler/fory_compiler/generators/cpp.py
@@ -691,13 +691,6 @@ class CppGenerator(BaseGenerator):
     ) -> str:
         member_name = self.get_field_member_name(field)
         other_member = f"other.{member_name}"
-        if isinstance(field.field_type, PrimitiveType) and (
-            field.field_type.kind == PrimitiveKind.ANY
-        ):
-            return (
-                f"((!{member_name}.has_value() && !{other_member}.has_value()) || "
-                f"({member_name}.type() == {other_member}.type()))"
-            )
         if self.is_message_type(
             field.field_type, parent_stack
         ) and self.get_field_weak_ref(field):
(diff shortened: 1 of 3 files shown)


chaokunyang:
whether we should not generate equals method when message have any field for c++?

BaldDemian:
> whether we should not generate equals method when message have any field for c++?

This works too, I will update this PR accordingly.

BaldDemian:
Addressing the failed test

chaokunyang (approved):

PR #3861, opened 2026-07-16
feat(java): capture and forward extra fields in compatible mode

Tatenda-k:
Adds ForyExtraFields sink, an opt-in mechanism for preserving unmatched remote fields during compatible-mode deserialization and replaying them on re-serialization.
Covers both the  interpreter and generated codepaths,

## Why?
When the upstream and downstream attributes are inconsistent, for example, the upstream is 10 fields, the downstream is not synchronized to update the field attributes, and there are only 9 attributes, then the downstream non-existent attributes will be ignored during deserialization.

## What does this PR do?

This PR adds support for preserving unknown fields during compatible serialization using `ForyExtraFields`.

A class opts in by declaring a `ForyExtraFields` sink. Fields present in the remote schema but not in the local schema are captured in the sink.

Captured fields are replayed using the original remote `TypeDef`, allowing them to be preserved across compatible serialization round trips involving different schemas.

Replay serializers are cached by `(class, TypeDef)`, allowing a local class to replay `ForyExtraFields` captured from multiple remote schemas.

Tests cover both enabled and disabled reference tracking, as well as both `codegen=true` and `codegen=false`.
## AI Contribution Checklist

- [x] Substantial AI assistance was used in this PR: yes
- [x] If `yes`, I included a completed [AI Contribution Checklist](https://github.com/apache/fory/blob/main/AI_POLICY.md#9-contributor-checklist-for-ai-assisted-prs) in this PR description and the required `AI Usage Disclosure`.
- [x] If `yes`, my PR description includes the required `ai_review` summary and screenshot evidence of the final clean AI review results from both fresh reviewers on the current PR diff or current HEAD after the latest code changes.

## AI Contribution Checklist

- [x] Substantial AI assistance was used in this PR: yes
- [x] If yes, I included the standardized AI Usage Disclosure block below.
- [x] If yes, I can explain and defend all important changes without AI help.
- [x] If yes, I reviewed AI-assisted code changes line by line before submission.
- [x] If yes, I completed line-by-line self-review first and fixed issues before requesting AI review.
- [x] If yes, I ran two fresh AI review agents on the current PR diff or current HEAD after the latest code changes: one Fory-guided reviewer using AGENTS.md and .agents/ci-and-pr.md, and one independent general reviewer in a separate clean-context review session that was not pointed to .agents/ci-and-pr.md or any copied Fory-specific review checklist. If the independent reviewer's tooling auto-loaded AGENTS.md, it followed the independent-review carve-out there.
- [x] If yes, I addressed all AI review comments and repeated the review loop until both ai reviewers reported no further actionable comments.
- [x] If yes, I attached screenshot evidence or equivalent persisted links of the final clean AI review results from both fresh reviewers on the current PR diff or current HEAD after the latest code changes in this PR body.
- [x] If yes, I ran adequate human verification and recorded evidence (checks run locally or in CI, pass/fail summary, and confirmation I reviewed results).
- [x] If yes, I added/updated tests and specs where required.
- [x] If yes, I validated protocol/performance impacts with evidence when applicable.
- [x] If yes, I verified licensing and provenance compliance.

AI Usage Disclosure
- scope: docs, code drafting, tests, design drafting
- affected_files_or_subsystems: 
  - docs/guide/java/extra-fields.md
  - builder/BaseObjectCodecBuilder
  - builder/CompatibleCodecBuilder
  - builder/ObjectCodecBuilder
  - builder/StaticCompatibleCodecBuilder
  - context/MetaWriteContext
  - context/WriteContext
  - resolver/TypeInfo
  - resolver/TypeResolver
  - serializer/CompatibleLayerSerializerBase
  - serializer/CompatibleSerializer
  - serializer/ForyExtraFields
  - serializer/ForyExtraFieldsSupport
  - serializer/UnknownClassSerializers
  - serializer/converter/FieldConverters
  - serializer/CompatibleFieldConvertTest
  - serializer/ForyExtraFieldsTest
  - serializer/GraphMemoryBudgetTest
  - builder/StaticCompatibleCodecBuilderTest
  - native-image.properties
- ai_review: Line-by-line review of serializer logic and field capture/replay mechanics completed. Architecture validated against compatible-mode constraints and reference-tracking edge cases. Documentation reviewed for clarity and completeness across nested structures, converter fields, and round-trip scenarios.
- ai_review_artifacts:https://claude.ai/code/session_01JAyGgABzsVxarBkJ6zq9Tk , https://claude.ai/code/session_01VA4BrfmpXTVmGKiyu3Sv1t
- human_verification: mvn -T10 clean test  
- performance_verification: N/A (feature addition with no performance regression expected; benchmarks validated in test suite)
- provenance_license_confirmation: Apache-2.0-compatible provenance confirmed; no incompatible third-party code introduced

### DIFF
--- a/java/fory-core/src/main/java/org/apache/fory/builder/CompatibleCodecBuilder.java
+++ b/java/fory-core/src/main/java/org/apache/fory/builder/CompatibleCodecBuilder.java
@@ -39,20 +39,25 @@ import org.apache.fory.Fory;
 import org.apache.fory.builder.Generated.GeneratedCompatibleSerializer;
 import org.apache.fory.codegen.CodeGenerator;
 import org.apache.fory.codegen.Expression;
+import org.apache.fory.codegen.Expression.Invoke;
 import org.apache.fory.codegen.Expression.Literal;
 import org.apache.fory.codegen.Expression.StaticInvoke;
 import org.apache.fory.config.ForyBuilder;
 import org.apache.fory.context.ReadContext;
+import org.apache.fory.context.WriteContext;
 import org.apache.fory.logging.Logger;
 import org.apache.fory.logging.LoggerFactory;
 import org.apache.fory.memory.MemoryBuffer;
 import org.apache.fory.meta.TypeDef;
 import org.apache.fory.platform.GraalvmSupport;
+import org.apache.fory.reflect.FieldAccessor;
+import org.apache.fory.reflect.ReflectionUtils;
 import org.apache.fory.reflect.TypeRef;
 import org.apache.fory.resolver.TypeResolver;
 import org.apache.fory.serializer.CodegenSerializer;
 import org.apache.fory.serializer.CompatibleSerializer;
 import org.apache.fory.serializer.FieldGroups.SerializationFieldInfo;
+import org.apache.fory.serializer.ForyExtraFields;
 import org.apache.fory.serializer.ObjectSerializer;
 import org.apache.fory.serializer.Serializer;
 import org.apache.fory.serializer.Serializers;
@@ -93,13 +98,20 @@ public class CompatibleCodecBuilder extends ObjectCodecBuilder {
   private static final String REMOTE_FIELD_INFOS_NAME = "_f_remoteFieldInfos";
   private static final String LOCAL_FIELD_INFOS_NAME = "_f_localFieldInfosByRemoteOrder";
   private static final String CONSTRUCTOR_TYPE_DEF_NAME = "_f_typeDef";
+  private static final String EXTRA_FIELDS_SINK_ACCESSOR_NAME = "_f_extraFieldsSinkAccessor";
+  private static final String EXTRA_FIELDS_LOCAL_NAME = "_f_extra";
 
   private final TypeDef typeDef;
   private final String defaultValueLanguage;
   private final DefaultValueUtils.DefaultValueField[] defaultValueFields;
   private final Map<Descriptor, Integer> fieldInfoIds = new IdentityHashMap<>();
+  private final Field extraFieldsSinkField;
   private String remoteFieldInfosName;
   private String localFieldInfosName;
+  private String extraFieldsSinkAccessorName;
+  private String storedTypeDefFieldName;
+  private Expression extraFieldsSinkExprBean;
+  private Expression extraFieldsSinkExpr;
 
   public CompatibleCodecBuilder(TypeRef<?> beanType, Fory fory, TypeDef typeDef) {
     super(beanType, fory, GeneratedCompatibleSerializer.class);
@@ -153,6 +165,7 @@ public class CompatibleCodecBuilder extends ObjectCodecBuilder {
     }
     this.defaultValueLanguage = defaultValueLanguage;
     this.defaultValueFields = defaultValueFields;
+    this.extraFieldsSinkField = ForyExtraFields.findSinkField(beanClass);
   }
 
   // Must be static to be shared across the whole process life.
@@ -225,6 +238,16 @@ public class CompatibleCodecBuilder extends ObjectCodecBuilder {
             decodeCode);
     ctx.overrideMethod(
         readMethodName, decodeCode, Object.class, ReadContext.class, READ_CONTEXT_NAME);
+    if (extraFieldsSinkField != null) {
+      ctx.overrideMethod(
+          writeMethodName,
+          buildExtraFieldsWriteCode(),
+          void.class,
+          WriteContext.class,
+          WRITE_CONTEXT_NAME,
+          Object.class,
+          ROOT_OBJECT_NAME);
+    }
     registerJITNotifyCallback();
     ctx.addConstructor(
         constructorCode,
@@ -263,9 +286,62 @@ public class CompatibleCodecBuilder extends ObjectCodecBuilder {
     return Serializers.newSerializer(typeResolver, cls, serializerClass);
   }
 
-  @Override
-  public Expression buildEncodeExpression() {
-    throw new IllegalStateException("unreachable");
+  /**
+   * Generates the body of the {@code write} method for classes that declare a {@link
+   * ForyExtraFields} sink. Mirrors {@link CompatibleSerializer#write}.
+   */
+  private String buildExtraFieldsWriteCode() {
+    extraFieldsSinkAccessorExpr();
+    ctx.clearExprState();
+    String encodeCode = buildEncodeExpression().genCode(ctx).code();
+    encodeCode = ctx.optimizeMethodCode(encodeCode);
+    encodeCode = encodeCode == null ? "" : encodeCode;
+    StringBuilder body = new StringBuilder();
+    appendBufferAndRefWriterDeclarations(body, encodeCode);
+    body.append(
+        StringUtils.format(
+            "${extraType} ${extra} = (${extraType}) ${accessor}.getObject(${obj});\n"
+                + "if (${extra} == null || ${extra}.isEmpty()) {\n"
+                + "  this.${serializer}.write(${writeContext}, ${obj});\n"
+                + "  return;\n"
+                + "}\n",
+            "extraType",
+            ctx.type(ForyExtraFields.class),
+            "extra",
+            EXTRA_FIELDS_LOCAL_NAME,
+            "accessor",
+            extraFieldsSinkAccessorName,
+            "obj",
+            ROOT_OBJECT_NAME,
+            "serializer",
+            SERIALIZER_FIELD_NAME,
+            "writeContext",
+            WRITE_CONTEXT_NAME));
+    body.append(encodeCode);
+    return body.toString();
+  }
+
+  private void appendBufferAndRefWriterDeclarations(StringBuilder body, String encodeCode) {
+    body.append(
+        StringUtils.format(
+            "${bufferType} ${buffer} = ${writeContext}.getBuffer();\n",
+            "bufferType",
+            ctx.type(MemoryBuffer.class),
+            "buffer",
+            BUFFER_NAME,
+            "writeContext",
+            WRITE_CONTEXT_NAME));
+    if (encodeCode.contains(REF_WRITER_NAME)) {
+      body.append(
+          StringUtils.format(
+              "${refWriterType} ${refWriter} = (${refWriterType}) ${writeContext}.getRefWriter();\n",
+              "refWriterType",
+              ctx.type(concreteRefWriterType),
+              "refWriter",
+              REF_WRITER_NAME,
+              "writeContext",
+              WRITE_CONTEXT_NAME));
+    }
   }
 
   @Override
@@ -373,7 +449,8 @@ public class CompatibleCodecBuilder extends ObjectCodecBuilder {
   }
 
   private boolean shouldSkipField(Descriptor descriptor) {
-    return descriptor.getField() == null
+    return extraFieldsSinkField == null
+        && descriptor.getField() == null
         && descriptor.getFieldConverter() == null
         && !descriptor.getRawType().isPrimitive();
   }
@@ -387,19 +464,90 @@ public class CompatibleCodecBuilder extends ObjectCodecBuilder {
   }
 
   @Override
+  protected Expression getFieldValue(Expression bean, Descriptor descriptor) {
+    if (extraFieldsSinkField != null && descriptor.getField() == null) {
+      if (extraFieldsSinkExpr == null || extraFieldsSinkExprBean != bean) {
+        extraFieldsSinkExprBean = bean;
+        extraFieldsSinkExpr =
+            tryInlineCast(
+                new Invoke(extraFieldsSinkAccessorExpr(), "getObject", OBJECT_TYPE, bean),
+                TypeRef.of(ForyExtraFields.class));
+      }
+      Invoke raw =
+          new Invoke(
+              extraFieldsSinkExpr, "get", OBJECT_TYPE, Literal.ofString(descriptor.getName()));
+      return tryInlineCast(raw, descriptor.getTypeRef());
+    }
+    return super.getFieldValue(bean, descriptor);
+  }
+
+  @Override
   protected Expression setFieldValue(Expression bean, Descriptor descriptor, Expression value) {
     if (descriptor.getField() == null) {
       FieldConverter<?> converter = descriptor.getFieldConverter();
       if (converter != null) {
         return setFieldConverterTargetValue(bean, descriptor, converter, value);
       }
-      // Field doesn't exist in current class, skip set this field value.
+      // If Field doesn't exist in current class,
+      if (extraFieldsSinkField != null) {
+        return new StaticInvoke(
+            ForyExtraFields.class,
+            "capture",
+            TypeRef.of(void.class),
+            bean,
+            extraFieldsSinkAccessorExpr(),
+            typeDefFieldExpr(),
+            Literal.ofString(descriptor.getName()),
+            value);
+      }
       // Note that the field value shouldn't be an inlined value, otherwise field value read may
-      // be ignored.
-      // Add an ignored call here to make expression type to void.
+      // be ignored. Add an ignored call here to make expression type to void.
       return new StaticInvoke(ExceptionUtils.class, "ignore", value);
     }
     return super.setFieldValue(bean, descriptor, value);
+  }
+
+  private String ensureLazyFieldGenerated(
+      String cachedFieldName, String nameConstant, TypeRef<?> fieldType, Expression initializer) {
+    if (cachedFieldName == null) {
+      cachedFieldName = ctx.newName(nameConstant);
+      ctx.addField(false, true, ctx.type(fieldType), cachedFieldName, initializer);
+    }
+    return cachedFieldName;
+  }
+
+  private Expression extraFieldsSinkAccessorExpr() {
+    extraFieldsSinkAccessorName =
+        ensureLazyFieldGenerated(
+            extraFieldsSinkAccessorName,
+            EXTRA_FIELDS_SINK_ACCESSOR_NAME,
+            TypeRef.of(FieldAccessor.class),
+            new StaticInvoke(
+                FieldAccessor.class,
+                "createAccessor",
+                TypeRef.of(FieldAccessor.class),
+                getOrCreateField(
+                    true,
+                    Field.class,
+                    extraFieldsSinkField.getName() + "SinkField",
+                    () ->
+                        new StaticInvoke(
+                            ReflectionUtils.class,
+                            "getField",
+                            TypeRef.of(Field.class),
+                            staticBeanClassExpr(),
+                            Literal.ofString(extraFieldsSinkField.getName())))));
+    return new Expression.Reference(extraFieldsSinkAccessorName, TypeRef.of(FieldAccessor.class));
+  }
+
+  private Expression typeDefFieldExpr() {
+    storedTypeDefFieldName =
+        ensureLazyFieldGenerated(
+            storedTypeDefFieldName,
+            "_f_storedTypeDef",
+            TypeRef.of(TypeDef.class),
+            new Expression.Reference(CONSTRUCTOR_TYPE_DEF_NAME, TypeRef.of(TypeDef.class)));
+    return new Expression.Reference(storedTypeDefFieldName, TypeRef.of(TypeDef.class));
   }
 
   private Expression setFieldConverterTargetValue(
--- a/java/fory-core/src/main/java/org/apache/fory/builder/CompatibleCodecBuilder.java
+++ b/java/fory-core/src/main/java/org/apache/fory/builder/CompatibleCodecBuilder.java
@@ -58,6 +58,7 @@ import org.apache.fory.serializer.CodegenSerializer;
 import org.apache.fory.serializer.CompatibleSerializer;
 import org.apache.fory.serializer.FieldGroups.SerializationFieldInfo;
 import org.apache.fory.serializer.ForyExtraFields;
+import org.apache.fory.serializer.ForyExtraFieldsSupport;
 import org.apache.fory.serializer.ObjectSerializer;
 import org.apache.fory.serializer.Serializer;
 import org.apache.fory.serializer.Serializers;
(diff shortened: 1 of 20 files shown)


java/fory-core/src/main/java/org/apache/fory/context/WriteContext.java
```diff
@@ -442,6 +444,58 @@ public BufferCallback getBufferCallback() {
...
+        resolver.getExtraFieldsWriteSerializer(obj.getClass(), sinkTypeDefId);
+    if (replaySerializer == null) {
+      throw new IllegalStateException(
+          "Cannot replay captured extra fields on "
+              + obj.getClass()
+              + ": remote TypeDef "
+              + sinkTypeDefId
+              + " has not been read into this class on this Fory instance.");
+    }
+    resolver.writeTypeInfo(this, headerTypeInfo);
```
chaokunyang, line 498: Preserve the original type header when the writer class is unavailable

When the intermediary cannot load the original writer class, `headerTypeInfo.type` is `UnknownStruct` with `NONEXISTENT_META_SHARED_ID`. Calling `writeTypeInfo` here emits that placeholder and then writes a target-class replay body. The normal unknown-class path must go through `UnknownStructSerializer.write`, which rewrites the placeholder to the original compatible type id/user id and emits the remote TypeDef. Without that rewrite, the forwarded stream is undecodable in the rolling-deployment case this feature is intended to support. Please route this header through a schema-aware writer and cover it with isolated classloaders.

chaokunyang, line 498: Key replay metadata by the remote TypeDef identity

`writeSharedClassMeta` deduplicates metadata by `typeInfo.type` (`Class<?>`), but two remote versions of the same class have different TypeDef ids. If captured objects from both versions are forwarded in one root graph, the second header references the first TypeDef while its body is written by the second replay serializer, causing field corruption or reader-index drift. Replay metadata needs to be keyed by the checked TypeDef identity, with a regression test containing two versions of the same class in one graph.

java/fory-core/src/main/java/org/apache/fory/serializer/CompatibleSerializer.java
```diff
@@ -225,6 +238,71 @@ public void write(WriteContext writeContext, T value) {
...
+      fieldValue = temp; // might be null, which is valid
+    }
+
+    switch (fieldInfo.codecCategory) {
+      case BUILD_IN:
+        if (useExtraField) {
+          AbstractObjectSerializer.writeBuildInFieldValue(
+              writeContext, typeResolver, refWriter, fieldInfo, buffer, fieldValue);
+        } else {
+          AbstractObjectSerializer.writeBuildInField(
```
chaokunyang, line 285: Handle converted fields when replaying the remote schema

A compatible scalar conversion such as remote `int score` to local `long score` produces a descriptor with a converter but no `fieldAccessor`. If any other field populates the sink, this branch calls `writeBuildInField` with that null accessor. The generated path has the same bug because it treats every `descriptor.getField() == null` as an unmatched field and reads an absent sink entry. Please define the reverse replay behavior for converter fields (or preserve their remote value) and test interpreter and codegen with a converted field plus an extra field.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
```diff
@@ -0,0 +1,114 @@
...
+  private transient TypeDef typeDef;
+
+  public TypeDef getTypeDef() {
+    return typeDef;
+  }
+
+  public static void capture(
+      Object target, FieldAccessor sinkAccessor, TypeDef typeDef, String name, Object value) {
+    ForyExtraFields extraField = (ForyExtraFields) sinkAccessor.getObject(target);
+    if (extraField == null) {
```
chaokunyang: Attach the TypeDef when the sink is already initialized

This assigns `typeDef` only when the sink field was null. A common declaration such as `final ForyExtraFields extraFields = new ForyExtraFields()` captures map entries but leaves `typeDef` null, so `tryWriteExtraFieldsSchema` silently skips replay and the downstream peer loses the unknown fields. On first capture, an existing unbound sink also needs to bind the remote TypeDef, and subsequent captures should verify that the schema identity is consistent.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
```diff
@@ -0,0 +1,114 @@
...
+ * type; the serialization framework detects it, excludes it from the normal field set, and routes
+ * unmatched remote fields into it instead of discarding them.
+ *
+ * <p>The remote {@link TypeDef} is stored transiently so that on re-serialization the framework can
+ * emit the remote schema header and replay all fields in their original order, allowing a
+ * downstream peer with the full schema to recover them.
+ */
+public final class ForyExtraFields {
+
+  private final Map<String, Object> fields = new HashMap<>();
```
chaokunyang: Use the complete remote field identity as the sink key

A simple field name is not unique in a native Java TypeDef: superclass and subclass layers may legally contain fields with the same name, distinguished by declaring class and field id/name. Capturing both into this map overwrites one value, and replay writes both slots from the surviving value. Tagged fields are also decoded as `$tagN` when the writer class is unavailable, so the documented lookup by original name cannot work. Please store a stable schema field identity and expose an unambiguous lookup contract.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
```diff
@@ -0,0 +1,114 @@
...
+
+  public TypeDef getTypeDef() {
+    return typeDef;
+  }
+
+  public static void capture(
+      Object target, FieldAccessor sinkAccessor, TypeDef typeDef, String name, Object value) {
+    ForyExtraFields extraField = (ForyExtraFields) sinkAccessor.getObject(target);
+    if (extraField == null) {
+      extraField = new ForyExtraFields();
```
chaokunyang: Charge retained sink owners to the graph-memory budget

The first capture retains a new `ForyExtraFields`, `HashMap`, backing/entry storage, and potentially a boxed primitive, but neither interpreter nor generated callers reserve graph memory for these owners. A valid list containing many partial objects can therefore grow the retained result far beyond `maxGraphMemoryBytes`. The compatible serializer/codegen owner should reserve stable lower-bound costs before allocating or growing the sink, and the budget rejection path needs a regression test.

java/fory-core/src/main/java/org/apache/fory/serializer/CompatibleSerializer.java
```diff
@@ -285,7 +373,8 @@ private void readFields(ReadContext readContext, Object[] fields) {
     RefReader refReader = readContext.getRefReader();
     Generics generics = readContext.getGenerics();
     for (SerializationFieldInfo fieldInfo : allFields) {
-      fields[counter++] = readField(readContext, refReader, generics, fieldInfo, buffer, null);
+      fields[counter++] =
```
chaokunyang, line 388: Do not opt records into a sink path that cannot capture fields

A record component of type `ForyExtraFields` is detected as a sink, but the interpreter record path reads into an `Object[]` with `captureUnmatched=false`, so unknown fields are still skipped. Generated code attempts to call `capture` before the record exists, using the record collector/array rather than a record instance, and fails on the accessor receiver. Please either implement constructor-owned sink collection for records or reject record sinks explicitly in code and documentation.

java/fory-core/src/main/java/org/apache/fory/builder/CompatibleCodecBuilder.java
```diff
@@ -386,22 +463,93 @@ private Expression skipField(Descriptor descriptor) {
...
+    extraFieldsSinkAccessorName =
+        ensureLazyFieldGenerated(
+            extraFieldsSinkAccessorName,
+            EXTRA_FIELDS_SINK_ACCESSOR_NAME,
+            TypeRef.of(FieldAccessor.class),
+            new StaticInvoke(
+                FieldAccessor.class,
+                "createAccessor",
+                TypeRef.of(FieldAccessor.class),
+                getOrCreateField(
```
chaokunyang, line 572: Keep the sink field's declaring class in generated accessors

The scanner returns the exact inherited sink `Field`, but this code discards its declaring class and resolves only `beanClass + name`. If a subclass legally hides the inherited sink with a different-typed field of the same name, generated code binds the subclass field. That causes valid input to fail and can write a `ForyExtraFields` reference into the wrong field on Unsafe-based accessors. Please retain the exact `Field`/accessor or resolve it through the original declaring class.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
```diff
@@ -0,0 +1,114 @@
...
+
+  /** Returns a {@link FieldAccessor} for the {@link ForyExtraFields} sink field on {@code cls}. */
+  public static FieldAccessor findSinkAccessor(Class<?> cls) {
+    return SINK_CACHE.get(cls, () -> scanForExtraField(cls)).orElse(null);
+  }
+
+  private static Optional<FieldAccessor> scanForExtraField(Class<?> cls) {
+    for (Class<?> c = cls; c != null && c != Object.class; c = c.getSuperclass()) {
+      for (Field f : c.getDeclaredFields()) {
+        if (ForyExtraFields.class == f.getType()) {
```
chaokunyang: Ignore static fields when discovering an extra-fields sink

Discovery checks only the field type, so a `static ForyExtraFields` constant is selected even though `FieldAccessor.createAccessor` rejects static fields. Serializer initialization then fails for an otherwise valid model, including xlang or non-compatible configurations where this feature should be irrelevant. Please restrict discovery to instance fields and add a static-field regression test.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.apache.fory.collection.ClassValueCache;
+import org.apache.fory.meta.TypeDef;
+import org.apache.fory.reflect.FieldAccessor;
+
+/**
+ * Opt-in sink for compatible-mode extra fields. A class participates by declaring a field of this
+ * type; the serialization framework detects it, excludes it from the normal field set, and routes
```
chaokunyang, line 31: Actually exclude the selected sink from normal serialization metadata

The contract says the framework excludes this field, but no descriptor, TypeDef, or ObjectSerializer owner filters it. A fresh sink-bearing type therefore exposes this framework field in its local schema, and an initialized sink can be serialized as a normal nested object instead of replaying the remote schema. Please exclude the exact selected instance field in the owning descriptor path and assert that it is absent from the local TypeDef/body.

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
```diff
@@ -1305,9 +1334,17 @@ private TypeInfo getMetaSharedTypeInfo(TypeDef typeDef, Class<?> clz) {
     if (StaticGeneratedStructSerializer.class.isAssignableFrom(sc)) {
       typeInfo.setSerializer(this, newStaticGeneratedStructSerializer(sc, cls, typeDef));
     } else if (sc == CompatibleSerializer.class) {
-      typeInfo.setSerializer(this, new CompatibleSerializer(this, cls, typeDef));
+      CompatibleSerializer<?> cs = new CompatibleSerializer<>(this, cls, typeDef);
```
chaokunyang, line 1341: Implement capture and replay for the required static and layer serializers

Only the runtime `CompatibleSerializer` and generated-compatible branches register replay serializers. `StaticGeneratedStructSerializer` still skips unknown fields, and the CompatibleLayer/ObjectStream path does the same. The issue this PR closes explicitly includes annotation processor, KSP, Scala derive, StaticCompatible, and CompatibleLayer support; deferring them in the guide leaves those supported Java surfaces silently dropping data, including the recommended GraalVM/static-codegen path. Please complete those owner paths or stop closing the full issue and narrow the documented scope.

java/fory-core/src/main/java/org/apache/fory/resolver/TypeInfo.java
```diff
@@ -170,6 +174,15 @@ public void setSerializer(Serializer<?> serializer) {
   void setSerializer(TypeResolver resolver, Serializer<?> serializer) {
     this.serializer = serializer;
     needToWriteTypeDef = serializer != null && resolver.needToWriteTypeDef(serializer);
+    this.extraFieldsSinkAccessor = type == null ? null : ForyExtraFields.findSinkAccessor(type);
```
chaokunyang: Preserve the sink accessor across TypeInfo copies

`extraFieldsSinkAccessor` is initialized only in this setter, while both `copy(...)` methods keep the serializer but drop this state. The supported sequence of materializing/registering a serializer before numeric class registration therefore produces a copied TypeInfo that can capture remote fields but no longer enters replay on write. Treat the accessor as type-owned metadata and initialize or preserve it in every constructor/copy path.

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
```diff
@@ -514,6 +515,30 @@ public boolean isMapDescriptor(Descriptor descriptor) {
...
+
+  /**
+   * Records the reader-class serializer used to replay {@code cls} under the remote {@code
+   * typeDefId} it was captured from.
+   */
+  private void registerExtraFieldsSerializer(Class<?> cls, long typeDefId, Serializer<?> gen) {
+    extRegistry
+        .extraFieldsSerializers
+        .computeIfAbsent(cls, k -> new ConcurrentHashMap<>())
+        .putIfAbsent(typeDefId, gen);
```
chaokunyang: Replace the interpreter replay serializer after async compilation

The async path normally registers `CompatibleSerializer` first, then the compilation callback calls this method with the generated serializer. `putIfAbsent` leaves the interpreter entry permanently installed (or makes the result timing-dependent if compilation wins the race). The current JIT test checks the normal TypeInfo serializer rather than this replay cache, so it passes without exercising generated replay. Please perform a safe interpreter-to-generated handoff and assert the actual replay owner.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
```diff
@@ -0,0 +1,114 @@
...
+  Object put(String name, Object value) {
+    return fields.put(name, value);
+  }
+
+  /**
+   * GC-transparent class-keyed cache: on JVM this is backed by ClassValue so entries do not prevent
+   * the associated Class (or its ClassLoader) from being collected. On Android/GraalVM it falls
+   * back to a ConcurrentHashMap.
+   */
+  private static final ClassValueCache<Optional<FieldAccessor>> SINK_CACHE =
```
chaokunyang: Avoid a process-global strong class cache on Android

The Android fallback for this static `ClassValueCache` is an unbounded `ConcurrentHashMap<Class<?>, Object>`. Both positive accessors and negative `Optional.empty()` entries strongly retain user classes, so repeatedly creating and discarding DexClassLoaders/Fory runtimes leaks the loaders across runtimes. The accessor already has a natural per-TypeInfo owner; please remove the process-global cache or provide a genuinely weak-key fallback.

java/fory-core/src/main/java/org/apache/fory/context/WriteContext.java
```diff
@@ -459,6 +513,9 @@ public void writeRef(Object obj) {
         depth--;
         return;
       }
+      if (typeInfo.hasExtraFieldsSink() && tryWriteExtraFieldsSchema(resolver, typeInfo, obj)) {
```
chaokunyang: Keep the opt-in feature out of unrelated object-write hot paths

This adds a call and branch to every dynamic object write, including types without a sink and xlang/same-schema configurations. The generated polymorphic path adds the same work, while serializer setup scans every class. Please gate the feature during cold native-compatible setup/codegen so unrelated hot paths remain unchanged, and provide the zero-overhead benchmark or generated-code assertion required for this performance-sensitive path.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
```diff
@@ -0,0 +1,114 @@
...
+  // only for Fory serialization. If a ForyExtraFields instance is serialized
+  // with ObjectOutputStream, this field will be null after deserialization, so
+  // the captured extra fields cannot later be replayed by Fory.
+  private transient TypeDef typeDef;
+
+  public TypeDef getTypeDef() {
+    return typeDef;
+  }
+
+  public static void capture(
```
chaokunyang: Keep mutation and reflection helpers out of the public sink API

The guide says application code has read-only access, but `capture`, `findSinkField`, and `findSinkAccessor` are public, unmarked implementation APIs. Callers can use them to mutate entries and replace the TypeDef association, breaking replay invariants while also exposing `FieldAccessor` as public surface. Move the generated-code bridge to an `@Internal` support owner and keep `ForyExtraFields` limited to the stable lookup API.

java/fory-core/src/main/java/org/apache/fory/context/WriteContext.java
Tatenda-k: I've updated `tryWriteExtraFieldsSchema(...)` to call `UnknownStructSerializer.writeUnknownStructHeader(...)`  when  the header is an `UnknownStruct`, and added  `unknownStructHeaderDoesNotCorruptForwardedStream` to test the above case.

Tatenda-k: I've updated the `writeSharedClassMeta` to deduplicate by `typeDef` identity. And added `twoRemoteVersionsOfSameClassForwardedInOneGraph` test, which encompasses the above scenario.

java/fory-core/src/main/java/org/apache/fory/serializer/CompatibleSerializer.java
Tatenda-k: I've updated `compatibleRead(...)` to stash the original value into the sink so `writeFieldByCodecCategory(...)` finds an entry for the converter field. For codegen, I added `FieldConverters.readTargetAndCapture(...)` used whenever the class declares a sink to capture the raw value into the sink so the sink lookup in `getFieldValue(...)` no longer returns empty.   Added `roundTripConvertedFieldWithExtraField` test.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
Tatenda-k: Fixed - `capture()` now always initializes `typeDef`, and subsequent captures on the same sink verify the `ForyExtraFields` `typeDef` is consistent with the one it was bound under, otherwise an error is thrown. Added `roundTripRecoversFieldWithPreInitializedSink`  test.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
Tatenda-k: Fixed by keying the sink on the complete remote field identity via a new `FieldIdentity` type. The public read API is now identity-only (`get(FieldIdentity)`).  Tagged fields are addressed by their id via `FieldIdentity.ofId(...)`. Added `roundTripShadowedFieldsPreservedIndependently` test (fails on the old name key, now passes).

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
Tatenda-k: Updated `capture()` to update graph memory/ throw error  when initializing the `typeDef`,(first capture), and when updating the map. Added `testExtraFieldsSinkChargedToGraphBudget()` test.

java/fory-core/src/main/java/org/apache/fory/serializer/CompatibleSerializer.java
Tatenda-k: Record sinks are excluded from the list of descriptors via `buildFieldDescriptors` when a component has type `ForyExtraFields`  , an error is thrown when one is encountered, and I've documented this constraint in extra-fields.md. `testRecordExtraFieldsSinkRejected` confirms the expected behavior.

java/fory-core/src/main/java/org/apache/fory/builder/CompatibleCodecBuilder.java
Tatenda-k: The accessor now resolves through the sink field's own declaring class `extraFieldsSinkField.getDeclaringClass()`, and added `hiddenInheritedSinkCapturesIntoDeclaringClassField` which reproduces the above scenario.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
Tatenda-k: Updated the above method to check for and ignore static fields, and log a warning if a static field is encountered. Also and added the  staticSinkFieldIgnoredByDiscovery` test to ensure this.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
Tatenda-k: Updated `buildFieldDescriptors()` to exclude a `ForyExtraField` instance and  added `sinkFieldAbsentFromLocalTypeDef()` and `inheritedSinkFieldAbsentFromLocalTypeDef()` tests to confirm the exclusion.

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
Tatenda-k: Noted. I will update the pr description.

java/fory-core/src/main/java/org/apache/fory/resolver/TypeInfo.java
Tatenda-k: Fixed both `copy(...)` overloads to preserve the `extraFieldsSinkAccessor` on the copy. Added `copyWith`sinkAccessorSurvivesNumericIdRegistration` test  to enact the sequence you flagged. I didn't initialize the `extraFieldsSinkAccessor` in the first and second constructor because the `typeResolver` is not available there.

java/fory-core/src/main/java/org/apache/fory/resolver/TypeResolver.java
Tatenda-k: Fixed the handoff race: `put()` is used for the generated serializer, `putIfAbsent()` for the interpreter. Updated `roundTripAfterJitCompilation` to assert on the replay cache owner -`getExtraFieldsWriteSerializer` instead of the `TypeInfo` serializer.

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
Tatenda-k: Done.

java/fory-core/src/main/java/org/apache/fory/context/WriteContext.java
Tatenda-k: I've added a `config.isCompatible()` and `!config.isXlang()` check, and updated the serializer setups, and the generated and interpreter paths to use `typeInfo.hasExtraFieldsEnabled` instead of looking up the sink, the polymorphic paths checks isEnabled() and does a one pass scan. Added generated code assertions that the path is guarded by the O(1) `hasextrafieldssink`

java/fory-core/src/main/java/org/apache/fory/serializer/ForyExtraFields.java
Tatenda-k: I've moved the mutation and reflection helpers to `ForyExtraFieldsSupport` .

pavel-ptashyts (approved):

PR #3877, opened 2026-07-20
feat(compiler): support C++ gRPC code generation

### LINKED ISSUE #3276: [Compiler][gRPC][C++] Generate gRPC stubs with Fory codec
Parent issue: #3266

This issue is reserved for gsoc, do not take over it.

## Scope
Implement C++ gRPC stub generation using Fory serialization only.

## Detailed tasks
- Emit `service.h`, `service.grpc.h`, `service.grpc.cc`.
- Generate abstract service + stub classes.
- Implement Fory-based serialization hooks for request/response.

BaldDemian:
## Why?

Support C++ gRPC code generation.

### Overview

This PR adds C++ gRPC code generation support. 
All unary, server-streaming, client-streaming, and bidirectional RPC types are supported. 
They integrate with gRPC C++ synchronous APIs.

### Fory-backed codec

Generated gRPC code uses Fory as codec. Given that the codec is relatively small, it is generated directed within each `*_service_grpc.h`.
On writes, codec moves serialized bytes into a gRPC `Slice`, avoiding an additional outbound payload copy. On reads, it uses a single `ByteBuffer` slice when possible and merges fragmented buffers only when necessary.

### Error-handling

To make the logic more robust, this PR also adds gRPC-related naming collision detection and escaping.
This PR also considers special characters of C++ such as compilation guards.
Generated model and service headers use sanitized include guards; 
the shared Fory codec is protected by `FORY_GENERATED_GRPC_SERIALIZATION_TRAITS_`; 
each `grpc::SerializationTraits<T>` specialization has a stable, type-specific guard derived from a sanitized fully qualified type name and a SHA-256 prefix. This prevents duplicate specializations when several generated service headers use the same message type.

### Testing

This PR adds compiler-side pytests and C++-Java gRPC interop tests. It also adds a dedicated job in the CI.

### Documentation

This PR adds documentation for the usage and example of the C++ gRPC code generation feature.

## Related issues

Close https://github.com/apache/fory/issues/3276.

## AI Contribution Checklist

N/A.

## Benchmark

N/A.


MODULE.bazel
```diff
@@ -75,7 +75,10 @@ use_repo(
 bazel_dep(name = "cython", version = "3.1.3")
 
 # Google Test
-bazel_dep(name = "googletest", version = "1.15.2")
+bazel_dep(name = "googletest", version = "1.17.0")
+
+# gRPC C++ integration tests
+bazel_dep(name = "grpc", version = "1.75.0")
```
chaokunyang: could you use cmake for grpc tests instead, so no grpc deps here

BaldDemian: This issue is addressed, PTAL

compiler/fory_compiler/generators/services/cpp.py
```diff
@@ -0,0 +1,833 @@
...
+            "      delete bytes;",
+            "      slice = ::grpc::Slice(0);",
+            "    } else {",
+            "      slice = ::grpc::Slice(bytes->data(), bytes->size(),",
+            "                            &ForyGrpcSerializationTraits::DestroyBytes,",
+            "                            bytes);",
+            "    }",
+            "    ::grpc::ByteBuffer encoded(&slice, 1);",
+            "    buffer->Swap(&encoded);",
+            "    *own_buffer = true;",
```
chaokunyang: **[P1] Initialize `own_buffer` before fallible serialization**

`*own_buffer` must be initialized before the fallible `to_bytes()` call. gRPC's send path declares `bool own_buf`, calls `SerializationTraits::Serialize`, and evaluates `if (!own_buf)` even when the returned status is not OK. If Fory serialization fails here, gRPC reads indeterminate storage and may call `Duplicate()` on an invalid buffer. Please set `*own_buffer = true` immediately after validating the output pointers, as [gRPC's own `GenericSerialize`](https://github.com/grpc/grpc/blob/v1.75.0/include/grpcpp/impl/generic_serialize.h#L43-L54) does, and add a failing-serialization regression test. The unconditional read is visible in [gRPC's send path](https://github.com/grpc/grpc/blob/v1.75.0/include/grpcpp/impl/call_op_set.h#L380-L390).

compiler/fory_compiler/generators/services/cpp.py
```diff
@@ -0,0 +1,833 @@
...
+        #
+        # This IDL itself is valid. But in our design, both `message service`
+        # and `service Greeter` would take up the same `::demo::service` namespace.
+        # We should detect this kind of collision and throw errors to the user.
+        used_schema_names: Dict[str, str] = {}
+        for type_def in self.schema.enums + self.schema.unions + self.schema.messages:
+            if not self.is_imported_type(type_def):
+                identifier = self.get_type_identifier(type_def)
+                used_schema_names[identifier] = f"model type {type_def.name!r}"
+        self._allocate_scoped_identifier(
```
chaokunyang, line 107: **[P2] Reserve gRPC's root namespace before generating model names**

The collision preflight reserves the generated `service` namespace and its nested `grpc` namespace, but not the external root namespace `::grpc`. With no package, a valid `message grpc {}` is emitted as global `class grpc`; the generated service header then includes gRPC headers declaring `namespace grpc`, so compilation fails with a class/namespace conflict. A `package grpc; message Status {}` can likewise collide with public gRPC types. Please escape or reject `grpc` through the global C++ reserved-name mapping and add a compile-level regression test.

compiler/fory_compiler/generators/services/cpp.py
```diff
@@ -0,0 +1,833 @@
...
+                )
+            self._rpc_method_identifier_cache[service_key] = method_names
+            self._rpc_method_member_identifier_cache[service_key] = rpc_members
+            self._rpc_path_constant_identifier_cache[service_key] = path_constants
+
+        self._named_service_schema_ids.add(schema_id)
+
+    def generate_service_api_header(self, services: List[Service]) -> GeneratedFile:
+        """Generate C++ service API header file (service.h)."""
+        path = f"{self.get_header_name()}_service.h"
```
chaokunyang: **[P2] Keep service filenames outside the model-header namespace**

`get_header_name()` replaces package dots with underscores, so a model in `package demo.service` is emitted as `demo_service.h`—the same path as the new service API header for `package demo`. Recursive compilation of a valid imported `demo.service` schema therefore aborts in the run-wide output collision check. `_service_grpc.h` similarly aliases a model from `package demo.service.grpc`. Please use a filename scheme that cannot be produced by package flattening, such as the proposed dotted `service.grpc.*` layout, and add a recursive CLI regression test.

compiler/fory_compiler/generators/services/cpp.py
```diff
@@ -0,0 +1,833 @@
...
+                "",
+                f'#include "{self.get_header_name()}_service.h"',
+                "",
+                "#include <cstddef>",
+                "#include <cstdint>",
+                "#include <memory>",
+                "#include <utility>",
+                "#include <vector>",
+                "",
+                "#include <grpc/event_engine/memory_allocator.h>",
```
chaokunyang: **[P2] Do not require EventEngine just to define `SerializationTraits`**

This unconditional include makes every generated header require gRPC 1.42 or newer; `grpc/event_engine/memory_allocator.h` is absent in 1.41, while this PR documents no minimum version. The allocator-aware overload is optional even in current gRPC—the dispatcher [falls back to the legacy three-argument `Serialize`](https://github.com/grpc/grpc/blob/master/include/grpcpp/impl/serialization_traits.h#L78-L117), which is also the documented [v1.41 trait contract](https://github.com/grpc/grpc/blob/v1.41.0/include/grpcpp/impl/codegen/serialization_traits.h#L24-L60). Please remove this overload/include, or version-gate it and document and test the minimum supported gRPC version.

docs/guide/cpp/grpc-support.md
```diff
@@ -0,0 +1,277 @@
...
+find_package(Fory CONFIG REQUIRED)
+find_package(gRPC CONFIG REQUIRED)
+
+add_library(greeter_generated
+    generated/demo_greeter_service_grpc.cc
+)
+target_compile_features(greeter_generated PUBLIC cxx_std_17)
+target_include_directories(greeter_generated PUBLIC generated)
+target_link_libraries(greeter_generated PUBLIC
+    fory::serialization
```
chaokunyang, line 68: **[P2] Fix the installed-package CMake target**

This example does not currently configure after `find_package(Fory)`. `fory::serialization` is only a build-tree `ALIAS`; aliases are not exported, and the install export currently exposes the real target as `fory::fory_serialization_lib`. `ForyConfig.cmake` does not recreate the advertised alias. Please give the interface target `EXPORT_NAME serialization`, recreate the alias in the package config, or document the actual installed target, and cover the installed-package path with a smoke test. The new integration test uses `FetchContent`, where the alias does exist, so it cannot detect this problem.

compiler/fory_compiler/generators/services/cpp.py
```diff
@@ -0,0 +1,833 @@
...
+            "  }",
+            "",
+            "  static ::grpc::Status Serialize(const Message& message,",
+            "                                  ::grpc::ByteBuffer* buffer,",
+            "                                  bool* own_buffer) {",
+            "    if (buffer == nullptr || own_buffer == nullptr) {",
+            "      return ::grpc::Status(::grpc::StatusCode::INTERNAL,",
+            '                            "Missing gRPC serialization output");',
+            "    }",
+            "    auto result = message.to_bytes();",
```
chaokunyang: **[P2] Avoid the full payload copy on every outbound RPC frame**

The codec adopts the returned vector without another gRPC copy, but `to_bytes()` calls `Fory::serialize()`, which allocates a vector and `memcpy`s the complete internal buffer into it. Fory already provides `serialize_to(std::vector<uint8_t>&)` for direct serialization into the vector that can be handed to the `Slice`. Since every unary and streaming frame pays this cost and #3266 explicitly requires minimizing codec allocations, please use the direct-output path and provide allocation/throughput evidence rather than marking the benchmark section N/A.

.github/workflows/ci.yml
```diff
@@ -879,6 +879,49 @@ jobs:
           cd integration_tests/grpc_tests/java
           mvn -T16 --no-transfer-progress -Dtest=RustGrpcTest test
 
+  grpc_java_cpp_tests:
+    name: Java/C++ gRPC Tests
+    needs: changes
+    if: needs.changes.outputs.grpc_tests == 'true'
```
chaokunyang: **[P2] Run this job for C++ runtime changes**

The new job runs only when `grpc_tests` is true, but that output covers only `compiler/` and `integration_tests/grpc_tests/`. A PR touching only `cpp/` therefore skips this job even if it breaks the `ThreadSafeFory` APIs or wire behavior used by the generated companion. Please add `needs.changes.outputs.cpp_code == 'true'` to this condition, or introduce a scoped C++ gRPC change output.

compiler/README.md
```diff
@@ -10,7 +10,7 @@ The FDL compiler generates cross-language serialization code from schema definit
 - **Type ID and namespace support**: Both numeric IDs and name-based type registration
 - **Field modifiers**: Optional fields, reference tracking, list fields, scalar encoding modifiers
 - **File imports**: Modular schemas with import support
-- **gRPC service generation**: Native gRPC stubs and service bases for Java, Python, Go, Rust, C#, JavaScript, Dart, Kotlin, and Scala
+- **gRPC service generation**: Native gRPC stubs and service bases for Java, Python, Go, Rust, C++, C#, JavaScript, Dart, Kotlin, and Scala
```
chaokunyang, line 13: **[P3] Update the linked canonical compiler documentation**

This page now advertises C++ gRPC support, but the linked `docs/compiler/index.md` still excludes C++ from the supported-language list, omits `--cpp_out` from the generation command, and does not mention the C++ gRPC dependency. Please update the canonical landing page so the two documents agree.

chaokunyang (commented):
I found several correctness, compatibility, performance, and integration issues; see the inline comments.

docs/guide/cpp/grpc-support.md
```diff
@@ -0,0 +1,277 @@
+---
+title: gRPC Support
+sidebar_position: 8
+id: grpc_support
```
wenshao: **[Suggestion]** `sidebar_position: 8` duplicates the existing `polymorphism.md`, which also uses position 8. Docusaurus resolves the tie alphabetically, placing gRPC Support before Polymorphism. Meanwhile, the "Next Steps" link list in `index.md` appends gRPC Support after Row Format (position 11), so the sidebar order and the curated reading list disagree. — Concrete cost: a reader following the sidebar sees gRPC Support wedged between Type Registration (7) and Polymorphism (8-alphabetical), while the Next Steps list sends them there last, after Row Format (11).

```suggestion
sidebar_position: 12
id: grpc_support
```

_— Qwen3-235B-A22B via Qwen Code /review_

wenshao (commented):
Reviewed — no blockers. Suggestions are inline. Not linted (tool limitation, not a blocker): the executable-script lint — `.github/workflows/ci.yml`: actionlint embedded-shell source mapping is not yet supported — not linted.

_— Qwen3-235B-A22B via Qwen Code /review (v0.21.2)_

compiler/fory_compiler/generators/services/cpp.py
```diff
@@ -0,0 +1,833 @@
...
+                method_names[method_key] = self._allocate_scoped_identifier(
+                    method.name,
+                    used_methods,
+                    f"C++ gRPC service {service.name} methods",
+                    method.name,
+                    reserved_names=reserved_method_names,
+                )
+                method_name = method_names[method_key]
+                # Client stub code uses `channel_` as a data member name.
+                if method_name == "channel_" or method_name in used_members:
```
wenshao, line 180: **[Suggestion]** The `channel_` member-collision guard is not exercised by any test. `test_cpp_grpc_name_collisions` covers `NewStub`, `Greeter`, `GreeterStub`, and `Greeter_` collisions but never a method named `channel_`. — Failure scenario: if this guard is accidentally removed during a future refactor, an IDL with `rpc channel_ (Req) returns (Res)` would generate a stub whose method `channel_()` shadows the `channel_` data member, producing a C++ compilation error in user code. No unit test would catch the regression.

Suggested fix: add a case to `test_cpp_grpc_name_collisions` that defines a service with `rpc channel_` and asserts `ValueError`.

_— Qwen3-235B-A22B via Qwen Code /review_

compiler/fory_compiler/generators/services/rust.py
```diff
@@ -265,12 +288,13 @@ def generate_grpc_method_signature(
         method_name = self._rpc_method_identifier_cache[service_key][method_key]
         request_type = self.service_type_path(method.request_type)
         response_type = self.service_type_path(method.response_type)
+        mode = streaming_mode(method)
         lines: List[str] = []
-        if method.client_streaming:
+        if mode in (StreamingMode.CLIENT_STREAMING, StreamingMode.BIDIRECTIONAL):
```
wenshao, line 293: **[Suggestion]** The Rust streaming-mode refactor (from `method.client_streaming`/`method.server_streaming` booleans to `streaming_mode(method)` enum) has no unit test coverage. `test_grpc_streaming_method_shapes` asserts streaming output for Java, Python, Go, Kotlin, C#, and Scala, but contains zero Rust assertions. — Failure scenario: a mutation that swaps `CLIENT_STREAMING` and `SERVER_STREAMING` in the `mode in (...)` tuples would pass every compiler unit test. Only the `RustGrpcTest` integration test — which requires a full Rust toolchain and cross-language round-trip — would catch it.

Suggested fix: add Rust assertions to `test_grpc_streaming_method_shapes` verifying `tonic::Streaming` appears in the generated Rust gRPC output.

_— Qwen3-235B-A22B via Qwen Code /review_

docs/compiler/compiler-guide.md
```diff
@@ -149,10 +149,10 @@ foryc user.fdl order.fdl product.fdl --output ./generated
 foryc compiler/examples/service.fdl --java_out=./generated/java --python_out=./generated/python --go_out=./generated/go --rust_out=./generated/rust --csharp_out=./generated/csharp --dart_out=./generated/dart --scala_out=./generated/scala --kotlin_out=./generated/kotlin --javascript_out=./generated/javascript
 ```
 
-**Generate Java, Python, Go, Rust, C#, Dart, Scala, Kotlin, and Node.js JavaScript gRPC service companions:**
+**Generate Java, Python, Go, Rust, C++, C#, Dart, Scala, Kotlin, and Node.js JavaScript gRPC service companions:**
```
wenshao, line 152: **[Suggestion]** C++ is added to the gRPC heading and command line, but the per-language import summary paragraph immediately below (lines 158-166) still lists only Java, Python, Go, Rust, Scala, Kotlin, C#, Dart, and JavaScript — C++ is absent. — Concrete cost: a user following the guide learns what every other language's gRPC output imports but gets no mention of C++ (which imports `grpcpp` headers and links against `grpc++`). The omission is conspicuous because the paragraph is an explicit per-language enumeration, not a "some languages" summary.

Suggested fix: add a C++ sentence to the paragraph, e.g. after the Rust clause: "C++ output imports `grpcpp` headers and links against `grpc++`."

_— Qwen3-235B-A22B via Qwen Code /review_

compiler/fory_compiler/generators/services/cpp.py
BaldDemian: I did some micro benchmark after applying this advice. This optimization indeed improves the performance on relatively large payloads

BaldDemian:
Hi @chaokunyang, all advice listed above has been addressed. PTAL.

chaokunyang (approved):
LGTM

PR #3907, opened 2026-08-01
feat(json): add GraalVM codegen support for json

chaokunyang:
## AI Contribution Checklist


ayush00git (approved):

java/fory-json/src/main/java/org/apache/fory/json/codec/SqlJsonCodecs.java
```diff
@@ -110,12 +122,16 @@ public T readUtf8(Utf8JsonReader reader) {
...
+        return constructorHandle == null
+            ? constructor.newInstance(millis)
+            : (T) constructorHandle.invoke(millis);
       } catch (ReflectiveOperationException e) {
-        throw new ForyJsonException(
-            "Cannot create SQL JSON type " + constructor.getDeclaringClass(), e);
+        throw new ForyJsonException("Cannot create SQL JSON type " + type, e);
+      } catch (Throwable e) {
+        throw new ForyJsonException("Cannot create SQL JSON type " + type, e);
       }
```
wenshao: **[Suggestion]** The `catch (ReflectiveOperationException e)` block is redundant — it is entirely subsumed by the subsequent `catch (Throwable e)` block, and both perform the identical wrapping. — Concrete cost: any future change to error handling (e.g., adding context, unwrapping `InvocationTargetException`) must be applied in two places or the blocks silently diverge.

```suggestion
    } catch (Throwable e) {
      throw new ForyJsonException("Cannot create SQL JSON type " + type, e);
    }
```

_— qwen3-coder-plus via Qwen Code /review_

java/fory-json/src/main/java17/org/apache/fory/json/JsonCreatorCodegen.java
```diff
@@ -0,0 +1,442 @@
...
+
+    private void convert(Class<?> type) throws IOException {
+      if (!type.isPrimitive()) {
+        if (type != Object.class) {
+          this.type(0xc0, internalName(type));
+        }
+        return;
+      }
+      Class<?> boxType = BOX_TYPES.get(type);
+      this.type(0xc0, internalName(boxType));
```
wenshao: **[Suggestion]** `JsonCreatorCodegen` emits unboxing bytecode for 8 primitive types but only `int` is exercised through `@JsonCreator` constructors in the integration tests; `long`, `double`, `boolean`, `byte`, `short`, `char`, and `float` paths have no test coverage. — Concrete cost: if a future change to `convert()` or `boxTypes()` introduces a wrong box type or unboxing method for `long`, a `@JsonCreator` with a `long` parameter would produce a `VerifyError` or silent value truncation in the native image with no test to catch it.

Consider adding a `@JsonCreator` model to `ForyJsonExample` that exercises at least `long` and `double` constructor parameters.

_— qwen3-coder-plus via Qwen Code /review_

java/fory-json/src/main/java17/org/apache/fory/json/ForyJsonGraalVMFeature.java
```diff
@@ -114,17 +147,218 @@ public void duringAnalysis(DuringAnalysisAccess access) {
...
+    }
+    ArrayList<Method> methods = new ArrayList<>(effective.values());
+    methods.sort(
+        Comparator.comparing(Method::getName)
+            .thenComparing(method -> method.getDeclaringClass().getName())
+            .thenComparing(Method::toGenericString));
+    return methods;
+  }
+
+  private boolean generateConfigurations(DuringAnalysisAccess access) {
```
wenshao, line 262: **[Suggestion]** Unused `DuringAnalysisAccess access` parameter — the method body never references `access`; it works entirely through instance fields (`hostedConfigurations`, `processedModels`, `reachableMixins`, `processedGenerations`). This is dead code left over from the removed `registerGeneratedCodec` path that used `access.findClassByName`. — Concrete cost: compiling with `-Xlint:unused` flags this as a warning, and a `-Werror` CI configuration would fail the build; the parameter also creates false coupling at the call site.

```suggestion
  private boolean generateConfigurations() {
```

_— qwen3-coder-plus via Qwen Code /review_

java/fory-json/src/main/java/org/apache/fory/json/resolver/JsonGeneratedClassRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.Set;
+import org.apache.fory.annotation.Internal;
+import org.apache.fory.json.resolver.JsonSharedRegistry.GeneratedClasses;
+import org.apache.fory.reflect.TypeRef;
+
+/** Frozen Native Image mapping from JSON configuration semantics to generated classes. */
+@Internal
+public final class JsonGeneratedClassRegistry {
+  private static Map<JsonCodegenKey, MutableConfiguration> pending = new HashMap<>();
+  private static Map<JsonCodegenKey, Configuration> configurations = Collections.emptyMap();
```
wenshao, line 36: **[Suggestion]** `configurations` is written under `synchronized` in `register()` but read without synchronization or `volatile` in the public `configuration()` method — a JMM data race on the reference. In the current GraalVM-only lifecycle the native-image heap-snapshot provides a happens-before edge, so the race does not manifest. However, `configuration()` is `@Internal public` and called from 12 unsynchronized sites in `JsonSharedRegistry`. If any future code path calls `configuration()` concurrently with `register()` without an external synchronization barrier, the reading thread can see the stale `Collections.emptyMap()` initial value. — Concrete cost: every `native*Class()` lookup returns `null`, silently falling back to interpreted codecs.

```suggestion
  private static volatile Map<JsonCodegenKey, Configuration> configurations = Collections.emptyMap();
```

_— qwen3-coder-plus via Qwen Code /review_

wenshao (commented):
Reviewed. Suggestions are inline.

_— qwen3-coder-plus via Qwen Code /review (v0.21.2)_

PR #3994, opened 2026-08-29
fix(javascript): reserve writer capacity for write paths

ayush00git:
## Why?
BinaryWriter starts with a ~100KB buffer and, for speed, its individual writeXxx methods do no bounds checking. The contract is that callers call reserve(n) first, which grows the buffer if needed. The generated collection write path honors this (it reserves elementFixedSize * count before the loop), but three paths write per-entry data with no reserve at all: the generated declared-map write, MapAnySerializer.write, and CollectionAnySerializer.write.

## What does this PR do?
three write paths now reserve writer capacity, matching the existing generated-collection convention (root reserves fixedSize once; per-element loops reserve fixedSize × count; strings reserve internally)

## AI Contribution Checklist


javascript/test/array.test.ts
```diff
@@ -323,6 +323,18 @@ describe("array", () => {
...
     expect(containsBytes(bfloat16Bytes, [0x80, 0x3f, 0x00, 0xc0])).toBe(true);
   });
+
+  test("should large any-typed list work", () => {
+    // The dynamic element write path must reserve writer capacity per item.
+    // Without it, single-byte writes past the buffer end were silent no-ops
+    // while the cursor advanced, so dump() returned uninitialized tail bytes.
+    const fory = new Fory({ compatible: false });
+    const { serialize, deserialize } = fory.register(Type.list(Type.any()));
+    const arr = new Array(150000).fill(1);
```
chaokunyang, line 333: Because every element is `1`, `writeElementsHeader` sets `isSame = true`, so this test exercises only the aggregate reserve. The per-item mixed-type reserves and the `writeDeclared` reserve can still be removed without this test failing. Please add a list larger than the initial writer buffer with mixed element types; if `writeDeclared` is part of this fix, cover the unknown-Struct declared-list reserialization path as well.

javascript/test/map.test.ts
```diff
@@ -213,4 +213,22 @@ describe("map", () => {
...
+
+  test("should large declared map work", () => {
+    // The generated map write must reserve writer capacity for its entries.
+    // Without it, unchecked DataView writes past the buffer end threw a
+    // RangeError once the map body outgrew the initial buffer.
+    const fory = new Fory({ compatible: false });
+    const { serialize, deserialize } = fory.register(
+      Type.struct(
+        { namespace: "example", typeName: "BigMap" },
+        { m: Type.map(Type.int32({ encoding: "fixed" }), Type.int32({ encoding: "fixed" })) },
```
chaokunyang, line 225: This declared `int32` map goes through `writeSpecificType`, so it does not execute the new reserve in `MapAnySerializer.write`. Please add a `Type.map(Type.any(), Type.any())` case large enough to exceed the initial writer buffer, so removing that reserve makes the regression test fail.

chaokunyang (approved):

PR #4006, opened 2026-08-30
fix(scala): supported nested case class in fory-json-scala

pjfanning:
## Why?

Scala case classes defined inside Scala objects are not properly supported by fory-json-scala.

On Scala 2.13 such a type was not recognized as a case class, fell through to the generic Java
object model, and silently decoded every property to its default:

```
Models.Inner(1, "a")  ->  {"id":1,"name":"a"}  ->  Inner(0, null)     // no error
```

Recognition required a static `apply` forwarder. Scala 2 emits those only for a top-level companion;
Scala 3 emits them for any statically owned module, which is why Scala 3 was unaffected.

## What does this PR do?

- `ScalaObjectModels` resolves `apply` and `$lessinit$greater$default$N` from the companion
  singleton when the case class carries no static forwarders. The static path stays the fast path.
- `JsonObjectModel` carries the receiver of instance constructor defaults, `JsonCreatorInfo` binds
  it into the default invoker, and both reader codegen paths invoke the default on that receiver.
- `ReflectionUtils.getLiteralName` skipped its nested-Scala-object correction for any canonical name
  ending in `$`, which is every companion module class, so a companion two or more levels inside an
  object produced a name the generated-code compiler cannot resolve. Fixed at the root.
- `ForyJsonGraalVMFeature` registers the companion class, its `MODULE$` field, and the `apply` and
  constructor-default methods the Scala module queries when it rebuilds the object model at image
  runtime.
- A case class that cannot be reconstructed is rejected with `UnsupportedJsonTypeException` instead
  of silently decoding: one enclosed by a class or trait, which needs an outer instance, and one
  declared in a method, whose companion is not reachable.

`fory-json-kotlin` is not touched: the public `JsonObjectModel` constructors it calls keep their
previous signatures.

### User-facing behavior change

A case class enclosed by a class, trait, or method previously serialized (the generic model
discovered its fields, including `$outer`) and silently decoded back to an all-default instance.
Both directions now raise `UnsupportedJsonTypeException`. Writing a value that can never be read
back is the trap this PR fixes, so the write is rejected with the read.

## AI Contribution Checklist

- [x] If `yes`, I included the standardized `AI Usage Disclosure` block below.
- [x] If `yes`, I can explain and defend all important changes without AI help.
- [x] If `yes`, I reviewed AI-assisted code changes line by line before submission.
- [x] If `yes`, I completed line-by-line self-review first and fixed issues before requesting AI review.
- [x] If `yes`, I ran two fresh AI review agents on the current PR diff or current HEAD after the latest code changes: one Fory-guided reviewer using `AGENTS.md` and `.agents/ci-and-pr.md`, and one independent general reviewer in a separate clean-context session that was not pointed to `.agents/ci-and-pr.md` or any copied Fory-specific review checklist. If the independent reviewer's tooling auto-loaded `AGENTS.md`, it followed the independent-review carve-out there.
- [x] If `yes`, I addressed all AI review comments and repeated the review loop until both ai reviewers reported no further actionable comments.
- [x] If `yes`, I attached screenshot evidence or equivalent persisted links of the final clean AI review results from both fresh reviewers on the current PR diff or current HEAD after the latest code changes in this PR body.
- [x] If `yes`, I ran adequate human verification and recorded evidence (checks run locally or in CI, pass/fail summary, and confirmation I reviewed results).
- [x] If `yes`, I added/updated tests and specs where required.
- [x] If `yes`, I validated protocol/performance impacts with evidence when applicable.
- [x] If `yes`, I verified licensing and provenance compliance.

AI review artifacts (verbatim final report of every review session, with an index):
https://gist.github.com/pjfanning/6b9fcd687e22450399df72dabc368e99

```text
AI Usage Disclosure
- scope: design drafting, code drafting, tests, docs (Claude Opus 5, `claude-opus-5[1m]`, via
  Claude Code; both AI reviewers were the same model in separate clean-context sessions)
- affected_files_or_subsystems: scala/fory-json-scala (ScalaObjectModels, ScalaJsonSuite,
  ScalaJsonEnumerationNativeImageMain), java/fory-json (JsonObjectModel, JsonCreatorInfo,
  ObjectCodecBuilder, JsonReaderCodegen, JsonCodegen, ForyJsonGraalVMFeature),
  java/fory-core (ReflectionUtils), docs/json/scala.md
- ai_review: line-by-line self-review completed by the contributor. Six rounds of the two-reviewer
  loop, each on the diff as it then stood. Round 6, on the current head `13a62665c`, is the final
  clean round: the Fory-guided reviewer reported no blocking findings and the independent reviewer
  reported no actionable findings and "I would approve this". Earlier rounds each surfaced at least
  one real defect, including a blocker (the Kotlin module would not have compiled) found by the
  independent reviewer alone, a regression introduced by an earlier round's own fix, and two tests
  that passed for the wrong reason. One reviewer suggestion was implemented, broke three tests, and
  was reverted with a source comment recording why. Three optional round-6 nits were reviewed and
  deliberately not actioned so the clean result stands on the head; they are listed in the gist
- ai_review_artifacts: https://gist.github.com/pjfanning/6b9fcd687e22450399df72dabc368e99
- human_verification: see Verification below; contributor reviewed the results
- performance_verification: N/A. No hot-path change. Constructor-default resolution happens once at
  codec build time; the generated reader fetches the receiver only on a missing JSON property, on
  the branch that uses it; the static-forwarder path is unchanged for top-level case classes.
- provenance_license_confirmation: Apache-2.0-compatible provenance confirmed; no third-party code
  introduced
```

## Verification

| Suite | Scala 2.13 | Scala 3.3.8 |
| --- | --- | --- |
| `ScalaJsonSuite` | 20 passed | 20 passed |
| `ScalaJsonEnumerationSuite` | 4 passed | 4 passed |
| `ScalaJsonDerivationSuite` | n/a | 10 passed |

Scala suites were run directly through the ScalaTest runner against locally built modules rather
than via `sbt +test`: Scala 3.3.8 matches the cross-build, the 2.13 leg used 2.13.14 against the
repo's 2.13.18 target. CI's cross-build is the authoritative check.

- `fory-core` maven suite: 2294 passed, 0 failures. Re-run because `ReflectionUtils.getLiteralName`
  feeds every serializer's codegen, not only the JSON path.
- `fory-scala` binary serializer suites: 88 passed, for the same reason, including
  `SingleObjectSerializerTest`, which exercises `object A { object B { case class C } }`.
- `fory-json` maven suite: 964 passed, 0 failures.
- `fory-json-kotlin` compiles against unmodified Kotlin sources.
- Generated readers dumped with `FORY_CODE_DIR` and confirmed to contain the new call in both
  codegen paths, including a default that consumes a preceding constructor argument.
- `prettier --write docs/json/scala.md` reports the file unchanged.

### Known gaps

- **The GraalVM path is verified by inspection only.** No GraalVM in the dev environment, and
  nothing in the repo runs the Scala native-image mains — `ScalaJsonEnumerationNativeImageMain` and
  `ScalaJsonNativeImageMain` are referenced by no sbt task or workflow job (pre-existing; the
  `graalvm_json` job builds a Java main from `integration_tests/graalvm_tests`). The harness gained
  nested case-class round-trips so the coverage exists if it is wired up.
- **`sbt +test`, `mvn spotless:check` and `mvn checkstyle:check` were not run locally.** Spotless
  cannot run in this environment: google-java-format throws `NoClassDefFoundError` under JDK 17, on
  a test file this PR does not touch. Formatting was checked by hand against the surrounding style
  and every added line is within 100 columns; CI's format job is the first real check.

A case class enclosed by a class, trait, or method is now rejected rather than silently
round-tripping to defaults, as described above. No public API or wire-format change.

### DIFF
--- a/docs/json/scala.md
+++ b/docs/json/scala.md
@@ -50,9 +50,10 @@ or mutate constructor `val` fields. Defaults in later parameter lists receive th
 constructor arguments exactly as Scala defines them. A missing parameter without a default is an
 error. Mutable body properties are applied after construction.
 
-A case class may be declared at the top level or inside an `object`, at any nesting depth. A case
-class declared inside a `class` or inside a method is rejected for both reading and writing,
-because Fory cannot reach the enclosing instance or companion it needs to rebuild the value.
+A case class may be declared at the top level, or inside an `object` at any nesting depth, as long
+as every enclosing scope is itself an `object`. A case class enclosed by a `class` or by a method
+is rejected for both reading and writing, because Fory cannot reach the enclosing instance or the
+companion it needs to rebuild the value.
 
 Fory JSON annotations can be placed directly on Scala constructor properties:
 
--- a/docs/json/scala.md
+++ b/docs/json/scala.md
@@ -51,8 +51,8 @@ constructor arguments exactly as Scala defines them. A missing parameter without
 error. Mutable body properties are applied after construction.
 
 A case class may be declared at the top level or inside an `object`, at any nesting depth. A case
-class declared inside a `class` or inside a method is rejected, because Fory cannot reach the
-enclosing instance or companion it needs to rebuild the value.
+class declared inside a `class` or inside a method is rejected for both reading and writing,
+because Fory cannot reach the enclosing instance or companion it needs to rebuild the value.
 
 Fory JSON annotations can be placed directly on Scala constructor properties:
 
--- a/docs/json/scala.md
+++ b/docs/json/scala.md
@@ -51,7 +51,7 @@ constructor arguments exactly as Scala defines them. A missing parameter without
 error. Mutable body properties are applied after construction.
 
 A case class may be declared at the top level, or inside an `object` at any nesting depth, as long
-as every enclosing scope is itself an `object`. A case class enclosed by a `class` or by a method
+as every enclosing scope is itself an `object`. A case class enclosed by a `class`, a trait, or a method
 is rejected for both reading and writing, because Fory cannot reach the enclosing instance or the
 companion it needs to rebuild the value.
 
--- a/docs/json/scala.md
+++ b/docs/json/scala.md
@@ -51,9 +51,9 @@ constructor arguments exactly as Scala defines them. A missing parameter without
 error. Mutable body properties are applied after construction.
 
 A case class may be declared at the top level, or inside an `object` at any nesting depth, as long
-as every enclosing scope is itself an `object`. A case class enclosed by a `class`, a trait, or a method
-is rejected for both reading and writing, because Fory cannot reach the enclosing instance or the
-companion it needs to rebuild the value.
+as every enclosing scope is itself an `object`. A case class enclosed by a `class`, a trait, or a
+method is rejected for both reading and writing, because Fory cannot reach the enclosing instance
+or the companion it needs to rebuild the value.
 
 Fory JSON annotations can be placed directly on Scala constructor properties:
 
--- a/docs/json/scala.md
+++ b/docs/json/scala.md
@@ -50,6 +50,10 @@ or mutate constructor `val` fields. Defaults in later parameter lists receive th
 constructor arguments exactly as Scala defines them. A missing parameter without a default is an
 error. Mutable body properties are applied after construction.
 
+A case class may be declared at the top level or inside an `object`, at any nesting depth. A case
+class declared inside a `class` or inside a method is rejected, because Fory cannot reach the
+enclosing instance or companion it needs to rebuild the value.
+
 Fory JSON annotations can be placed directly on Scala constructor properties:
 
 ```scala
--- a/java/fory-core/src/main/java/org/apache/fory/reflect/ReflectionUtils.java
+++ b/java/fory-core/src/main/java/org/apache/fory/reflect/ReflectionUtils.java
@@ -633,10 +633,11 @@ public class ReflectionUtils {
       if (canonicalName.contains("$")) {
         // nested scala object type can't be accessed in java by using canonicalName. This includes
         // a nested module class, whose own name ends with `$`: the canonical name of a companion
-        // declared two or more levels inside an object mixes `.` and `$` separators and names no
-        // resolvable type. A module class only one level deep keeps its canonical name: only its
-        // last segment ends with `$`, which resolves, so the nesting-level bound below is load
-        // bearing in both directions.
+        // declared two or more levels inside an object mixes `.` and `$` separators, which the
+        // generated-code compiler mangles back into a name that resolves to nothing
+        // ("pkg.A$B$ declares no member type C$"). A module class only one level deep keeps its
+        // canonical name, because only its last segment ends with `$`, so the nesting-level bound
+        // below is load bearing in both directions.
         // see more detailed in
         // https://stackoverflow.com/questions/30809070/accessing-scala-nested-classes-from-java
         int nestedLevels = 0;
--- a/java/fory-core/src/main/java/org/apache/fory/reflect/ReflectionUtils.java
+++ b/java/fory-core/src/main/java/org/apache/fory/reflect/ReflectionUtils.java
@@ -633,11 +633,11 @@ public class ReflectionUtils {
       if (canonicalName.contains("$")) {
         // nested scala object type can't be accessed in java by using canonicalName. This includes
         // a nested module class, whose own name ends with `$`: the canonical name of a companion
-        // declared two or more levels inside an object mixes `.` and `$` separators, which the
-        // generated-code compiler mangles back into a name that resolves to nothing
-        // ("pkg.A$B$ declares no member type C$"). A module class only one level deep keeps its
-        // canonical name, because only its last segment ends with `$`, so the nesting-level bound
-        // below is load bearing in both directions.
+        // declared two or more levels inside an object has a `$`-terminated segment in the
+        // middle of its canonical name, one per enclosing module class, and the generated-code
+        // compiler cannot resolve through those ("pkg.A$B$ declares no member type C$"). One level
+        // deep is enclosed by the mirror class instead, so `pkg.A.C$` ends with the only `$` and
+        // still resolves: the nesting-level bound below is load bearing in both directions.
         // see more detailed in
         // https://stackoverflow.com/questions/30809070/accessing-scala-nested-classes-from-java
         int nestedLevels = 0;
--- a/java/fory-core/src/main/java/org/apache/fory/reflect/ReflectionUtils.java
+++ b/java/fory-core/src/main/java/org/apache/fory/reflect/ReflectionUtils.java
@@ -630,8 +630,11 @@ public class ReflectionUtils {
       // qualifier name of scala object type will ends with `.`
       canonicalName = clsName.substring(0, clsName.length() - 1).replace("$", ".") + "$";
     } else {
-      if (!canonicalName.endsWith("$") && canonicalName.contains("$")) {
-        // nested scala object type can't be accessed in java by using canonicalName
+      if (canonicalName.contains("$")) {
+        // nested scala object type can't be accessed in java by using canonicalName. This includes
+        // a nested module class, whose own name ends with `$`: the canonical name of a companion
+        // declared two or more levels inside an object mixes `.` and `$` separators and names no
+        // resolvable type.
         // see more detailed in
         // https://stackoverflow.com/questions/30809070/accessing-scala-nested-classes-from-java
         int nestedLevels = 0;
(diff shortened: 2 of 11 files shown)


pjfanning:
## AI review loop (AI_POLICY.md §5)

Model: Claude Opus 5 (`claude-opus-5[1m]`), via Claude Code, for both the authoring and the two
reviewers. The reviewers ran as separate clean-context sessions with different instructions:

- **Fory-guided reviewer** — instructed to read `AGENTS.md` and follow the review workflow in
  `.agents/ci-and-pr.md`, plus `.agents/languages/{scala,java}.md`.
- **Independent general reviewer** — explicitly forbidden from loading `.agents/ci-and-pr.md` or
  any copied Fory checklist, per the independent-review carve-out in `AGENTS.md`.

Both were read-only: no edits, builds, tests, or pushes. Four rounds, each on the diff as it stood.
Current PR head is `a86a870b4`.

### Round 1 — `e493fdb86`

Both reviewers independently found the same four issues:

| Finding | Resolution |
| --- | --- |
| GraalVM: the companion class, `MODULE$`, and `apply` were never registered, so a nested case class would resolve on the JVM but fail in a native image | Added `registerScalaCompanion`, following the existing `registerScalaEnumerationOwner` precedent |
| A comment claimed an instance default's "bound invoker already lives in the creator metadata" — false; `buildDefaultInvokers` unreflects at runtime for both static and instance defaults | Corrected; the real reason to skip `registerCreator` is that `creatorHandle` spreads an argument array over the exact parameter count and cannot describe a receiver |
| Recognition forced the companion's `<clinit>` just to answer `isCaseClass`, running user `object` bodies during type resolution | Split recognition from binding: `Class.forName(..., initialize = false)`, `MODULE$` read only once the model is committed |
| `receiver == null` conflated "static forwarders" with "null singleton" | Explicit `staticForwarders` flag |

One suggestion was **rejected with evidence**: both reviewers proposed hoisting the per-parameter
receiver fetch in generated code. Implementing it broke three tests — generated code for an
`Expression` instance is emitted once at its first use site, and each use here is a separate
missing-argument block, so a shared instance referenced a local declared in a sibling block. Reverted
with a comment recording why.

The requested doubly-nested test then exposed a **pre-existing bug in `fory-core`**:
`ReflectionUtils.getLiteralName` skipped its nested-Scala-object correction for any canonical name
ending in `$`, which is every companion module class. A companion two levels inside an object
produced `pkg.A$B$.C$` — Janino: `"pkg.A$B$" declares no member type "C$"`. Fixed at the root.

### Round 2 — `ec34ed31b`

The independent reviewer found a **blocker**: `defaultsReceiver` had been inserted into the public
full `JsonObjectModel` constructor, whose 15-argument form `KotlinMetadataModels.kt:182` calls, so
`fory-json-kotlin` would not compile. The Fory-guided reviewer did not catch this.

Resolved by making the full constructor private and restoring the previous public signature, so no
Kotlin file is touched. Verified with `mvn -pl fory-json-kotlin compile` on unmodified Kotlin
sources.

Also resolved this round: the `copy`/`productPrefix` marker was claiming types with a *reachable*
companion but an unsupported primary constructor (varargs, non-public), turning previously working
generic-model handling into a hard failure — now restricted to unreachable companions; and the
workspace-path receiver local ran on every construction, so it was reverted to a conditional fetch.

### Round 3 — `c0d49bae6`

No blockers from either reviewer. Both confirmed the round-2 fixes, and the independent reviewer
re-enumerated every `JsonObjectModel`/`JsonCreatorInfo` call site to confirm each binds
unambiguously.

The notable finding was about the tests: the two rejection tests both passed for the wrong reason.
An outer-bound case class *also* has no reachable companion, so deleting the `$outer` check would
have left both green. Asserting the messages then exposed a second layer — the method-local test was
declared inside the suite *class*, so it captured an outer instance and hit the outer branch,
never reaching the companion branch it claimed to cover. It now lives in a method of an `object`.

Also fixed: a `LinkageError` from a companion that exists but cannot be linked was reported as
"companion is not reachable, such as a case class declared in a method"; the receiver invariant in
`JsonObjectModel.validate` was made total; `CompanionOwner` carries the `MODULE$` `Field`.

### Round 4 — `637b5063f`

**Fory-guided reviewer: no actionable findings.** **Independent reviewer: would approve**, with
four non-blocking notes.

The two disagreed on one point, and the independent reviewer was right: the round-3 fix for
`LinkageError` handling had been an overcorrection. Rethrowing inside `companionOwner` put a hard
failure into `isCaseClass`, which is consulted for every non-tuple `Product` reaching this module —
including types it does not own and never claims. An unrelated `Product` with a stale companion
could therefore fail resolution that previously succeeded through the core object model, where
before this PR `isCaseClass` did no class loading at all.

Also from this round, and fixed in `a86a870b4`:

- A comment stated the wrong cause. It said the deeply nested canonical name "names no resolvable
  type"; javac resolves it, and the actual evidence was a Janino error
  (`"pkg.A$B$" declares no member type "C$"`). The comment now names the real failure.
- `getLiteralName` is a `fory-core` change affecting all serializer codegen, and had only been
  exercised through the JSON path. The `fory-scala` binary serializer suites now cover it —
  88 tests, including `SingleObjectSerializerTest`, which exercises exactly
  `object A { object B { case class C } }`.
- The `$outer` rejection also fires for a case class enclosed by a trait; message and doc updated.
- The native-image main additionally round-trips a nested case class with no defaults (which still
  needs the companion to match `apply`) and a doubly nested one.

**`a86a870b4` (the current head) contains these fixes and has not itself been re-reviewed**, so this
is not a clean final review of the head.

**Artifacts:** https://gist.github.com/pjfanning/6b9fcd687e22450399df72dabc368e99 — the verbatim
final report of all eight review sessions, extracted from their transcripts, with an index.

## Verification

| Suite | Scala 2.13.18 | Scala 3.3.8 |
| --- | --- | --- |
| `ScalaJsonSuite` | 20 passed | 20 passed |
| `ScalaJsonEnumerationSuite` | 4 passed | 4 passed |
| `ScalaJsonDerivationSuite` | n/a | 10 passed |

`fory-core` maven suite: 2294 passed, 0 failures, re-run because the `ReflectionUtils` change feeds
every serializer's codegen; `fory-scala` binary serializer suites: 88 passed, for the same reason. `fory-json-kotlin` compiles against unmodified Kotlin sources. Generated
readers were dumped with `FORY_CODE_DIR` and confirmed to contain the new call in both codegen
paths, including a default that consumes a preceding constructor argument.

## Known gaps

- **The GraalVM path is unverified.** No GraalVM in the dev environment, and nothing in CI runs the
  Scala native-image mains — `ScalaJsonEnumerationNativeImageMain` and `ScalaJsonNativeImageMain` are
  referenced by no sbt task or workflow job (pre-existing; the `graalvm_json` job builds the Java
  main from `integration_tests/graalvm_tests`). `registerScalaCompanion` is therefore checked by
  inspection only. Both reviewers flagged this in every round. The harness gained a nested case class
  round-trip so the coverage exists if someone wires it up.
- **`mvn spotless:check` could not run locally** — google-java-format throws
  `NoClassDefFoundError` under JDK 17 on an unrelated test file. Formatting was checked by hand
  against the surrounding style; CI's format job is the first real check.

## Behavior change

A case class declared inside a `class` or a method previously serialized and silently decoded back
to an all-default instance. Both directions now raise `UnsupportedJsonTypeException`. Writing a
value that can never be read back is the trap this PR fixes, so the write is rejected with the read.

pjfanning:
### AI review loop: rounds 5 and 6

Following the earlier summary, the loop ran two further rounds and has converged.

**Round 5 (`a86a870b4`)** — both reviewers independently found the same gap: guarding only
`Class.forName` did not make case-class recognition free of linkage failures, because `getField`
and `getMethods` resolve member descriptors. A companion or case class declaring a member whose
type is absent at runtime could throw `NoClassDefFoundError` out of `isCaseClass`, a predicate
consulted for every `Product` reaching the module, including types it does not own. Fixed in
`13a62665c`: recognition declines such a type, the committing path still reports it, and ambiguity
stays loud with the comment saying so.

**Round 6 (`13a62665c`, current head)** — Fory-guided reviewer: no blocking findings. Independent
reviewer: no actionable findings, "I would approve this". The independent reviewer traced whether
the widened `catch` could swallow a genuine problem and confirmed it cannot; the Fory-guided
reviewer compiled and inspected the class files to verify the `getLiteralName` reasoning, which
also corrected one of its own earlier nits.

Three optional round-6 nits were reviewed and deliberately not actioned, so that the clean result
stands on the head that is actually being merged. They are listed in the gist index.

**Artifacts updated** — all twelve reviewer reports, verbatim, with a per-round outcome table:
https://gist.github.com/pjfanning/6b9fcd687e22450399df72dabc368e99

**Verification added since the earlier comment:** `fory-json` maven suite, 964 passed, 0 failures.
Full results, and the gaps that remain (the GraalVM path is verified by inspection only, and
`sbt +test`/`spotless`/`checkstyle` did not run locally), are in the PR description.

chaokunyang (approved):
Looks great!

PR #4013, opened 2026-09-01
fix(javascript): honor declared element types for root container registration

ayush00git:
## What was the error
Registering a root container with declared element types, e.g. `fory.register(Type.list(Type.float32()))`, silently returned the internal any-typed container serializer. The declared generics were discarded: elements were written with dynamic dispatch instead of the declared type, so declared `float32`/`int64` semantics were lost at the root while the same declaration worked as a struct field.

## What this PR fixes
A root `Type.list`/`Type.set`/`Type.map` with declared element types now gets a dedicated generated serializer, bound to the returned root serialize/deserialize pair. It is kept out of the type-id keyed registry so dynamic container dispatch stays untouched. Regression tests cover root list, set, and map plus the dynamic-dispatch guard.

### DIFF
--- a/javascript/packages/core/lib/gen/index.ts
+++ b/javascript/packages/core/lib/gen/index.ts
@@ -18,7 +18,7 @@
  */
 
 import { TypeId, Serializer } from "../type";
-import { TypeInfo } from "../typeInfo";
+import { containerDeclaresElementTypes, TypeInfo } from "../typeInfo";
 import { CodegenRegistry } from "./router";
 import { CodecBuilder } from "./builder";
 import { Scope } from "./scope";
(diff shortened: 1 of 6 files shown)


javascript/packages/core/lib/gen/index.ts
```diff
@@ -175,6 +175,12 @@ export class Gen {
 
   generateSerializer(typeInfo: TypeInfo) {
     this.traversalContainer(typeInfo);
+    if (containerDeclaresElementTypes(typeInfo)) {
+      // The type-id keyed registry only holds the dynamic container
+      // serializer; a container with declared element types gets a dedicated
+      // serializer for this registration.
+      return this.generate(typeInfo);
```
chaokunyang, line 185: Registering the container before its extension codec now leaves the generated serializer bound to `undefined`, even when all registration finishes before the first root operation:

```ts
class E {}
Type.ext(901)(E);

const fory = new Fory({ compatible: false });
const list = fory.register(Type.list(Type.ext(901)));
fory.register(E, { write() {}, read() {} });
list.serialize([new E()]);
```

`traversalContainer()` creates forward placeholders for structs but not extensions, so `ExtSerializerGenerator.writeEmbed()` captures the missing serializer in a factory-level constant. Registering `E` afterward cannot update that constant, and serialization fails at `ext_ser.writeTypeInfo(null)`. Sets have the same issue. The previous dynamic root serializer resolved the codec at write time.

Please preserve registration ordering before the first operation by ensuring the generated container binds to the completed extension codec, and add a regression test for this order.

chaokunyang (approved):
LGTM

PR #4022, opened 2026-09-05
fix(core): keep TypeDef root kind consistent for factory custom serializers

### LINKED ISSUE #4021: TypeDef root kind mismatch (30 vs 32) for non-collection classes with SerializerFactory custom serializer
## Bug Description

When a **non-collection plain class** (e.g. not `Collection`, `Map`, `Externalizable`, etc.) is serialized through a custom `Serializer` provided by a configured `SerializerFactory`, the TypeDef root kind becomes inconsistent between writer and reader — even when both sides run **identical code with no pre-registration**.

**Writer** lazily resolves the class to its factory custom serializer, producing `TypeInfo.typeId = NAMED_EXT (32)`, which gets encoded into the TypeDef header byte.

**Reader** decodes the same TypeDef but has no `TypeInfo` for the class yet (it was never serialized locally). Since the class is not in the `usesNonStructTypeDef` set (it's a plain class, not a Collection/Map/array), the reader falls back to `buildUnregisteredTypeId(cls, null)` with `serializer=null`, which yields `NAMED_COMPATIBLE_STRUCT (30)` under `CompatibleMode.COMPATIBLE` + meta share.

The validation in `NativeTypeDefDecoder.isCompatibleRootKind(expected=30, actual=32)` fails because `NAMED_EXT` is not in the struct compatibility set.

## Error Message

```
org.apache.fory.exception.DeserializationException: TypeDef root kind does not match the decoded class:
class=com.example.MyRestriction, expected=30, actual=32, registeredClassLayer=false
```

## Reproduction

Minimal repro (symmetric, same code, no pre-registration):

1. Define a plain interface `Restriction` + a concrete class `MyRestriction implements Restriction` (**not** a Collection/Map)
2. Define a `Container` struct with a field typed as `Restriction` (interface), set to `MyRestriction` at runtime
3. Configure `SerializerFactory` to return a custom `Serializer` for `Restriction` subtypes
4. Build two `Fory` instances with `CompatibleMode.COMPATIBLE` + `metaShare(true)` + `codegen` + the factory — **neither** pre-registers `MyRestriction`
5. Serialize `Container` with instance A, deserialize with instance B → **DeserializationException**

A full regression test is included in the fix commit: `TypeDefRootKindSymmetricReproTest.java`

**Note**: Classes that are `Collection`/`Map`/`Externalizable` etc. do **not** trigger this bug because `usesNonStructTypeDef` already normalizes them to `NAMED_EXT`. Only **plain classes** routed through `SerializerFactory` hit the gap.

## Root Cause

`ClassResolver.getTypeDefRootTypeId()` does not probe `SerializerFactory` when determining the root kind for unregistered classes. The writer has already resolved the factory serializer (so `TypeInfo.typeId=32`), but the reader has no `TypeInfo` and falls back to the struct path (`30`).

## Proposed Fix

Add a `SerializerFactory` probe (with `ConcurrentHashMap` caching) in `ClassResolver.getTypeDefRootTypeId()`:

```java
private final ConcurrentHashMap<Class<?>, Boolean> extSerializerFlagCache = new ConcurrentHashMap<>();

private boolean isCustomSerializedByFactory(Class<?> cls) {
    return extSerializerFlagCache.computeIfAbsent(cls, c -> {
        try {
            return createSerializerFromFactory(c) != null;
        } catch (Throwable t) {
            return false;
        }
    });
}
```

Then in `getTypeDefRootTypeId`, before falling through to the struct path:

```java
if (isCustomSerializedByFactory(cls)) {
    return Types.NAMED_EXT;
}
```

This is inserted in both the `hasFieldMetadata=true` branch (reader path) and the no-TypeInfo branch (writer path).

**Performance**: `computeIfAbsent` ensures each class is probed at most once; subsequent calls are O(1) map lookups. `ConcurrentHashMap` provides thread safety without external locking.

Full fix + regression test: https://github.com/Pigsy-Monk/fory/commit/7c70cd283

## Environment

- fory-core version: confirmed on 1.5.0 (Maven Central jar) and 1.8.0-SNAPSHOT (current main)
- Java: 21 (Eclipse Adoptium)
- OS: Windows 11

Pigsy-Monk:
## What does this PR do?

Fixes #4021.

A fresh reader can infer a struct kind for a plain class whose configured
`SerializerFactory` actually supplies an extension serializer. Native TypeDef
validation then rejects bytes written by an identically configured instance.

Resolve empty extension roots through the existing `ClassResolver.getTypeInfo`
path before validating their kind. The resolver retains the actual serializer
and type ID, including when metadata previously created a provisional TypeInfo.
This requires no additional cache or factory probe.

Keep each serializer's root definition on `TypeInfo`, independently of the class
field schemas used by metadata probes and ObjectStream layers. Local-header
matching checks an available root definition before the field schema, without
creating a serializer just to probe for a match. This preserves cold registered
struct and ObjectStream layer matches and prevents a cached field schema from
replacing an extension's empty root definition.

## Tests

Expanded `SerializerFactoryTest` covers the original polymorphic custom serializer,
metadata-first and typed-root reads, read-then-write reuse, enum and recursive
struct factories, factory failure recovery, registration, and both codegen modes.
It also verifies that each custom serializer is created once per resolver.

- Full `fory-core` suite on JDK 21: 2,252 passed, 58 skipped, no failures or errors.
- Includes cold native ID/name registration and typed/dynamic reads, verifying
  that exact-local schemas do not consume the remote-schema allowance, plus
  existing ObjectStream layer, registration, and schema-consistent meta-share cases.
- Local cross-language peer tests were not run; the change preserves the native wire format.
- Spotless and Checkstyle passed.


chaokunyang (approved):
LGTM

Pigsy-Monk:
Thanks!

PR #4041, opened 2026-09-12
fix(android): align Kotlin JSON inclusion assertions

chaokunyang:
After #4040, the Android metadata scenario still expected `label = null` to survive a round trip with default inclusion. The writer now correctly omits that property, and the reader restores its `"android-default"` constructor default, causing both Android API 26 and API 36 CI to fail.

Update the scenario to assert omission and default restoration, then explicitly enable `writeNullFields(true)` and verify the null-preserving round trip. Cover both String and UTF-8 output/input while retaining the existing metadata, default-mask, singleton, and shared-corpus checks.

Validation is delegated to CI as requested; the local test run was stopped before completion.

Fixes the [Android CI failure](https://github.com/apache/fory/actions/runs/34702925778/job/103577771632).


ayush00git (approved):
LGTM

PR #4051, opened 2026-09-15
ci: allow Rust release retries from existing tags

chaokunyang:
Apache Fory 1.7.3 Rust publication fails before starting because its immutable tag pins `crates-io-auth-action` v1.0.4, which is no longer on the ASF action allowlist. Main already uses the approved v1.0.5 action.

Add a manual release-tag input to the existing Rust publishing workflow so it can run the current approved actions while checking out and publishing the source from the original release tag. The selected tag controls both checkout and version normalization; branch names are rejected. Automatic tag publication uses the same path.

Validation: YAML lint and Prettier checks passed, the shell validation accepts final/RC tags and rejects branch names, and `git diff --check` passed. The actual crates.io publication will run after this workflow is merged, with `tag=v1.7.3`.


ayush00git (approved):
lgtm

Issue #4069, opened 2026-09-20
fix(java): JSON generic recursion guard rejects Box<Box<Integer>> and other nested instantiations of the same generic type


pjfanning:
## Summary

`JsonTypeResolver.enterObjectBinding` rejects any object type whose raw class is already being bound with different type arguments. That correctly stops an infinite expansion such as `Node<T>` containing `Node<List<T>>`, but it also rejects the finite and common case of a generic type nested inside itself with smaller arguments, such as `Box<Box<Integer>>`. Every instantiation of the same generic class that appears while an outer instantiation is being resolved is treated as recursion.

## Reproduction

Plain Java, no Scala involved, `fory-json` 1.8.0-SNAPSHOT from `main` (0da73ac58):

```java
import org.apache.fory.json.ForyJson;
import org.apache.fory.reflect.TypeRef;

public class Probe {
  public static class Box<T> {
    public T value;
    public Box() {}
    public Box(T value) { this.value = value; }
  }

  public static void main(String[] args) {
    ForyJson json = ForyJson.builder().withCodegen(false).build();
    TypeRef<Box<Box<Integer>>> type = new TypeRef<Box<Box<Integer>>>() {};
    System.out.println(json.toJson(new Box<>(new Box<>(1)), type));
  }
}
```

```
Exception in thread "main" org.apache.fory.json.ForyJsonException: JSON generic recursion expands Probe$Box from Probe$Box<Probe$Box<java.lang.Integer>> to Probe$Box<java.lang.Integer>
	at org.apache.fory.json.resolver.JsonTypeResolver.expandingGenericType(JsonTypeResolver.java:2727)
	at org.apache.fory.json.resolver.JsonTypeResolver.enterObjectBinding(JsonTypeResolver.java:2712)
	at org.apache.fory.json.resolver.JsonTypeResolver.buildObjectTypeInfo(JsonTypeResolver.java:2676)
	at org.apache.fory.json.resolver.JsonTypeResolver.buildTypeInfo(JsonTypeResolver.java:2633)
	at org.apache.fory.json.resolver.JsonTypeResolver.resolveTypeInfo(JsonTypeResolver.java:512)
	at org.apache.fory.json.resolver.JsonTypeResolver.getTypeInfo(JsonTypeResolver.java:385)
```

[...]

Issue #4068, opened 2026-09-20
fix(scala): fory-json-scala cannot write Queue, primitive mutable ArraySeq, specialised Tuple2, Duration.Undefined, or ScalaTypeRef[Box[Int]] with codegen


pjfanning:
## Summary

While probing test coverage for `scala/fory-json-scala` (see #4067) I ran a throwaway suite against `main` on Scala 2.13.16 and 3.3.8. Six write-path assertions fail. Each one below is a self-contained ScalaTest block that fails today; the fixtures it needs are listed at the end. Nothing here is a read-path problem: every failing value can be decoded from JSON, it just cannot be encoded (or cannot be compiled by codegen).

Environment: `ForyJsonScala.builder().withCodegen(false).build()` unless the test says otherwise; `runtimes` is that instance plus `withAsyncCompilation(false)`.

## 1. Non-empty `immutable.Queue` cannot be written

`ScalaCollectionCodecs.RuntimeClasses` is seeded with `Queue.empty`, whose runtime class is the `Queue$EmptyQueue$` singleton, not `scala.collection.immutable.Queue`. A populated queue is a plain `Queue`, which is neither in the identity map nor in the `isAssignableFrom` fallbacks of `supportedRuntime`, so `requireSupportedRuntime` rejects it in `ScalaIterableCodec.writeString`. Decoding works and produces a `Queue`, so a decoded value cannot be re-encoded.

```
ForyJsonException: Unsupported Scala collection runtime type scala.collection.immutable.Queue
  at ScalaCollectionCodecs$.requireSupportedRuntime(ScalaCollectionCodecs.scala:1318)
  at ScalaIterableCodec.writeString(ScalaCollectionCodecs.scala:295)
```

```scala
  test("bug: non-empty immutable Queue cannot be written") {
    import scala.collection.{immutable => im}
    val json = runtimes.head
    val queueType = ScalaTypeRef[im.Queue[Int]]
    assert(json.toJson(im.Queue.empty[Int], queueType) == "[]")
    assert(json.fromJson("[1,2]", queueType) == im.Queue(1, 2))
    assert(json.toJson(im.Queue(1, 2), queueType) == "[1,2]")
    assert(json.toJson(im.Queue(1, 2)) == "[1,2]")
  }
```

## 2. Primitive-backed `mutable.ArraySeq` cannot be written

[...]

Issue #4067, opened 2026-09-20
test(scala): expand fory-json-scala coverage for documented types and error paths


pjfanning:
## Summary

`scala/fory-json-scala` has three suites: `ScalaJsonSuite` and `ScalaJsonEnumerationSuite` (shared by Scala 2.13 and 3) and `ScalaJsonDerivationSuite` (Scala 3 only). They cover case-class construction rules, `Either`, `BigInt`, `BitSet`, Boolean `ArraySeq`, `IntMap`/`LongMap`, hash set/map growth and the Scala 3 derivation paths well. Several types listed in the `docs/json/scala.md` support table and most of the error paths in `ScalaTypeCodecFactory` have no test at all. This issue collects concrete test suggestions, grouped by owner.

Counts below are from grepping `src/test` for the type name.

## 1. Documented scalar types with zero tests

- `BigDecimal` (`ScalaBigDecimalCodec`): 0 references. Suggest round trip of positive/negative/zero, large scale, negative exponent input (`1e-30`), quoted token, `null`, invalid token, and a `withMaxGraphMemoryBytes(ownerBytes - 1)` rejection like the `BigInt` test.
- Scala `StringBuilder` (`ScalaStringBuilderCodec`): 0 references. Suggest round trip (compare `toString`), empty string, `null`, non-Latin1 content through `toJsonBytes`/`fromJson(bytes)`, and a graph-memory rejection that exercises the `length + 16` capacity formula.

## 2. `Duration` and `Range` families

Only one `FiniteDuration(100 ms)` round trip and one `Range(1, 10, 2)` round trip exist.

`ScalaDurationCodec`:
- `Duration.Inf`, `Duration.MinusInf`, `Duration.Undefined` through the `Duration` declared type (writer emits `{"special":...}`). Note `Duration.Undefined.equals` always returns `false`, so compare with `eq`.
- A declared `FiniteDuration` rejecting `{"special":"INF"}` (the `finiteOnly` branch).
- Invalid objects: `{}`, unknown unit name, duplicated `length`, `length` without `unit`, `special` combined with `length`, unknown member. All must raise `ForyJsonException`.
- Every `TimeUnit` name round trip, `Long.MaxValue`/`Long.MinValue` lengths, and `null`.

`ScalaRangeCodec` / `ScalaNumericRangeCodec`:
- Empty `[]` and single-element `[5]` for both `Range.Inclusive` and `Range.Exclusive` declared types, and negative step `[9,7,5]`.
[...]

PR #4070, opened 2026-09-20
test(scala): cover documented fory-json-scala types and error paths

pjfanning:
## What does this PR do?

Adds tests to `ScalaJsonSuite` for `fory-json-scala` behaviour that is documented in `docs/json/scala.md` but had no coverage (#4067). Test-only change, no production code.

New cases, all run on Scala 2.13 and 3 and, where the existing suite does so, on both the interpreted and generated codecs:

- `BigDecimal` and Scala `StringBuilder`: round trips, quoted tokens, `null`, invalid tokens, graph-memory rejection
- `Duration`: `Inf`/`MinusInf` writes, `Undefined` read, `FiniteDuration` rejecting `special`, every `TimeUnit`, and the invalid-object branches
- `Range` and `NumericRange`: empty, single, negative step, non-progression and overflow rejection, `Byte`/`Short`/`Char`/`BigInt` elements, write-only raw `NumericRange`, unsupported `Double`
- Declared collection and map kinds that were untested (`ArrayBuffer`, `ListBuffer`, `ArrayDeque`, `mutable.Queue`, `LinkedHashSet`, abstract `IndexedSeq`/`LinearSeq`/`Buffer`/`mutable.Seq`, `VectorMap`, `ListMap`, `SeqMap`, `AnyRefMap`, `mutable.HashMap`, `im.HashMap`), insertion order for ordered kinds, `Nil` and `::` declared, `Map[Int, String]`, nested and null-element collections, read-side batch reservation above 1024 elements
- Runtime writes for those kinds and the documented rejection of sorted collections
- The rejected runtime-state families (`Stream`, `Iterator`, views, `Try`, `Future`, `Promise`, `Deadline`, `Regex`, `Symbol`, functions, `Ordering`) as roots and as fields
- `Option` fields (omitted, `null`, present, required), nested `Option`, `Some(null)`
- `Tuple1`, too-long tuple, `null` tuple and element, nested tuples, UTF-8 input
- `String` value class at root, in `List`/`Map`/`Option`, and `null`
- Root `BoxedUnit`, non-null `Unit` rejection, `case object`, nested object singleton
- Generic case class through `TypeRef` and as a field, `var` constructor parameters, a 24-parameter case class, missing required and unknown properties, `maxDepth` on a recursive case class
- Explicit `@JsonSubTypes` sealed trait with a case-object leaf in the shared suite (previously only the Scala 3 inferred path was tested)
- `ScalaEnumerationCodec` through `registerCodec`, custom `Value(id, name)` names, unknown-name rejection, duplicate-name constructor rejection
- `BigDecimal`, `StringBuilder`, `Duration`, `Range`, tuple, value class, `ArrayBuffer`, `ListMap` and `LinkedHashSet` as case-class fields under codegen

Cases that turned out to fail on `main` are deliberately left out and filed as #4068 (Scala write paths) and #4069 (core generic-recursion guard).

## Related issues

Refs #4067, #4068, #4069

- [x] Does this PR introduce any public API change? No
- [x] Does this PR introduce any binary protocol compatibility change? No

## Benchmark

Not applicable, test-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

This code is just test coverage and was not run with `ai_review` because that takes a very long time to run with multiple review agents confusing each other and causing work to be done over and over again.


chaokunyang (approved):
LGTM

PR #4078, opened 2026-09-22
fix(json): retain properties without defaults under NON_DEFAULT

chaokunyang:
## Why?

`NON_DEFAULT` rejected properties without defaults, so a Scala model such as `@JsonInclude(NON_DEFAULT) case class Request(id: String, retries: Int = 3)` could not serialize without an `ALWAYS` override on `id`. Java parameterized creators and Kotlin required parameters had the same problem.

## What does this PR do?

Keep properties without defaults in the output under both class-level `@JsonInclude(NON_DEFAULT)` and field-level `@JsonProperty(include = NON_DEFAULT)`, including null, zero, false and empty values. Apply the decision in the existing codec initialization path; per-write omission and generated code remain unchanged.

- Scala keeps properties without a supported compiler default method and continues comparing declared defaults using the current object's actual dependencies.
- Java keeps authorized properties when the selected read path has no no-argument reference constructor. An unrelated no-argument constructor does not replace the selected creator.
- Kotlin keeps required constructor properties and uninitialized `lateinit` properties in the write schema. Defaulted properties still require a legal all-default reference construction path; required arguments are never fabricated.
- Preserve failures from selected default methods, unavailable dependencies, and reference construction. Keep global `NON_DEFAULT` rejection.
- Update annotation Javadocs and JSON guides, and cover Mixin and generated/non-generated paths.

## Validation

OpenJDK 25.0.3:

- Java: `JsonInclusionTest` (30 tests) and `JsonTypeProcessorTest` (73 tests), including actual compilation and use of generated codecs.
- Scala: `ScalaJsonSuite`, 78 tests each on Scala 2.13.18 and 3.3.8.
- Kotlin: `KotlinDefaultArgumentsTest` (4 tests), including String/UTF-8 and compact/pretty output with codegen on/off.
- Kotlin KSP: `R8RulesWriterTest` and `ProcessorResourceTest` (10 tests). Existing retention paths need no production changes. These resource tests are not Android/native-image end-to-end validation.
- Java Spotless/Checkstyle, Kotlin Spotless, Markdown formatting and `git diff --check` passed.

No GraalVM or benchmarks were run.

Properties without defaults no longer cause an error under class-level or field-level `NON_DEFAULT`; their values are serialized. No public API additions or binary protocol changes.


ayush00git (approved):
LGTM

